PR #5, opened 2013-06-23
Update to Clipper 5.1.6

hroncok:


BagelOrb:
It might be the case that this PR is outdated :D

PR #28, opened 2013-09-19
added setting for bottom layer flow

dreimers:
Hi,

I added a setting to adjust the filament flow for the bottom layer as there was too much material at the beginning. This resulted in bubbles on the printed surface.

This pull request is according to request https://github.com/daid/Cura/issues/563

Best regards


BagelOrb:
outdated

PR #41, opened 2013-11-04
First travel movement towards initial position is not slowed down

demestav:
Travel speed at layer 0 is slowed down by a factor, as with the extrude speed. Initially, the extruder may need to travel a relatively large distance before it starts printing and because of the slowing down, its possible to have oozing, no matter how much you retracted before the move.

This commit, prevents slowing down the very first travel movement, i.e. the movement just before printing.


BagelOrb:
Would have been nice to have this now, but this PR is outdated.

PR #54, opened 2014-02-04
stretch algorithm (second try)

electrocbd:
This is a patch I used to fix incorrect diameters of vertical holes and, in general, insufficient inner dimensions. I created a new parameter "stretchDistance". When this parameter is zero (the default), nothing new occurs. When the parmeter is 600 - for my printer - the system increases diameters in a manner all diameters are accurate, even small ones.

The idea behind the algorithm is that the command sent to the printer is like the position of the front wheel of a bike, and the real position of the plastic is the position of the rear wheel. stretchDistance is the distance, in microns, between the front and the rear wheel.  Maths : http://www.mathcurve.com/courbes2d/tractrice/tractoiredecercle.shtml

### DIFF
--- a/fffProcessor.h
+++ b/fffProcessor.h
@@ -101,7 +101,7 @@ private:
         for(unsigned int n=1; n<MAX_EXTRUDERS;n++)
             gcode.setExtruderOffset(n, config.extruderOffset[n].p());
         gcode.setFlavor(config.gcodeFlavor);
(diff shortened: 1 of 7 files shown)


derekhe:
@electrocbd Any update about this patch? Seems it is quite useful

BagelOrb:
For further discussion, see https://github.com/Ultimaker/CuraEngine/pull/94

PR #58, opened 2014-02-22
Rewrite bridge angle detection

josefpavlik:
I rewrite the function for detecting bridges.
Now I'm searching the longest non attached edge of the bridge and report the angle of this edge.
In the case of 2 or more edges that are not attached, the angle between 2 ends of the chain of non attached edges is got.


BagelOrb:
Whatever happened to this pull request?
Why did you remove the bridging code?

josefpavlik:
My bridging code was buggy, in the meantime a working bridging code was added to mainstream. I don't know how it's implemented now, but seems that now works fine.

PR #59, opened 2014-03-11
Add perimeter inset setting to the Advanced settings.

GregFrost:
Added the inset setting as suggested by the discussion in this thread: http://forums.reprap.org/read.php?263,273929

The associated updates to Cura are in another pull request (pull/780).

This allows the flow rate to set so that 100% filled objects are perfectly filled whilst having an independent setting to adjust for the minor oversizing of objects because of the way the plastic is laid down (as discussed in the forum thread).

### DIFF
--- a/fffProcessor.h
+++ b/fffProcessor.h
@@ -221,7 +221,7 @@ private:
                 if (config.spiralizeMode && int(layerNr) < config.downSkinCount && layerNr % 2 == 1)//Add extra insets every 2 layers when spiralizing, this makes bottoms of cups watertight.
                     insetCount += 5;
                 SliceLayer* layer = &storage.volumes[volumeIdx].layers[layerNr];
-                generateInsets(layer, config.extrusionWidth, insetCount);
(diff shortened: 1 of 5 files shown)


derekhe:
I think I need this feature and can guys merge?

nallath:
Well, this is a pull request that is more then a year old. I'm not working on the engine, but I doubt this is going to be merged.

BagelOrb:
I think this is solved better by the XY-offset setting of the new Cura.

derekhe:
@BagelOrb
I think this is different to XY-offset. This patch looks like only treat inset which means only affect holes, not the outer side. XY offset will makes holes bigger while makes the outer side smaller (which is not expected)

BagelOrb:
@derekhe I've looked at the code and it does not treat holes diffferent from outsides.

derekhe:
@BagelOrb 
ok, thanks for that. I may be misleading by the function name "generateInsets". 
I will wait for the lastest cura version :) thanks

PR #70, opened 2014-03-22
Adding anchor line

Gregwar:
This adds the possibility of drawing an anchor line at the begining of the print on the X axis

It introduces following parameters:
* `anchorThickness` and `anchorLinewidth`, the size of the anchor line
* `anchorSpeed`, the speed of the anchor line
* `anchorY`, the Y position of the anchor line
* `anchorFromX` and `anchorToX`, the X interval on which the line have to be

It can be disabled with anchorThickness=0

### DIFF
--- a/fffProcessor.h
+++ b/fffProcessor.h
@@ -346,6 +346,24 @@ private:
         unsigned int totalLayers = storage.volumes[0].layers.size();
         gcode.writeComment("Layer count: %d", totalLayers);
 
+        if (config.anchorThickness > 0)
+        {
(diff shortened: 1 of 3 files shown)


BagelOrb:
I've never reviewed this PR. What is an anchor line for? Perhaps I will reintroduce the setting myself sometime.

Gregwar:
An anchor line is a line that is extruded for instance in the front of the plate to begin the extrusion process (you get this for instance using current Makerbot Desktop)
However, @daid introduced another system in Cura after this PR (when the printer begins extrusion before starting the print)

PR #74, opened 2014-03-31
Up skin generation error when the raft support is used.

Ryoko:
When the raft used, up skin is generated thinner than configurated by settings.
In my case, with layer height 0.25 and bottom/top thickness = 0.5, without raft - generated 3 bottom and top skin layers. But when I set raft support - only one up skin layer generated.

So I've made this fix, which resolve this issue.

P.S. This pull request contain one line changed in my previous pull request. It's line 390 in fffProcessor.h file. I don't know how to take it out of this pull request.


BagelOrb:
Not a problem in Cura 2

PR #78, opened 2014-04-08
C++11 range-based for loops and minor cleanup

Skeen:
This pull request adds C++11 for loops instead of indexed loops.
Also there are some minor changes, to utilize the algorithm standard header, rather than duplicate code functions.
Lastly, fffprocessor and multivolume has been split into CPP/H files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ BUILD_DIR = build
 SRC_DIR = src
 LIBS_DIR = libs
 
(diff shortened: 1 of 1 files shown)


BagelOrb:
lambda's are in now, range based for loops are used and files we are touching also get split into .h and .cpp files.

Do note that index based loops are pretty handy. You can have vectors of the same size to map from a item in the list to some property which you want all items in the original vector to have.

PR #88, opened 2014-04-27
Kisslicer like 50/50 infill under top shells.

Skeen:
Added Kisslicer like 50/50 infill under top shells, to better support these.


BagelOrb:
This is possible using Gradual infill. It's not totally the same: gradual infill does solve the underlying problem, but in a different way.

PR #89, opened 2014-04-27
Made polygon iterable, and moved it's implementation into a seperate file

Skeen:
Making polygon iterable will allow us to do a lot more range-based for loops.
Having the implementation in a seperate file ensures the file is compiled whenever it is changed, and should shorten rebuild times.


BagelOrb:
Most of this PR has already been done, but because our maxime is to only fix the problems where we encounter them, some implementation might still be in the header file.

Issue #47, opened 2013-12-11
Won't compile, missing timeEstimate.cpp


TopherMan:
CuraEngine won't compile since the addition of the time estimate. It looks like it's just missing from the list of  SOURCES, so if I add timeEstimate.cpp to the end of the SOURCES line in the Makefile, it works properly.

PR #94, opened 2014-05-13
Stretch algorithm

electrocbd:
This is a patch I used to fix incorrect diameters of vertical holes and, in general, insufficient inner dimensions. I created a new parameter "stretchDistance". When this parameter is zero (the default), nothing new occurs. When the parmeter is 600 - for my printer - the system increases diameters in a manner all diameters are accurate, even small ones.
The idea behind the algorithm is that the command sent to the printer is like the position of the front wheel of a bike, and the real position of the plastic is the position of the rear wheel. stretchDistance is the distance, in microns, between the front and the rear wheel. Maths : http://www.mathcurve.com/courbes2d/tractrice/tractoiredecercle.shtml

The following pictures show the effect of the stretch algorithm. A simple object with 1, 2, 3, 4, 5, 6, 7 and 8 mm holes diameter is a perfect fit for corresponding drills
stretch2
stretch1

Previous (closed) pull requests are #53 and #54


BagelOrb:
What's the rationale for using the bike model?

At http://www.mathcurve.com/courbes2d/tractrice/tractoiredecercle.shtml we see that a small circle will result in a weird looking flower. Why would we want such a thing?

electrocbd:
Hi
The good picture is when the bike runs forward, not backward.: http://www.mathcurve.com/courbes2d/tractrice/tractoire-GrandBi-cercle.gif
The rationale is that the plastic acts as a string which "take a shortcut' when the head turns,
exactly like the back Wheel (the deposited filament)  compared to the front Wheel (the head)
My measures of the actual diameter of a hole versus the wanted diameter are similar.

nophead:
What happens when you turn a sharp corner, for example a polygon with a low number of sides. Doesn't the head have to overshoot the corner and then instantly jump to new line direction? Since there are no real curves doesn't this happen on every corner?

BagelOrb:
I suspect this pull request severely impacts performance. Have you tested the computing time difference?

electrocbd:
For the sharp corners, the algorithm doesn't change movement if angles are too sharp, so it is "stable", and for performance, it is slighly slower.

BagelOrb:
I don't understand why you needed to add so much code. I would expect that in order to compensate for the effect, you just make each segment longer by the stretchDistance and introduce a point on the next segment around 1.5 times the stretchDistance, if the next segment is is long enough to contain that point.

That would include just a couple lines of code, IMHO.

Wouldn't my idea be simpler and also as effective as yours?

Could you explain how you did it?

electrocbd:
The main issue is to have a reliable behavior when an inset have few points. Ideally, a circle should be converted into a new circle with an higher radius, even if this circle has only 4 points. Other shapes like gears have also to be fixed and functional, not shifted in any direction.
Another issue is the amount a plastic extruded. When a circle has an higher radius because we anticipate its retraction, we shall not extrude more plastic than for its initial radius (if we do, the extra amount of plastic fills the hole, which is counter effective).
Another issue is the precision of the computation of the radius of the part of the curve we a modifying. When using only two successive points, when the points are very near, the estimated radius is wrong, so the correction is bad.

electrocbd:
One of my first tries used an algorithm similar than the one you describe. For very simple shapes like circles, the result was good, but for a normal objects with many different curves, it had too many side effects.

BagelOrb:
"Another issue is the amount a plastic extruded."
Ah yes, compensation such that the amount of material extruded is the same is also a required aspect.

It seems your fix has gotten out of hand. I think the math is quite correct for the simple algorithm; the fact that it had many side effects seems to me to prove that the 'bike model' is not such a good model for how the plastic behaves.

Indeed, when thinking about it I can imagine that the inside of a curved line is better modelled by the bike model than the outside. >>

"Another issue is the precision of the computation of the radius of the part of the curve we a modifying."
Wait, what do you use the radius for? As I understand it, you are saying that you fit a circle to any part in the curve and you use the radius of this circle for something I don't know.

electrocbd:
"the fact that it had many side effect"
The current algorithm does not have side effects (I haven't seen one since one year). The side effects were with the first version.
"what do you use the radius for?"
In each point of any inner or outer wall, I compute the radius of the nearest circle. Then, I find the center of this circle, and I increase the radius to move the point. So, each point is moved in the opposite direction than the direction of its "natural retraction".
The factor by which the radius is multiplied is related to the "bike model maths". It is more important for small circles than for big circles.
The advantage of this method is that it works with any kind of curve.

BagelOrb:
You misunderstood; I was quoting you when talking about side effects.
"One of my first tries used an algorithm similar than the one you describe. For very simple shapes like circles, the result was good, but for a normal objects with many different curves, it had too many side effects."

It seems you started with the simple algorithm I described, but when you saw it didn't work so well, you started tweaking and made the code quite complex, instead of concluding that the approach wasn't that good. It still seems like that to me..

I still don't understand the thing about circles. You talk about "the nearest circle", while you could find any circle arbitrarily close to one given point. Perhaps you can explain by a picture?

Here's a rough drawing of my simple algorithm.
bike_drag_compensation

electrocbd:
The following picture shows the effect of the current algorithm, with a high value for stretchDistance.
stretch_effect
The blue points are the initial points, and the yellow points are the new points. When the yellow points are not visible, at the beginning and the end, they are exactly at the same position than the blue ones.

electrocbd:
The initial idea is really simple. I first printed many holes with different sizes (from 1mm to 8 mm), and measured the real inner diameter. I then searched a formula to "explain" the difference between the expected and the obtained diameter. The formula that fits the best is the one of the "bike model".

electrocbd:
To have a vertical hole with the real diameter D, it is necessary to print a hole with the diameter sqrt(D<sup>2</sup>+S<sup>2</sup>) where S is a constant, the "stretch distance". There is no need to compensate horizontal holes.

electrocbd:
To extend this algorithm for every kind of curve, I search in each point the nearest circle (looking at previous and next point to have three point) and I move the point as if it was a point of this circle. I mean I move the point in the direction opposite to the center of the circle. If the distance between the point and the center of the circle is D, I move it at the distance sqrt(D<sup>2</sup>+S<sup>2</sup>) from the center.

electrocbd:
circle
In this figure, P1 P2 and P3 and three successive points of the curve. C is the center of the circle, and NP is the new point, which will be used in place of P2. The distance between NP and C is sqrt(D<sup>2</sup>+S<sup>2</sup>) where D is the distance between C and P2

This operation is done for every point a the curve, in fact each point of any inner or outer wall.

BagelOrb:
Alright. That's pretty clear!

Your algorithm is symmetric. The yellow curve you plotted is rotationally symmetric because the blue curve is also rotationally symmetric. This is not true for the bike-model, however.
In the bike model the upper part of the yellow line would never be above the blue line, because any positive Y-component would also cause the back wheel to have a (smaller) positive Y-component, while the blue curve has a negative Y-component everywhere!

Your algorithm doesn't follow the bike model; however it is still quite interesting!
I will think about it.

derekhe:
@BagelOrb @electrocbd 
Any update about this patch? Seems this could fix some small hole problem. XY offset is bad because the outer dimension will be affected as well.

BagelOrb:
This PR will not be merged because it is not theoretically sound.

Having the very same model in higher poly causes it to be printed differently. 
Lines can start to collide, with themselves and all kinds of horrible may happen.
Moreover, The circle only determines the direction in which a point is moved - even straight lines will get shifted.

See below for a little visual feedback on unwanted results.
stretch_alg_is_bs

electrocbd:
You say that straight lines get shifted, but for straight line, the radius is infinite, and in this case, there is no correction
I stop applying correction when angles are too sharp, because in this case the circle is not a good approximation of the real movement. Your last figure is a good example, without additional points, it is very difficult to determine what *should* by the movement.
With the "security" I added in the source code to stop modifying the path for sharp angles, my patch is, for the moment, the one which does the best job for inner holes without affecting outer dimensions.

derekhe:
@electrocbd 
I'm thinking about another way. Could we only detect inside holes and stretch a certain distance if the area is small enough? 
For example, we slice and find a hole in side (maybe it is not circle, can be any shape), then stretch this area to bigger. 
If it is a open hole, just ignore it.

BagelOrb:
@electrocbd Yeah of course, but if the line is infinitesimally close to being a straight line it will get shifted.

I thought the bike model had some logic behind it which your current algorithm has lost. Why shouldn't we implement the algorithm I proposed (20 apr)?

Swonkie:
I am not familiar with the slicer code, but I would like to give my insight to this interesting issue. It seems to me, that it's simply a matter of ofsetting the wall paths, because the problem arises from the nozzle not being infinitely small, but having a certain diameter.

This illustration shows what I mean:
extruder

BagelOrb:
@swonkie, nice illustration, but that is how it's done already

Quick note: my bike algorithm actually also models what has been called jerk. The stretch of the material counters the inertia of the printing head. Presumably the stretch has more effect than the overshoot by inertia, though. Otherwise the stretch hypothesis doesn't hold.

electrocbd:
@BagelOrb I am afraid that your algorithm induces a distortion. The shape will be globally turned in the same direction as the head. If the head turns clockwise, the shape will be turned clockwise. It was an undesirable effect I noticed during my preliminary tests. This is why I tried to have a symmetric algorithm. This is the same with a real bike. One have to anticipate turns to have the rear wheel at the desired position.

electrocbd:
@derekhe I think it is important to have a kind of symmetric approach. When printing a gear, the teeth have to be rectified the same way in the bottom and in the top. Material has to be removed from the bottom, and material has to be added to the top. A normal - not corrected - print will tend to create triangular teeth because plastic "take a shortcut"

BagelOrb:
@electrocbd Of course my algorithm introduces a distortion; one that counters distortions induced by filament stretch. 

Your example is incorrect, my algorithm would not just turn it, since it would introduce new points just after the corners it moved. 

The stretching of the filament is not a symmetric process. The plastic takes a shortcut from where it was to where it is now, not to where it is going to be. 

This is similar with a bike. You don't anticipate the turn of the back wheel. It's hard to explain. Imagine you are riding right beside the curb and the road makes a corner right. If you anticipate the move of your back wheel your front wheel will first crash into the curb and then after that the back wheel will. When your front wheel exactly follows the curb the backwheel will grind against the curb, since a straight line from the back wheel to the front wheel intersects with the curb.

AKSoapy29:
What ever happened with this? Has it been officially implemented? I'd like to use it / figure it out.

electrocbd:
On Sun, Apr 09, 2017 at 10:22:14AM -0700, AKSoapy29 wrote:
> What ever happened with this? Has it been officially implemented? I'd like to
> use it / figure it out.
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or mute the thread.*
> 
No, because my algorithm had been rejected (too many side effects)

The general idea could by good, but it is necessary to find the nearest
circle to increase its radius, and there are often too few points in each wall's polygon to find a reliable radius.

Using a brand new ultimaker 2+ with all default settings, the effect
seems to be far less important than the global constant offset of about 0,1mm for the radius of each hole.

It seems that each time a wall turn left with nothing at its left, it is
globally moved by a fixed distance to the left (and the same thing to the right of
course)

If I had time for this, I would try to implement this algorithm: For
each triangle (3 consecutive points of a polygon's wall), if there is no
plastic inside this triangle (deposited before) then move the middle
point far from the two others by a constant distance.

Regards

C.Baribaud

DDDirk:
I have a similar issue with another project, where I make a curve extra wide. And in general, I find this rather important for mechanical parts. I regularly have to adapt models to make sure the print comes out right. That's not the way it should be. Either the slicer or the firmware should handle this.

A general solution that I see is similar to BagelOrb's proposal (and drawing) of 20 april 2015. Extend both line segments of the turn a bit and connect those two vertices. So the nozzle overshoots the original vertex, then moves sideways until it crosses the extended other line segment and then continues from there.

So this:

..........
..../\....
.../  \...
../    \..

becomes that:
   ___
...\.../....
... \./.....
..../\......
.../  \.....
../    \....

(This is for a concave turn; the printed part is shown by the dots.)

The move sideways is because of precisely the problem this is meant to deal with, namely that the nozzle drags the plastic into the turn. So now the nozzle drags it back a bit first. This prevents the problem that electrocbd mentioned on 15 june 2015, that if it is not symmetrical the shape gets turned in the direction of the turn.
Of course there should be no extrusion during the extended bit, in other words those have to be G0's. But there might still be a bit of ooze.

So there are now 3 vertices instead of 1. And there are 3 extra lines of gcode (3 G0's), which may significantly increase the size of the file, but that should not be a problem, I assume. And the maths are quite simple, so I don't think this will significantly increase calculation time.
(Btw, I assume here that it is dealt with by Cura, but it could also be done by the firmware. I don't know which is best.)

How far to extend the lines will or may depend on:
- Nozzle size (the size of the extrusion hole). This is probably the most important factor. (My guess is the extension length should be about the nozzle size.)
- The size of the physical nozzle (the flat bit around the hole), because that is what the plastic sticks to.
- How sharp the curve is. For a very shallow curve it might be left out completely.
- The speed and mass of the printhead. (In other words jerk?)
- The 'stickyness' or viscosity (and therefore the temperature) of the material, but I think that will be less of an issue here than it normally is.
Also, the extended paths don't have to be exacly like in the drawing. Finding out which direction is best is a matter of experimentation. With some luck, it might turn out that following the bisectrix between the two line segments gives tjhe best results. If that is true for both extensions, then only 1 vertex and 2 lines of gcode need to be added.

It becomes more complicated when turns are close to each other. For example, if it is a semi-circle that is well defined (many vertices). Or maybe it becomes simpler then. Just do it for the first turn, but not the following ones. That could at the very least be a first implementation (to experiment with).


Btw, I wonder if something similar could be done with convex turns (where the model has a sticky out bit). Especially a sharp turn, like in the drawing, tends to be rounded off by the nozzle, but this might make it more 'pointy'. Of course it shouldn't be 'blobby'. But that might be prevented by extruding less material, especially on re-entry (resuming the normal print after the extension).
This way the stickyness is put to good use (turning disadvantages into advantages is one of my great hobbies :) ).

electrocbd:
I am currently using a post processing program (which translate gcode into gcode)  to have good inner and outer diameters. It is not necessary to add more points, but only to move part of them without changing the extrusion amount.
So far, the best results I obtained use two consecutive algorithms. The first move each point of a curve a little to the outside of the turn, and the second move each point of any segment to the direction of wall which is on one and only one of its sides, to "push the wall".
This two algorithms, when used together, allow to have good inner and outer diameters

DDDirk:
Do I understand correctly that you now use something completely different? If so, could you share those programs? (I am rather new to github, so I don't know how you could do that.)

I assume that you do the first step only for concave turns. Or do you move all points towards the printed area? In other words, do you use an extra inset?
The second part I don't understand.

electrocbd:
I just created a repository : https://github.com/electrocbd/post_stretch
There are a lot of work remaining: Translate it in English, add documentation etc.
It could be interesting to add it as a post processing option in Cura, but I have no idea how to do this.

Ghostkeeper:
If you think it works nicely and relatively bug-free, then you can consider making a pull request into our repository for the post-processing plug-in: https://github.com/nallath/PostProcessingPlugin/pulls

electrocbd:
I just created pull request "New stretch script #47". It is slower than the C++ code (python is slooow) but it gives the same results

BagelOrb:
Reopening issue because of the info in here which is still valuable for when anything like this will ever get implemented.

Search terms: [dimensional accuracy, inner outer corners holes compensate inward outward]

ChrisTerBeke:
Closing because super duper old. Relevant info can still be found even though it's archived.

BagelOrb:
Being old shouldn't be a criterion for closing an issue. It's about whether this issue is relevant now or in the near future.

ChrisTerBeke:
It hasn't been relevant since May, but if the PR author (or someone else) would update the PR to make it relevant again for modern Cura, we can still review it of course. If no one takes that initiative, I'd call it not relevant anymore.

BagelOrb:
oh wait this is the PR... never mind.

PR #97, opened 2014-05-20
Adds support for multi extrusion print based on color flag in the STL

zuzkins:
- this feature must be enabled by the '-e' flag
- supports only binary STL files
- what this does, is that it groups all faces with same color into same
  SimpleVolume. Model then has as many volumes, as there were distinct
  colors for the faces, so it requires you to color your model correctly.
  The effect of grouping the faces is that at the end multiple extruders
  will be used.
- the volumes get sorted by the (numerical) value of the color (so color
  1 will be printed with extruder 1 and volume with color 2 with
  extruder 2 no matter what was the order of faces discovered in the STL file

### DIFF
--- a/src/fffProcessor.h
+++ b/src/fffProcessor.h
@@ -17,6 +17,7 @@ class fffProcessor
 private:
     int maxObjectHeight;
     int fileNr;
+    bool coloredStlSupport = false;
     GCodeExport gcode;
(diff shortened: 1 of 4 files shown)


BagelOrb:
There are better formats than STL for achieving this.

Also this is hopelessly outdated.

PR #103, opened 2014-05-29
Add option to Reverse Insets Print Order

### LINKED ISSUE #98: Outer Perimeter First
Hi,
Could an option to print the outer Perimeter before any other Perimeter be added.
I find that although overhangs are more difficult with outer perimeter first an objects dimensions tend to be more accurate.

For reference KISSlicer calls this option "Loops go from inside to Perimeter" and Slic3r calls this option "External perimeters first"

Thanks
Neil

totalretribution:
"this commit closes #98"

### DIFF
--- a/src/gcodeExport.cpp
+++ b/src/gcodeExport.cpp
@@ -233,7 +233,7 @@ void GCodeExport::writeMove(Point p, int speed, int lineWidth)
                 isRetracted = true;
             }
         }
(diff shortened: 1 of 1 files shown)


paranojik:
I think you merged the wrong commit. This has nothing to do with the order of perimeters. @totalretribution I think you should create a new merge request against the correct branch.

PR #109, opened 2014-06-21
Increase adhesion of support structure with raft by filling the air gap

Bevaz:
The base of support structure does not adhere properly to the raft because of the air gap.
This patch leaves air gap for the model, but fills for the support structure.


BagelOrb:
Oh wow, a forgotten pull request from ancient times when @daid was in the process of handing CuraEngine over to me. I guess that's why this PR never got merged.

This has been fixed in the meantime, though. The filler layers have a layer thickness closest to the normal layer thickness, such that the airgap is an integer multiple of the filler layer height...

PR #113, opened 2014-07-08
Print blobs reduction

michaelDomingues:
Printing with latest versions was causing blobs during prints:
before

To reduce them it was inserted a clipping algorithm in the writeGCode method to clip points for a limit distance (new SETTING). Depending if it is a INNER or OUTER wall a retraction or a fast move is done.
After results show improvements:

after

Algorithm behavior:

Planner (previous)

cura_3
cura_4

Planner (after)

cura_13_5

Clipping with retractions (after)

cura_13_3

cura_13_4

Summary:
- New setting - pointsClipDistance  (settings.cpp)
- New GCode path configuration data constructor -  void setData(int speed, int lineWidth, int clip, const char\* name)   (gcodeExport.h)
- New GCode export algorithm when clipping and spiralize is off (gcodeExport.cpp)
- GCodePathConfig (inset0Config and insetXConfig) using setData constructor to support clip  (fffProcessor.h)
- New GCode moves for Inner and Outer walls with clip, if not spiralize.

### DIFF
--- a/src/fffProcessor.h
+++ b/src/fffProcessor.h
@@ -96,8 +96,8 @@ private:
     void preSetup()
     {
         skirtConfig.setData(config.printSpeed, config.extrusionWidth, "SKIRT");
-        inset0Config.setData(config.inset0Speed, config.extrusionWidth, "WALL-OUTER");
-        insetXConfig.setData(config.insetXSpeed, config.extrusionWidth, "WALL-INNER");
+        inset0Config.setData(config.inset0Speed, config.extrusionWidth, config.pointsClipDistance, "WALL-OUTER");
+        insetXConfig.setData(config.insetXSpeed, config.extrusionWidth, config.pointsClipDistance, "WALL-INNER");
         fillConfig.setData(config.infillSpeed, config.extrusionWidth, "FILL");
         supportConfig.setData(config.printSpeed, config.extrusionWidth, "SUPPORT");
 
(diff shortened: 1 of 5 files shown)


BagelOrb:
It would certainly be nice to have this, but this PR is outdated.

BagelOrb:
for my own reference: CURA-3442

thinkyhead:
Evidently this breaks `LIN_ADVANCE` (by @SebastianV650) in Marlin 1.1 (RC) and it may break other pressure-based extrusion methods that calculate volume on the fly.

The problem seems to stem from the small extrusion values used in these moves (rather than using 0.0 which, I understand, would trigger non-extrusion speed in the firmware). We're trying to come up with a workaround, but it's not exactly elegant. Our "hack" sets the **W**idth-over-**H**eight-over-**D**istance to a constant value instead of calculating it. This, unfortunately, may potentially break those prints that use more than one layer height or variable extrusion width.

So, I'm just popping by to ask, is this Cura behavior something that can be disabled? If so, then it may be a better option for use to just document the issue rather than incorporate yet another variable factor.

Ghostkeeper:
We have no setting that guarantees that small extrusion moves are not generated.

There is a setting though that generates a lot of them: Fill Gaps Between Walls. If you set that to Nowhere, it won't try to fill little gaps between walls and in corners.

There's a few more features that use more than one layer height, such as infill layer thickness, but those are not used by default.

thinkyhead:
Thanks @Ghostkeeper. The point of Advance is to reduce blobs by a similar principle (pressure already sufficient). I see slicers taking on this role more. But I feel there is still some need for pressure-based extrusion in the firmware. We will try to figure out a method that can adapt to these moves. At the moment it's all-or-nothing. Either the firmware always uses Advance, or it never does. Perhaps we can make it switchable for cases like this.

PR #117, opened 2014-07-19
Fix PolygonRef::inside

electrocbd:
The function PolygonRef::inside fails when the tested point is at the same Y value than an edge of the polygon, like in the following example.

inside_fix

The red point is considered in the yellow triangle, this is obviously wrong.
It is possible to test the issue with the following test case:
<pre>
void inside_testcase()
{
    Point p;
    p.X = 0;
    p.Y = 1;
    Polygon poly;
    {
        Point pt;
        pt.X = 1;
        pt.Y = 0;
        poly.add(pt);
    }
    {
        Point pt;
        pt.X = 2;
        pt.Y = 1;
        poly.add(pt);
    }
    {
        Point pt;
        pt.X = 3;
        pt.Y = 0;
        poly.add(pt);
    }
    if (poly.inside(p))
        printf("Point in triangle (this is wrong!)\n");
    else
        printf("Point out of triangle\n");
}
</pre>
My fix consists in using the function <i>PointInPolygon</i> provided by Clipper.

### DIFF
--- a/src/utils/polygon.h
+++ b/src/utils/polygon.h
@@ -134,28 +134,10 @@ public:
         return ret;
     }
     
-    //Check if we are inside the polygon. We do this by tracing from the point towards the negative X direction,
-    //  every line we cross increments the crossings counter. If we have an even number of crossings then we are not inside the polygon.
+    //Check if we are inside the polygon.
     bool inside(Point p)
     {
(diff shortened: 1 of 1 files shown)


BagelOrb:
Has been fixed in the mean time.

PR #121, opened 2014-08-11
Add a bulge correction factor to the outer contour to make outer dimensions accurate

OskarLinde:
Yet another dimension correction patch. This one doesn't address filament stretching for rounded contours, but merely adjusts for the bulging of the outer surface of the outer perimeter. Assuming a circular bulge profile, the correction factor is easy to calculate and no parameter is needed.

Here is an example of how, with this patch, a 10x10 mm square peg (with rounded corners) fits tightly into a 10x10 mm square hole with quite good dimensional accuracy:

h_id
h_od
p_od
fit2
fit1


BagelOrb:
Horizontal Offset is the option to configure this. This option has been implemented in the meantime.

PR #151, opened 2014-12-26
Fix wrong pull-request #134

repetier:
After updating to newer CuraEngine I got error reports that position was not honored as it should. Reason was Pull-Request #134 which rendered the disabling centering useless. This patch undoes the line position change that caused it.

### DIFF
--- a/src/optimizedModel.h
+++ b/src/optimizedModel.h
@@ -62,6 +62,7 @@ public:
         vMax = model->max();
 
         Point3 vOffset((vMin.x + vMax.x) / 2, (vMin.y + vMax.y) / 2, vMin.z);
+        vOffset -= center;
         if(ConfigSettings::config->autoCenter != 1)
         {
             vOffset.x = 0;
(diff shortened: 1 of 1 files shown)


zwbrush:
In my mind，constructor parameter("center") which according by command line parameters(-s posx=?, -s posy=?) mean the offset which user assigned.  if set offset to 0 , posx,posy will disabled even set autocenter=0.

repetier:
@zwbrush If I understand you correctly, you want to define a different center position, right? That might be interesting for some things, but it would still be centering only around a different point.

I added the autocenter parameter to disable the centering, simply to get the stl printed where it is accoring to stl coordinates, so the user gets them where they were set and not around some other point. What you do might be done if autocenter = 1 as it is centering.

zwbrush:
@repetier, I think I understand what you want now, you only want to print the file at its own coordinates, right? But if I only want to offset the mode to an specified value(posx, posy) without autocenter , I have no way to reach it. I think you can set posx =0, posy=0, autocenter=0 to get what you want.

zwbrush:
@repetier, I think our difference of opinion is what's the mean of posx & posy, I think it should be "Object Position" like its name in setting.h, and maybe you think it should be bed center.

repetier:
You are right. I used posxy as center position. With your advice I reverted to your fix and set pos to 0,0 and it also gets the correct result.

So I revoke the patch, now that I have rewritten the host to the new version. No problem as long a users do not start to mix binaries.

Thanks for showing what the reason for your change was. I did not realize this, only the user reports about wrong positioned objects after the update.

zwbrush:
:)

Date: Sun, 1 Feb 2015 08:38:59 -0800

PR #152, opened 2015-01-03
New option: Add chamfer to bottom of models

OskarLinde:
This merge request adds an option to add a (45°) chamfer to the bottom of each model. 0.5 mm is usually a good value. This can be used to eliminate oversized bottom layers due to squishing ("Elephant's foot"). A slightly larger chamfer also makes the model easier to remove from the bed.

### DIFF
--- a/src/fffProcessor.h
+++ b/src/fffProcessor.h
@@ -254,7 +254,11 @@ private:
                 int extrusionWidth = config.extrusionWidth;
                 if (layerNr == 0)
                     extrusionWidth = config.layer0extrusionWidth;
-                generateInsets(layer, extrusionWidth, insetCount);
+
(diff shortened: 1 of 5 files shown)


daid:
Patch looks nice, small and simple. And adds an interesting feature that people might like.

Small comment tough, layer->printZ should be layer->sliceZ I think. As the printZ is higher when the raft is added (there is only a difference between printZ and sliceZ when the raft is used)

OskarLinde:
Thank you! Updated the patch to use sliceZ.

BagelOrb:
PR is outdated. Also I think this should be fixed by levelling your bed correctly, but indeed people might like it, so I am willing to accept a new PR to this effect.

PR #153, opened 2015-01-04
Use retract on skins

presslab-us:
Using combing on skins causes unsightly scarring on flat surfaces.
By forcing a retract between travel moves we can lift the head
above the surface to avoid this.

### DIFF
--- a/src/fffProcessor.h
+++ b/src/fffProcessor.h
@@ -100,7 +100,7 @@ private:
         inset0Config.setData(config.inset0Speed, config.extrusionWidth, "WALL-OUTER");
         insetXConfig.setData(config.insetXSpeed, config.extrusionWidth, "WALL-INNER");
         infillConfig.setData(config.infillSpeed, config.extrusionWidth, "FILL");
-        skinConfig.setData(config.skinSpeed, config.extrusionWidth, "FILL");
+        skinConfig.setData(config.skinSpeed, config.extrusionWidth, "SKIN", true);
         supportConfig.setData(config.printSpeed, config.extrusionWidth, "SUPPORT");
 
         for(unsigned int n=1; n<MAX_EXTRUDERS;n++)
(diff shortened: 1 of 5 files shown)


BagelOrb:
A lot of retraction can cause the feeder to dig into the filament, leading to underextrusion or no extrusion at all. Did you implement the change as an option? Or did you change the code such that restraction always occurs?

daid:
I don't think this is the proper solution at all. First off, it's not configurable as @BagelOrb mentions.
Secondly, it forces retraction on every single move, which only isn't grinding your filament to a total flat pieces because of the "retractionMinimalDistance". Which some people set to 0.

presslab-us:
How would you like it to work?

BagelOrb:
There are a number of possible solutions.
A good solution would be to perform combing along the inside of the polygon, instead of the outside border. 
However, it is very hard to think of a suitable algorithm to find a path on the inside which can't fail and which doesn't take an extraordinary amount of time (like O^2).

As of yet there is no workable alternative. Can you think of any?

presslab-us:
Is combing preferred to retractions?  I realize each has it's own advantages...  But I think retractions work pretty good these days, and are often faster than combing.  I think finding a way to make retractions sensical on skin layers would be a workable solution.

In response to @daid, what would happen if combing were turned off with a retractionMinimalDistance of 0?  It seems like this would produce a retraction for every line of skin, as the code is now.  A way around this is to have a different minimal distance for the skin.  But this is a bad solution for a few reasons, and I think adding more options to the slicer is bad unless absolutely necessary.  Any additions should "just work" if possible.

presslab-us:
I have updated my branch, allowing this change to be configured.  Please let me know what you guys think, and if it's not acceptable let me know how you would like it to look.  I have modified Cura also, and can submit a pull request for that once this one is completed.

PR #157, opened 2015-02-05
Add wipe after retract feature.

presslab-us:
The setting retractionWipe > 0 enables this feature and defines the
length of the wipe move.  After a retract the head will retrace the
print path before doing the z-hop.  This can help with stringing on
difficult materials (e.g. nylon) and also allows for a larger z-hop
to keep the head from hitting the print on travel moves.
Reference this issue from a while ago: https://github.com/daid/Cura/issues/291

### DIFF
--- a/src/fffProcessor.h
+++ b/src/fffProcessor.h
@@ -107,7 +107,7 @@ private:
             gcode.setExtruderOffset(n, config.extruderOffset[n].p());
         gcode.setSwitchExtruderCode(config.preSwitchExtruderCode, config.postSwitchExtruderCode);
         gcode.setFlavor(config.gcodeFlavor);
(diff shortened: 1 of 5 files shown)


daid:
Intresting. Not merging yet, because I need to do a maintaince release right now. But after that I'm pretty sure I want to merge this and look more into it.

presslab-us:
Hi @daid I'll be gone for a long time coming up in 2 weeks.  So now is the time if there is anything you want to discuss on this.

BagelOrb:
@daid: you should take a look at this.

hakimio:
@daid Any idea when you'll take a look at the patch? It would be really nice feature for most people using bowden extruders.

BagelOrb:
By now we're pretty close to releasing the new Cura, which resides in the PinkUnicorn branch. This Cura has a totally revised settings system.

If you convert your code and create a PR for the PinkUnicorn branch, I will look into merging it instead of daid.

One question: is the wipe length based on the length of the past couple of line segments? Or do you wipe a given distance along the number of segments necessary to get that distance?

PR #158, opened 2015-02-09
Wipe Tower Offset

reprappro:
This commit of our CuraEngine fork allows the user optionally to reposition the wipe tower.  This is useful for big prints that go near to the edge of the build bed becuase it allows you to put the tower in any unused place in the print.

### DIFF
--- a/src/fffProcessor.h
+++ b/src/fffProcessor.h
@@ -1,3 +1,6 @@
+// Altered by Adrian Bowyer 6 February 2015
+// Code added to allow the user to position the wipe tower
+
 #ifndef FFF_PROCESSOR_H
 #define FFF_PROCESSOR_H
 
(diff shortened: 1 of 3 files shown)


Dim3nsioneer:
Good idea. What about a GUI realisation?

adamarumjahn:
David, I'm wondering if it's possible to add a feature for a real physical wiper (not a printed one), like the one in the CubeX or Stratasys printers below.  I figured this would be the best place to add it if you're working on this already.  It would be a nice solution for many of the multi-extruder printers out there to avoid wasting material on the wipe/prime towers.

cubexwiper

stratasyswiper

BagelOrb:
@innovo3d you can easily insert the wipe code as a pre extruder switch gcode

innovo3d:
Thanks Tim.  That's a really great suggestion.  I'll try that and see how
it goes.

Is there any provision for cooling down the extruder before an Extruder
Switch? to help prevent oozing?

*Adam Rumjahn*

* <http://www.innovo3d.co>*

*adam@innovo3d.co <adam@innovo3d.co>Cel: +57 304 643 1248*
*Fijo: +57 2 374 2473*

BagelOrb:
There is no such provision yet.

Cooling down can either happen during printing or it should be in the
extruder switch gcode.

It's not yet possible to do it during printing, but I will be looking into
that in the near future.
Op 28 sep. 2015 19:40 schreef "Adam Rumjahn" <notifications@github.com>:

innovo3d:
The problem by doing it in the G-code is that it still takes some time to
cool down, and meanwhile the filament will keep oozing.

If you place it in the slicer engine during printer, I think it would be
helpful to be able to cool down the active extruder in anticipation of the
extruder switch, say "begin cool down X seconds before extruder switch" and
the user can select time X.  This will help minimize oozing.

The same should go for heating up again, to begin pre-heating the inactive
extruder Y seconds before switching.

Those are just some thoughts/suggestions.  Feel free to comment.

*Adam Rumjahn*

* <http://www.innovo3d.co>*

*adam@innovo3d.co <adam@innovo3d.co>Cel: +57 304 643 1248*
*Fijo: +57 2 374 2473*

BagelOrb:
The prime tower position can now be positioned in the frontend.

PR #177, opened 2015-03-27
Add option to Reverse Inset Print Order

totalretribution:
Now that you seem to be accepting new features im trying again to get this merged.

This adds the option to reverse the print order of the insets so as to print the outer layer first and increase the precision of x,y dimensions.


BagelOrb:
I'm sorry to say this has not yet been approved. The problem is the name of the option. "Reverse inset order" doesn't describe the actual order of the perimeters. Please change the option to something like "Outer perimeter first", or something of the like.

totalretribution:
Would "printOuterInsetFirst" be ok. In CuraEngine it looks as if the term "perimeter" is not used and instead the outer walls of an object are called insets? Maybe "outerInsetFirst" would be even better?

BagelOrb:
Yes, that name is indeed better! :)

Note that the order of the outer inset w.r.t. the rest is separate from the order in which the non-outer insets will be printed. With your outerInsetFirst option, do you print the remaining insets in the normal order, that is: from the inside to the outside?

totalretribution:
The current default option is to print the insets from the inside out. This option reverses that so that all the insets are printed from the outside in which is what I believe most other slicers do. I know that KISSlicer and slic3r does it this way.

totalretribution:
In fact thinking about it if you printed the very outer inset and then moved to printing the very inner inset the outer inset would not have as good a print quality as your extruder would have to stop when moving from outer to inner. 
You all ways want the insets to print continually in a loop without stopping for best quality and strength.

daid:
Outside in printing reduces print quality, especially on overhangs. If Slic3r and KISS are doing it this way by default, then they are pretty much retarded IMHO. I did quite some testing on this before the first version of the new CuraEngine was released.

totalretribution:
Both Slic3r and KISS give you the option to print in either direction and I personally found the oposite to you in my testing and I have read of others who also found this. 

I agree that you sacrifice overhang quality but actual print quality and precision dimensions are far improved for my printer when I print outside in. I personally only choose inside out when I have overhangs otherwise I allways print outside in when using cura.

The default would still be inside out.

lerouxb:
I've [explained my findings elsewhere before](https://github.com/Ultimaker/CuraEngine/issues/98#issuecomment-87049791), but to summarise:

I find that outer perimeters first makes a huge difference when it gets to the dimensional accuracy of holes or small gaps. It would be nice to be able to choose. Sometimes you're printing a pretty colourful object with overhangs, other times you're printing something functional where you just want things to fit together or print-in-place features with narrow gaps to not fuse together.

nophead:
Yes Cura's dimensions are miles off when printing ABS because high
viscosity plastics don't pack together fully, leaving voids. If they are
done from the inside out the volume of these voids causes a cumulative
offset error. To get the right dimensions the flow would need to be reduced
by the volume of the voids, but that depends on plastic type, temperature
and layer height / width so is never going to be accurate without lots of
empirical calibration.

On the other hand the first perimeter extruded spreads symmetrically from
the nozzle, so only needs to have the correct volume to give a predictable
dimension.

BagelOrb:
Is there an accompanying PR for the introduction of the setting to the frontend as well?

I need to merge this on the engine and frontend at the same time.

src/FffGcodeWriter.cpp
```diff
@@ -811,34 +811,69 @@ void FffGcodeWriter::processInsets(GCodePlanner& gcode_layer, SliceMeshStorage*
...
-                    gcode_layer.addPolygonsByOptimizer(part.insets[inset_number], &mesh->insetX_config);
-                }
-                else
-                {
-                    Polygons& outer_wall = part.insets[inset_number];
-                    WallOverlapComputation wall_overlap_computation(outer_wall, mesh->getSettingInMicrons("wall_line_width_x"));
-                    gcode_layer.addPolygonsByOptimizer(outer_wall, &mesh->insetX_config, &wall_overlap_computation);
-                }
-            }
+			for(unsigned int inset_number=0; inset_number<part.insets.size(); inset_number++)
```
BagelOrb: Why the extra indent?

src/FffGcodeWriter.cpp
```diff
@@ -811,34 +811,69 @@ void FffGcodeWriter::processInsets(GCodePlanner& gcode_layer, SliceMeshStorage*
...
+						gcode_layer.addPolygonsByOptimizer(outer_wall, &mesh->insetX_config, &wall_overlap_computation);
+					}
+				}
+			}
+        }
+        else
+        {
+			for(int inset_number=part.insets.size()-1; inset_number>-1; inset_number--)
+			{
+				if (inset_number == 0)
```
BagelOrb: This looks like a lot of code duplication.
Could you maybe rewrite the code such that the code isn't duplicated, but reused?

src/FffGcodeWriter.cpp
totalretribution: Sorry missed that, I will ammend this.

src/FffGcodeWriter.cpp
totalretribution: I can give it a go however im not much of a programmer so any ideas on the best method of this.

totalretribution:
Will you want this added to cura in the Experimental category or somewhere else i.e Material.
Also I fixed the initial Indent issue that was caused by using tab instead of spaces.

Finally does anyone know the best way to reduce code duplication for this? I was considering moving the duplicated code in each for loop to a single function but wondered if there was a better way to solve code duplication in this particular case.

src/FffGcodeWriter.cpp
Ghostkeeper: Perhaps something like this:

    const bool compensate_overlap = inset_number == 0 ? compensate_overlap_0 : compensate_overlap_x;
    ?? inset_config = inset_number == 0 ? &mesh->inset0_config : &mesh->insetX_config;
    unsigned int line_width = inset_number == 0 ? mesh->getSettingInMicrons("wall_line_width_0") : mesh->getSettingInMicrons("wall_line_width_x");
    if (!compensate_overlap)
    {
        gcode_layer.addPolygonsByOptimizer(part.insets[inset_number], inset_config);
    }
    else
    {
        Polygons& outer_wall = part.insets[inset_number];
        WallOverlapComputation wall_overlap_computation(outer_wall, line_width);
        gcode_layer.addPolygonsByOptimizer(outer_wall, inset_config, &wall_overlap_computation);
    }

Ghostkeeper:
I've made a suggestion at the code that was duplicated.

totalretribution:
Hi Ghostkeeper, How does this reduce the two for loops i have added? I think i should of explained my self better, sorry.

The code that needs duplication removed is in function "processInsets" (Line 796) here https://github.com/totalretribution/CuraEngine/blob/71bfafbc0fdd91747ff03abfcf1a6b7dd889070b/src/FffGcodeWriter.cpp

As you can see the variable "outer_inset_first" reverses the for loop and the way I did this meant that the code inside the For loop is duplicated. I could move that code to a seperate function however I wonder if theres a better way?

Thanks
TotalRetribution

Ghostkeeper:
Ah, I only looked at the sorta-duplicated code within one of the for loops.

I guess the best way to prevent that duplication is to move it to a function. Otherwise you could loop over a separate integer that counts from `0` to `part.insets.size() * 2` and compute something like `inset_number = part.insets.size() - std::abs(i - part.insets.size())` to convert it into a counter that counts up first and back down afterwards, but I think that is not really clear either.

totalretribution:
I have attempted to move the duplicated code to a seperate function. 
Not sure if its much better than before. If this is acceptable i will create a pull request for cura as well.

src/FffGcodeWriter.cpp
BagelOrb: Or simply something like
```
if (outer_inset_first)
{
    real_inset_number = part.insets.size() - 1 - inset_number;
}
```
just below the normal `for`

BagelOrb: @Ghostkeeper your code is to optimize out the duplication which was already present in the original code. So I'm not such a good coder either :stuck_out_tongue_winking_eye: 
My comment was about this whole commit; it's one big duplication of the whole for loop you are looking at...

BagelOrb:
I think the best place for the setting in the frontend is just above Alternate Extra Wall in the Shell category.
The insets together with the top/bottom make up the shell ;)

I think @Ghostkeeper is kinda missing the point.
I've added a comment in the code of how you can do this in a really simple way, by just adding only a couple of lines to my original code.

totalretribution:
Thanks for the sugestion BagelOrb, Thats much cleaner now.

BagelOrb:
Looking good, my man!
Now only the frontend setting in fdmprinter.def.json and we're ready to merge the code!

totalretribution:
The front end Pull Request is all ready done at https://github.com/Ultimaker/Cura/issues/991
I changed it to move the setting to the Shell Category as per your request.

BagelOrb:
I'm sorry to say that we recently sent out all strings to be translated. That means that your settings cannot be part of the  final 2.3 release.

Let's hope the release after that will come soon!

Ghostkeeper:
Just got word from tha boss that the string changes for this feature should not block the merge. He finds it acceptable that there are some untranslated strings in the release as long as they are well hidden. So we can consider merging this again. @BagelOrb, the decision to merge is on you.

BagelOrb:
This feature will be in the final 2.3 release, even though it will not be translated.

Ghostkeeper:
In fact, it might get translated due to a mistake on our side.

PR #178, opened 2015-04-03
Fix retracts with dual extruder

presslab-us:
Just before moving to the wipe tower, the filament change retract is used
for the travel move instead of the normal retract.  This is
especially a problem with the wipe/prime tower.  Fix this so the
correct retract is used for the travel move, and the filament
retract difference is set after that.

### DIFF
--- a/src/fffProcessor.h
+++ b/src/fffProcessor.h
@@ -764,7 +764,8 @@ private:
         if (config.wipeTowerSize < 1)
             return;
         //If we changed extruder, print the wipe/prime tower for this nozzle;
-        gcodeLayer.setAlwaysRetract(true);
+        if (config.enableCombing)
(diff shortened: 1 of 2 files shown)


BagelOrb:
Your code looks dirty. A variable named 'zero' which is set to zero, but then later 'zero' becomes something different from zero. Please make your variable names descriptive.

I do not understand the problem. Just before changing filaments, the filament change retract is exactly the retract that is defined to be used right then!

presslab-us:
Here is the description for the G92 command: http://reprap.org/wiki/G-code#G92:_Set_Position
Would you prefer the variable is called absoluteZero? newZero? offsetZero?

BagelOrb:
What about extrusionZeroPoint?

PR #184, opened 2015-04-16
adding clipper_convex_hull.{h,cpp}

gsmith-to:
See comments in clipper_convex_hull.h for usage.


BagelOrb:
We already have a convex hull algorithm atm. This PR is outdated.

PR #189, opened 2015-04-23
fixes in PolygonRef.inside, and in PolygonRef.centerOfMass calculations

gsmith-to:
centerOfMass generates incorrect result when correct result has x < 0. This fixes the issue. No 'bug' at application level that I know of, but centerOfMass is used only in bridgeAngle(), problems may pop up in rare cases (such as 'center1' and 'center2', calculated in bridgeAngle, being on opposite sides of x=0, so that only one of them gets flipped).

### DIFF
--- a/src/utils/polygon.h
+++ b/src/utils/polygon.h
@@ -129,9 +129,18 @@ public:
         return ret;
     }
     
-    //Check if we are inside the polygon. We do this by tracing from the point towards the negative X direction,
+    //Check if we are inside the polygon. We do this by tracing from the point towards the positive X direction,
     //  every line we cross increments the crossings counter. If we have an even number of crossings then we are not inside the polygon.
-    bool inside(Point p)
+    // Care needs to be taken, if p.Y exactly matches a vertex to the right of p, then we need to count 1 intersect if the
+    // outline passes vertically past; and 0 (or 2) intersections if that point on the outline is a 'top' or 'bottom' vertex.
+    // The easiest way to do this is to break out two cases for increasing and decreasing Y ( from p0 to p1 ).
+    // A segment is tested if pa.Y <= p.Y < pb.Y, where pa and pb are the points (from p0,p1) with smallest & largest Y.
+    // When both have the same Y, no intersections are counted but there is a special test to see if the point falls
+    // exactly on the line.
+    //
+    // Returns false if outside, true if inside; if the point lies exactly on the border, will return 'border_result'.
+    //
+    bool inside(Point p, bool border_result=false)
     {
         if (polygon->size() < 1)
             return false;
@@ -141,12 +150,45 @@ public:
         for(unsigned int n=0; n<polygon->size(); n++)
         {
             Point p1 = (*polygon)[n];
-            
-            if ((p0.Y >= p.Y && p1.Y < p.Y) || (p1.Y > p.Y && p0.Y <= p.Y))
+            // no tests unless the segment p0-p1 is at least partly at, or to right of, p.X
+            if ( std::max(p0.X, p1.X) >= p.X )
             {
-                int64_t x = p0.X + (p1.X - p0.X) * (p.Y - p0.Y) / (p1.Y - p0.Y);
-                if (x >= p.X)
-                    crossings ++;
+                int64_t pdY = p1.Y-p0.Y;
+                if (pdY < 0)      // p0->p1 is 'falling'
+                {
+                    if ( p1.Y <= p.Y && p0.Y > p.Y )    // candidate
+                    {
+                        // dx > 0 if intersection is to right of p.X
+                        int64_t dx = (p1.X - p0.X) * (p1.Y - p.Y) - (p1.X-p.X)*pdY;
+                        if (dx == 0)                    // includes p == p1
+                            return border_result;
+                        if (dx > 0)
+                            crossings ++;
+                    }
+                }
+                else if (p.Y >= p0.Y)
+                {
+                    if (p.Y < p1.Y)    // candidate for p0->p1 'rising' and includes p.Y
+                    {
+                        // dx > 0 if intersection is to right of p.X
+                        int64_t dx = (p1.X - p0.X) * (p.Y - p0.Y) - (p.X-p0.X)*pdY;
+                        if (dx == 0)                // includes p == p0
+                            return border_result;
+                        if (dx > 0)
+                            crossings ++;
+                    }
+                    else if (p.Y == p1.Y)
+                    {
+                        // some special cases here, points on border:
+                        //  - p1 exactly matches p (might otherwise be missed)
+                        //  - p0->p1 exactly horizontal, and includes p.
+                        //    (we already tested std::max(p0.X,p1.X) >= p.X )
+                        if (p.X == p1.X ||
+                              pdY==0 && std::min(p0.X,p1.X) <= p.X )
+                            return border_result;
+                        // otherwise, count no crossings
+                    }
+                }
             }
             p0 = p1;
         }
@@ -175,7 +217,7 @@ public:
 
     friend class Polygons;
     friend class Polygon;
-};
+}; // class PolygonRef
 
 class Polygon : public PolygonRef
 {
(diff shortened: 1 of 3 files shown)


gsmith-to:
I've also found & fixed a more significant issue in PolygonRef.inside() - there are cases where it will make errors.  There are three issues:
  (1) points exactly on the boundary of the polygon may or may not be identified as 'inside'. Probably not a concern.
  (2) points very close to the edge ( <1 unit) may produce the wrong result due to truncation in the division used to calculate the intersection. Again, may not be a concern.
  (3) Some points can give the wrong result even though they may be far from the polygon. This occurs when the supplied point has the same Y coordinate , and lower X coordinate, as a polygon point; and the polygon outline changes Y direction at that point (from moving up to moving down, or vice versa).  In the previous code, one intersection would be counted in this case (one of the two edges and not the other). leading to an incorrect result.

The new implementation fixes all of these and I've added a small test suite (which checks this function, and also centerOfMass and area functions).

Points exactly on the polygon now always return false from inside(p) - I've added an optional parameter, so e.g. poly.inside(pt,true) will give true if pt is exactly on the polygon  -- and otherwise is the same result as poly.inside(pt)

BagelOrb:
It's a shame you didn't write this code for the new CuraEngine, which resides in the PinkUnicorn branch. 

Some further considerations:
- why are you calling `double RefArea` a reference?
- please start a variable with a lower case letter. variables should be camelCased
- I don't think we should include your testing code at such a global level as the Makefile.

I've corrected the implementation in the PinkUnicorn branch to my own insight. I did it in a more concise manner, but didn't include the nice border_result functionality, since I feel we don't need it.
See https://github.com/Ultimaker/CuraEngine/commit/a7335a5caa6f7ab991328dd0bf766aabd24ac760

BagelOrb:
@daid: why did you have that invert code at the end of the computation of centerOfMass??

daid:
Most likely because I was getting negative points.

Now that I look at the code, the reason for that is quite possibly because the Area function gives a negative number for holes, so the result of area should be taken abs()

daid:
Note, if there are bugs in the inside() function, then those bugs are most likely also in the line infill code, as they are build the same way.

BagelOrb:
I've already overhauled the line infill code some time ago. Now the code is not so similar anymore.

Just did a fix for centerOfMass like you said...

gsmith-to:
I'm new to git. I determined that PinkUnicorn still has the same polygon.h
as the master branch, and presumed that a PR could be pulled to either,
apologies if that's not true.

 - RefArea is the "reference" area, calculated in the test suite, used to
compare the value calculated by area(); not a C++ reference. I apologize,
I'm not used to Java-like conventions for C++; use of uppercase for global
variables (and lowercase for local) is an older style and still very widely
used. Since that is a test suite and not part of the code proper, I was not
as careful about following the project's style.

Please suggest where test should be, I will move it.

Did you test  your modification with this suite?   The test suite doesn't
contain any points for inside() which are exactly on the polygon, so it
doesn't matter if border_result is implemented, it should still pass the
tests.

As I see, it doesn't actually fix the problems. It will still incorrectly
classify points which are very close (due to the division).

Also, consider case where polygon is    (2000,2000),  (1000,1000),  ( 1100,
100)

For point (-2000, 2000), your new implementation, from what I can see, now
gives the correct result by counting 2 intersections.
But if you test  point ( -2000, 1000), the new implementation will count 3
intersections, and give an incorrect result (and for any (x,1000) where x <
1000).

I would not offer a fix to such an important routine without both carefully
analyzing the problem and testing it. I spent several hours on the weekend
doing this and the result I've submitted is a correct implementation, and
the test suite is quite thorough and includes many specifically chosen
special cases which could cause problems.

I understand the importance of being concise, since the functions are in
the class declaration and nobody wants to scroll through a bunch of long
functions when reading a class declaration. I would suggest that it's more
important to take the number of lines needed to produce a a correct and
efficient implementation, and then perhaps, if it's too long, the
implementation could be placed after the class declaration as an  inline?

Also, I see that someone has changed my fix to centerOfMass, in another PR,
by adding (in effect) area= abs(area). This will lead to an incorrect
result when the polygon area is negative (again, I considered that, and the
test suite includes cases where the polygon has negative area). When the
polygon has a negative area, the centerOfMass routine *must* divide by the
negative area or the result will be incorrect.

Greg

BagelOrb:
First of all: I'm ashamed to say I didn't take a good enough look and rushed the coding. I should never have committed that code. I have been naive for thinking that would do.

The PinkUnicorn branch is quite different from the master already; I don't know whether your PR can be pulled to it. I will figure it out.

I find it hard to read your test suite, and I don't know whether we would want a test suite like that. I think we should think how we would like a test suite to look like by also considering the rest of the engine. Me and daid will have a look at it.
For now I think it's best if you leave out the test suite from the PR.

I believe you have tested your `inside(.)` function thoroughly. Can you perhaps make a PR with only that function? 

The centerOfMass issue is a separate issue. I don't see why the center of mass should depend on whether it is a hole or not. A negative area means the winding order is reversed, but why would the winding order influence the center of mass of the polygon?

BagelOrb:
Hi Greg,

I just hacked your code into the PinkUnicorn branch: https://github.com/Ultimaker/CuraEngine/commit/600eed61fc1c84dddd7c471474bdc074b76ec4c8

What do you think about the centerOfMass issue? (see my comments above)

BagelOrb:
For centerOfMass: https://github.com/Ultimaker/CuraEngine/commit/d073dda5c41830c73b88daf28a705bb1f74f086a

BagelOrb:
Hopefully we can make the PinkUnicorn branch the master branch soon, so that we don't have such issues anymore that people create PRs for the old CuraEngine!

gsmith-to:
Sure, of course I will work on PinkUnicorn from now on. And I see there is a style guide there.

For sure, my test suite was driven by getting the job done in the time I had available, and it isn't very well set up to expand to other tests or anything like that. And it doesn't provide very much information when tests fail...

Regarding centerOfMass, the winding order doesn't influence the center of mass, but it changes the way the sum is calculated inside centerOfMass. Suppose you find the centerOfMass of a polygon having a positive area; then you reverse the order of vertices, and call the function again. This time the same pairs of vertices (p0,p1) are visited (in a different order) but this time the pairs are reversed (p0,p1) -> (p1,p0). As a result, the 'second_factor' values have opposite signs, whereas the (p0.X + p1.X) and (p0.Y+p1.Y) sums are the same as before; and so the final sums of  x and y will have opposite signs. So, by dividing this time by the negative area, the same result is obtained.

The method works by considering each triangle ((0,0), p0, p1) in sequence; the area of the triangle is 'second_factor/2' and its center of mass is ( (p0.X+p1.X)/3,  (p0.Y + p1.Y)/3), and then it finds the sums of these centers, weighted by the areas (factoring the /2 and /3 to the end) and divides by the total weight (total area). Any (or all) of the triangle areas may be positive, zero, or negative (in fact, you could accumulate 'second_factor' in the loop to get the area). The implementation of 'area' in the clipper lib is a bit different - it accumulates area of quadrilaterals formed between p0,p1 and the x-axis, which is mathematically equivalent, and numerically, somewhat less susceptible to pathological rounding problems. But the conditions in cura - with all the vertex values being reasonable integers, and summations done with doubles - essentially eliminates such rounding concerns in quadratic calculations like this one. I.e. we shouldn't see any rounding at all when finding sums of products of [sums and differences of coordinates]. If you made a full-size model of the Empire State building, to 1 micron resolution, it would need 25 bit coordinates, that might cause a bit of rounding, but smaller things, no :-)

BagelOrb:
Thanks for the explanation!

Note that I had already corrected centerOfMass in the same way you had corrected it.

Thanks for finding these bugs and solving them!

:D

gsmith-to:
OK, I see that PinkUnicorn has been merged; if I'm working on convex hull, should I work on master or on PinkUnicorn?
I've updated my master to match Ultimaker, but I'll create a new branch for any new work in any case.
If I delete this branch gsmith-to:patch-2, will it make this PR page go away? since I'd rather not do that just now, because I need to refer to the discussion above.

BagelOrb:
On master.

New Cura is comming....... soon...... :D

PR #213, opened 2015-06-01
use deault path for data file

yhfudev:
to avoid the error when run the command line without arguments:

ERROR: Failed to load json file: fdmprinter.json

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,0 +1,158 @@
+
+DEFS+=-std=c++11
+
+ACLOCAL_AMFLAGS=-I m4
+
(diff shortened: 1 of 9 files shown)


BagelOrb:
Sorry for the late reply. This PR slipped through the cracks.

It seems you have included too many changes in your PR; for example you have included a makefile.am and changed includes in some header files.

Can you please isolate your fix?

Then again, I don't see the real problem. The fdmprinter.json file should be found at Cuta/resources/settings and not anywhere within the CuraEngine directory.

PR #221, opened 2015-07-19
Crash during slicing

electrocbd:
Polygons::splitIntoPartsView_processPolyTreeNode use a pointer to the vector partsView which may be reallocated during the recursive call to splitIntoPartsView_processPolyTreeNode, invalidating iterators and pointers


BagelOrb:
You are totally right! Thank you for finding this bug!

PR #222, opened 2015-07-20
Infinite loop in cura::FffPolygonGenerator::processDraftShield

electrocbd:
When specifying Platform Adhesion = Brim and Draft Shield Height = 0.8 (which is the height of the first layer), the system enters an infinite loop.

If I understood the source code, `layer_skip` is an optimization to speed up process, but there is no increment after the "skip" in the following loop. This is why I added one to the value of `layer_skip`.

I also changed an `int` into `unsigned int`, to suppress a compiler warning.

### DIFF
--- a/src/FffPolygonGenerator.cpp
+++ b/src/FffPolygonGenerator.cpp
@@ -341,7 +341,7 @@ void FffPolygonGenerator::processDraftShield(SliceDataStorage& storage, unsigned
     
     unsigned int max_screen_layer = (draft_shield_height - layer_height_0) / layer_height + 1;
     
-    int layer_skip = 200 / layer_height + 1;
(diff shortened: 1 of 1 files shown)


BagelOrb:
How are you finding these bugs?

electrocbd:
@BagelOrb I am planning to work - again - on a stretch algorithm :innocent:, this is why I updated my CuraEngine version. When trying to activate brim, I encountered this issue.

By the way, why do you test `draft_shield_height` when printing the Brim? (method `FffGcodeWriter::processSkirt`). I had to set a small positive value (0.01) to have my brim - and only my brim - printed.

BagelOrb:
Because I thought that that code only disabled the skirt, not the brim as well. I've fixed this bug now.

PR #244, opened 2015-09-21
always retract when combing is disabled

fxtentacle:
Due to my new travel path display (https://github.com/Ultimaker/Cura/pull/410), I noticed that when I disable combing, that CuraEngine will refuse to do any retraction at all. So that means that "combing=no" and "retraction=yes" would yield exactly the same result as "retraction=no". Attached is a small patch so that the GCodePlanner will always advise a rectraction if combing is disabled. If both retraction and combing are disabled, the GCodeExporter will then ignore the advised retractions and issue moves.

Old behaviour:
retraction=yes combing=no 
=> no retractions at all

New behaviour:
retraction=yes combing=no
=> will retract for every move

### DIFF
--- a/src/gcodePlanner.cpp
+++ b/src/gcodePlanner.cpp
@@ -165,6 +165,12 @@ void GCodePlanner::addTravel(Point p)
         was_combing = is_going_to_comb;
     }
     
+    if(comb == nullptr) {
+        // no combing? always retract!
+        path = getLatestPathWithConfig(&travelConfig);
+        path->retract = true;
+    }
+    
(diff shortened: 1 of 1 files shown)


fxtentacle:
I think this one should maybe go into the 15.10 beta, because otherwise there would be no way for a user to force retracting for every move.

nallath:
Agreed. Could you make a pull request that merges this into 15.10?

BagelOrb:
Note: this has been merged into 15.10 as well: https://github.com/Ultimaker/CuraEngine/pull/247

PR #245, opened 2015-09-21
Keep Z-Hop active for multi-step moves

fxtentacle:
I noticed that even with an extremely high z-hop like 1mm, my nozzle would sometimes:
1. retract
2. move up the 1mm
3. then move diagonally down and towards the retraction target point
4. if the retraction target point is close to the wall, the nozzle would already be very low there, as if no z-hop was enabled, and basically scrub over the wall

I tracked this down to the fact that if a non-extrusion move with Z change is issued during the retraction, then that Z value is used directly, which causes the diagonal down move I described in (3). So the fix is super easy: store the current Z-hop height and apply it to every non-extrusion move. 

That way, the nozzle will move over the retraction target point and then move down on top of the target point. This prevents the nozzle from scratching the wall.

Old GCODE example (with 2mm zhop to show the effect):
G10
G1 Z6.200
G0 X112.242 Y102.096 Z6.100
G1 Z6.100
G11
(as you can see, the G0 moves down diagonally)

New GCODE:
G10
G1 Z6.200
G0 X112.242 Y102.096 Z6.300
G1 Z6.100
G11
(here, G0 moves up diagonally to ensure that the zhop distance is kept from the surface)

### DIFF
--- a/src/gcodeExport.cpp
+++ b/src/gcodeExport.cpp
@@ -19,7 +19,7 @@ GCodeExport::GCodeExport()
     currentSpeed = 1;
     retractionPrimeSpeed = 1;
     isRetracted = false;
-    isZHopped = false;
+    isZHopped = 0;
     last_coasted_amount_mm3 = 0;
     setFlavor(EGCodeFlavor::REPRAP);
 }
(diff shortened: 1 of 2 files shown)


nallath:
Excellent work!

PR #246, opened 2015-09-21
fix retraction for isles < 0.8mm radius

fxtentacle:
Please take a look at this screenshot:
screen shot 2015-09-21 at 11 50 31

The retraction minimum travel and minimum extrusion distance are both set to 0. Combing is enabled. So with those settings, I would expect CuraEngine to retract for any move that cannot be combed. But if you look at the area that I marked with the purple box, you see that there is a thin dark blue line between the red column and the main tower. I added the blue lines with https://github.com/Ultimaker/Cura/pull/410 and https://github.com/Ultimaker/CuraEngine/pull/242 so the thin blue line there shows is that the GCodeExport class from CuraEngine did NOT emit a retraction.

So to summarize: I setup the settings so that CuraEngine should retract for ANY jump that cannot be combed.. but it clearly failed to do so.

Here's a screenshot of the same slice after applying this patch:
screen shot 2015-09-21 at 14 16 11
As you can see, CuraEngine is now correctly doing retracted jumps (thick light blue lines) for everything that moves to or from the small side columns. 

I tracked the erroneous behavior in the unpatched CuraEngine down to two issues:

1. There was a hardcoded Comb::max_comb_distance_ignored = 1.5mm threshold in the Comb class. This alone prevents CuraEngine from handling any retraction shorter than 1.5mm correctly, which kind of defeats the whole purpose of the retraction_min_travel_distance user-configurable setting. So I replaced the hardcoded constant with a variable, which is now initialized to retraction_min_travel_distance. After this change, CuraEngine now correctly analyzes retractions, if the min travel distance is set below 1.5mm by the user. But that alone wasn't sufficient to get a correct slicing for my test object.

2. Comb::getLayerSecondWalls was implemented to either use the 2nd perimeter line OR to move offset_from_outlines from the outline of the object. offset_from_outlines is hardcoded to be 2*nozzle_size, which in my case would be 0.8mm. So the getLayerSecondWalls function would try to move the outline inside by 0.8mm for every isle that doesn't have at least 2 perimeter lines. In my object, the column radius is about 0.5mm, which means that getLayerSecondWalls reduced it to.. NOTHING. That's why CuraEngine wasn't emitting correct retraction jumps for my columns, even after fix (1).. because the getLayerSecondWalls was removing my small isles. So I reworked getLayerSecondWalls to:
2a. first try to use the 2nd perimeter line, like before
2b. then try to use the 1st perimeter line, if the object is too thin to have 2 perimeter lines
2c. third try to move the object outline inside by 0.8mm, like before. But only use this result, if there's any object left.
2.d if the isle is so small that we didn't find any point inside to jump to.. just use the outline directly

For any user that retains the default value of 1.5mm for retraction min travel, this patch should not change anything, because the constant that I replaced also had a hardcoded value of 1.5mm. But for retraction min travel values that are below 1.5mm, this patch will prevent CuraEngine from "overlooking" air gaps. And as you can see in the screenshots, for my test object, that makes the difference between randomly "combing" through thin air and retracting as it should.

### DIFF
--- a/src/comb.cpp
+++ b/src/comb.cpp
@@ -20,14 +20,26 @@ Polygons Comb::getLayerSecondWalls()
     {
         for (SliceLayerPart& part : mesh.layers[layer_nr].parts)
         {
-            if (part.insets.size() >= 2)
-            {
+            // we want the 2nd inner walls
+            if (part.insets.size() >= 2) {
                 layer_walls.add(part.insets[1]);
+                continue;
             }
-            else 
-            {
-                layer_walls.add(part.outline.offset(-offset_from_outlines));
+            // but we'll also take the inner wall if the 2nd doesn't exist
+            if (part.insets.size() >= 1) {
+                layer_walls.add(part.insets[0]);
+                continue;
             }
+            // and if there is no walls, we'll try to move inside from the outline
+            Polygons newOutline = part.outline.offset(-offset_from_outlines);
+            if(newOutline.polygonLength() > 0) {
+                layer_walls.add(newOutline);
+                continue;
+            }
+            // offset_from_outlines was so large that it completely destroyed our isle,
+            // so we'll just use the regular outline
+            layer_walls.add(part.outline);
+            continue;
         }
     }
     return layer_walls;
--- a/src/comb.cpp
+++ b/src/comb.cpp
@@ -64,7 +64,7 @@ Comb::~Comb()
         delete boundary_outside;
 }
 
-bool Comb::calc(Point startPoint, Point endPoint, CombPaths& combPaths, bool startInside, bool endInside)
+bool Comb::calc(Point startPoint, Point endPoint, CombPaths& combPaths, bool startInside, bool endInside, int64_t max_comb_distance_ignored)
 {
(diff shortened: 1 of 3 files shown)


fxtentacle:
Oops I just noticed that retraction min travel seems to default to 4.5mm on the unpatched Cura GUI.. so then this patch would change stuff by increasing the retraction threshold from a hardcoded 1.5mm to the user-configurable 4.5mm. So I think maybe the best way is to merge this and to simultaneously lower the default value from 4.5mm to 1.5mm in the GUI.

nallath:
Nice work. We just started a IRC chat for cura development on freenode (#cura-dev). If you have any questions, feel free to ask them.

nallath:
Just a FYI; We are in feature freeze right now. As a result most of your changes will not be in the 15.10 release, but in the 16.01. If you have any specific bug fixes for the current beta, push them to the 15.10 branch.

BagelOrb:
Indeed, nice work!

I hadn't really put a lot of thought into damage reduction in case the combing algorithm fails.

It's nice to see someone building on my code. :)

Especially the combing code is quite horrific to understand, as I recall.

PR #247, opened 2015-09-22
always retract when combing is disabled

fxtentacle:


nallath:
And as always, thanks for the work :) It's much appreciated.

BagelOrb:
Note: this has been merged into master as well: https://github.com/Ultimaker/CuraEngine/pull/244

BagelOrb:
Currently this PR causes there to be retracts on moves as small as one micron.

You forgot the part which said
```if (!shorterThen(lastPosition - p, last_retraction_config->retraction_min_travel_distance))```

The code you introduced was already there about 10 lines above in the else-case for when combing failed. The only problem was that that piece of code was never executed when combing wasn't enabled, so now I just moved the code down and introduced a boolean variable which kept track of whether combing had occurred.

BagelOrb:
BTW: I now fixed both our bugs ;)

PR #249, opened 2015-09-25
Re-nable layer count using a consistent tag name

oliv3r:
Patch a9db2813 disabled the Layer count comment as it was 'just a comment'. Re-enable it with a proper cased tag name so it can be used for very rough print time estimation purely based on layers.

Signed-off-by: Olliver Schinagl <o.schinagl@ultimaker.com>

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -36,7 +36,7 @@ void FffGcodeWriter::writeGCode(SliceDataStorage& storage, TimeKeeper& time_keep
     meshgroup_number++;
 
     unsigned int total_layers = storage.meshes[0].layers.size();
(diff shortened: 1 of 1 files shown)


BagelOrb:
As I see it, the removal of the later count is a bug. For compatibility
with firmware which usually this count, new gcode will not work properly.

Did you change it such that the layer count comment is the same as it used
to be?

Moreover, after looking into the diff I see that the reason I first removed
the layer count comment is because the value used there is not the actual
amount of layers printed. The raft introduces complexity.

It's not hard to have a gcode comment with the time estimate at every
layer...

How do you now think the gcode comments should look like?
Op 25 sep. 2015 12:07 schreef "Olliver Schinagl" <notifications@github.com>:

PR #255, opened 2015-10-06
Explicitly write a \n when finding a new line in a comment

nicanor-romero:
This way the start_gcode and end_gcode that includes '\n's will not
have their commands executed since they now appear as commented out, within the same line.

Solves https://github.com/Ultimaker/Cura/issues/459 .


BagelOrb:
Thanx for the fix!

I've removed the profile string at the end of the gcode altogether, though.

There's a bug in the firmware of a lot of printers where a line longer than 80 characters will result in executing the rest as if it was a new line. A very long comment line may result in gcode commands.

Issue #240, opened 2015-09-06
Unable to choose heat order or wait for temperature in start gcod


markwal:
For reprap flavor gcode:

I believe (see issue #67) that there used to be a way to have tokens in start gcode plus if the start gcode contained the set temperatures then curaengine wouldn't set the temperatures itself.  Now, however, a machine profile with "machine_start_gcode" doesn't appear to get token replacement and cura engine insists on emitting first wait for bed then wait for extruder temperature if "material_bed_temperature" is a setting that is shown.

So, if you'd like to be able to specify the temperature in the UI and have the heat up be simultaneous and after some portion of the gcode, I don't believe there is an easy way to do that.

BagelOrb:
I find it highly unlikely the start_gcode was ever parsed. @daid ?

This feature request seems low priority. If you work with a custom start_gcode you could also customize it such that you change the temperature, right?

markwal:
Yes, but I'd like to choose the temperature setting in the UI and not choose a different printer profile for every temperature I might want to try.  I could rig up a PR that does these two things:

1. simple token replacement on the start/end gcode (just straight pattern substitution, no parsing)
2. if any of the following exist (simple text match ignoring whether they are in comments or not) in the start gcode then the built-in temperatures are not emitted at the beginning: M104, M140, M109, M190

markwal:
It wasn't in the engine before it was in Cura, but so was the M190 emit:
https://github.com/daid/Cura/blob/9b5aa5c182a6c2285f1f5f362ac0eeb31d3ff222/Cura/util/profile.py#L1320

BagelOrb:
I think this should be done in the new frontend as well.

It would then be possible to do token replacements in each gcode setting.

Where the string contains "{...}" the text should be replaced by the setting which named within the braces.

markwal:
Ok, but there is still the problem that the engine now emits the M190 & M109. Should that move to the front end as well?
Would a pull request be considered and how is a change done across both the back end and front end?

BagelOrb:
A pull request would certainly be welcome!

You are right to say that the first heat up gcode commands generated by the engine should be suppressed (in the engine) if the start gcode already contains the heat up commands. However, what would you do if the temperature in the start gcode doesn't coincide with the generated temperature command? I would say the generated temperature command should 'win', since it's more likely that that temperature has been tweaked by the user.

The frontend would still have to replace the `{temperature}` by the actual temperature.

[...]

PR #266, opened 2015-10-18
Omit start temp if already in starting gcode

markwal:
so that it can participate in token expansion. See #240 

This change has a dependent corresponding change to Cura which is pull request Ultimaker/CuraEngine#490

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -200,15 +200,21 @@ void FffGcodeWriter::processStartingCode(SliceDataStorage& storage)
     }
     if (gcode.getFlavor() != EGCodeFlavor::ULTIGCODE)
     {
-        if (getSettingBoolean("machine_heated_bed") && getSettingInDegreeCelsius("material_bed_temperature") > 0)
-            gcode.writeBedTemperatureCommand(getSettingInDegreeCelsius("material_bed_temperature"), true);
-        
-        for(SliceMeshStorage& mesh : storage.meshes)
-            if (mesh.getSettingInDegreeCelsius("material_print_temperature") > 0)
-                gcode.writeTemperatureCommand(mesh.getSettingAsIndex("extruder_nr"), mesh.getSettingInDegreeCelsius("material_print_temperature"));
-        for(SliceMeshStorage& mesh : storage.meshes)
-            if (mesh.getSettingInDegreeCelsius("material_print_temperature") > 0)
-                gcode.writeTemperatureCommand(mesh.getSettingAsIndex("extruder_nr"), mesh.getSettingInDegreeCelsius("material_print_temperature"), true);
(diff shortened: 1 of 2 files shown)


BagelOrb:
You just deleted all the logic around temperatures!
If you just include the temperature commands in the start gcode setting, then dual extrusion printers will end up with both nozzles heated all the time, unless you change the start gcode setting per print job.

The way it is now only nozzles which are used are heated.

Did you include such functionality in the frontend?

markwal:
I tried to.  It loops across the objects and only generates commands for referenced extruders.  But I'm not sure I followed how the per object settings get set properly.  https://github.com/Ultimaker/Cura/pull/490/files#diff-81af33b9376386b0e7e43b8470fb209aR357

markwal:
Updated Ultimaker/Cura#490 to move logic to slice job created by @awhiemstra and rebased to master

markwal:
@BagelOrb Ultimaker/Cura#490 got merged.  And now this change is the simpler one that you proposed there.

BagelOrb:
Feedback:
- I had to fix some style issues concerning braces (https://github.com/Ultimaker/CuraEngine/commit/9c8e32f37bcfb055597b0554cf82abcf1b5db1fa)
- I don't really see why you introduced material_print_temp_wait. Could it be that strictly viewed this is a separate issue?
- material_print_temp_wait and material_bed_temp_wait were not in the json. ***Every*** setting passed to the backend should be in the json.

BagelOrb:
correction:
- material_print_temp_prepend and material_bed_temp_prepend were not in the json

markwal:
I added the wait ones because you told me to in your first comment on Ultimaker/Cura#490.  You may have offered that then because the original proposed change introduced new placeholders separately for temp set and waits.

I didn't add the prepends because I was always computing them, based on whether temperature expansions occur in the start/end gcode and didn't realize that every setting should be in the json.

BagelOrb:
Every setting should be in the json, so that CuraEngine knows they exist. When calling CuraEngine directly from the commandline, instead of via Cura frontend, CuraEngine uses the json file to get the default values of each setting.

As I understand your frontend code material_print_temp_prepend is set to false when the start gcode contains "{material_print_temperature}", whether it would be in an M104 or in an M109.
I would like it if CuraEngine would still prepend a wait-till-temp-is-reached command even though the start code already contained a set-temp-and-dont-wait command.
I think that's why I wanted the material_print_temp_prepend setting.

markwal:
Ok, I'll edit the json to default prepend to true and do another PR. I see, definitely needed for the command line case.

BagelOrb:
I included the settings in the json already, but thanks for the gesture!

BagelOrb:
see https://github.com/Ultimaker/Cura/commit/94ee3882ad738d2de4c25b9c4cf7c8deb8f3d473

PR #270, opened 2015-10-27
Accepts 'help' as a known command

nicanor-romero:
'help' was being considered an unknown command and it also returned 1 as a return code when running './CuraEngine help'. If you agree, I believe that it should recognize it a known command and return 0 instead.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -317,6 +317,11 @@ int main(int argc, char **argv)
     {
         slice(argc, argv);
     }
+    else if (stringcasecompare(argv[1], "help") == 0)
+    {
+        print_usage();
+        exit(0);
+    }
     else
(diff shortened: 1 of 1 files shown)


nallath:
@Ghostkeeper @BagelOrb What are your ideas on this? I'd even go as far as saying that help should you know, give actual help on how to use the engine.

BagelOrb:
@nallath It seems you haven't used CuraEngine as a command line tool in a while, since it already does that. The only problem was that `CuraEngine help` resulted in a 1 as exit code.

nicanor-romero:
Thanks guys.

BagelOrb:
No no; thank you, sir!
Thanks for contributing to the project!

PR #281, opened 2015-12-01
Adding ENABLE_ARCUS option in CMakeLists

Gregwar:
This adds the `ENABLE_ARCUS` option in the CMakeLists that allow you to compile cura without the Arcus and protobuf dependency, and allowing you an easier use as a standalone library

I know this is a little bit complicated since it adds much defines check and maybe not easy to maintain

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,7 @@ option (ENABLE_ARCUS
     "Enable support for ARCUS" ON)
 
 if (ENABLE_ARCUS)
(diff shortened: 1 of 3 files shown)


Ghostkeeper:
I've tested this and it works good. Thanks! I'm going to merge this.

But first, it'd be good if you could add a message to the if-statement like `message(STATUS "Building with Arcus.")`. This would make it easier to debug the build.

Gregwar:
OK, It's done

PR #297, opened 2016-01-29
Changes needed for CuraEngine to work with the new Arcus API

awhiemstra:


src/commandSocket.cpp
```diff
@@ -93,20 +93,29 @@ void CommandSocket::connect(const std::string& ip, int port)
...
+    private_data->socket->registerMessageType(&cura::proto::SlicedObjectList::default_instance());
+    private_data->socket->registerMessageType(&cura::proto::Progress::default_instance());
+    private_data->socket->registerMessageType(&cura::proto::GCodeLayer::default_instance());
+    private_data->socket->registerMessageType(&cura::proto::ObjectPrintTime::default_instance());
+    private_data->socket->registerMessageType(&cura::proto::SettingList::default_instance());
+    private_data->socket->registerMessageType(&cura::proto::GCodePrefix::default_instance());
 
     private_data->socket->connect(ip, port);
+
+    log("Connecting to %s:%i", ip.c_str(), port);
```
BagelOrb: I suppose this message is superfluous given the message below.

src/commandSocket.cpp
```diff
@@ -158,9 +167,9 @@ void CommandSocket::connect(const std::string& ip, int port)
             //sendPrintTime();
         }
 
-        if(!private_data->socket->errorString().empty()) 
+        if(private_data->socket->getLastError().isValid())
         {
-            logError("%s\n", private_data->socket->errorString().data());
+            logError("%s\n", private_data->socket->getLastError().toString());
```
BagelOrb: Perhaps change `"%s\n"` to `"Socket Error: %s\n"`

BagelOrb:
Otherwise changes look ok. This can be merged as soon as the corresponding PR for Cura is merged.

src/commandSocket.cpp
awhiemstra: The Arcus error is converted to string in the following format:

`Arcus <Fatal?> Error (<ErrorCode>): <Error Description>`

If that is not sufficient I can add the socket error bit.

BagelOrb: Nah; it's okay like this.

src/commandSocket.cpp
awhiemstra: Having the Connecting and Connected messages separately helps with debugging, so that you can properly distinguish whether it just fails to connect or something else is going on.

BagelOrb: Ah I didn't notice the small difference between "connectING" and "connectED"....

PR #298, opened 2016-01-31
Registering "SlicingFinished" message

### LINKED ISSUE #622: Debian: Using rapidjson
Just found this, while working on #619:

https://anonscm.debian.org/cgit/3dprinter/packages/cura-engine.git/tree/debian/patches/03-system-rapidjson.patch

thopiekar:
"The reason is that CuraEngine doesn't have the SlicingFinished message registered and so it sends it with the id 0 instead of 8. Then libArcus throws a UnknownMessageError exception because CuraEngineBackend didn't register the message SlicingFinished with id 0 but with 8."
- https://github.com/Ultimaker/Cura/issues/622#issuecomment-177592468
* Fixes https://github.com/Ultimaker/Cura/issues/622


thopiekar:
Tested this fix by @ricardopvz and it works well.

Ghostkeeper:
This was already fixed in https://github.com/Ultimaker/CuraEngine/commit/0e662d7d675eb9074424c1a8a281a6119113d1f3, which I forgot to push. Sorry!

PR #299, opened 2016-02-01
Fix Cura not receiving print time

ricardopvz:
SlicingFinished must be sent after ObjectPrintTime because Cura restarts the Engine as soon as it receives the SlicingFinished message and never receives ObjectPrintTime message.


Ghostkeeper:
Was already fixed last Friday in https://github.com/Ultimaker/CuraEngine/commit/f763edfb05ae6a95b9679c92b84846842c580e16, which I forgot to push. Sorry!

PR #304, opened 2016-02-11
fix build warnings

soyersoyer:


Ghostkeeper:
Seems to work well. Thanks!

BagelOrb:
The settingsRegistry had a parent class member in order to generate more informative error messages in the future.

We'll see whether we need a parent member there when we fix the issue of better warning messages...

PR #309, opened 2016-02-24
make some methods const

codehero:
I am trying to understand the source code better and the lack of const tags on function parameters makes it confusing. Therefore I have started tagging these parameters const myself.

### DIFF
--- a/src/settings.cpp
+++ b/src/settings.cpp
@@ -72,27 +72,28 @@ void SettingsBase::setSetting(std::string key, std::string value)
     }
 }
 
-std::string SettingsBase::getSettingString(std::string key)
+std::string SettingsBase::getSettingString(std::string key) const
 {
     if (setting_values.find(key) != setting_values.end())
     {
-        return setting_values[key];
+        return setting_values.at(key);
     }
     if (parent)
     {
         return parent->getSettingString(key);
     }
     
+		SettingsBase& nonConstThis = const_cast<SettingsBase&>(*this);
     if (SettingRegistry::getInstance()->settingExists(key))
     {
-        setting_values[key] = SettingRegistry::getInstance()->getSettingConfig(key)->getDefaultValue();
+        nonConstThis.setting_values[key] = SettingRegistry::getInstance()->getSettingConfig(key)->getDefaultValue();
     }
     else
     {
-        setting_values[key] = "";
+        nonConstThis.setting_values[key] = "";
         cura::logError("Unregistered setting %s\n", key.c_str());
     }
-    return setting_values[key];
+    return setting_values.at(key);
 }
 
 void SettingsMessenger::setSetting(std::string key, std::string value)
@@ -100,7 +101,7 @@ void SettingsMessenger::setSetting(std::string key, std::string value)
     parent->setSetting(key, value);
 }
 
-std::string SettingsMessenger::getSettingString(std::string key)
+std::string SettingsMessenger::getSettingString(std::string key) const
 {
     return parent->getSettingString(key);
 }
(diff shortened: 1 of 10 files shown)


src/settings.cpp
```diff
@@ -72,27 +72,28 @@ void SettingsBase::setSetting(std::string key, std::string value)
...
     if (parent)
     {
         return parent->getSettingString(key);
     }
     
+		SettingsBase& nonConstThis = const_cast<SettingsBase&>(*this);
     if (SettingRegistry::getInstance()->settingExists(key))
     {
-        setting_values[key] = SettingRegistry::getInstance()->getSettingConfig(key)->getDefaultValue();
+        nonConstThis.setting_values[key] = SettingRegistry::getInstance()->getSettingConfig(key)->getDefaultValue();
```
BagelOrb: Why make getSettingString const when it clearly changes the container?

BagelOrb:
Good work.
I have been hesitant to use const_cast in the past; because in many cases the use of a const somewhere enforced the use of const at other places, I wasn't happy to start to introduce const tags at new places.

Now I know better. The right way to start to introduce const correctness to the program is by having const_casts to limit the amount of const-creep.

Good to see you working on this.

src/settings.cpp
BagelOrb: I've been struggling with whether to make it const or not for quite a while. What's your reasoning?

soyersoyer:
Why don't you just read the default values from SettingRegistry if not set? I think that you shouldn't modify the setting in a getter.

codehero:
@soyersoyer I don't know the code well enough to understand why the author is not just falling to SettingRegistry each time (as opposed to caching in the local map member). Without deeply understanding the use of setting_values[] everywhere in the file the behaviour should not be changed in the same pull request.

src/settings.cpp
codehero: getSettingString is a "get" function and is surprising that it is not const. It is the foundation of all other getSetting* functions, which are interpreting data into a specific data type and not modifying the Settings. If these get functions cannot be made const, nothing else can.

BagelOrb:
Reading from the SettingsRegistry would be less efficient than reading from the unordered map. 

Perhaps it's possible to refactor the code such that all default settings from the SettingsRegistry are put into setting_values of the root SettingsBase before they are overwritten by the actual setting values.

For the time being it's maybe best to have the const cast in there.

src/settings.cpp
BagelOrb: Sounds like reason enough! The fact that the setting_values is changed should be addressed somewhere in the future, after merging your PR.

BagelOrb:
@codehero are you finished tagging stuff with 'const'? Or should I wait with merging this PR until you are finished?

codehero:
@BagelOrb Basically my plan is to add a feature to Cura to directly use presliced polygons as a basis for motion planning. So I was tagging things const to better understand how to modify the code. I do have a few more things to tag and will be putting them up over the next few days, so I will let you know when I am finished.

codehero:
const correctness will definitely have to be done in stages.

Resolving const issues in PolygonRef in utils/polygon.h (which combines const and nonconst methods) is a whole different const branch. This is especially worrisome segment:

https://github.com/Ultimaker/CuraEngine/blob/5941d2252c898cbdf3e740ca018a1feb988077cc/src/utils/polygon.h#L41-L44

A const copy constructor that copies a modifiable pointer! Not exactly logical const.

Will keep this PR posted with more findings.

BagelOrb:
Ouch! That's embarrassing! Quickly! Remove it!

codehero:
I think this does it for this round of const changes. However there is one thing I noticed in MeshGroup:

`https://github.com/Ultimaker/CuraEngine/blob/5941d2252c898cbdf3e740ca018a1feb988077cc/src/MeshGroup.h#L23-L30`

How come extruder_count is initialized to -1 and not to what's in the settings? Can the extruder count change during execution?

I bring this up because when I had no -eX parameters on the command line CuraEngine segfaulted in the middle of execution.

BagelOrb:
The extruder count is set by the json file for your machine, so it's set as
soon as that file is read, which is after the meahgroup is initialised.
Op 26 feb. 2016 17:43 schreef "David Bender" <notifications@github.com>:

BagelOrb:
The MeshGroup object is initialised before settings are assigned because the settings are assigned to the MeshGroup. At construction time the MeshGroup cannot access the settings yet because they haven't been assigned yet.

BagelOrb:
Thanks for the help!

Please notify us when you finished implementing feature of sending presliced layers to Cura.
Are you using the command socket?
What format do you use for sending the data? Just a vector of 2D integer points for each polygon?

codehero:
I've really just hacked it in at this point as members of the Mesh and have not used the command socket. I am using a quickly rolled JSON format to send the data. My vision for this feature is to break out of the (CAD->STL->Slice) mold in which the community is indoctrinated and see how presliced data can fix some of the issues STL brings. For example, in OpenSCAD due to performance constraints imposed by CGAL you must use cylinders and spheres with roughly triangulated faces. This results in undersized cylindrical holes (segments cut across the circle hole). Exporting slice geometry instead (either as very many way points or even as a perfect circle primitive) would mitigate this issue. Then CuraEngine would just perform the support, motion planning, etc.

So I believe discussion and a stage of proof of concept in a topic branch is necessary to prove the value of this feature, determine the exact file format, etc is in order.

PR #313, opened 2016-03-01
refactor CommandSocket

soyersoyer:
use a null object as a default command socket
refactor arcus specific command socket to a separate compilation unit


src/FffGcodeWriter.cpp
```diff
@@ -176,18 +173,18 @@ void FffGcodeWriter::initConfigs(SliceDataStorage& storage)
...
+    //    std::ostringstream prefix;
+    //    prefix << "FLAVOR:" << toString(gcode.getFlavor());
+    //    gcode.writeComment(prefix.str().c_str());
+    //    if (gcode.getFlavor() == EGCodeFlavor::ULTIGCODE)
+    //    {
+    //        gcode.writeComment("TIME:666");
+    //        gcode.writeComment("MATERIAL:666");
+    //        gcode.writeComment("MATERIAL2:-1");
+    //    }
+    //}
```
BagelOrb: Why did you comment this out?
This should be in there to ensure there is a header in the gcode file for UM2 machines when using CuraEngine directly via the terminal.

src/Wireframe2gcode.cpp
```diff
@@ -548,10 +548,10 @@ void Wireframe2gcode::processStartingCode()
...
     if (gcode.getFlavor() == EGCodeFlavor::ULTIGCODE)
     {
-        if (!CommandSocket::isInstantiated())
-        {
-            gcode.writeCode(";FLAVOR:UltiGCode\n;TIME:666\n;MATERIAL:666\n;MATERIAL2:-1\n");
-        }
+        //if (!CommandSocket::isInstantiated())
+        //{
+        //    gcode.writeCode(";FLAVOR:UltiGCode\n;TIME:666\n;MATERIAL:666\n;MATERIAL2:-1\n");
+        //}
```
BagelOrb: This code should be here to make sure the gcode has a proper header for UM2 machines when calling CuraEngine directly from the command line.

src/commandSocket.cpp
```diff
@@ -1,472 +1,17 @@
...
-    else
-    {
-        layer = current_sliced_object->add_layers();
-        layer->set_id(id);
-        current_layer_count++;
-    }
-
-    return layer;
-}
-#endif
```
BagelOrb: It's a shame git didn't see this commit as a rename of commandSocket.cpp to commandSocketArcus.cpp ....

src/commandSocket.h
```diff
@@ -7,95 +7,62 @@
...
 namespace cura
 {
 
 class CommandSocket
 {
 private:
-    static CommandSocket* instance; //!< May be a nullptr in case it hasn't been instantiated.
-
-    CommandSocket(); //!< The single constructor is known only privately, since this class is similar to a singleton class (except the single object doesn't need to be instantiated)
+    static std::unique_ptr<CommandSocket> instance_; //!< always instantiated
```
BagelOrb: Why does instance_ have an underscore at the end?
There is no code convention to that end.

BagelOrb:
Took me some time to see that while all socket functions have become virtual they all have an empty default implementation.

src/FffGcodeWriter.cpp
soyersoyer: I tried to make the code independent whether a commandSocket instantiated or not.
Is it ok if it runs in the instantiated case too?

awhiemstra:
What is the point of this? It adds a bunch of abstraction and complicated logic for no apparent gain.

soyersoyer:
It removes a lot of ifdefs and ifs, which are complicate the code. These abstractions complicates the code too, but in a better way (IMO)

nallath:
I kinda like the changes soyersoyer made. However, I do think there are parts of the engine that could use more love than the command socket.

BagelOrb:
I also liked the PR.

However the way it was didn't impact the runtime, while your implementation
introduces a virtual table which is accesses every time a function from the
command socket is used, right?
Op 1 mrt. 2016 17:30 schreef "Jaime van Kessel" <notifications@github.com>:

src/FffGcodeWriter.cpp
BagelOrb: No it should only run when CuraEngine outputs directly to the terminal instead of via the command socket.

BagelOrb:
The idea was nice, but there were a couple of problems with this and I don't think anybody will solve those and use this PR as a basis.

It's a bit of a shame, but that's life I guess.

PR #323, opened 2016-03-26
Fix typo. downSkinCount -> upSkinCount

sean041:
This typo causes flaky crash when downSkinCount < upSkinCount.


BagelOrb:
Good find! Thanks!

Will merge this ASAP after easter.

PR #335, opened 2016-05-05
Compare in PolygonRef::shorterThan is strange.

sean041:
PolygonRef::shorterThan will always return false?


BagelOrb:
Thanks for making the PR when you spotted this mistake!

PR #350, opened 2016-06-11
Ignore generated Makefile and archives of code

thopiekar:
When running CMake on Linux, Cygwin (and also OSX?) it produces Makefiles to build CuraEngine via "make". Additionally, when building via Make archives of compiled code are created (*.a files). These should be ignored, too.


BagelOrb:
Shouldn't we just ignore the build folder, rather than specific contents of it?

thopiekar:
If you run "cmake ." and "make" they are all placed at the root of the
project folder. Additionally, I don't think that we will ever ship *.a
files with the sources ( in that case we have cura-binary-data for that).

BagelOrb:
What we do is we make a dir called "build" go into it and run "cmake .." and "make"...

awhiemstra:
I generally consider not building in a separate directory a bad practice, which is one of the reasons these excludes were never added.

thopiekar:
I absolutely agree that building software that way is bad (also packaging
tools, like debhelper create a build directory automatically). But I don't
see any reason why adding this could hurt 🤔
I just hit that situation, when trying to set up my windows installation
for development and just wanted to try whether it is possible to build the
engine at all or not.
In short: I guess it won't be 'useful' in most situations, but I think, if
we are already ignoring .o files, we can also ignore .a files as well.

awhiemstra:
There is nothing wrong with adding them (the PR was already merged) I am just pointing out my opinion. :)

BagelOrb:
You should never have hit that situation if you had tried to build in a separate directory.

PR #353, opened 2016-06-13
Adding some references

Johan3DV:
Not passing these variables by reference in some cases leads to really slow slicing because of excessive copying of data.


awhiemstra:
Since they are already passed as const I think this was just an oversight. So I am merging this.

BagelOrb:
Oh dear! How foolish of me!
Thank you so much!
:D

PR #354, opened 2016-06-14
Adding polygon packing feature and change the structure of the layer …

Johan3DV:
…data sent to the front-end to speed up the layer view

Needs the corresponding pull requests to Uranium and Cura to work.


sedwards2009:
I'm curious, what kind of a speed up can I expect with this? Have you got any numbers to share?

Johan3DV:
Sorry for not giving more of a background here yet. I have no specific nummers but I can share a video: https://www.dropbox.com/s/zla2em8atnzxzq1/Kazam_screencast_00004.mp4?dl=0

The slicing is done @4:00. Don't mind the slicing time,  the engine was compiled without optimizations.

BagelOrb:
You, sir, are my hero. That layer view is awesomely quick.

Does this PR come with a corresponding PR on the frontend?

We've been planning to change sending polygons to sending lines, but just don't have the time :/

I'm inclined to merge this just because of the results shown in the video, but I guess I'll first review your code...

src/PrintFeature.h
```diff
@@ -1,12 +1,14 @@
 #ifndef PRINT_FEATURE
 #define PRINT_FEATURE
 
+#include <cstdint>
+
 namespace cura
 {
 
-enum class PrintFeatureType
+enum class PrintFeatureType: std::uint8_t
```
BagelOrb: Why not (unsigned) char?

src/PrintFeature.h
```diff
@@ -1,12 +1,14 @@
...
 {
-    NoneType, // unused, but libArcus depends on it
+    NoneType, // used to mark unspecified jumps in polygons. libArcus depends on it
```
BagelOrb: What do you mean by unspecified jumps? I'm curious.

src/commandSocket.cpp
```diff
@@ -91,11 +91,59 @@ class CommandSocket::Private
 
     std::unordered_map<int, std::shared_ptr<cura::proto::Layer>> sliced_layers;
 };
+
+class CommandSocket::PolygonCompiler
```
BagelOrb: Could you please document this class and all its members?

It's unclear to me what the code is for.

src/commandSocket.cpp
```diff
@@ -91,11 +91,59 @@ class CommandSocket::Private
...
+public:
+	PolygonCompiler(CommandSocket::Private& cs_private_data):
+		_cs_private_data(cs_private_data),
+		_layer_nr( 0 ),
+		_line_widths(),
+		_poly()
+	{}
+	~PolygonCompiler(){
+		if( _poly.size() )
+			flushPolygon();
```
BagelOrb: Please take note of our code conventions (https://github.com/Ultimaker/Meta/blob/master/code_conventions.md)
Most notable: now I will put your braces on new lines and introduce braces for one-line blocks of code which removes your credits from the git blame ;)

src/commandSocket.cpp
```diff
@@ -91,11 +91,59 @@ class CommandSocket::Private
 
     std::unordered_map<int, std::shared_ptr<cura::proto::Layer>> sliced_layers;
 };
+
+class CommandSocket::PolygonCompiler
+{
+	static_assert(sizeof(PrintFeatureType) == 1, "To be compatible with the Cura frontend code PrintFeatureType needs to be of size 1");
```
BagelOrb: Indentation mistake: please indent by 4 spaces instead of tabs.

src/commandSocket.h
```diff
@@ -62,6 +62,16 @@ class CommandSocket
     void sendPolygons(cura::PrintFeatureType type, int layer_nr, cura::Polygons& polygons, int line_width);
 
     /*! 
+     * Send a polygon to the engine with variable line widths. This is used for the layerview in the GUI
```
BagelOrb: To the _frontend_, this is the engine..

Johan3DV:
The way I see it the problem with the current layer view implementation is that the data is sent from the engine to the front-end in polygon packets that are really small. This leads to for-loops with many iterations in the front-end code which are really slow.

The current implementation of sending polygons to the front-end sends polygons that contain a single line segment each. This is not only inefficient in terms of memory usage but also make the processing in Cura slow. To start out with I added a class in the back-end called PolygonCompiler which buffered polygons of the same type and sent them to the front-end packed together into larger polygons to mitigate the problem. 

It turned out to only help so much because of the structure of the actual data. The way a polygon is currently defined, it's a collection of lines of the same type (infill, skin, etc) with a common line width. This is not really a good way of representing the data CuraEngine produces. Line widths vary and only the insets contain continuous segments of the same type. It seems more appropriate to pack the layer data in another way. The polygon message sent from the engine to the front-end in this PR has been redefined so that for each line segment the polygon defines the line width and the type. This also allows the computations in the front-end to be expressed much more efficiently using numpy.

**The big problem with the implementation in this PR is that it's currently a protocol breaking change**. CuraEngine and Cura from before and after this PR won't be compatible for viewing polygons.

I think the alternative to this protocol breaking change is to make another type of layer message in the protocol which allows one to define another type of polygon according to the one in this PR. A front-end supporting these later protocol version would then need to notify the back-end which type of message to use and if no such notification is received the back-end falls back to the currently used type of layer/polygon message. These cases needs to be handled both in the front-end and in the back-end. 

Resolving the protocol breaking changes would add more complexity to the code. I don't see any reason for using the current implementation of the polygon message in the future. Even if the code responsible for handling the data in the front-end was rewritten using C++ the current polygon is still not flexible enough. Another protocol breaking change would be sending the point coordinates of the polygon not as int64 but as floats, which are actually used for displaying them, which would reduce message size. The question arises whether the precision offered by int64 will be beneficial to Cura in the future or just be a memory overhead. There might be other ways of specifying the line data that would be better from a memory point of view.

Cura.proto
```diff
@@ -44,21 +44,9 @@ message Layer {
...
-        SupportInfillType = 7;
-        MoveCombingType = 8;
-        MoveRetractionType = 9;
-    }
-    Type type = 1; // Type of move
-    bytes points = 2; // The points of the polygon, or two points if only a line segment (Currently only line segments are used)
-    float line_width = 3; // The width of the line being laid down
+    bytes line_type = 1;
+    bytes points = 2;
+    bytes line_width = 3;
```
BagelOrb: I think this should be something like:
```
message Paths {
    repeated PathSegment
}
message PathSegment {
    enum Type { ... {
    Type type
    bytes points
    float line_width
}
```

BagelOrb:
(See the accompanying Cura PR: https://github.com/Ultimaker/Cura/pull/800)

Did you maybe mean "polyline" instead of "polygon"? The difference is that a polyline is not closed, while a polygon will have a line segment between the last and the first point.

A 'polygon' of two points is actually just a single line segment and not a proper polygon. There is special casing everywhere to check whether a list of points is a polygon or a line segment..

I can't imagine how you assemble the line segments into polygons, given that zigzag support consists of non-closed polylines.

Just to be clear: the current protocol is a remainder of a past when line widths didn't change and whole polygons were sent to the frontend. It's good that you have been given the time to fix such issues.

> The polygon message sent from the engine to the front-end in this PR has been redefined so that for each line segment the polygon defines the line width and the type.
If you are really sending the data per polygon proper, the type won't change and it doesn't make sense to have a list of types. If however you are sending the data of different types, then you should rename 'Polygon' to 'Paths'.

Perhaps it's best to leave the current Polygons message handling as it is and introduce new handling for your new message which should be called something like 'Paths' then. We could use a polygons message in the future for sending the infill _area_ instead of the printed lines, so that we can view the lower layers faster and at an earlier stage in slicing.

Johan3DV:
The polygon is a misnomer on my behalf. The "Polygons" I refer to are the ones named as such in the commandSocket and Cura.proto not the internal structures of the Cura engine. I agree the distinction should be made between these path segments and proper polygons. The naming should be changed to Paths and the old Polygon message restored.

When it comes to the varying line widths I'm wondering whether all of them are actual line width differences or they are an artefact of varying the extrusion based on line overlap. If it's the latter case I guess the line widths should not change for visualisation purposes.

I was not aware of the code conventions document and will read up on it. ( Going through it quickly I must say I don't agree with in the conventions that using NULL as an integer is acceptable. NULL should never be used as an integer )

I will look at all comments this evening as I can't access the code right now.

BagelOrb:
The code conventions say that NULL should _never_ be used...

Johan3DV:
Sorry, I'm a bit tired. I interpreted the phrase "NULL is an integer, not a pointer." not as a statement of the non-pointer nature of NULL but as a declaration that it's an integer. :)

nickthetait:
The performance in the video does look impressive. The real 'proof' would be two videos of the same object being rendered, with and without this change introduced.

src/PrintFeature.h
Johan3DV: I mean there is a movement in the path that has not been specified by added data. This could for instance be if you were to send the line data before the export step where you only have isolated infill lines. If they were added to a path segment the NoneType would indicate this jump.

BagelOrb: I think a better example would be the extruder switch. When you switch extruder (in place) you don't actually have a travel move to the location of the new nozzle, so the next path is not connected to the previous.

Is that an example of when NoneType should be used?

Cura.proto
```diff
@@ -61,6 +61,23 @@ message Polygon {
     float line_width = 3; // The width of the line being laid down
 }
 
+message LayerOptimized {
+    int32 id = 1;
+    float height = 2; // Z position
+    float thickness = 3; // height of a single layer
+
+    repeated PathSegment path_segment = 4; // layer data
```
BagelOrb: Why are the paths in a layer split into `PathSegment`s?

You could suffice with just the byte arrays below, right?

For what kind of optimization do you use this?

Johan3DV: A very good question. I think the answer is to be safe and flexibility for future changes. The PathCompiler flushes the data when the layer changes, if the layer were to jump back to a previous layer you would get two PathSegment messages for that layer. I've not seen this happen but I'm not certain that it can't happen or won't happen in the future.

BagelOrb: But would the layer be flushed then? Wouldn't it be better to just send two
optimisedLayer s at the same height?

There is a feature which might just possibly be implemented at some time:
printing a couple of layers of one object before printing a couple of
layers of the other object, etc.

Anyway I think simplicity of the current code is more important than
flexibility for future possibly-might-be's.
Op 16 jun. 2016 00:29 schreef "Johan" <notifications@github.com>:

Johan3DV: One idea when this is beneficial, related to your earlier comment, is when using multiple extruders. Perhaps a new field should be added to the PathSegment message to indicate which extruder is active. Then you would get as many PathSegments as you have extruders. Could of course be implemented in another way using multiple layers or line type offsets.

BagelOrb: That seems like a good idea!
Sending the extruder along to the frontend had still to be done anyway.
Visualizing the gcode differently depending on the extruder would be a very nice feature.

src/commandSocket.cpp
```diff
@@ -447,25 +588,108 @@ void CommandSocket::sendGCodePrefix(std::string prefix)
...
+    }
+    points.clear();
+    line_widths.clear();
+    line_types.clear();
+}
+
+//TODO: Reason about these functions when points is empty or the polygon entered is empty
+void CommandSocket::PathCompiler::sendLine(PrintFeatureType print_feature_type, Point from, Point to, int width)
+{
+    handleInitialPoint( from );
```
BagelOrb: I think it's better to expose this. What I suggest is to rewrite quite a bit of code.

Rather than registering line segments, register moves to a new location, assuming the from-location is the last provided to-location. Besides, gcode also works that way.
Then handle jumps/initial point from outside this class.

So instead of exposing `sendLine(.)`, expose a function `setCurrentLocation(.)` (which is similar to `PathCompiler::handleInitialPoint(.)`) and a function `sendMoveTo(.)`.

Then the functions `GCodePlanner::sendPolygon` (and the one in `MergeInfillLines`) can simply not use the `from` parameter.

The only places where `setCurrentLocation` should be called are:
- at the very start of the gcode it 'jumps' to the first location
- when switching extruder it jumps a relative distance or to an absolute point, based on some machine settings. For the moment I suggest we neglect this case, because it doesn't really matter that the extruder switch shows up as a travel move.

src/commandSocket.cpp
```diff
@@ -57,45 +57,135 @@ class Listener : public Arcus::SocketListener
...
     
     // Print object that olds one or more meshes that need to be sliced. 
     std::vector< std::shared_ptr<MeshGroup> > objects_to_slice;
 
-    std::unordered_map<int, std::shared_ptr<cura::proto::Layer>> sliced_layers;
+    SliceDataStruct<cura::proto::Layer> sliced_layers;
+    SliceDataStruct<cura::proto::LayerOptimized> optimized_layers;
+};
+
+class CommandSocket::PathCompiler
```
BagelOrb: Please add documentation for this class.

Cura.proto
```diff
@@ -44,11 +44,40 @@ message Layer {
...
+}
+
+
+message PathSegment {
+	int32 ordinal = 1; // The absolute ordinal of this line segment
```
BagelOrb: indentation: please replace tabs by four spaces.

src/commandSocket.cpp
```diff
@@ -57,93 +57,135 @@ class Listener : public Arcus::SocketListener
     }
 };
 
+template <typename T>
+class SliceDataStruct
```
BagelOrb: Please document this class.

src/commandSocket.cpp
```diff
@@ -57,93 +57,135 @@ class Listener : public Arcus::SocketListener
...
+        line_widths(),
+        points()
+    {}
+    ~PathCompiler()
+    {
+        if( points.size() )
+            flushPathSegments();
+    }
+
+    /*!
```
BagelOrb: I'd rather see the documentation with the declaration, rather than the definition.
However, I'm not blocking merging this PR on that..

BagelOrb:
The diffs went a bit haywire on your last commit. Please try to commit your code in smaller chunks, so that they are easier to review.

BagelOrb:
For example: I usually do renaming of variables, functions or classes in a separate commit, so that the reviewer doesn't really need to review that at all.

BagelOrb:
I'm sorry for letting you wait so long before I reviewed your second commit. Maybe I overlooked that you committed new stuff or maybe I didn't get a notification. To be sure next time, please email me (t.kuipers@ultimaker.com) when you have new stuff to review.

BagelOrb:
You still need to perform the refactor such that `sendLine(.)` only takes a destination point, rather than also the departure point, right?

Johan3DV:
There are a couple of reasons I didn't change the interface functions. 

My main concern is finding everywhere the initial `setCurrentLocation(.)` should be called. The interface changes does simplify the code and reduce runtime but it's not strictly necessary. And there will possibly be issues to root out.

Also the current layer doesn't need to be passed along each call, only set when it's changed.

There are still places where polygons are sent and there you still need to use `PathCompiler::handleInitialPoint(.)`.

BagelOrb:
I think it's a bit weird I hear this only now, while I already gave this feedback over 2 weeks ago.
After I have reviewed it's good practice to always comment on every comment I have made. Either say you have fixed it, or leave a reply of why you think that my comment shouldn't be fixed.

1) Since sendPolygons in master is always called with gcode.getPositionXY() you could simply setCurrentLocation in the very first meshgroup at the start of slicing to how that position is initialized.

2) Good point! It should also be fairly easy to set the layer_nr via a separate function.

3) Where polygons are sent `sendPolygons` is called instead of `sendLine` right?


The way the code is now is a remainder of how sendPolygons was misused. I don't want to leave overcomplex and longer running code, just because we might encounter issues. If we do that each time the code base will soon be a mess. We shouldn't only implement what's strictly necessary.


I think you should be able to figure out how the proposed refactor could work in a good way. I will review your work afterwards anyway. I don't think we should be too scared of bugs - we've got code review, component test and quality assurance for that!


If you find it too hard to find out what stuff needs to change, I could try to do the refactor myself. No worries.

Johan3DV:
I've done the refactoring you suggested.

src/commandSocket.cpp
```diff
@@ -138,13 +138,14 @@ class CommandSocket::PathCompiler
...
         data_point_type(cura::proto::PathSegment::Point2D),
+        line_types(),
         line_widths(),
         points(),
         last_point{0,0}
     {}
     ~PathCompiler()
     {
-        if( points.size() )
+        if( line_types.size() )
```
BagelOrb: According to the code conventions there should be a space after the keyword `if` and it's not conventional to put extra spaces after an before the `(` and the `)`.

src/commandSocket.cpp
```diff
@@ -470,35 +488,75 @@ void CommandSocket::sendPolygons(PrintFeatureType type, int layer_nr, const Poly
...
+    if (CommandSocket::isInstantiated())
+    {
+        auto& path_comp = CommandSocket::getInstance()->path_comp;
+        path_comp->setCurrentPosition( position );
+    }
 #endif
 }
 
-void CommandSocket::sendLine(cura::PrintFeatureType type, int layer_nr, Point from, Point to, int line_width)
+void CommandSocket::setLayerForSend( int layer_nr )
```
BagelOrb: Spaces around function argumnents is unconventional.

src/commandSocket.cpp
```diff
@@ -684,7 +742,7 @@ std::shared_ptr<cura::proto::LayerOptimized> CommandSocket::Private::getOptimize
 #ifdef ARCUS
 void CommandSocket::PathCompiler::flushPathSegments()
 {
-    if( points.size() > 0 && CommandSocket::isInstantiated() )
+    if( line_types.size() > 0 && CommandSocket::isInstantiated() )
```
BagelOrb: Spaces around function argumnents is unconventional.
Space after if

src/commandSocket.cpp
```diff
@@ -707,20 +765,19 @@ void CommandSocket::PathCompiler::flushPathSegments()
 }
 
 //TODO: Reason about these functions when points is empty or the polygon entered is empty
-void CommandSocket::PathCompiler::sendLine(PrintFeatureType print_feature_type, Point from, Point to, int width)
+void CommandSocket::PathCompiler::sendLineTo(PrintFeatureType print_feature_type, Point to, int width)
 {
-    handleInitialPoint( from );
+    assert( points.size() > 0 );
```
BagelOrb: When doing an assert, it would be nice to see what the assertion means.
One simple way to do this is like:
```
assert(points.size() > 0 && "Sending a line to a new location presupposes we already have a current location set!");
```

BagelOrb: never mind

src/commandSocket.cpp
```diff
@@ -57,45 +57,188 @@ class Listener : public Arcus::SocketListener
...
+ */
+class CommandSocket::PathCompiler
+{
+    typedef cura::proto::PathSegment::PointType PointType;
+    static_assert(sizeof(PrintFeatureType) == 1, "To be compatible with the Cura frontend code PrintFeatureType needs to be of size 1");
+    //! Reference to the private data of the CommandSocket used to send the data to the front end.
+    CommandSocket::Private& _cs_private_data;
+    //! Keeps track of the current layer number being processed. If layer number is set to a different value, the current data is flushed to CommandSocket.
+    int _layer_nr;
+    int extruder;
```
BagelOrb: Please document each class member.

src/commandSocket.cpp
```diff
@@ -57,45 +57,188 @@ class Listener : public Arcus::SocketListener
...
+        data_point_type(cura::proto::PathSegment::Point2D),
+        line_types(),
+        line_widths(),
+        points(),
+        last_point{0,0}
+    {}
+    ~PathCompiler()
+    {
+        if( line_types.size() )
+            flushPathSegments();
```
BagelOrb: Code conventions: space after `if`, always use braces `{`, `}`

src/commandSocket.cpp
```diff
@@ -57,45 +57,188 @@ class Listener : public Arcus::SocketListener
...
+    int getLayer() const
+    {
+        return _layer_nr;
+    }
+    /*!
+     * Used to set which extruder will be used for printing the following layer data is intended for.
+     */
+    void setExtruder(int new_extruder)
+    {
+        if(extruder != new_extruder ){
```
BagelOrb: code convention: opening brace `{` always on a new line.

BagelOrb: Also space after `if`.

You seem to introduce spaces after the `(` and before the `)` quite inconsistently. I suggest you check your code and remove those spaces everywhere.

BagelOrb:
I think your code is generally ok to be merged, but I'd like for it to be up to the code conventions standard.

https://github.com/Ultimaker/Meta/blob/master/code_conventions.md

There's still a lot of cases of wrong spacing and missing braces `{ }`

src/commandSocket.cpp
Johan3DV: Sorry about that, it's an old habit as I think the extra spaces makes the code easier to read. 

I will clean up the code this evening.

BagelOrb:
I've solved the code style problems for you.

Also I've solved a crash which happened when using one-at-a-time mode. See https://github.com/Ultimaker/CuraEngine/commit/39e1c5ffd92f9243ccd13c1479a627aac6c4aa61

BagelOrb:
Please add documentation for the undocumented stuff you introduced.

PR #361, opened 2016-06-24
[CURA-1738] Time estimates way off

thopiekar:
Just applied on the code my own formula, which assumes that the acceleration is always constant during delta_t. The derivation of the formula is on JIRA.

Contributes to CURA-1738


BagelOrb:
Your formula is equavalent, but uses more operations. Moreover, you shouldn't use pow(.) when squaring. That's why the old function just used a function square(.)

PR #364, opened 2016-06-27
[CURA-1738] Time estimates way off

thopiekar:
Zero'ing the discriminant is physically incorrect.


BagelOrb:
The documentation of the function clearly needs to mention that the sign of the acceleration and the sign of the distance should be the same.

Also you can add an assert to check whether that is the case or the values are so small that the difference in sign might be due to a rounding error.

As Ruben said, however, the max(0, ...) is due to rounding errors happening inside the function itself.

PR #368, opened 2016-07-06
Slicer speedup

scottlenser:
This branch has speedups for the slicer in Cura.

For one (probably nearly worst case) model I tested, the changes
result in the slicing time (without optimizations on) going from
73.348 seconds to 0.04 seconds.  That's a speedup of 1833x.  I think
that's a new persona best. :) I'm attaching the bad case model.  It is
important to slice it with all layer height parameters set to 0.25mm.
This makes the slicing height equal to the height of some of the
vertices in the model.  It should only be a single layer high.  This
is a scaled in z version of a model I found on thingiverse.  The
original model is at http://www.thingiverse.com/thing:559047 but I
don't expect this to cause the performacne problem in the same way.  I
was looking at it because it is a difficult model for perimeters /
insets.  I expect the speedups to be more significant when the polygon
has more vertices coincident with slicing heights or more errors that
can not be handled in the makeBasicPolygonLoops() stage.

Here are the major speedups related to slicing:

- When constructing basic polygon loops, it now only checks the face
  that shares the edge that created the end point.
- When constructing basic polygon loops, if the segment ended on a mesh
  vertex, then it checks all the faces connnected to that vertex to try
  to find the continuation.  This slows down basic polygon looping an
  immeasurably small amount but increases overall performance.  This
  drastically reduces the number of open polylines for some model /
  slicing height combinations.
- The connectOpenPolylines() function has been replaced with a new
  implementation.  This changes the run time from O(n^2) to O(n) and
  also connects many more segments.  The previous implementation
  would miss some connections.  For instance, if the connections should
  be 0 -> 10 -> 5, the previous implementation would connect 0 and 10
  but fail to connect that result to 5.
- The stitch() function has been replaced with a new implementation.
  This changes the run time from O(n^3) to O(n).
- These implementations are O(n) where n is the number of polylines for
  the merge decisions.  The actual merging process is still O(i*j) I think
  where i is the number of polylines merged into the first one and j
  is the number of points in those polylines.

Some implementation notes related to the slicing speedups:

- connectOpenPolylines() and stitch() now share an implementation that
  is parameterized on max match distance and whether polylines can be
  reversed in direction
- A spatial data structure provides efficient searching for nearby polyline
  end points.
- A priority queue is used to remember and prioritize the end points that
  were close enough to consider joining.
- A directional mapping via vectors is used to track how end points of the
  polylines move around as polylines are merged.
- The new implementations use a new spatial data structure class called
  SparseGrid.  This provides very similar functionality to BucketGrid2D.
  Major differences are:
  - SparseGrid doesn't currently support a findNearest function.  This
    could be added pretty trivially if desired.
  - SparseGrid can take arbitrary search radii and still provide correct
    results.  It is only efficient for search radii near the grid cell
    size, but still produces correct results away from this.  This makes
    the API easier to describe.  It also offers finer control.  By changing
    the cell size with a given search radius, the tradeoff between number
    of grid cells that have to be searched and extra volume beyond radius
    that might return results can be managed.
  - SparseGrid never returns the same stored value multiple times.  It
    returns each item within the search radius exactly once.  It returns
    some items beyond the search radius, but for each item returned it
    is returned exactly once.
  - SparseGrid is invasive.  The value's location must be stored inside the
    stored value.  This makes it slightly less memory efficient but is also
    what allows it to be precise on returned values only being returned once.
  - The implementation is based on unordered_multimap instead of unordered_map.

Currently this branch also has a number of other changes that may or
may not be desired.  These can all be removed on request, but I didn't
know what you might want and what you wouldn't.  Here are the other
changes (in order of most likely to be wanted):

- More const correctness in places
- Fixed logging of number of layers
- Added error message when it can't find an inherited JSON file.
- Aligned bridge direction with calculated direction.
  - The bridging is still terrible, but this at least gives it a chance of working.
- Removed sending of extra SlicingFinished message
  - This was being sent in 2 places which was confusing my version of the front end
- Added convexHull() function
  - Renamed old function to approxConvexHull()
- Rounding during interpolation when calculating vertices from slicing
- Updated command_line_settings JSON file to version 2 format to match
  expected format of other JSON files.

This branch also has some changes that might be temporarily desired to
see the effect of other deltas.  These are:

- Printouts of various timings from slicing
- Printouts of stats regarding creation of polygons from slicing.

I imagine you'll probably want some other cleanup / changes, but rather than
try to guess what you might want I figured I'd just ask.  I'm just working on
this as a hobby because I think the process is interesting.  I am a professional
programmer during the day.  My work is sort of a combination of robots and
applied math.


scottlenser:
Here is aforementioned bad case stl file.

[tree_of_life_thin.zip](https://github.com/Ultimaker/CuraEngine/files/349051/tree_of_life_thin.zip)

scottlenser:
I've just done some more testing on random models.  For well behaved models with no special cases, this actually seems to be a small slowdown, but these models are fast to begin with.  The patch might drastically improve the worst case but cost a little in the normal case.  In any case, I'd like your thoughts on what I have so far.

src/MeshGroup.cpp
```diff
@@ -283,13 +284,17 @@ bool loadMeshSTL(Mesh* mesh, const char* filename, const FMatrix3x3& matrix)
...
+
     const char* ext = strrchr(filename, '.');
     if (ext && (strcmp(ext, ".stl") == 0 || strcmp(ext, ".STL") == 0))
     {
         Mesh mesh = object_parent_settings ? Mesh(object_parent_settings) : Mesh(meshgroup); //If we have object_parent_settings, use them as parent settings. Otherwise, just use meshgroup.
         if(loadMeshSTL(&mesh,filename,transformation)) //Load it! If successful...
         {
             meshgroup->meshes.push_back(mesh);
+            std::cout << "loading " << filename << " took " <<
+                load_timer.restart() << " seconds." << std::endl;
```
Ghostkeeper: You'll want to do this with the logger, not directly in cout. Especially since slicing with the command line by default outputs g-code to cout as well.

Ghostkeeper:
Sounds good! I reckon that BagelOrb would like to take a good look at this.

nallath:
Very awesome!

daid:
I'm already amazed that you managed to make head and/or tails from this piece of my code. Great job!

BagelOrb:
The first thing I notice is the fact that you introduced SparseGrid which is very similar to BucketGrid2D.
When I would merge your PR as is then I would have two classes who perform a _very_ similar task.

I would think that either one of the classes should remain. Perhaps you could merge the two into one (and use your name, which is better than my class name). Also you should take a look at the branch `feature_bucket_`feature_bucket_grid_2d_fix_2d_fix`. It contains an alternative implementation of BucketGrid2D and I haven't been able to decide on which one is better: the one on master or the one on the branch.

BucketGrid2D on master makes use of the fact that an unordered_map uses buckets internally. Multiple cells will map to the same bucket and the bucket may contains multiple objects. Also the same location may map to multiple objects, because the unordered_map uses an equality constraint which always returns false. This implementation is a bit awkward due to the fact that multiple cells will be mapped to the same bucket. The findNearest function iterates over all objects in the bucket and so iterates over points which may be far away from the search location.

BucketGrid2D on `feature_bucket_grid_2d_fix` works more like a multimap, but instead uses an unordered_map to map from a CellIdx to a vector of Elem. I don't see any reason why I didn't use a multimap there. At may be easiest for you to merge your SparseGrid with the BucketGrid2D on that branch. I.e. maybe you could merge that branch to your branch and then merge the two classes into one.

I don't really understand why you have made SparseGrid invasive. An `std::unordered_multimap<From, To>` already expects a `std::pair<From, To>` so it already contains `From` - there's no need to put `From` inside `To`. Am I misunderstanding what you mean by invasiveness?
Also you say that the invasiveness allows you to have a uniqueness constraint on the outputted values. That is to say that you have a uniqueness constraint on the pairs of object and location, but not on objects themselves. If you map multiple locations to the same object, you will still have that object in your result multiple times. I would therefore think you shouldn't put the object's location inside the object.

However, I say all of this without having looked at the code, so I might be mistaken about how SparseGrid actually works. I think I will be checking how well your PR works and reviewing your code this afternoon.

scottlenser:
I agree that it makes sense to only have 1 of SparseGrid or BucketGrid2D.  To me, it makes sense to consolidate on the SparseGrid implementation over BucketGrid2D.  I'll just need to add the nearest functions and switch over the code that uses BucketGrid2D.  That all should be pretty trivial.  I just didn't want to put into the effort to do that until I got some indication that that was acceptable.

I just took a quick look at the feature_bucket_grid_2d_fix branch.  The implementation there is more similar to the SparseGrid implementation.  We have 3 basic approaches at this point:
master:BucketGrid2D - unordered_map from Point to T with all keys unique
bucket_grid_2d_fix:BucketGrid2D - unordered_map from CellIdx to vector<Elem>, Elem contains Point and T
slicer-speedup:SparseGrid - unordered_multimap from GridPoint (equivalent to CellIdx) to Val, Val contains Point and T

Making the mapping from GridPoint to values gets rid of the problem with returning uninteresting points due to bucket collisions.  This means that the map value for std::unordered_{multi}map<GridPoint, T> doesn't automatically contain the Point location which we need since it is a std::pair<GridPoint, T> that is stored.  In the bucket_grid_2d_fix version, it solves this problem by making a type that is similar to a std::pair<Point, T> and storing that in the map instead.  I solved this problem by taking an extra template argument PointAccessor that can get a point from T.  I did it this way for extra flexibility.  If the objects that you want to store happen to already contain a Point internally, then an extra copy is not introduced.  If the objects that you want to store (say of type T) do not already contain a Point internally, then you can pass std::pair<Point, T> as the type argument to SparseGrid.  I haven't really thought through if there is a way to get the flexibility of the accessor approach with the convenience of the bucket_grid_2d_fix approach.  I did think about whether to use std::unordered_map<GridPoint, std::vector<Val>> or std::unordered_multimap<GridPoint, Val>.  It's not clear to me which would be better.  The multimap version basically uses a singly linked list in the implementation for each bucket instead of a vector.  The vector version I think ends up using a singly linked list that has an element pointing to the vector.  My guess is that the vector version is faster for large number of collisions and the multimap is faster for small numbers of collisions, but I have absolutely no data to back that up.  I ended up going with multimap simply because I could reuse more implementation and I figured that the folks writing std::unordered_multimap had a good reason for implementing it that way as they could have used a vector if they wanted to.

What I meant by the invasiveness, is that the passed type to be stored has to contain a Point for its location (or be able to calculate one from what it stores).  The From for the map is a GridPoint in this case representing the grid cell coordinates which is not the same thing at all.  For instance, with a cell size of 0.2 and a Point value of (1.2, 0.5), the corresponding GridPoint is (6, 2).  Relative to the approach of master, the mapping from GridPoint instead of Point is what offers the uniqueness even in the case of bucket collisions.  I'm assuming master:BucketGrid2D.h can return the same object multiple times since there are big warning comments stating this.  All I meant by uniqueness in SparseGrid is that the returned values only gives you each object once.  bucket_grid_2d_fix also has this property.  Once the map is switched from Point to GridPoint, then the Point has to go somewhere.  I did this through invasiveness (requiring the user to have an appropriate field) and bucket_grid_2_fix did this through its Elem type.  Both are valid approaches I think.  My approach is a little more flexible and the bucket_grid_2_fix approach is a little more convenient.

With SparseGrid, it is not possible to map multiple locations to the same object of type Val.  The locations are extracted from the Val type and so are always the same for a given object of type Val.  What you can do, is set Val = std::pair<Point,T> and then map several instances with the same T and different Point locations.  I guess uniqueness is not a great way to express the property, because you can add the same Val multiple times if you want to (although this seems useless) and it would be returned multiple times in the result.  I was just trying to express that you get what you would expect out from the map with no duplicates due to the implementation.

scottlenser:
Once you get a change to review, let me know what parts you want to keep and which you want to cut out.  Then I can remove the unwanted parts.

src/MeshGroup.cpp
scottlenser: Good point.  I'm not sure if this should even stay in the output at all as it mostly useful for performance testing.  I only left it in the diff in case someone wanted to run performance tests to verify the patch.

BagelOrb:
Indeed I think it's best to implement the `findNearby(.)` function (which is just `getNearby(.)` called with `cell_size`, I think)
Then either of us could just delete BucketGrid2D and use SparseGrid instead.

I'm sorry what I said about the From being the location anyway; I wasn't paying enough attention. Of course it's the CellIdx and not the location.

I'm now looking at your code and re-evaluating what you meant with the invasiveness / uniqueness. The reason I brought up my points was because I use BucketGrid2D to map locations on line segments to line segments of polygons. I would like to get a uniqueness constraint such that the same line segment which goes through a couple of cells will be returned only once. Now that I understand SparseGrid better I think it's better to enforce that uniqueness outside of SparseGrid, since that class isn't enforcing it anyway.

Hopefully I'll have time tomorrow to really review the code.

Maybe you can have a look at our code conventions in the meant time: https://github.com/Ultimaker/Meta/blob/master/code_conventions.md
The document is not perfect yet, but should make the most important practices clear.

src/MeshGroup.cpp
nickthetait: I would say that keeping this info would definitely be useful for testing purposes in the future.

scottlenser:
I was looking for that document on code conventions.  Thanks for the link.  I tried to make it match the ones that were obvious from the nearby code.  I'll try to make it match the conventions you linked.

src/settings/SettingRegistry.cpp
```diff
@@ -121,7 +121,7 @@ bool SettingRegistry::getDefinitionFile(const std::string machine_id, const std:
         char* path = strtok(paths,";:,"); // search for path delimited by ';', ':', ','
         while (path != NULL)
         {
-            result = std::string(path) + std::string("/") + machine_id + std::string(".def.json");
+            result = std::string(path) + std::string("/") + machine_id;
```
BagelOrb: I don't understand this commit. In Cura 2.2 the `inherit`field of a json file doesn't include the extension, so it must be added here to get the full file name.

BagelOrb: never mind

src/settings/settings.h
```diff
@@ -260,7 +260,11 @@ class SettingsBase : public SettingsBaseVirtual
                 sstream << "\"" << pair.first << "\": {\"default_value\":\"" << out_val << "\"},\n";
             }
         }
-        return sstream.str();
+        std::string out_str = sstream.str();
+        if (!out_str.empty()) {
+            out_str = out_str.substr(0,out_str.size()-2) + "\n";
+        }
+        return out_str;
```
BagelOrb: I think this is a bit of an ugly way to do it. You could do it by adding `,\n` _before_ the string except for the first iteration.
I don't really mind much, though. This is debugging code.

Please do document what this piece of code is for (or change it).

src/FffGcodeWriter.cpp
```diff
@@ -785,6 +785,7 @@ void FffGcodeWriter::processSkin(GCodePlanner& gcode_layer, SliceMeshStorage* me
         if (bridge > -1)
         {
             pattern = EFillMethod::LINES;
+            skin_angle = bridge;
```
BagelOrb: Seeing this commit I conclude that bridging simply didn't work at all.
Thanks for this fix!

src/FffProcessor.cpp
```diff
@@ -135,6 +135,9 @@ bool FffProcessor::processMeshGroup(MeshGroup* meshgroup)
         gcoder.writeGCode();
         log("finished Neith Gcode generation...\n");
         
+    } else if (meshgroup->getSettingBoolean("lenser_backend_enabled"))
```
BagelOrb: Perhaps a more descriptive setting name?

BagelOrb: I find it a bit weird to log this message here. Maybe you should log it at the place you use the setting.

BagelOrb:
I'm trying to review each commit and found stuff about a whole new MasonBackend, which wasn't in the eventual code. This makes reviewing quite hard for me.

What I would suggest is to make a new branch and cherry-pick all commits which are of interest for the slicer speedup. Could you please try to clean up your history in some way and make a new PR?

Perhaps you could make a separate PR for all the code cleanup you did including the bridging fix, const-correctness, json output, etc?

scottlenser:
I think you'll find it much easier to just review the diff of the end result compared to master.  Trying to review commit by commit probably won't work well in this case.  I was using:

git diff `git merge-base master slicer-speedup`

with the slicer-speedup branch checked out in order to get a diff.  I can't really cherry-pick out the commits of interest for the slicer speedup because they aren't split up cleanly like that.

I can split up the diffs into separate branches, but that isn't going to make the history look clean.  It would make the overall diff from start to finish simpler.  I could make a branch which is just one commit that just has the net changes.  Would that help?  Or if there is another way to clean things up, please let me know.

BagelOrb:
Ok I will just keep to the diff then.
Op 8 jul. 2016 16:47 schreef "scottlenser" <notifications@github.com>:

scottlenser:
I think I might have figured out a way to split up the history to make it cleaner.  I'll see what I can do.  I'm also working on converting BucketGrid2D uses to SparseGrid.

BagelOrb:
I'll hopefully squeeze in some time to review this weekend.
Op 8 jul. 2016 17:22 schreef "scottlenser" <notifications@github.com>:

scottlenser:
This PR is now obsolete.  I have split up the contents of this PR and made the history look nicer.  See PR #370, PR #371, and PR #372.  The union of those PR requests is the same as this PR as of this comment.

PR #370, opened 2016-07-08
Slicer speedup clean

scottlenser:

This branch has only the slicer speedup portions of the previous PR.  It also has a cleaned up (i.e. manufactured history).

This branch has speedups for the slicer in Cura.

For one (probably nearly worst case) model I tested, the changes
result in the slicing time (without optimizations on) going from
73.348 seconds to 0.04 seconds. That's a speedup of 1833x. I think
that's a new persona best. :) I'm attaching the bad case model. It is
important to slice it with all layer height parameters set to 0.25mm.
This makes the slicing height equal to the height of some of the
vertices in the model. It should only be a single layer high. This
is a scaled in z version of a model I found on thingiverse. The
original model is at http://www.thingiverse.com/thing:559047 but I
don't expect this to cause the performacne problem in the same way. I
was looking at it because it is a difficult model for perimeters /
insets. I expect the speedups to be more significant when the polygon
has more vertices coincident with slicing heights or more errors that
can not be handled in the makeBasicPolygonLoops() stage.

Here are the major speedups related to slicing:

- When constructing basic polygon loops, it now only checks the face that shares the edge that created the end point.
- When constructing basic polygon loops, if the segment ended on a mesh vertex, then it checks all the faces connnected to that vertex to try to find the continuation. This slows down basic polygon looping an immeasurably small amount but increases overall performance. This drastically reduces the number of open polylines for some model / slicing height combinations.
- The connectOpenPolylines() function has been replaced with a new implementation. This changes the run time from O(n^2) to O(n) and also connects many more segments. The previous implementation would miss some connections. For instance, if the connections should be 0 -> 10 -> 5, the previous implementation would connect 0 and 10 but fail to connect that result to 5.
- The stitch() function has been replaced with a new implementation. This changes the run time from O(n^3) to O(n).
- These implementations are O(n) where n is the number of polylines for the merge decisions. The actual merging process is still O(i*j) I think where i is the number of polylines merged into the first one and j is the number of points in those polylines.

Some implementation notes related to the slicing speedups:

-  connectOpenPolylines() and stitch() now share an implementation that is parameterized on max match distance and whether polylines can be reversed in direction
- A spatial data structure provides efficient searching for nearby polyline end points.
- A priority queue is used to remember and prioritize the end points that were close enough to consider joining.
- A directional mapping via vectors is used to track how end points of the polylines move around as polylines are merged.
- The new implementations use a new spatial data structure class called SparseGrid. This provides very similar functionality to BucketGrid2D. Major differences are:
- SparseGrid doesn't currently support a findNearest function. This could be added pretty trivially if desired.
- SparseGrid can take arbitrary search radii and still provide correct results. It is only efficient for search radii near the grid cell size, but still produces correct results away from this. This makes the API easier to describe. It also offers finer control. By changing the cell size with a given search radius, the tradeoff between number of grid cells that have to be searched and extra volume beyond radius that might return results can be managed.
- SparseGrid never returns the same stored value multiple times. It returns each item within the search radius exactly once. It returns some items beyond the search radius, but for each item returned it is returned exactly once.
- SparseGrid is invasive. The value's location must be stored inside the stored value. This makes it slightly less memory efficient but is also what allows it to be precise on returned values only being returned once.
- The implementation is based on unordered_multimap instead of unordered_map.


scottlenser:
I still need to do the cleanup to convert uses of BucketGrid2D to SparseGrid.

scottlenser:
The cleanup to convert uses of BucketGrid2D to SparseGrid is done.

src/utils/polygon.h
```diff
@@ -378,6 +378,10 @@ class Polygons
     {
         paths.push_back(*poly.path);
     }
+    void add(Polygon&& other_poly)
+    {
+        paths.emplace_back(std::move(*other_poly));
+    }
```
BagelOrb: I'm still not accustomed to using rvalues and move semantics. I should really use it more often.

For example the `add` function below could use a rvalue reference variant.

Errr.. This comment is more of a remark than it is a comment on your code.

src/utils/SparseGrid.h
```diff
@@ -0,0 +1,170 @@
...
+
+    /*! \brief Compute the grid coordinates of a point.
+     *
+     * \param[in] point The actual location.
+     * \return The grid coordinates that correspond to point.
+     */
+    GridPoint toGridPoint(const Point &point) const;
+
+    /*! \brief Map from grid locations (GridPoint) to values (Val). */
+    GridMap m_grid;
```
BagelOrb: Putting `m_` in front of member variables is not really a code convention we are following...

src/utils/SparseGrid.h
```diff
@@ -0,0 +1,170 @@
...
+     * Appends values from cell that might have values within \p
+     * radius of \p query_pt.  Appends all values that are within
+     * radius of query_pt.  May append values that are up to radius +
+     * cell_size from query_pt.
+     *
+     * \param[in] query_pt The point to search around.
+     * \param[in] radisu The search radius.
+     * \param[out] ret The values in the cell are appended to ret.
+     */
+    void addNearby(const Point &query_pt, coord_t radius,
```
BagelOrb: We generally have the ampersand with the type, rather than the variable, i.e. `Point& query_pt` rather than `Point &query_pt`...

src/utils/SparseGrid.h
```diff
@@ -0,0 +1,170 @@
...
+
+#include <unordered_map>
+#include <vector>
+
+namespace cura {
+
+/*! \brief Sparse grid which can locate spatially nearby values efficiently.
+ *
+ * \tparam Val The value type to store.
+ * \tparam PointAccess The functor to get the location from Val.  PointAccess
```
BagelOrb: How about the following type parameter name? `Locator`

src/utils/SparseGrid.h
```diff
@@ -0,0 +1,170 @@
...
+SG_TEMPLATE
+typename SG_THIS::GridPoint SG_THIS::toGridPoint(const Point &point)  const
+{
+    // This mapping via truncation results in the cells with
+    // GridPoint.x==0 being twice as large and similarly for
+    // GridPoint.y==0.  This doesn't cause any incorrect behavior,
+    // just changes the running time slightly.  The change in running
+    // time from this is probably not worth doing a proper floor
+    // operation.
+    return point / m_cell_size;
```
BagelOrb: We had a lil discussion about this and indeed, since most printers have their origin at the side of the printer this will never really be a problem.

src/utils/SparseGrid.h
```diff
@@ -0,0 +1,170 @@
...
+void SG_THIS::addNearby(const Point &query_pt, coord_t radius,
+                        std::vector<Val> &ret) const
+{
+    Point min_loc(query_pt.X-radius, query_pt.Y-radius);
+    Point max_loc(query_pt.X+radius, query_pt.Y+radius);
+
+    GridPoint min_grid = toGridPoint(min_loc);
+    GridPoint max_grid = toGridPoint(max_loc);
+
+    for (coord_t grid_y=min_grid.Y; grid_y<=max_grid.Y; ++grid_y)
```
BagelOrb: One of our code conventions is to put spaces around binary operators...

src/slicer.cpp
```diff
@@ -82,122 +84,305 @@ int SlicerLayer::getNextSegmentIdx(const Mesh* mesh, const SlicerSegment& segmen
...
+
+    struct PossibleStitch
     {
-        PolygonRef open_polyline = open_polylines[open_polyline_idx];
-        
-        if (open_polyline.size() < 1) continue;
-        for(unsigned int open_polyline_other_idx = 0; open_polyline_other_idx < open_polylines.size(); open_polyline_other_idx++)
+        int64_t dist2;
+        // polyline_idx*2 + {0 | front, 1 | back}
+        unsigned int terminus_0_idx;
```
BagelOrb: Why not have a separate boolean member?

This seems to be an optimization which favours small RAM usage over small computation time.
If that's the case I would like to split out this member into two members for readability.

src/slicer.cpp
```diff
@@ -82,122 +84,305 @@ int SlicerLayer::getNextSegmentIdx(const Mesh* mesh, const SlicerSegment& segmen
...
+        bool in_order() const
         {
-            PolygonRef open_polyline_other = open_polylines[open_polyline_other_idx];
-            
-            if (open_polyline_other.size() < 1) continue;
-            
-            Point diff = open_polyline.back() - open_polyline_other[0];
+            // in order if using back of line 1 and front of line 2
+            return (terminus_0_idx & 1) &&
+                !(terminus_1_idx & 1);
```
BagelOrb: Wouldn't a stitch also be in order if using front of line 1 and back of line 2?

src/slicer.cpp
```diff
@@ -82,122 +84,305 @@ int SlicerLayer::getNextSegmentIdx(const Mesh* mesh, const SlicerSegment& segmen
...
+            Point operator()(const StitchGridVal &val) const
+            {
+                return val.polyline_end;
+            }
+        };
+
+        SparseGrid<StitchGridVal,StitchGridValPointAccess> grid(cell_size);
+
+        // populate grid
+        for(unsigned int polyline_0_idx = 0; polyline_0_idx < open_polylines.size(); polyline_0_idx++)
```
BagelOrb: Why is there a zero in the variable name?

src/slicer.cpp
```diff
@@ -82,122 +84,305 @@ int SlicerLayer::getNextSegmentIdx(const Mesh* mesh, const SlicerSegment& segmen
...
-        bool reversed = false;
+        int64_t max_dist2 = max_dist * max_dist;
+
+        struct StitchGridVal
+        {
+            unsigned int polyline_idx;
+            Point polyline_end;
+        };
+
+        struct StitchGridValPointAccess
```
BagelOrb: Why not have something like
```
[&open_polylines](const StitchGridVal& val) { return open_polylines[val.polyline_idx].back(); }
```
?

src/slicer.cpp
```diff
@@ -82,122 +84,305 @@ int SlicerLayer::getNextSegmentIdx(const Mesh* mesh, const SlicerSegment& segmen
...
-}
+    };
+
+    std::priority_queue<PossibleStitch> stitch_queue;
 
-void SlicerLayer::stitch(Polygons& open_polylines)
-{ // TODO This is an inefficient implementation which can run in O(n^3) time.
-    // below code closes smallest gaps first
-    while(1)
     {
```
BagelOrb: Could you document what the following code block is about? Why is it in a separate block?

src/slicer.cpp
```diff
@@ -82,122 +84,305 @@ int SlicerLayer::getNextSegmentIdx(const Mesh* mesh, const SlicerSegment& segmen
...
+    connectOpenPolylinesImpl(open_polylines, largest_neglected_gap_second_phase, cell_size, allow_reverse);
+}
+
+void SlicerLayer::stitch(Polygons& open_polylines)
+{
+    bool allow_reverse = true;
+    connectOpenPolylinesImpl(open_polylines, max_stitch1, max_stitch1, allow_reverse);
+}
+
+void SlicerLayer::connectOpenPolylinesImpl(Polygons& open_polylines, coord_t max_dist, coord_t cell_size, bool allow_reverse)
```
BagelOrb: This function is about 300 lines. Could you maybe try to cut it up into smaller pieces?

Breaking the code up into bit size chunks and adding some documentation might help me a lot in understanding the code.

src/slicer.cpp
```diff
@@ -82,122 +84,305 @@ int SlicerLayer::getNextSegmentIdx(const Mesh* mesh, const SlicerSegment& segmen
...
+{
+    // below code closes smallest gaps first
+
+    struct PossibleStitch
```
BagelOrb: I think it's better to define this struct as a private class of SlicerLayer, also in light of breaking up this function into smaller pieces.

src/slicer.h
```diff
@@ -115,6 +115,19 @@ class SlicerLayer
...
      */
     void stitch_extensive(Polygons& open_polylines);
+
+private:
+    /*!
+     * Connecting polylines that are not closed yet.
+     *
+     * Any polylines that are closed by this function are added to this->polygons.
+     *
+     * \param[in,out] open_polylines The polylines which couldn't be closed into a loop
```
BagelOrb: Can `open_polylines` overlap? Can a line segment be included in multiple open polylines after this function has been called?

src/slicer.cpp
```diff
@@ -82,122 +84,305 @@ int SlicerLayer::getNextSegmentIdx(const Mesh* mesh, const SlicerSegment& segmen
...
+        PolygonRef polyline_1 = open_polylines[best_polyline_1_idx];
+
+        // append polygons according to plan
+        if (reverse[0])
+        {
+            // reverse polyline_0
+            size_t size_0 = polyline_0.size();
+            for (size_t idx=0U; idx!=size_0/2; ++idx)
+            {
+                std::swap(polyline_0[idx], polyline_0[size_0-1-idx]);
```
BagelOrb: It seems to me that this reordering is unnecessary if `reverse[1]`.

We could keep the ordering reversed and simply append `polyline_0` to `polyline_1` and clear `polyline_0`.

What do you think?

src/slicer.cpp
```diff
@@ -82,122 +84,305 @@ int SlicerLayer::getNextSegmentIdx(const Mesh* mesh, const SlicerSegment& segmen
...
+        {
+            // if we already used this terminus, then this stitch is no longer usable
+            continue;
         }
-        else
-        { // connect two polylines
-            if (reversed)
+        unsigned int old_terminus_1_idx = next_stitch.terminus_1_idx;
+        unsigned int terminus_1_idx = terminus_old_to_cur_map[old_terminus_1_idx];
+        if (terminus_1_idx == ~0U)
```
BagelOrb: Should be `INVALID_TERMINUS`

src/slicer.cpp
```diff
@@ -671,20 +710,53 @@ Slicer::Slicer(Mesh* mesh, int initial, int thickness, int layer_count, bool kee
...
             
             SlicerSegment s;
+            int end_edge_start_pt = -1;
             if (p0.z < z && p1.z >= z && p2.z >= z)
+            {
                 s = project2D(p0, p2, p1, z);
+                end_edge_start_pt = 0;
+                if (p1.z == z)
+                {
+                    s.endOtherFaces = v1.connected_faces;
```
BagelOrb: Wouldn't it be better to just store an (optional) pointer to the vertex, rather than copying a vector?

The check at line 88 would then be `segment.end_vertex == nullptr` rather than `segment.endOtherFaces.empty()`

Also note that a segment could end in a vertex of a face which is not properly connected to other faces in case of a bad model.

src/slicer.cpp
```diff
@@ -53,32 +53,68 @@ void SlicerLayer::makeBasicPolygonLoop(const Mesh* mesh, Polygons& open_polyline
...
 {
     int next_segment_idx = -1;
-    const MeshFace& face = mesh->faces[segment.faceIndex];
-    for (unsigned int face_edge_idx = 0; face_edge_idx < 3; face_edge_idx++)
-    { // check segments in connected faces
-        decltype(face_idx_to_segment_idx.begin()) it;
-        if (face.connected_face_index[face_edge_idx] > -1 && (it = face_idx_to_segment_idx.find(face.connected_face_index[face_edge_idx])) != face_idx_to_segment_idx.end())
+
+    const std::vector<uint32_t> &faces_to_try = segment.endOtherFaces;
+    bool segment_ended_at_edge = faces_to_try.empty();
```
BagelOrb: Note that it could also be empty because the face was disconnected on that vertex.
That doesn't really seem to be a real performance hit, though, because it will return immediately on line 94.

src/slicer.cpp
```diff
@@ -671,20 +710,53 @@ Slicer::Slicer(Mesh* mesh, int initial, int thickness, int layer_count, bool kee
             if (layer_nr < 0) continue;
             
             SlicerSegment s;
+            int end_edge_start_pt = -1;
```
BagelOrb: I find the naming a bit confusing here. I think this should be `end_edge_index`

src/utils/SparseGrid.h
```diff
@@ -0,0 +1,360 @@
...
+
+    /*! \brief Map from grid locations (GridPoint) to elements (Elem). */
+    GridMap m_grid;
+    /*! \brief Accessor for getting locations from elements. */
+    PointAccess m_point_access;
+    /*! \brief The cell (square) size. */
+    coord_t m_cell_size;
+};
+
+namespace SparseGridImpl {
```
BagelOrb: I'm curious. Why is this in a separate namespace?

src/utils/SparseGrid.h
```diff
@@ -84,25 +86,80 @@ class SparseGrid
...
+    Point operator()(const SparseGridElem<T> &elem)
+    {
+        return elem.point;
+    }
+};
+
+} // namespace SparseGridImpl
+
+template<class Val>
+class SparseGrid : public SparseGridInvasive<SparseGridImpl::SparseGridElem<Val>,
```
BagelOrb: I will move this class to its own file..

src/utils/SparseGrid.h
```diff
@@ -27,8 +27,10 @@ class SparseGridInvasive
      *
      * \param[in] cell_size The size to use for a cell (square) in the grid.
      *    Typical values would be around 0.5-2x of expected query radius.
+     * \param[in] elem_reserve Number of elements to research space for.
+     * \param[in] max_load_factor Maximum average load factor before rehashing.
      */
-    SparseGridInvasive(coord_t cell_size);
+    SparseGridInvasive(coord_t cell_size, size_t elem_reserve=0U, float max_load_factor=1.0f);
```
BagelOrb: Note: I included the max_load_factor because in the (horrible) master implementation of BucketGrid2D I needed a rather high load factor to prevent constant rehashing. I misused the bucket implementation of unordered_map and introduced hash collisions on purpose. I therefore had a high load factor on purpose. 
Now that each cell maps to one container of multiple items, rather than all those items themselves we don't have that problem any more.

tests/utils/SparseGridTest.cpp
```diff
@@ -0,0 +1,270 @@
+//Copyright (c) 2015 Ultimaker B.V.
```
BagelOrb: This diff went horribly wrong :(

It didn't see it as a rename of the file.

I suppose this is just a simple refactor, so I won't review it.

src/utils/SparseGrid.h
```diff
@@ -68,7 +69,7 @@ class SparseGridInvasive
      * \return True if and only if an object has been found within the radius.
      */
     bool getNearest(const Point &query_pt, coord_t radius, Elem &elem_nearest,
-                    std::function<bool(const Elem& elem)> precondition = no_precondition) const;
+                    const std::function<bool(const Elem& elem)> precondition = no_precondition) const;
```
BagelOrb: I don't see a reason this is not passed by const reference, but that was a problem of my original BucketGrid2D...

src/pathOrderOptimizer.cpp
```diff
@@ -188,14 +188,16 @@ void LineOrderOptimizer::optimize()
         int best_line_idx = -1;
         float best_score = std::numeric_limits<float>::infinity(); // distance score for the best next line
 
-        for(unsigned int close_line_poly_idx :  line_bucket_grid.findNearbyObjects(prev_point)) /// check if single-line-polygon is close to last point
+        /// check if single-line-polygon is close to last point
+        for(typename SparseGrid<unsigned int>::Elem close_line_elem :
+                line_bucket_grid.getNearby(prev_point, gridSize))
```
BagelOrb: I think you should just use getNearbyVals here, since you introduced a function for that.

BagelOrb:
I must say that in general I'm very satisfied with your code.
I have learned quite a bit just from looking at your code!

However the lack of documentation makes it quite a hard read. I'm not referring to doxygen documentation, but implementation explanation.

scottlenser:
I'll probably go through your comments and implement the feedback on Friday.

src/utils/SparseGrid.h
scottlenser: It looks like most of the code uses the same convention for member variables and local variables.  I think it is important to distinguish between the two because member variables have a much larger scope and extra care needs to used when accessing them because they could affect other functions.  Without a naming convention difference, the only way to figure out if a variable assignment leaves the scope of the local function is to lookup the variable declaration for the variable being assigned to.  That makes it so assignments to member variables aren't obvious in the code.

src/utils/SparseGrid.h
scottlenser: Ah.  Habit.  I can change it if you want.  I started doing that a long time ago so that things like:

    int *p, x, y;

Look more similar to what they do.  I pretty much never declare multiple variables on the same line, but the habit stuck.  Let me know if you want me to bother to change it.

src/utils/SparseGrid.h
scottlenser: Hmm, I do think that is a better name.  I'll change it.

src/slicer.cpp
scottlenser: I honestly mostly did it because it made it easier to swap than a separate variable.  I think I should have made a small class instead.  I'll look at changing it.

src/slicer.cpp
scottlenser: If polyline 0 is the polyline for terminus_0_idx and polyline 1 is the polyline for terminus_1_idx, then this class represents the possibility of appending polyline 1 starting at terminus_1_idx onto the terminus_0_idx end of polyline 0.  For example, if the polylines are (A -> B) and (C -> D) and terminus 0 is B and terminus 1 is C, then this represents the possibility A -> B -> C -> D.  With the same polylines, if terminus 0 is C and terminus 1 is B, then this represents the possibility of C -> D -> B -> A.  The class encodes which direction to append which is why this test doesn't consider the other order.  If the other order is relevant, there will be another StitchPossibility instance to represent the other order.  in_order() basically means that neither segment needs to be reversed to do the stitch.

src/slicer.cpp
scottlenser: The implementation uses 0 for the polyline that is the first part of the stitched polyline and 1 for the polyline that is the second part of the stitched polyline.  This is the nominal order although when reversing segments it can use the other order.

src/slicer.cpp
scottlenser: It could have been done that way.  I feel like this way avoids repeating the array indexing multiple times at a trivial memory cost.

src/slicer.cpp
scottlenser: It's in a separate block only because it is a separate logical step.  I think I need to make some private functions for the various steps.  I also need to add some comments.  Basically, this block is creating the set of possible stitches (in terms of the end points involved).

src/slicer.cpp
scottlenser: I agree, it is too big.  I'll split it up some and add some documentation.

src/slicer.h
scottlenser: What do you mean by overlap?  All joins that meet the thresholds will be performed.  For the second question, it never creates segments and a segment can only be used once.

src/slicer.cpp
scottlenser: It is not possible for both reverse[0] and reverse[1] to be true.  In that case, it simply swaps 0 and 1 in the preceding code that plans how to append polygons.

src/slicer.cpp
scottlenser: Yes it should. :)

src/slicer.cpp
scottlenser: I'll have to look at it more, but my initial reaction is yes.  I'll look into changing it.

src/utils/SparseGrid.h
scottlenser: I wanted to make it private, but didn't have a good way to do that.  Putting it in a namespace gave me a way to at least discourage use from outside the module.

src/utils/SparseGrid.h
scottlenser: It could still be potentially useful so I left it in.  If because of the input data you are expecting to have pairs of points near each other, it might be useful to set the load factor as 2.0 as otherwise about half the buckets will be empty.  At least if I understand the unordered_map implementation well enough.

src/utils/SparseGrid.h
scottlenser: I probably just forget to add the reference symbol.  I'll make it a reference.

src/pathOrderOptimizer.cpp
scottlenser: Yep.  I'll change it.

scottlenser:
I think I'm done incorporating the feedback.  I also found a bug in the simplification process and another bit that was sub-optimal.  Please check over the changes when you get a chance.

src/utils/SparseGrid.h
BagelOrb: Oh dear, I didn't know that line would be interpreted that way. That's horrible!

Perhaps we should forbid one-line multiple declarations.

Anyway the code base buts the ampersand and start with the type, so it would be nice to keep that consistent. If you don't change it I will...

src/slicer.h
```diff
@@ -136,7 +136,11 @@ class SlicerLayer
     /*!
      * Connecting polylines that are not closed yet.
      *
-     * Any polylines that are closed by this function are added to this->polygons.
+     * Any polylines that are closed by this function are added to
+     * this->polygons.  All possible polyline joins that meet the
+     * distance and reversal criteria will be performed.  This
+     * function will not introduce any copies of the same polyline
+     * segment.
```
BagelOrb: Just to be clear: I was asking for this because there is an option to also print the open_polylines.

`open_polylines` is not just a local temp variable to compute `this->polygons`

src/slicer.h
```diff
@@ -109,12 +116,350 @@ class SlicerLayer
...
+     * polyline in a polyline vector.
+     *
+     * The location records the index in the polyline vector and
+     * whether this is the vertex at the start of the polyline or the
+     * vertex at the end.
+     */
+    class Terminus
+    {
+    public:
+        /*! A representation of Terminus that can be used as an array index. */
```
BagelOrb: I find this explanation a bit unclear. Each integer can be used as an array index.

Is there a particular array in which this is actually indexing?

I thought that the Index type only retained the relative ordering of the polyline indices, but doesn't give any other guarantee. Isn't that how it's meant?

If that's true it would be nice to rename `asIndex(.)` to something like `getOrder(.)`

src/slicer.cpp
```diff
@@ -173,92 +172,101 @@ bool SlicerLayer::PossibleStitch::operator<(const PossibleStitch &other) const
...
-void SlicerLayer::connectOpenPolylinesImpl(Polygons& open_polylines, coord_t max_dist, coord_t cell_size, bool allow_reverse)
+std::priority_queue<SlicerLayer::PossibleStitch>
+SlicerLayer::findPossibleStitches(
+    Polygons& open_polylines, coord_t max_dist, coord_t cell_size, bool allow_reverse)
 {
-    // below code closes smallest gaps first
-
     std::priority_queue<PossibleStitch> stitch_queue;
 
+    int64_t max_dist2 = max_dist * max_dist;
```
BagelOrb: When I know I'm going to commit code which indents differently I first commit with wrong indentation and then make a separate commit for the indentation fix.

Git doesn't understand indentation, which makes for hard reviewing.

Just a little tip ;)

src/slicer.cpp
```diff
@@ -189,28 +189,32 @@ SlicerLayer::findPossibleStitches(
...
         Point operator()(const StitchGridVal &val) const
         {
-            return val.polyline_end;
+            return val.polyline_term_pt;
         }
     };
 
     // Used to find nearby end points within a fixed maximum radius
-    SparseGridInvasive<StitchGridVal,StitchGridValPointAccess> grid(cell_size);
+    SparseGridInvasive<StitchGridVal,StitchGridValPointAccess> grid_ends(cell_size);
```
BagelOrb: Shouldn't this now just be a `SparseGrid` rather than a `SparseGridInvasive`? You've now changed the element type to include the point, which is precisely what SparseGrid does.

src/slicer.h
scottlenser: The Terminus is encoding 2 things.  The first is the index into open_polylines.  This is what is returned by getPolylineIdx().  The second is whether it is referring to the start (polyline[0]) or end (polyline.back()) of the polyline.

No, that's not quite correct.  asIndex() is used to index the std::vector's which implement the maps between old and current terminus.  The Index type is non-negative, contiguous, and has a well defined past the end index.  It doesn't strictly need to be contiguous, but it needs to be mostly so to be efficient.  It doesn't matter that the Index type keeps the same ordering as the polyline indices.

BagelOrb: Ok, so there is a particular array into which this is indexing: `TerminusTrackingMap::m_terminus_old_to_cur_map` and `TerminusTrackingMap::m_terminus_cur_to_old_map`

Saying that the Index can be used as "an array index" is rather non-specific. Since you expose the `Index` functionality specially to be used with a TerminusTrackingMap, it would be nice to document that here.

src/slicer.cpp
```diff
@@ -51,153 +52,488 @@ void SlicerLayer::makeBasicPolygonLoop(const Mesh* mesh, Polygons& open_polyline
...
-            
-            if (open_polyline_other.size() < 1) continue;
-            
-            Point diff = open_polyline.back() - open_polyline_other[0];
+            return val.polyline_term_pt;
+        }
+    };
+
+    // Used to find nearby end points within a fixed maximum radius
+    SparseGridInvasive<StitchGridVal,StitchGridValLocator> grid_ends(cell_size);
```
BagelOrb: Shouldn't this now just be a `SparseGrid` rather than a `SparseGridInvasive`?
You've now changed the element type to include the point, which is precisely what SparseGrid does.

BagelOrb:
In general I am quite happy with the changes you've made. The quality of slicer.cpp is miles, no kilometers better than the current version on master.

Thanks for more documentation!

src/slicer.cpp
scottlenser: I usually pass the "-b" option to select the git diff formatting if there are a lot of whitespace changes.  If you know the commit id, then "git show -b 1a8b4c6655af66b9954a96148e2383e673148c7f" shows the diff treating all whitespace sequences with 1 or more spaces as equivalent.  It also works with "git log -p -b" and the regular "git diff" for changes in the local working copy.

src/slicer.cpp
scottlenser: It could be, but this lets me use a slightly more meaningful name.  It also gives a convenient place to add documentation.

src/slicer.h
scottlenser: I'll see if I can make it a little bit clearer.  I'm having trouble coming up with good descriptions for some of the concepts even though they are clear in my head.  I think it is because I am thinking of them spatially and so don't have a ready made verbal representation.

BagelOrb: I can fully understand that. In these cases it's hard to come up with good
names.

Op 25 jul. 2016 15:53 schreef "scottlenser" <notifications@github.com>:

src/slicer.cpp
BagelOrb: I think that we can do away with the invasive variant if you would retractor it into a SparseGrid.

The thing is that I'm thinking about implementing a sparse grid which works on objects which can cover multiple cells rather than having one location. It makes sense to add another layer of template classes to the inheritance tree for that. I worry that having three SparseGrid-like classes would clutter understanding of the code. I will keep the sparsegridinvasive for the moment, though, cause it is more efficient for some cases.

Also the code here would be easier to understand once you use SparseGrid because you don't need to define the stitchgridval or the locator.

src/utils/SparseGrid.h
Ghostkeeper: We knew and considered that use case when defining that code style standard, but we decided it was not common enough to sacrifice the logic of the pointer/reference belonging to the type rather than the variable name. The world is just not agreeing with us in this case :wink: We'd write the multiple declaration case like this:

    int* p, * x, y, z;

src/slicer.cpp
scottlenser: Feel free to change it if you like.  I consider this a neutral change as opposed to a bug fix or clear improvement.  I can understand why you might want something like a sparse grid that works on objects with non-zero size and why you might not want 3 similar classes.

BagelOrb:
We noticed some bugs in the code.

You have a tendency not to initialize member variables of helper classes when the object is constructed.

This caused crashes when SlicerSegment::endVertex wasn't initialized to nullptr in the constructor.

More specifically: you set the endVertex to nullptr for a SlicerSegment s, but let s be overwritten so that endVertex wasn't nullptr any more.

We've fixed this in https://github.com/Ultimaker/CuraEngine/pull/374

PR #371, opened 2016-07-08
Misc cleanups

scottlenser:

This branch has only the misc cleanups that are not required for the slicer speedup to function.

Here are the changes (in order of most likely to be wanted):

- More const correctness in places
- Fixed logging of number of layers
- Added error message when it can't find an inherited JSON file.
- Aligned bridge direction with calculated direction.
        The bridging is still terrible, but this at least gives it a chance of working.
- Removed sending of extra SlicingFinished message
        This was being sent in 2 places which was confusing my version of the front end
- Rounding during interpolation when calculating vertices from slicing
- Updated command_line_settings JSON file to version 2 format to match expected format of other JSON files.
- Printouts of various timings from slicing

### DIFF
--- a/resources/command_line_settings.def.json
+++ b/resources/command_line_settings.def.json
@@ -1,40 +1,50 @@
 {
-    "version": 1, 
+    "version": 2,
     "name": "Command line setting defaults CuraEngine",
-    "author": "Ultimaker B.V.",
-    "categories": {
+    "metadata":
+    {
+        "author": "Ultimaker B.V."
+    },
+    "settings": {
         "command_line_settings": {
             "label": "Command Line Settings",
-            "settings": {
+            "type": "category",
+            "children": {
                 "center_object": {
                     "description": "Whether to center the object on the middle of the build platform (0,0), instead of using the coordinate system in which the object was saved.",
-                    "type": "boolean",
-                    "default": false
+                    "type": "bool",
+                    "label": "Center object",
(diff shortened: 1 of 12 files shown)


src/FffGcodeWriter.cpp
```diff
@@ -778,6 +778,7 @@ void FffGcodeWriter::processSkin(GCodePlanner& gcode_layer, SliceMeshStorage* me
         if (bridge > -1)
         {
             pattern = EFillMethod::LINES;
+            skin_angle = bridge;
```
BagelOrb: Thank you very much.
We didn't think it was a bug; we thought the bridging algorithm was just really poor.
It's a good thing you noticed this!

resources/command_line_settings.def.json
```diff
@@ -1,40 +1,50 @@
...
+                    "label": "Mesh position z",
+                    "default_value": 0
                 },
                 "prime_tower_dir_outward": {
                     "description": "Whether to start printing in the middle of the prime tower and end up at the perimeter, or the other way around. This is only used for certain types of prime tower.",
-                    "type": "boolean",
-                    "default": false
+                    "type": "bool",
+                    "label": "Prime tower direction outward",
+                    "default_value": false
```
BagelOrb: Thank you very much.
This didn't have high enough priority because command-line slicing isn't really a typical user scenario.

PR #372, opened 2016-07-08
Added convexHull function for Polygons.

scottlenser:

This branch has only the addition of the convex hull function, convexHull().  The old function was renamed to approxConvexHull().

### DIFF
--- a/src/utils/polygon.cpp
+++ b/src/utils/polygon.cpp
@@ -156,6 +156,89 @@ Polygons PolygonRef::offset(int distance, ClipperLib::JoinType joinType, double 
     return ret;
 }
 
+Polygon Polygons::convexHull() const
+{
+    // Implements Andrew's monotone chain convex hull algorithm
+    // See https://en.wikibooks.org/wiki/Algorithm_Implementation/Geometry/Convex_hull/Monotone_chain
+
+    using IntPoint = ClipperLib::IntPoint;
+
+    std::vector<IntPoint> all_points;
+    for (const ClipperLib::Path &path : paths)
+    {
+        // Reserve big enough for result, but also maintain O(1)
+        // amortized resize cost
+        size_t reserve_size = std::max(all_points.size() + path.size(),
+                                       2*all_points.size());
+        all_points.reserve(reserve_size);
+        for (const IntPoint &point : path)
+        {
+            all_points.push_back(point);
+        }
+    }
+
+    struct HullSort
+    {
+        bool operator()(const IntPoint &a,
+                        const IntPoint &b)
+        {
+            return (a.X < b.X) ||
+                (a.X == b.X && a.Y < b.Y);
+        }
+    };
+
+    std::sort(all_points.begin(), all_points.end(), HullSort());
+
+    // positive for left turn, 0 for straight, negative for right turn
+    auto ccw = [](const IntPoint &p0, const IntPoint &p1, const IntPoint &p2) -> int64_t {
+        IntPoint v01(p1.X - p0.X, p1.Y - p0.Y);
+        IntPoint v12(p2.X - p1.X, p2.Y - p1.Y);
+
+        return
+            static_cast<int64_t>(v01.X) * v12.Y -
+            static_cast<int64_t>(v01.Y) * v12.X;
+    };
+
+    size_t num_points = all_points.size();
+
+    Polygon hull_poly;
+    ClipperLib::Path &hull_points = *hull_poly;
+    hull_points.resize(2*num_points);
+    // index to insert next hull point, also number of valid hull points
+    size_t hull_idx = 0;
+
+    // Build lower hull
+    for (size_t pt_idx=0U; pt_idx!=num_points; ++pt_idx)
+    {
+        while (hull_idx >= 2 &&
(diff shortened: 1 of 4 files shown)


src/utils/polygon.cpp
```diff
@@ -156,6 +156,89 @@ Polygons PolygonRef::offset(int distance, ClipperLib::JoinType joinType, double
...
+
+    using IntPoint = ClipperLib::IntPoint;
+
+    std::vector<IntPoint> all_points;
+    for (const ClipperLib::Path &path : paths)
+    {
+        // Reserve big enough for result, but also maintain O(1)
+        // amortized resize cost
+        size_t reserve_size = std::max(all_points.size() + path.size(),
+                                       2*all_points.size());
```
BagelOrb: I think std::vector already does that. http://www.cplusplus.com/reference/vector/vector/reserve/
"the function causes the container to reallocate its storage increasing its capacity to n (or greater)."

Why don't you just reserve once and compute the amount of points needed in a separate loop?
```
reserve_size = 0;
for (Path path : paths)
    reserve_size += path.size();
all_points.reserve(reserve_size);
```

nallath: You could even use  std::vector<IntPoint> all_points(num_points_to_reserve);

src/utils/polygon.cpp
```diff
@@ -156,6 +156,89 @@ Polygons PolygonRef::offset(int distance, ClipperLib::JoinType joinType, double
...
+        return
+            static_cast<int64_t>(v01.X) * v12.Y -
+            static_cast<int64_t>(v01.Y) * v12.X;
+    };
+
+    size_t num_points = all_points.size();
+
+    Polygon hull_poly;
+    ClipperLib::Path &hull_points = *hull_poly;
+    hull_points.resize(2*num_points);
```
BagelOrb: Why `2 *`? The convex hull can only contain less points, right?

src/utils/polygon.cpp
scottlenser: I didn't think of the separate loop to compute the size.  I like that solution.  I'll change it.

src/utils/polygon.cpp
scottlenser: I had just copied that from the wikipedia article and hadn't really thought about it that much.  Thinking more about it, I think that is just a really loose upper bound.  I think it only needs "num_points + 1".  I'll change it.

scottlenser:
I cleaned up the things you noted.

src/utils/polygon.cpp
BagelOrb: I actually think the size is correct.

When you have points like this:
```
'   :
    :
.   :
```
The lower part of the hull would include all points on the right and the upper part of the hull would also include all those points until it reaches the last point in the left corner.

I wouldn't change it after all. (That's why I deleted my comment ;)

scottlenser: I don't think your example case produces more than "num_points + 1" points.  This convex hull implementation does not include points that are colinear on the hull due to == being in the decisions on backing out a point.  So when producing the lower hull, it would end up with the lower left point, the lower right point, and the upper right point only.  All the intermediate points on the right middle wouldn't be in the hull.  Similarly, when constructing the upper hull, it will only have the upper right point and the lowest point it has reached so far until it gets to the upper left point.  It never has more than 5 points in the convex hull I think for the case you mentioned.

src/utils/polygon.cpp
```diff
@@ -156,6 +156,89 @@ Polygons PolygonRef::offset(int distance, ClipperLib::JoinType joinType, double
...
+    // index to insert next hull point, also number of valid hull points
+    size_t hull_idx = 0;
+
+    // Build lower hull
+    for (size_t pt_idx=0U; pt_idx!=num_points; ++pt_idx)
```
Ghostkeeper: Codestyle: Spaces around binary operators.

src/utils/polygon.cpp
```diff
@@ -156,6 +156,92 @@ Polygons PolygonRef::offset(int distance, ClipperLib::JoinType joinType, double
...
+    using IntPoint = ClipperLib::IntPoint;
+
+    size_t num_points = 0U;
+    for (const ClipperLib::Path &path : paths)
+    {
+        num_points += path.size();
+    }
+
+    std::vector<IntPoint> all_points;
+    all_points.reserve(num_points);
```
BagelOrb: In the comments of previous commits we've discussed that this should be `num_points + 1`

Consider the case
```
.   .
.   .
```

In an intermediate step it would than have one of the four points twice, resulting in 5 points, which is one more than there are in the original poly.

scottlenser: all_points is just the points from all the polygons concatenated, so it is always exactly num_points in size.  It is only there because we need to sort the points from the original polygons.  The hull_points needs to be of size "num_points + 1" and it is.

BagelOrb: Ah yes of course

BagelOrb:
I'm sorry for the late response.

We had a deadline yesterday, so I had some other matters to attend to.

The code looks all right. Thanks for this!

PR #374, opened 2016-07-27
Properly initialize member variables of classes in slicer.h

awhiemstra:
I get crashes because SlicerSegment::endVertex is used uninitialized.

### DIFF
--- a/src/slicer.h
+++ b/src/slicer.h
@@ -31,15 +31,15 @@ class ClosePolygonResult
 public:
     Point intersectionPoint;
     int polygonIdx = -1;
-    unsigned int pointIdx = 0;
+    unsigned int pointIdx = -1;
 };
 class GapCloserResult
 {
 public:
     int64_t len = -1;
     int polygonIdx = -1;
-    unsigned int pointIdxA = 0;
-    unsigned int pointIdxB = 0;
+    unsigned int pointIdxA = -1;
+    unsigned int pointIdxB = -1;
     bool AtoB = false;
 };
 
@@ -235,7 +235,7 @@ private:
          *
          * The polyline_idx and end flags are calculated from this on demand.
          */
-        Index m_idx = 0;
+        Index m_idx = -1;
     };
 
     /*!
--- a/src/slicer.h
+++ b/src/slicer.h
@@ -16,13 +16,13 @@ class SlicerSegment
 {
 public:
     Point start, end;
-    int faceIndex;
+    int faceIndex = -1;
     // The index of the other face connected via the edge that created end
-    int endOtherFaceIdx;
+    int endOtherFaceIdx = -1;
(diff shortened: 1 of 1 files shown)


daid:
Note that this could introduce a slight slowdown in the slicer, as a lot of these segments are created and filled with data. Now the data is set twice. (Do not know how well the optimizer deals with this and how bad it is). Wouldn't it be better to have a constructor to force the creator to set the data?

BagelOrb:
If it's used uninitialised then initialising it arbitrarily won't solve the underlying problem.
That's pulling a Javier: it won't give a crash any more, but bad output instead, which is even harder to trace.

Where was it used without being initialised? Which model did you slice?
Could you maybe make an issue for this?

Ghostkeeper:
I'd think GCC and MinGW would not optimize out the double initialisation if they cannot verify that a member is always overwritten (and apparently they can't, since the value was sometimes used uninitialised). I think the slight slowdown is a very minor concern in comparison with using uninitialised values though.

Currently it seems like the `SlicerSegment`s are constructed first and then filled with values: https://github.com/Ultimaker/CuraEngine/blob/c0d59627eb6c3e40db7e3c0a19fe2d0145d9a75a/src/slicer.cpp#L827 . But not all values. And `endVertex` is set before possibly getting overwritten by the `project2D` calls. I'd say that is where the real problem lies, and we need to guarantee that all values of `SlicerSegment` are set properly at least by the end of that function. A constructor seems like a good way to do this. I haven't looked into the other two classes that awhiemstra initialised.

I think initialising to a value we can easily recognise makes it easier to find bugs than leaving it uninitialised, especially since we have four platforms to support and these things might work fine and dandy with Linux' and Windows' memory management but break on OSX or Win32. As a bonus, initialising them makes it easier to find _other_ uninitialised values as well with a memory checker.

daid:
But I do agree with BagelOrb here. Initializing it to "valid but wrong" values will give subtle bugs which will be extremely hard to find&fix. It's better to initialize them on invalid values then, so you always crash, and have a better debugging scenario.

BagelOrb:
I'm merging this PR as per request of Jasper. I'm not supposed to put work into this and we should just see whether this merge makes things better.

I'm going to change all all initializations of valid values to invalid values (unsigned int x = ~0U i.e.1111111111...)

BagelOrb:
Note that boolean values cannot be initialized to invalid values.

awhiemstra:
> It's better to initialize them on invalid values then, so you always crash, and have a better debugging scenario.

That's what I thought I did. But I guess opinions on what "invalid values" are differ.

BagelOrb:
0 is a valid value for an unsigned int which is used to refer to an index in an array!

-1 is an invalid array index and will very likely result in a crash when an array is indexed with it.

I don't think this is a matter of opinion, but just you making a small mistake.

PR #376, opened 2016-08-04
MergeInfillLines changes

markwal:
- Add setting to enable/disable automatic speed adjustment
- Add setting to remove hardcoded max speed
- Retrieve machine nozzle size from GCodeExport instead of hardcode 400 microns
- Moved some settings retrieval out of a nested loop
- Use the move_path's speed and travelSpeedFactor for the move path merge
- Use the last extrusion path's speed and extrudeSpeedFactor as the base speed for the extrusion merged path

Prior to this change it used the first of four path's (always a travel move) speed and factor as the base speed for the merged paths (regardless of whether they were extrusions or travel).

### DIFF
--- a/src/gcodePlanner.cpp
+++ b/src/gcodePlanner.cpp
@@ -623,6 +623,7 @@ void GCodePlanner::writeGCode(GCodeExport& gcode)
     bool jerk_enabled = storage.getSettingBoolean("jerk_enabled");
     bool speed_for_pressure_enabled = storage.getSettingBoolean("speed_for_pressure_enabled");
     double speed_pressure_maximum = storage.getSettingInMillimetersPerSecond("speed_pressure_maximum");
+    int64_t nozzle_size = gcode.getNozzleSize(extruder);
 
     for(unsigned int extruder_plan_idx = 0; extruder_plan_idx < extruder_plans.size(); extruder_plan_idx++)
     {
@@ -656,8 +657,6 @@ void GCodePlanner::writeGCode(GCodeExport& gcode)
         extruder_plan.inserts.sort([](const NozzleTempInsert& a, const NozzleTempInsert& b) -> bool { 
                 return  a.path_idx < b.path_idx; 
             } );
-
-        int64_t nozzle_size = gcode.getNozzleSize(extruder);
 
         for(unsigned int path_idx = 0; path_idx < paths.size(); path_idx++)
         {
--- a/src/gcodePlanner.cpp
+++ b/src/gcodePlanner.cpp
@@ -621,9 +621,6 @@ void GCodePlanner::writeGCode(GCodeExport& gcode)
     int extruder = gcode.getExtruderNr();
     bool acceleration_enabled = storage.getSettingBoolean("acceleration_enabled");
     bool jerk_enabled = storage.getSettingBoolean("jerk_enabled");
-    bool speed_for_pressure_enabled = storage.getSettingBoolean("speed_for_pressure_enabled");
-    double speed_pressure_maximum = storage.getSettingInMillimetersPerSecond("speed_pressure_maximum");
-    int64_t nozzle_size = gcode.getNozzleSize(extruder);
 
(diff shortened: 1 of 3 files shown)


BagelOrb:
I'm planning to overhaul the merge infill lines completely in the near future, but that might take a while.
The idea is to generate separate small lines before generating the infill lines using the PolygonProximityLinker in the bugfix_overlap_compensation branch.

Meanwhile this PR seems quite nice! Thanks!

BagelOrb:
I coincidentally just had a conversation with a colleague about what the right speeds should be. He argued that because these thinner lines are surrounded by already printed lines the pressure was higher, which would automatically affect the flow. Moreover it might be harder to get the filament into small crevices, which can lead one to think the speed should be slower, rather than higher.

I think I will set your setting to false by default.

src/gcodePlanner.cpp
```diff
@@ -619,6 +619,11 @@ void GCodePlanner::writeGCode(GCodeExport& gcode)
     GCodePathConfig* last_extrusion_config = nullptr; // used to check whether we need to insert a TYPE comment in the gcode.
 
     int extruder = gcode.getExtruderNr();
+    bool acceleration_enabled = storage.getSettingBoolean("acceleration_enabled");
+    bool jerk_enabled = storage.getSettingBoolean("jerk_enabled");
+    bool speed_for_pressure_enabled = storage.getSettingBoolean("speed_for_pressure_enabled");
+    double speed_pressure_maximum = storage.getSettingInMillimetersPerSecond("speed_pressure_maximum");
+    int64_t nozzle_size = gcode.getNozzleSize(extruder);
```
BagelOrb: The extruder changes on line 636 below; you should set the nozzle size there.
On the other hand you might simply inline this variable, or put it back where it was initially (around line 695) because the performance hit is negligible.

BagelOrb:
This PR is accompanied by a Cura PR: https://github.com/Ultimaker/Cura/pull/907

src/gcodePlanner.cpp
```diff
@@ -619,6 +619,11 @@ void GCodePlanner::writeGCode(GCodeExport& gcode)
     GCodePathConfig* last_extrusion_config = nullptr; // used to check whether we need to insert a TYPE comment in the gcode.
 
     int extruder = gcode.getExtruderNr();
+    bool acceleration_enabled = storage.getSettingBoolean("acceleration_enabled");
+    bool jerk_enabled = storage.getSettingBoolean("jerk_enabled");
+    bool speed_for_pressure_enabled = storage.getSettingBoolean("speed_for_pressure_enabled");
```
BagelOrb: I would change this so that this is at least configurable per extruder. Enabling or disabling this might heavily depend on the type of extruder you have and the type of material.

src/gcodePlanner.cpp
```diff
@@ -619,6 +619,11 @@ void GCodePlanner::writeGCode(GCodeExport& gcode)
...
+    double speed_pressure_maximum = storage.getSettingInMillimetersPerSecond("speed_pressure_maximum");
```
BagelOrb: Then this should also be moved down in order to be retrieved from the current extruder train settings base.

markwal:
@BagelOrb I believe I incorporated all of your feedback.  Though I didn't change the default for the setting since you said "I think I will" so I'm leaving that up to you, correct?

BagelOrb:
Looks good!

Did you test your code? Did it work as expected?

markwal:
Yes. And it works nicely both on (extrude speed  instead  of  travel speed) and off for my test models. Though I tested setting it for a single extruder printer only. I don't see any multi extruder definitions in the repo?

BagelOrb:
The UMO dual extrusion definition should be in there (shortly).

Op 4 aug. 2016 21:08 schreef "Mark Walker" <notifications@github.com>:

BagelOrb:
Again: thanks for your contribution! :)

PR #383, opened 2016-08-23
Fixes for Microsoft Visual Studio builds

sevaa:
Now builds with MSVC 2015. Changes are:
- Global define NOMINMAX in CMakeLists.txt
- #ifndef around strings.h, unistd.h, sys/time.h, libgen.h
- Variable-length arrays replaced with alloca() (it's not a C++ feature)
- Homegrown implementation of dirname() using Win32's GetFileFullPath()
- Rogue "or" replaced with ||
- References to macros __WIN32, _WIN32 replaced with WIN32

There will be a HUGE lot of warnings, since the cmake-generated project file specifies the "all warnings" option. Feel free to reduce the warning level to your own liking.

You still need to patch and rebuild protobuf, or you'll get three "unresolved symbol" errors. Functions google::protobuf::internal::GetEmptyString(), GetEmptyStringAlreadyInited() in generated_message_util.h should be declared as non-inline and moved to a source file (generated_message_util.cc is a good place).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,6 +42,10 @@ if(NOT APPLE AND NOT WIN32)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -static-libstdc++")
 endif()
 
+if(WIN32)
+    add_definitions(/DNOMINMAX)
+endif()
+
 include_directories(${CMAKE_CURRENT_BINARY_DIR} libs)
 
 add_library(clipper STATIC libs/clipper/clipper.cpp)
--- a/src/PrimeTower.cpp
+++ b/src/PrimeTower.cpp
@@ -51,7 +51,7 @@ void PrimeTower::computePrimeTowerMax(SliceDataStorage& storage)
         
     extruder_count = storage.getSettingAsCount("machine_extruder_count");
     
-    int max_object_height_per_extruder[extruder_count]; 
+    int *max_object_height_per_extruder = (int*)alloca(sizeof(int)*extruder_count); 
     std::fill_n(max_object_height_per_extruder, extruder_count, -1); // unitialize all as -1
     { // compute max_object_height_per_extruder
         for (SliceMeshStorage& mesh : storage.meshes)
--- a/src/MeshGroup.cpp
+++ b/src/MeshGroup.cpp
@@ -1,6 +1,8 @@
 /** Copyright (C) 2013 David Braam - Released under terms of the AGPLv3 License */
 #include <string.h>
+#ifndef WIN32
 #include <strings.h>
+#endif
 #include <stdio.h>
 
 #include "MeshGroup.h"
--- a/src/Weaver.cpp
+++ b/src/Weaver.cpp
@@ -2,7 +2,9 @@
 
 #include <cmath> // sqrt
 #include <fstream> // debug IO
+#ifndef WIN32
 #include <unistd.h>
+#endif
 
 #include "progress/Progress.h"
 #include "weaveDataStorage.h"
--- a/src/commandSocket.cpp
+++ b/src/commandSocket.cpp
@@ -14,7 +14,7 @@
 
 #include <string> // stoi
 
-#ifdef _WIN32
+#ifdef WIN32
 #include <windows.h>
 #endif
 
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2,7 +2,9 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#ifndef WIN32
 #include <sys/time.h>
+#endif
 #include <signal.h>
 #if defined(__linux__) || (defined(__APPLE__) && defined(__MACH__))
 #include <execinfo.h>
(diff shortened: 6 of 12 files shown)


Ghostkeeper:
Code looks fine to me. And I verified that it still compiles and slices on Ubuntu.

I'd like one of our developers to test this with MinGW on a Windows machine. Perhaps @jackha?

BagelOrb:
Unfortunately this PR ended up on the list of forgotten PRs.

Would you still like to update this PR, so that we can merge it?

sevaa:
I could revisit this. Or you could resolve the conflict manually. All the changes I've done are written down.

BagelOrb:
I am directly contacting colleagues over this. Building on Windows isn't really my cup of tea...

src/PrimeTower.cpp
```diff
@@ -51,7 +51,7 @@ void PrimeTower::computePrimeTowerMax(SliceDataStorage& storage)
         
     extruder_count = storage.getSettingAsCount("machine_extruder_count");
     
-    int max_object_height_per_extruder[extruder_count]; 
+    int *max_object_height_per_extruder = (int*)alloca(sizeof(int)*extruder_count);
```
LipuFei: This function has been refactored. It's now moved to `src/FffPolygonGenerator.cpp`, named as `FffPolygonGenerator::computePrintHeightStatistics`, and it's not necessary to have this change any more.

CMakeLists.txt
```diff
@@ -42,6 +42,10 @@ if(NOT APPLE AND NOT WIN32)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -static-libstdc++")
 endif()
 
+if(WIN32)
```
LipuFei: Perhaps `if(MSVC)` would be better? Otherwise this will also affect MinGW.

LipuFei:
Hi @sevaa, I have tested your branch. Unfortunately it doesn't work on my laptop. I got the ton of warning and it fails with the following messages. Could you check it as well? Thanks.
```
%WORKSPACE%\install\protobuf-3.2.0\msvc14-w64\include\google/protobuf/arena.h(632): note: see declaration of 'google::protobuf::Arena::CreateMaybeMessage'
%WORKSPACE%\repositories\CuraEngine\src\bridge.cpp(62): warning C4710: 'T *google::protobuf::Arena::CreateMaybeMessage<cura::proto::Setting>(google::protobuf::Arena *,...)': function not inlined
        with
        [
            T=cura::proto::Setting
        ]
%WORKSPACE%\install\protobuf-3.2.0\msvc14-w64\include\google/protobuf/arena.h(632): note: see declaration of 'google::protobuf::Arena::CreateMaybeMessage'
%WORKSPACE%\repositories\CuraEngine\src\bridge.cpp(62): warning C4710: 'T *google::protobuf::Arena::CreateMaybeMessage<cura::proto::Object>(google::protobuf::Arena *,...)': function not inlined
        with
        [
            T=cura::proto::Object
        ]
%WORKSPACE%\install\protobuf-3.2.0\msvc14-w64\include\google/protobuf/arena.h(632): note: see declaration of 'google::protobuf::Arena::CreateMaybeMessage'
NMAKE : fatal error U1077: 'C:\PROGRA~2\MICROS~1.0\VC\bin\amd64\cl.exe' : return code '0x2'
Stop.
NMAKE : fatal error U1077: '"C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\BIN\amd64\nmake.exe"' : return code '0x2'
Stop.
NMAKE : fatal error U1077: '"C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\BIN\amd64\nmake.exe"' : return code '0x2'
Stop.
```

BagelOrb:
Hey @LipuFei,

I think you are having problems with protobuf which are unrelated.

Also we were a bit lacking on our side. The PR was from August!
I don't think @sevaa is still working on this issue.
I am revisiting this PR to see whether we can use the code changes and to see what we should do in order to get MSVC to work.

@awhiemstra told me that compiling with MSVC would make the build system a bit more simple, so it might be worth the trouble to fix the problems with this PR ourselves.

LipuFei:
@BagelOrb yes, looks like it too, but I didn't encounter any issue when compiling protobuf with MSVC. Do you have any idea what the problem might be? Could it be some settings in `CMakeLists.txt`?

LipuFei:
@BagelOrb Found out why this afternoon. MSVC doesn't like some C++ and OpenMP `pragma` statements. It will take some time to fix them all.

awhiemstra:
Note that I strongly disagree with the solution to variable-length arrays here. Since varlength arrays are a C99 feature, we should find a C++11 solution instead of coming up with an even uglier C solution. The simple answer is to use a std::vector but there might be solutions with less overhead.

BagelOrb:
Then I'll close this PR. If we really want this I expect a jira ticket to be created by one of our colleagues.

Ghostkeeper:
Well, I had made a ticket, but our product manager just closed it for being inactive for 3 months.

PR #385, opened 2016-08-29
Fix constexpr function warning for C++14

markwal:
Building using gcc 5.3.0 gives a warning for a bunch of the constexpr functions in optional.h because in C++14, constexpr functions will no longer be assumed to be const.  Also, the compiler warns that if statements aren't allowed in constexpr functions in some versions of C++.
 
See https://akrzemi1.wordpress.com/2013/06/20/constexpr-function-is-not-const/ for more info.

### DIFF
--- a/src/utils/optional.h
+++ b/src/utils/optional.h
@@ -114,11 +114,11 @@ public:
         }
         return *this;
     }
-    constexpr T* operator->()
+    constexpr T* operator->() const
     {
         return instance;
     }
-    constexpr T& operator*() &
+    constexpr T& operator*() const&
     {
         return *instance;
     }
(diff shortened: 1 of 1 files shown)


BagelOrb:
Somehow the seemingly simple concept of *const* boggles the steam out of my brain and the fact that *constexpr* in c++11 implies *const*, but c++14 doesn't only adds to my confusion levels.

After some mind grinding my conclusion is that your code changes shouldn't change anything, so it's ok.

If you do understand *constexpr* better than me, can you explain how `optional::value_or(.)` could be *constexpr* ? Its evaluation depends on whether `instance` is a nullptr or not!

markwal:
optional::value_or to be properly used should be passed a particular constant instance.  That is instance is computable at compile time for the particular use.  It shouldn't be called on a variable ptr.

PR #386, opened 2016-09-03
Use Bresenham's line algorithm to emplace the element into the sparse grid

markwal:
Don't merge this (obviously).  It's for discussion.

@BagelOrb 

So here's an attempt to use Bresenham's line algorithm to populate the sparse grid.  This doesn't color as much of the grid as the algorithm it replaces since that one appears to do the full triangle that is the left/lower half of the grid rectangle containing the line.  Is that intentional?

It seems to work better for printers with (0,0) at the center than the algorithm it replaces.

My first rough implementation assumed it should start the error half way through the cycle (just simple div by 2) and it appeared to "work" as well.  I don't know how precise this needs to be at emplacing the element, but this current attempt tries to pick the "pixels" that actually contain the line, but errs on coloring too many pixels, not sure that's correct for the purpose.  Do you have a moment to attempt to enlighten me?

### DIFF
--- a/src/utils/SparseLineGrid.h
+++ b/src/utils/SparseLineGrid.h
@@ -10,6 +10,7 @@
 #include "intpoint.h"
 #include "SparseGrid.h"
 #include "SVG.h" // debug
+#include "logoutput.h"
 
 namespace cura {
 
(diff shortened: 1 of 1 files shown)


BagelOrb:
I think you misunderstood my algorithm. It colours every pixel through
which the line goes. Bresenhams algorithm colours less than all pixels
through which the line goes so that the line will look thinner.
See http://i.stack.imgur.com/j19KD.png

Maybe this post is helpful:
http://stackoverflow.com/questions/24679963/precise-subpixel-line-drawing-algorithm-rasterization-algorithm

I couldn't find an algorithm which did 'my algorithm' in the image above,
and i thought it would be faster to just think of it myself. Figures that I
made some mistake somewhere.

The sparse grid classes have some debugging functions which produce an svg
in an HTML, so u can look which cells are actually included.

Op 3 sep. 2016 10:55 schreef "Mark Walker" <notifications@github.com>:

markwal:
Ah.  That is helpful.  Ok, I'll take another tack.

PR #388, opened 2016-09-05
SparseGridLine for negative coordinates

markwal:
@BagelOrb ok, this attempt is with a better understanding of the algorithm and the intent.  Hopefully.

There are a couple of changes here.  For handling the negative coordinate space, there are two things needed: one is that nearest_next_y is one cell away in y_dir when y_dir matches the sign of cell_y (or at y == 0).  In particular, sometimes it needs to be -1, but the max with 0 always defeated.

The other thing is that if the real corresponding_x has a fraction in the negative x, it can end up off by one.  This is because the algorithm always walks left to right and assumes that truncation would always back off, but in the negative space, will advance.  This wouldn't happen if we walked in the direction of the sign of x, but that wouldn't work for lines that cross the line x=0, so there's a bit of messy fixing that off by one.

But...  having to compute whether there is a modulo makes it not that hard to also pick up the case where the line crosses precisely at the cell corner, so I did that too.  This fix is in two places, one when computing x_cell_end and then again when deciding whether to advance x_cell_start by one.

### DIFF
--- a/src/utils/SparseLineGrid.h
+++ b/src/utils/SparseLineGrid.h
@@ -86,20 +86,16 @@ void SGI_THIS::insert(const Elem &elem)
         // nearest y coordinate of the cells in the next row
         coord_t nearest_next_y = SparseGrid<ElemT>::toLowerCoord(cell_y + ((nonzero_sign(cell_y) == y_dir || cell_y == 0) ? y_dir : coord_t(0)));
         grid_coord_t x_cell_end; // the X coord of the last cell to include from this row
-        coord_t corresponding_x; // the x coordinate corresponding to nearest_next_y
-        bool is_corresponding_x_whole; // does the real corresponding x not have a fractional portion
         if (y_diff == 0)
         {
-            is_corresponding_x_whole = false;
-            corresponding_x = end.X;
             x_cell_end = end_cell.X;
         }
         else
         {
             coord_t area = (end.X - start.X) * (nearest_next_y - start.Y);
-            is_corresponding_x_whole = ((area % y_diff) == 0);
-            corresponding_x = start.X + area / y_diff;
-            x_cell_end = SparseGrid<ElemT>::toGridCoord(corresponding_x + ((corresponding_x < 0) && !is_corresponding_x_whole));
+            // corresponding_x: the x coordinate corresponding to nearest_next_y
+            coord_t corresponding_x = start.X + area / y_diff;
+            x_cell_end = SparseGrid<ElemT>::toGridCoord(corresponding_x + ((corresponding_x < 0) && ((area % y_diff) != 0)));
             if (x_cell_end < start_cell.X)
             { // process at least one cell!
                 x_cell_end = x_cell_start;
--- a/src/utils/SparseLineGrid.h
+++ b/src/utils/SparseLineGrid.h
@@ -50,7 +50,7 @@ protected:
 
     /*! \brief Accessor for getting locations from elements. */
     Locator m_locator;
-    grid_coord_t sign(grid_coord_t z);
+    grid_coord_t nonzero_sign(grid_coord_t z);
 };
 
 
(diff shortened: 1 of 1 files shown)


markwal:
I think this addresses https://github.com/Ultimaker/Cura/issues/894

BagelOrb:
Reviewing this right now...

src/utils/SparseLineGrid.h
```diff
@@ -50,6 +50,7 @@ class SparseLineGrid : public SparseGrid<ElemT>
 
     /*! \brief Accessor for getting locations from elements. */
     Locator m_locator;
+    grid_coord_t sign(grid_coord_t z);
```
BagelOrb: This would be a nice candidate to go into utils/math.h, but I can also do that.

The general form is:
```
template <typename T> int sgn(T val)
{
    return (T(0) < val) - (val < T(0));
}
```

src/utils/SparseLineGrid.h
```diff
@@ -76,22 +77,29 @@ void SGI_THIS::insert(const Elem &elem)
...
 
     grid_coord_t y_dir = (y_diff >= 0)? 1 : -1;
     grid_coord_t x_cell_start = start_cell.X;
     for (grid_coord_t cell_y = start_cell.Y; cell_y * y_dir <= end_cell.Y * y_dir; cell_y += y_dir)
     { // for all Y from start to end
-        coord_t nearest_next_y = SparseGrid<ElemT>::toLowerCoord(cell_y + std::max(coord_t(0), y_dir)); // nearest y coordinate of the cells in the next row
+        // nearest y coordinate of the cells in the next row
+        coord_t nearest_next_y = SparseGrid<ElemT>::toLowerCoord(cell_y + ((sign(cell_y) == y_dir || cell_y == 0) ? y_dir : coord_t(0)));
         grid_coord_t x_cell_end; // the X coord of the last cell to include from this row
+        coord_t corresponding_x;
```
BagelOrb: `// the x coordinate corresponding to nearest_next_y`

src/utils/SparseLineGrid.h
```diff
@@ -76,22 +77,29 @@ void SGI_THIS::insert(const Elem &elem)
...
+        bool is_corresponding_x_whole;
```
BagelOrb: What is this variable exactly?
Please comment the meaning in the code.

BagelOrb:
A few minor points, but never mind those.

I'm ready to merge this.

Going to test it first...

src/utils/SparseLineGrid.h
markwal: My implementation gives +1 output for 0 input which is more useful for this case, but I think your general form is more common.  So maybe I should call it something else when it doesn't return 0 as the sign of 0?

BagelOrb:
Testing failed.

```
    line_grid.insert(std::make_pair<Point, Point>(Point(160, 190), Point(220, 170)));
```

This line still hits the assert, probably because it ends in a cell corner.

You should run `SparseLineGrid<int,int>::debugTest();` from somewhere.

src/utils/SparseLineGrid.h
BagelOrb: Ah ok. Yes please.
Maybe something like "nonzero_sign" ?

markwal:
Huh. I did... But I just remembered, I had made a flag in it that turned off the assert so I could get all the drawings and I forgot about it when I got all the drawings I was looking at to look right. :-p

markwal:
Dang it.  That one is a regression from trying to remove the TODO about precise corner crossing.  I leave the x_cell_end one short when it hits precisely so the next row can get it and have adjacent cells connected only by their corner.

But perhaps this is wrong since the precise point on the line is actually in that next box, so a -45 degree angle should actually be a stair case like this:

```
XXOO
OXXO
OOXX
```

Because if you ask what cell contains the point on the line that is precisely at the corner you get the upper right box (in the positive quadrant).  So maybe your TODO wasn't actually a TODO.  It's correct to keep those extra cells that overlap on each row, right?  At least for negative slope lines in the positive quadrant.  positive slope 45 degrees that crosses the cell corners correctly looks like this:

OOOX
OOXO
OXOO
XOOO

Because the intersection point is contained in the cell above and to the right of the point.

markwal:
So now it's really only down to two tweaks:

1. nearest_next_y gets advanced by one (+y_dir) when y's sign matches y_dir or when cell_y is 0 (because that row is double wide).
2. going left to right in -x space means that a fractional corresponding_x at the boundary means you need another cell. (in +x you already have the cell to the right because the boundaries are on the left).

Sorry about wasting your time with the test I should have caught.

BagelOrb:
No problem. Everyone makes mistakes ;)

I've also fixed the visualization of cells around the origin.

I've merged your fix to 2.3 and to master.

Thanks for your work!

PR #394, opened 2016-09-27
Remove non-applicable 'evil' clause from rapidJSON license

onitake:
JSON_checker is not part of the distribution, so the license does not apply.
See https://github.com/miloyip/rapidjson/issues/377 for more information.
Also added the clarification from upstream about the license.

### DIFF
--- a/libs/rapidjson/license.txt
+++ b/libs/rapidjson/license.txt
@@ -3,7 +3,7 @@ Tencent is pleased to support the open source community by making RapidJSON avai
 Copyright (C) 2015 THL A29 Limited, a Tencent company, and Milo Yip.  All rights reserved.
 
 If you have downloaded a copy of the RapidJSON binary from Tencent, please note that the RapidJSON binary is licensed under the MIT License.
-If you have downloaded a copy of the RapidJSON source code from Tencent, please note that RapidJSON source code is licensed under the MIT License, except for the third-party components listed below which are subject to different license terms.  Your integration of RapidJSON into your own projects may require compliance with the MIT License, as well as the other licenses applicable to the third-party components included within RapidJSON.
+If you have downloaded a copy of the RapidJSON source code from Tencent, please note that RapidJSON source code is licensed under the MIT License, except for the third-party components listed below which are subject to different license terms.  Your integration of RapidJSON into your own projects may require compliance with the MIT License, as well as the other licenses applicable to the third-party components included within RapidJSON. To avoid the problematic JSON license in your own projects, it's sufficient to exclude the bin/jsonchecker/ directory, as it's the only code under the JSON license.
 A copy of the MIT License is included in this file.
 
 Other dependencies and licenses:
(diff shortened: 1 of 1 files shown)


BagelOrb:
I don't understand. What is evil about it?

Also: if the licence doesn't apply, why does it need to change?

onitake:
Because of this: https://wiki.debian.org/qa.debian.org/jsonevil

Strictly speaking, changing it is not absolutely necessary, but it will confuse people as the license doesn't mention exactly which files it applies to. Oh, and it makes Lintian choke and spit when it sees the phrase: 'The Software shall be used for Good, not Evil.'

nallath:
No legal court ever defined the concept of "evil". As a result, the phrase "use it for Good, not for Evil" means absolutely nothing.
I don't think that changing the licence like this is not even allowed. If he wants to tell us to use it for good (and those are the conditions) we will have to adhere to these (regardless of how meaningless the phrase is in the first place).
Changing the licence just because lintian cant handle it also seems a bit weird imho.

sedwards2009:
onitake is saying that we don't need to have that troublesome "good, not evil" clause anywhere near the license information contained in CuraEngine's copy of rapidJSON. Fixing this is a good idea. Debian in particular are very sensitive about this kind of thing.

The "use for good, not evil" clause was a modification to the MIT license which appeared in one of the original JSON tools. It has become a (bad) running joke in the industry and just gives everyone a headache. Avoid it.

I vote merge.

Ghostkeeper:
> No legal court ever defined the concept of "evil". As a result, the phrase "use it for Good, not for Evil" means absolutely nothing.

Not true. Companies have avoided this license because they cannot guarantee that nobody will use their software for evil. Except IBM, who was given the exclusive rights by Douglas Crockford to use JSON for evil as well.

If it's not applicable to us, I'd say remove it. We aren't allowed to modify a license, but if the license indeed only applies to a part of RapidJSON that we don't distribute, then we should be able to remove that license.

Ghostkeeper:
The license file also says this:
> To avoid the problematic JSON license in your own projects, it's sufficient to exclude the bin/jsonchecker/ directory, as it's the only code under the JSON license.

onitake:
Question: Would it be ok to just pull in the upstream license modification (i.e. the stanza @Ghostkeeper posted) and not modify it any other way?
I think it's perfectly fine to add a Lintian override in this case, and then it would also be clear from the text why the 'evil' JSON license does not apply.

Ghostkeeper:
The clause that I quoted is not included in the license file as it was in the state when we copied the code to CuraEngine. We have to keep to the license as it was in the version that we included.

I think the best course of action for us right now is to update RapidJSON, so that we get their new license, can remove the JSONChecker part of the license and our users are allowed to use Cura for evil.

Ghostkeeper:
I've discussed this with others and higher up. We're not able to modify the license as it is now until we update RapidJSON to a version that has the aforementioned clause. This update brings some risk with it, so we're going to update it later. It's currently planned for release 2.5.

onitake:
Fair enough. Even if it's not 100% obvious from the license, the fact that the json.org test scripts are missing from the source tree make it clear enough that the section in question does not apply.
I'll add a Lintian override for now and see what the Debian devs think.
Thanks for checking!

Ghostkeeper:
This just got shot down by our product owner because it's been inactive for a few months.

PR #396, opened 2016-10-05
Allow firmware retract on machines without requiring Volumetric

cbxbiker61:
Marlin is quite fine doing firmware retractions without being in volumetric mode.  This allows one to use this by setting:

```
 "machine_settings": {
    "machine_firmware_retract": { "default": true }
}
```

### DIFF
--- a/src/gcodeExport.cpp
+++ b/src/gcodeExport.cpp
@@ -14,6 +14,8 @@ namespace cura {
 GCodeExport::GCodeExport()
 : output_stream(&std::cout)
 , currentPosition(0,0,MM2INT(20))
+, is_volumatric(false)
+, firmware_retract(false)
 , layer_nr(0)
 {
     *output_stream << std::fixed;
@@ -42,6 +44,7 @@ GCodeExport::~GCodeExport()
 
 void GCodeExport::preSetup(const MeshGroup* meshgroup)
 {
+    firmware_retract = settings->getSettingBoolean("machine_firmware_retract");
     setFlavor(meshgroup->getSettingAsGCodeFlavor("machine_gcode_flavor"));
     use_extruder_offset_to_offset_coords = meshgroup->getSettingBoolean("machine_use_extruder_offset_to_offset_coords");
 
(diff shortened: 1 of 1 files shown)


src/gcodeExport.cpp
```diff
@@ -246,10 +249,6 @@ void GCodeExport::setFlavor(EGCodeFlavor flavor)
     {
         firmware_retract = true;
     }
-    else 
-    {
-        firmware_retract = false;
-    }
```
BagelOrb: I suggest you also remove the conditions which make `firmware_retract` true in the lines above.
If a firmware is volumetric then that should be set accordingly in the machine definition and not be concluded from other parameters.

BFB is probably an exception, because retraction is handled in a really weird way.

Given that the volumetric condition is separated from the firmware type, it seems logical to me that `REPRAP_VOLUMETRIC` would cease to exist, because it would just be `REPRAP` with  `"machine_firmware_retract": false` in the .def.json.

BagelOrb:
I will merge this after the 2.3 release. I don't want to risk introducing an error to the 2.3 release at this stage.

I've been meaning to do this for as long as I can remember, but there is just no value for Ultimaker customers, so it has been far down on my priority list.

There's just a couple of small changes I would like to see before I would merge.

src/gcodeExport.cpp
cbxbiker61: Let me give you some concrete examples of how Marlin behaves.

Firmware retract can be enabled at compile time for a specific machine, but that does not imply volumetric.  Therefore without this patch the extrusion rates are wrong without this patch since currently the only way to enable firmware retract is to switch on volumetric.

Volumetric mode on Marlin is enabled by specifying the filament diameter with, e.g. "M200 D1.75" and turned off with "M200 0" .  I have another patch to Cura that looks at the machine_start_code for an M200... and then switches then switches the machine_gcode_flavor to volumetric which allows CuraEngine to do the right thing.

Bottom line is machine_firmware_retract and volumetric have to be treated independently.

cbxbiker61: I'd say the best solution would be to eliminate the REPRAP volumetric machine type and then have machine_firmware_retract and machine_volumetric in the machine definitions.  The patch to Cura would still be useful for dynamic determination of volumetric mode.

BagelOrb:
Closed due to inactivity.

PR #400, opened 2016-10-11
documentation for settings

Ocarthon:


BagelOrb:
I think the settings documentation is outdated by now.
Cura development is faster than a freight train! :D

Issue #381, opened 2016-08-16
Recursively Sub-divided Cubic Infill


mboerwinkle:
I am working on an infill similar to the Cubic infill but which recursively subdivides each cube based on its distance from the shell. The reasoning behind this infill is that by having a higher infill density near the shell and a lower density in the core, a higher strength to weight/print time/cost ratio can be obtained.
 Currently, the infill uses the distance from the center of the cube to the current layer's bounding polygon as its method of determining if a cube should be subdivided (by using PolygonUtils::moveInside2 and finding the distance to the resulting point). What would be optimal is if it could find the distance from the center of the cube to the shell of the 3d model so that floors and ceilings would also be reinforced.

In short: I am looking to write a function that takes as an argument a 3d point, and returns its distance from the model's shell.

Any guidance in how to cleanly implement this would be greatly appreciated.

Thanks,
Martin Boerwinkle

Attached is a screenshot of the current version of the infill in repetier and the gcode file that the screenshot is taken from (cube.txt).

[cube.txt](https://github.com/Ultimaker/CuraEngine/files/419333/cube.txt)

![screenshot](https://cloud.githubusercontent.com/assets/7738815/17685175/248e1ca8-6351-11e6-8ac7-3c706f33843c.png)

Ghostkeeper:
Looks interesting!

The 3D distance to the boundary would be fairly difficult. Here's the obstacles you're facing:
- The 3D polygon data is discarded after the slicing stage.
- The `PolygonUtils::moveInside2` function is only implemented for 2D polygons (like all other functions in that class. Implementing it for 3D polygons would be an order of magnitude more difficult.
- Computation time.
- Dealing with non-watertight meshes and meshes with extra vertices. CuraEngine's error correction is done at the 2D level, iirc (and then separately for Z).

We currently have an implementation of what's called "Gradual Infill" though, which does something similar but in a different manner. It defines areas in every layer that are close to the ceiling (like it also does for skin) and gives those layer parts a different infill density based on its proximity to the top. The current implementation halves the infill density every 5mm by default. You could try something similar, but it will sort of be the Chebyshev distance to the border rather than Euclidean. Currently, the Gradual Infill implementation is always exponential (always divides the density by half, since that aligns the lines on top of each other nicely) and it only counts from the top.

[...]

PR #401, opened 2016-10-14
Recursively Sub-Divided Cubic Infill Pull Request

mboerwinkle:
This is the pull request.
Thanks,
Martin Boerwinkle


BagelOrb:
Hey Martin,

I'm sorry I haven't taken much time to look at your code yet. We've been quite occupied with launching the Ultimaker 3!

Could you maybe give an outline of how your algorithm works?

I've  been thinking a lot on how I would implement it efficiently, but that lead me to an approach which is bottom-up instead of top down:
- use a 3D variant of utils/SparseGrid to register where faces are. Rotate all data so that the grid aligns with the direction of the cubes.
- For each cube in which faces occur: mark alll sides as sides which need to be printed.
- Inspect bottom-up which other faces of cubes need to be marked as such, by inspecting blocks of cubes.

I think using the SparseGrid data type could lead to a significant performance increase. It works by hashing coordinates and using an unordered_map. That way lookup times aren't logarithmic any more.
Perhaps the SparseGrid datatype could also be used in your top-down approach, but I don't know yet how your algorithm works in global lines.

BagelOrb:
Also see https://github.com/Ultimaker/Cura/tree/feature_recursive_cubic

BagelOrb:
I did some testing and found out that having `sub_div_rad_mult < 1.0` causes no subdivision at all. I didn't really expect that, though it might be as intended. What do you think?

NicholasSeward:
@BagelOrb It seems like the line generation time is about as small as can be expected and is very comparable to regular infill. [ @mboerwinkle gave a few numbers in issue #381 which have since been improved upon] Path planning on the other hand is where the main performance hit is due to the increase in number of line segments which is unavoidable.  Assuming that improving path planning is off the table, the theoretical max performance gain that is left on the table is probably around 5%.

mboerwinkle:
Sorry I took so long to respond. I have accepted your pull request pull request (pull pull request request?). Looks like all good stuff. 
  I also figured out why having sub_div_rad_mult < 1.0 causes problems. The algorithm begins with an arbitrarily large cube that is then subdivided recursively. The recursions happen if the circumscribed sphere of the cube, whose radius is multiplied by sub_div_rad_mult, encloses or touches a part of the model's surface. When the radius multiplier is set below a certain amount, the model actually fits between these now smaller and separated spheres and does not trigger any subdivisions.

mboerwinkle:
Though my algorithm uses a tree structure, it does not do any lookups, so logarithmic lookup times are not a concern. When it draws a specific layer, it does not search in the octree for specific cubes. Instead, it recurses through the octree drawing all of the cubes that exist on the target layer. It ignores branches of the octree that are completely above or below the target layer.

@BagelOrb I am not sure I understand the algorithm you are describing. Does it basically determine the set of smallest-size cubes that intersect the shell and from these generate the set of all bigger cubes that need to exist?

Algorithm Outline:
Precomputation: This happens once (precomputeOctree)
 -Generate Precomputed Values
   Starting at the smallest size cube defined by the line distance until an arbitrary size, various values are precomputed that are common to every cube at a certain recursion depth. (Side Length, distance between opposite corners, etc...)
 -Recursive cube generation (SubDivCube constructor)
  The initial arbitrarily large subdivcube is constructed.
  The constructor checks each of its eight potential subdivisions for touching the shell (isValidSubdivision) and calls the recursive constructor on which of the eight need to be subdivided.

Line Generation: This happens once every layer (generateSubdivisonLines)
  Line generation for a given layer is a recursive operation. The octree represents the set of cubes that are split into their eight subdivisions, not the set of cubes that exist in total. As a result, in order to draw the octree, we do not draw the outline of the cubes, we draw the subdivision lines of the cubes.
  To begin on a specific layer, the subdivision lines for the largest root cube are drawn. Since a cube is divided into eight cubes by three planes, three line segments are generated at the intersections of the three dividing planes with the target draw layer plane.
  For each of the cube's eight children, if the child cube touches the target draw layer plane, then generateSubdivisionLines is called on it.

Does this answer your questions, or were you more looking for info about a specific part of the algorithm?

Thanks,
Martin Boerwinkle

BagelOrb:
Thanks for the explanation! Helped a lot!

I've made some changes to the code and created a new meta-PR ^^

I still don't really understand the `inside_somewhere` vs `outside_somehwere` logic. Can you explain a bit?

Earlier I mentioned a max recursion peth setting. You could easily add it by adding the following to `isValidSubdivision`:
```
if (depth > max_depth)
{
    return true;
}
```

I didn't really process the `generateSubdivisionLines` function yet..

I want to do some profiles to verify the claims of Nicholas.

NicholasSeward:
@BagelOrb As you go through all the layers that intersect a cube you need to decide if it is to be subdivided.  Cubes that are close enough to the shell on any particular layer obviously get subdivided.  However, it is possible to have a cube that is always far from the shell on every layer but still needs to be subdivided.  For instance, a cube close to the top of a print could be completely inside the perimeter or completely outside the perimeter for each layer.  Therefore, a cube that is inside the perimeter completely on one layer and outside the perimeter completely on another layer should trigger a subdivision.

BagelOrb:
Ah yes that makes sense!

Still doing some profiling.
There were some memory leeks in the code which I solved. Checking whether
everything is fixed now.

Op 31 okt. 2016 20:50 schreef "NicholasSeward" <notifications@github.com>:

BagelOrb:
generateSubdivisionLines actually takes up to 5% of the total slice time, and generateCubicSubDivInfill takes 12% of the total slice time!

I've tested with default settings on the stanford dragon... 

There is certainly more optimization which can be done.

The biggest bottleneck is in dinstanceFromPointToMesh, which calls moveInside2 which takes 7.5% of the total slice time!
If we put all points in a 3D SparseGrid I bet we can reduce that to 1%

`generateSubdivisionLines` actually calls `Polygons::remove` which I think smells like a bug.
Also quite some time is spent in `operator[]` which also seems supicious to me.
I don't know whether these bugs were introduced here or merely exposed here, though.

NicholasSeward:
What is the same time for a normal cubic infill?  12% is not nothing but normal infill is also not nothing. :-)

I am not sure how SparseGrid can speed up finding the distance from a point to a Polygon.  Just having efficient lookup of points doesn't help you find the closest point among all the lines of a polygon.  One improvement could be to make PolygonUtils::findClosest use an R-tree or a [Quad-tree](http://donar.umiacs.umd.edu/quadtree/lines/pmr.html).  (It seems like most people prefer the PMR Quadtree over the R-tree approach.)  However, that improvement can be done independent of this infill. 

Should we fix findClosest to be O(log(n)) instead of O(n) for this pull request or independently in the future?

BagelOrb:
With the SparseGrid implementation, lookup times are expected O(1), while building the SparseGrid is O(n). It uses a hash table under the hood.

I am working on an implementation of this, so you don't have to. Please review my PR on your PR; there's some pre-work in there which might be interesting.

I definitely want to have the `dinstanceFromPointToMesh` function optimized more. I don't know whether my SparseGrid adventures will succeed in time, so I might ask you to have a look at that function.

Note that some `moveInside` functions already check whether the point is inside, so the least you could do is to merge the call to `isinside` with the call to `moveInside`.

Ghostkeeper:
I'm assigned to review this once again, to check BagelOrb's code properly and to pass an additional check on the original work of mboerwinkle.

src/infill/subDivCube.cpp
```diff
@@ -0,0 +1,279 @@
...
+    int32_t z_diff = abs(z - center.z); //!< the difference between the cube center and the target layer.
+    if (z_diff > height[depth] / 2) //!< this cube does not touch the target layer. Early exit.
+    {
+        return;
+    }
+    if (z_diff < max_draw_z_diff[depth]) //!< this cube has lines that need to be drawn.
+    {
+        Point relative_a, relative_b; //!< relative coordinates of line endpoints around cube center
+        Point a, b; //!< absolute coordinates of line endpoints
+        relative_a.X = (square_height[depth] / 2) * ((double)(max_draw_z_diff[depth] - z_diff) / (double)max_draw_z_diff[depth]);
```
Ghostkeeper: We should keep these computations in the fixed-point `long long`s, to prevent accumulation of floating point errors. That probably means casting `square_height` and such to `long long`s instead of casting the Z coordinates to `double`s.

src/infill/subDivCube.h
```diff
@@ -0,0 +1,81 @@
...
+    /*!
+     * Precompute the octree of subdivided cubes
+     * \param mesh contains infill layer data and settings
+     */
+    static void precomputeOctree(SliceMeshStorage& mesh);
+    /*!
+     * Generates the lines of subdivision of the specific cube at the specific layer. It recursively calls itself, so it ends up drawing all the subdivision lines of sub-cubes too.
+     * \param z the specified layer height
+     * \param result (output) The resulting lines
+     * \param directional_line_groups Should be nullptr. Used internally to keep track of line segments that are all pointing the same direction for line segment combining
```
Ghostkeeper: The typical way to guard against this is to make a public function that has no `directional_line_groups` parameter, which calls a private recursive function that has the parameter. You won't need all that hassle with the `calloc` and freeing the memory then.

src/infill/subDivCube.cpp
```diff
@@ -0,0 +1,279 @@
...
+{
+    int64_t distance2;
+    long int sphere_slice_radius2;//!< squared radius of bounding sphere slice on target layer
+    bool inside_somewhere = false;
+    bool outside_somewhere = false;
+    int inside;
+    double part_dist;//what percentage of the radius the target layer is away from the center along the z axis. 0 - 1
+    const long int layer_height = mesh.getSettingInMicrons("layer_height");
+    long int bottom_layer = (center.z - radius) / layer_height;
+    long int top_layer = (center.z + radius) / layer_height;
```
Ghostkeeper: A lot of these can be defined inside the for loop, where they are first used. That keeps the code more localised, and makes it immediately clear what it is. It may also help with putting variables on a lower-level register in some compilers.

src/infill/subDivCube.cpp
```diff
@@ -0,0 +1,279 @@
...
+    for (long int test_layer = bottom_layer; test_layer <= top_layer; test_layer += 3) // steps of three. Low-hanging speed gain.
+    {
+        part_dist = (double)(test_layer * layer_height - center.z) / radius;
```
Ghostkeeper: Is this still correct if `infill_sparse_thickness` > `layer_height`?

src/infill/subDivCube.cpp
```diff
@@ -0,0 +1,279 @@
...
+int SubDivCube::distanceFromPointToMesh(SliceMeshStorage& mesh, long int layer_nr, Point& location, int64_t* distance2)
+{
+    if (layer_nr < 0 || (unsigned long int)layer_nr >= mesh.layers.size()) //!< this layer is outside of valid range
+    {
+        return 2;
+    }
+    Polygons collide;
+    mesh.layers[layer_nr].getSecondOrInnermostWalls(collide);
+    Point centerpoint = location;
+    bool inside = collide.inside(centerpoint);
```
Ghostkeeper: If it's inside, you wouldn't need to compute the distance, right? I'd expect `distance2` would become 0 anyway.

src/infill/subDivCube.cpp
```diff
@@ -0,0 +1,279 @@
...
+void SubDivCube::rotatePointInitial(Point& target)
+{
+    target = infill_rotation_matrix.apply(target);
+}
+
+void SubDivCube::rotatePoint120(Point& target)
+{
+    int64_t x;
+    x = (-0.5) * target.X - sqrt_three_fourths * target.Y;
+    target.Y = (-0.5)*target.Y + sqrt_three_fourths * target.X;
```
Ghostkeeper: Could you define `sqrt_three_fourths` in here instead of statically? It's a compile-time variable anyway. Or as a macro. Took some time to figure out where this came from.

CMakeLists.txt
```diff
@@ -84,6 +84,7 @@ set(engine_SRCS # Except main.cpp.
     src/infill/ZigzagConnectorProcessorDisconnectedEndPieces.cpp
     src/infill/ZigzagConnectorProcessorEndPieces.cpp
     src/infill/ZigzagConnectorProcessorNoEndPieces.cpp
+    src/infill/subDivCube.cpp
```
BagelOrb: fixed

src/FffPolygonGenerator.cpp
```diff
@@ -421,6 +421,11 @@ void FffPolygonGenerator::processDerivedWallsSkinInfill(SliceMeshStorage& mesh,
     // create gradual infill areas
     SkinInfillAreaComputation::generateGradualInfill(mesh, mesh.getSettingInMicrons("gradual_infill_step_height"), mesh.getSettingAsCount("gradual_infill_steps"));
 
+    //SubDivCube Pre-compute Octree
+    if(mesh.getSettingAsFillMethod("infill_pattern") == EFillMethod::CUBICSUBDIV){
```
BagelOrb: fixed

src/infill.cpp
```diff
@@ -97,6 +100,34 @@ void Infill::generateTriangleInfill(Polygons& result)
     generateLineInfill(result, line_distance, fill_angle + 120, 0);
 }
 
+void Infill::generateCubicSubDivInfill(Polygons& result, SliceMeshStorage* mesh)
+{
+    if (mesh == nullptr)
+    {
+        logError("Cannot generate Cubic Subdivision infill without a mesh!\n");
+        return;
+    }
```
BagelOrb: fixed

src/infill.cpp
```diff
@@ -97,6 +100,34 @@ void Infill::generateTriangleInfill(Polygons& result)
...
+}
+
+void Infill::addLineSegmentsInfill(Polygons& result, Polygons& input)
+{
+    auto addLine = [&](Point from, Point to)
+    {
+        PolygonRef p = result.newPoly();
+        p.add(from);
+        p.add(to);
+    };
```
BagelOrb: fixed

src/infill.cpp
```diff
@@ -97,6 +100,34 @@ void Infill::generateTriangleInfill(Polygons& result)
...
+    ClipperLib::PolyTree interior_segments_tree = in_outline.lineSegmentIntersection(input);
+    ClipperLib::Paths interior_segments;
+    ClipperLib::OpenPathsFromPolyTree(interior_segments_tree, interior_segments);
+    for(uint64_t idx = 0; idx < interior_segments.size(); idx++){
```
BagelOrb: fixed

src/infill/subDivCube.cpp
```diff
@@ -0,0 +1,279 @@
...
+#include "../sliceDataStorage.h"
+
+namespace cura
+{
+
+std::vector<int64_t> SubDivCube::side_length;
+std::vector<int64_t> SubDivCube::height;
+std::vector<int64_t> SubDivCube::square_height;
+std::vector<int64_t> SubDivCube::max_draw_z_diff;
+std::vector<int64_t> SubDivCube::max_line_offset;
```
BagelOrb: fixed

src/infill/subDivCube.cpp
```diff
@@ -0,0 +1,279 @@
...
+    }
+}
+
+void SubDivCube::precomputeOctree(SliceMeshStorage& mesh)
+{
+    radius_multiplier = mesh.getSettingAsRatio("sub_div_rad_mult");
+    radius_addition = mesh.getSettingInMicrons("sub_div_rad_add");
+    double infill_angle = M_PI / 4.0;
+    int curr_recursion_depth = 0;
+    for (int64_t curr_side_length = mesh.getSettingInMicrons("infill_line_distance") * 2; curr_side_length < 25600000; curr_side_length *= 2) //!< 25600000 is an arbitrarily large number. It is imperative that any infill areas are inside of the cube defined by this number.
```
BagelOrb: fixed

src/infill/subDivCube.cpp
```diff
@@ -0,0 +1,279 @@
...
+    //                 Z                                                        .
+    //                /:\                              Y                        .
+    //               / : \                             ^                        .
+    //              /  :  \                            |                        .
+    //             /  .O.  \                           |                        .
+    //            /.~'   '~.\                          O---->X                  .
+    //          X """"""""""" Y                                                 .
+    double one_over_sqrt_3 = 1.0 / sqrt(3.0);
+    double one_over_sqrt_6 = 1.0 / sqrt(6.0);
+    double sqrt_two_third = sqrt(2.0 / 3.0);
```
BagelOrb: fixed

src/infill/subDivCube.cpp
```diff
@@ -0,0 +1,279 @@
...
+}
+
+void SubDivCube::generateSubdivisionLines(int64_t z, Polygons& result, Polygons** directional_line_groups)
+{
+    auto addLine = [&](Point from, Point to)
+    {
+        PolygonRef p = result.newPoly();
+        p.add(from);
+        p.add(to);
+    };
```
BagelOrb: fixed

src/infill/subDivCube.cpp
```diff
@@ -0,0 +1,279 @@
...
+    /*!
+     * Adds the defined line to the specified polygons. It assumes that the specified polygons are all parallel lines. Combines line segments with touching ends closer than epsilon.
+     * \param group the polygons to add the line to
+     * \param from the first endpoint of the line
+     * \param to the second endpoint of the line
+     */
+    auto addLineAndCombine = [&](Polygons& group, Point from, Point to)
+    {
+        int epsilon = 10;
```
BagelOrb: fixed

src/infill/subDivCube.cpp
```diff
@@ -0,0 +1,279 @@
...
+        for (unsigned int idx = 0; idx < group.size(); idx++)
+        {
+            if (abs(from.X - group[idx][1].X) < epsilon && abs(from.Y - group[idx][1].Y) < epsilon)
+            {
+                from = group[idx][0];
+                group.remove(idx);
```
BagelOrb: Fixed: made polygons:remove constant time complexity

src/infill/subDivCube.cpp
BagelOrb: fixed

src/infill/subDivCube.cpp
BagelOrb: I disagree. putting them inside the for loop makes it less readable and defining them before the for-loop helps the compiler in realizing that those values don't change over each iteration. This makes sure those values aren't recomputed each iteration, rather than leaving it up to the compiler.

src/infill/subDivCube.cpp
BagelOrb: fixed

PR #412, opened 2016-11-16
Feature precool

Ghostkeeper:
Starts cooling the nozzle slightly before going to the extruder switch and starts printing when the nozzle isn't quite at full temperature yet.

Implements CURA-1932.


src/LayerPlanBuffer.h
```diff
@@ -27,12 +27,14 @@ class LayerPlanBuffer : SettingsMessenger
 
     static constexpr const double extra_preheat_time = 1.0; //!< Time to start heating earlier than computed to avoid accummulative discrepancy between actual heating times and computed ones.
 
+    std::vector<bool> extruder_used_in_meshgroup; //!< For each extruder whether it has already been planned once in this meshgroup. This is used to see whether we should heat to the initial_print_temp or to the printing_temperature
```
Ghostkeeper: We now have:
* `ExtruderTrain::getIsUsed`
* `GCodeExport::getExtruderIsUsed`
* `LayerPlanBuffer::extruder_used_in_meshgroup`
* `SliceDataStorage::getExtrudersUsed`

src/Preheat.cpp
```diff
@@ -0,0 +1,160 @@
+#include "Preheat.h"
+
+namespace cura 
+{
+
+void Preheat::setConfig(MeshGroup& settings)
```
Ghostkeeper: It's a bit misleading to call a mesh group `settings`!

src/LayerPlanBuffer.cpp
```diff
@@ -203,8 +351,18 @@ void LayerPlanBuffer::insertPreheatCommands()
             continue;
         }
         
-        double avg_flow = extruder_plan.estimates.getMaterial() / time; // TODO: subtract retracted travel time
-        extruder_plan.required_temp = preheat_config.getTemp(extruder_plan.extruder, avg_flow, extruder_plan.is_initial_layer);
+        double avg_flow = extruder_plan.estimates.getMaterial() / time;
+        extruder_plan.printing_temperature = preheat_config.getTemp(extruder, avg_flow, extruder_plan.is_initial_layer);
```
Ghostkeeper: Eh, shouldn't we only do this if auto-temperature is enabled?

src/Preheat.h
```diff
@@ -153,76 +139,62 @@ class Preheat
...
+    //     ,temp0     /                                     .
+    //      \        /                                      .
+    //       \______/                                       .
+    //               "-> temp1
      * \param window_time The time window within which the cooldown and heat up must take place.
      * \param extruder The extruder used
-     * \param temp The temperature to which to heat
+     * \param temp0 The temperature from which to start cooling down
+     * \param temp1 The temeprature to which we try to cool down
+     * \param temp2 The temperature to which we need to have heated up at the end of the \p time_window
```
Ghostkeeper: Terrible variable names.

src/Preheat.cpp
```diff
@@ -0,0 +1,160 @@
...
+    //                 /                                    .
+    //     ,temp0     /                                     .
+    //      \ ' ' ' '/ ' ' '> outer_temp                    .
+    //       \______/                                       .
+    //               "-> temp1
+    //      ^^^^^^^^^^
+    //      limited_time_window
+    double outer_temp;
+    double limited_time_window;
+    if (temp0 < temp2)
```
Ghostkeeper: Shouldn't this be comparing `temp0` with `temp1`?

I guess that's precisely the problem with naming a variable `temp1`...

nickthetait:
Just curious, by about how much? A few seconds and <= 1 degree C?

src/Preheat.cpp
```diff
@@ -0,0 +1,160 @@
...
+    }
+    else 
+    {
+        result.heating_time += limited_time_window * time_to_heatup_1_degree / (time_to_cooldown_1_degree + time_to_heatup_1_degree);
+        result.lowest_temperature = std::max(temp1, temp2 - result.heating_time / time_to_heatup_1_degree);
+    }
+    return result;
+}
+
+Preheat::CoolDownResult Preheat::timeBeforeEndToInsertPreheatCommand_warmUpCoolDown(double time_window, unsigned int extruder, double temp0, double temp1, double temp2, bool during_printing)
```
Ghostkeeper: `warmUpCoolDown` doesn't describe anything in the context of regulating the temperature.

Ghostkeeper: `warmUpCoolDown` doesn't describe anything in the context of regulating the temperature.

Ghostkeeper: `warmUpCoolDown` doesn't describe anything in the context of regulating the temperature.

src/LayerPlanBuffer.cpp
```diff
@@ -38,41 +38,51 @@ void LayerPlanBuffer::insertPreheatCommand(ExtruderPlan& extruder_plan_before, d
...
     for (unsigned int extruder_plan_before_idx = extruder_plan_idx - 1; int(extruder_plan_before_idx) >= 0; extruder_plan_before_idx--)
     { // find a previous extruder plan where the same extruder is used to see what time this extruder wasn't used
-        ExtruderPlan& extruder_plan = *extruder_plans[extruder_plan_before_idx];
-        if (extruder_plan.extruder == extruder)
+        ExtruderPlan& extruder_plan_before = *extruder_plans[extruder_plan_before_idx];
+        if (extruder_plan_before.extruder == extruder)
         {
-            Preheat::WarmUpResult warm_up = preheat_config.timeBeforeEndToInsertPreheatCommand_coolDownWarmUp(in_between_time, extruder, required_temp);
+            double temp_before = preheat_config.getFinalPrintTemp(extruder);
+            if (temp_before == 0)
```
BagelOrb: 0 means: 'don't use the final_print_temp feature'. This is different than just putting the final print temperature to the same value as the print temperature for the initial layer and for when flow dependent temperatures are used.

BagelOrb: fixed the description..

src/LayerPlanBuffer.cpp
```diff
@@ -162,16 +175,149 @@ void LayerPlanBuffer::insertPreheatCommand(std::vector<ExtruderPlan*>& extruder_
...
+    int extruder = last_extruder_plan.extruder;
+
+    double final_print_temp = preheat_config.getFinalPrintTemp(extruder);
+    if (final_print_temp == 0)
+    {
+        return;
+    }
+
+    double heated_post_travel_time = 0;
+    { // compute heated_post_travel_time
```
BagelOrb: f

src/LayerPlanBuffer.cpp
```diff
@@ -162,16 +175,149 @@ void LayerPlanBuffer::insertPreheatCommand(std::vector<ExtruderPlan*>& extruder_
...
+                break;
+            }
+            heated_post_travel_time += path.estimates.getTotalTime();
+        }
+    }
+
+    double time_window = 0;
+    double average_print_temp = 0;
+    double initial_print_temp = -1;
+    { // compute time window and print temp statistics
```
BagelOrb: f

src/LayerPlanBuffer.cpp
```diff
@@ -162,16 +175,149 @@ void LayerPlanBuffer::insertPreheatCommand(std::vector<ExtruderPlan*>& extruder_
...
+            if (prev_extruder_plan.extruder != extruder)
+            {
+                break;
+            }
+            time_window += prev_extruder_plan.estimates.getTotalTime();
+            heated_pre_travel_time = prev_extruder_plan.heated_pre_travel_time;
+
+            if (prev_extruder_plan.estimates.getTotalUnretractedTime() > 0 && prev_extruder_plan.estimates.getMaterial() > 0)
+            { // handle temp statistics
+                assert(prev_extruder_plan.printing_temperature != -1 && "Previous extruder plan should already have a temperature planned");
```
BagelOrb: -1 is a stub which is used as invalid temperature by programmers.
0 is a stub which is used as 'don't use this temp variable' by users.
So: -1 is a programmer special case, while 0 is a special case which could be initiated by the user.

Should I use -1337 as invalid temp? Then it's not a valid temp anyhow...

Do you think I should introduce extra settings for the 0-option?
I don't like introducing too many settings, but I could introduce a 'Use Initial Printing Temperature' and a 'Use Final Printing Temperature'...

src/LayerPlanBuffer.cpp
```diff
@@ -162,16 +175,149 @@ void LayerPlanBuffer::insertPreheatCommand(std::vector<ExtruderPlan*>& extruder_
...
+                average_print_temp += prev_extruder_plan.printing_temperature;
+                print_temp_count++;
+                initial_print_temp = prev_extruder_plan.initial_printing_temperature;
+            }
+        }
+        time_window -= heated_pre_travel_time + heated_post_travel_time;
+        average_print_temp /= print_temp_count;
```
BagelOrb: f

src/LayerPlanBuffer.cpp
BagelOrb: The flow-temp graph returns the standard printing_temperature if flow based temp is disabled....
So the preheat_config already takes care of this.

src/LayerPlanBuffer.h
```diff
@@ -118,13 +121,32 @@ class LayerPlanBuffer : SettingsMessenger
...
+     */
+    void insertPrintTempCommand(ExtruderPlan& extruder_plan);
+
+    /*!
+     * Insert the temp command to start cooling from the printing temperature to the final print temp
+     * 
+     * The command is inserted at a timed offset before the end of the last extrusion move
+     * 
+     * \param extruder_plans The extruder plans in the buffer, moved to a temporary vector (from lower to upper layers)
+     * \param extruder_plan_idx The index of the last extruder plan in \p extruder_plans with the same extruder as previous extruder plans
```
BagelOrb: f

src/Preheat.cpp
BagelOrb: f

src/Preheat.cpp
BagelOrb: f

BagelOrb: f

BagelOrb: f

src/Preheat.h
BagelOrb: f

BagelOrb:
@NicholasTait that's up to our test department. I think it will be only a
couple of degrees, but you never know!

Op 16 nov. 2016 15:56 schreef "Ghostkeeper" <notifications@github.com>:

PR #415, opened 2016-11-28
Implementing std::reference_wrapper as the main member of PolygonRef

fuchstraumer:
As was discussed in one of the issues, I attempted to change PolygonRef to use a std::reference_wrapper instead of a raw pointer. This makes things generally feel safer and cleaner, imo, but also has the advantage of being generally easier to use: a std::reference_wrapper object can be inserted into the standard libraries container objects, unlike a regular reference.

HOWEVER, this means nullptr is no longer a check that works for PolygonRef. A std::reference_wrapper also has no default constructor - no way to set a null reference, after all. Instead, I defined a static instance of the ClipperLib::Path class called "nullPath". It should be initialized as empty - I'm not sure exactly how comparisons in the rest of the code would work, but it could be as simple as checking if the input path/path ref returns true for "path.empty()". That's me postulating, however. 

The ref can also be made constant - and constructing a new ref can be done with std::ref() or std::cref() for constant references. Seeing as how the polygonref class does allow for modification of the referred object though, I don't think that'll work or be necessary (as safe as const-ness can feel). 

EDIT: On looking further, it appears I used an older version of the polygon files as there are some differences between the two, even when I didn't have to change anything to make the new ref type work. Apologies about that! not sure what to do now, with that in mind...

### DIFF
--- a/src/utils/polygon.cpp
+++ b/src/utils/polygon.cpp
@@ -8,1223 +8,1109 @@
 namespace cura 
 {
 
-bool PolygonRef::shorterThan(int64_t check_length) const
-{
-    const PolygonRef& polygon = *this;
-    const Point* p0 = &polygon.back();
-    int64_t length = 0;
-    for (const Point& p1 : polygon)
-    {
-        length += vSize(*p0 - p1);
-        if (length >= check_length)
-        {
-            return false;
-        }
-        p0 = &p1;
-    }
-    return true;
+bool PolygonRef::pShorterThan(int64_t check_length) const {
+	const PolygonRef& polygon = *this;
+	const Point* p0 = &polygon.back();
+	int64_t length = 0;
+	for (const Point& p1 : polygon) {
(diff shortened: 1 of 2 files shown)


BagelOrb:
It seems like you have replaced all 4 spaces by a tab.
Please make a new PR which doesn't change the whole Polygon files.

This way a git blame will always point to you and it's very hard for me to review what changes you have done.

Just pushing a new commit which undoes the spaces-to-tabs won't do, because the history will then still be tainted by this wall which is your commit. Please make a new Pull Request.

BagelOrb:
What process are you using to create such a pull request? If you use the git command line tool to submit your changes you will never have this problem of having differences between your own file and the file in our repository. Your changes will then be recorded and inserted before any changes which we have made in the meantime.

fuchstraumer:
I used the tool here, on the website, to submit the request. I'll fix the issues and submit a new request through the command line tool. Sorry about that!

PR #417, opened 2016-12-07
Reinstate filament used comment for RepRap flavour gcode.

smartavionics:
This little addition writes a comment containing the total amount of filament used by the first extruder to RepRap flavour gcode files. The format of the comment is similar to that used by Cura 1.15.04 except that it doesn't include the weight estimate.

It has been tested with CuraEngine 2.3.1 and DuetWifi 1.16.

### DIFF
--- a/src/gcodeExport.cpp
+++ b/src/gcodeExport.cpp
@@ -182,6 +182,10 @@ std::string GCodeExport::getFileHeader(const double* print_time, const std::vect
             prefix << ";NOZZLE_DIAMETER:" << float(INT2MM(getNozzleSize(0))) << new_line;
             // TODO: the second nozzle size isn't always initiated! ";NOZZLE_DIAMETER2:"
         }
+        else if (flavor == EGCodeFlavor::REPRAP)
+        {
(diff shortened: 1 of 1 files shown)


BagelOrb:
This PR was a consequence of https://github.com/Ultimaker/CuraEngine/issues/416

smartavionics:
Thanks for that.

BagelOrb:
Just in time for the 2.4 release!

Hopefully my colleagues agree that merging this PR is super low risk, so that we will keep it in 2.4

smartavionics:
The only possible problem would be if extruder_attr[0].filament_area was zero because then we would get a div by zero. However, it's a double so it would just give a NaN? Can this code ever be reached if filament_area is zero? If so, is it worth either adding a very small amount (e.g. 1e-6) to avoid the /0 or testing for filament_area being zero.

BagelOrb:
You could do that. You could also add a check for filament Area being Zero
and handleiding that separately.

But if the filament area is zero there is a bigger problem, which should
already be handles in the front end, so I don't think it's high priority to
do so.

Op 12 dec. 2016 11:59 schreef "Mark Burton" <notifications@github.com>:

PR #419, opened 2016-12-12
Add comment that reports layer height in RepRap flavour gcode.

smartavionics:
The DuetWifi firmware expects the Cura gcode to include a comment that reports the basic layer height. This PR adds that feature.


src/gcodeExport.h
```diff
@@ -121,6 +121,8 @@ class GCodeExport : public NoCopy
 
     unsigned int layer_nr; //!< for sending travel data
 
+    double layer_height; //!< report basic layer height in RepRap gcode file.
```
BagelOrb: This variable doesn't need to be defined in the class

smartavionics:
Yes, good point, i've moved layer_height out of the class definition and it's now local to the gcodeExport.cpp file.

Ghostkeeper:
It works! Layer height is now reported in the header for RepRap flavour g-code.

It might've been better to just put this in the start g-code of the printers in question, but because of the nature of DuetWifi and because it's also sorta handy for some human reading the g-code, this solution is fine too.

src/gcodeExport.cpp
```diff
@@ -197,6 +201,7 @@ std::string GCodeExport::getFileHeader(const double* print_time, const std::vect
         else if (flavor == EGCodeFlavor::REPRAP)
         {
             prefix << ";Filament used: " << ((filament_used.size() >= 1)? filament_used[0] / (1000 * extruder_attr[0].filament_area) : 0) << "m" << new_line;
+            prefix << ";Layer height: " << layer_height << new_line;
```
BagelOrb: The `layer_height` variable should be declared and defined just above this line.

Note that when you are printing in one-at-a-time mode you can have different layer heights per meshgroup, so this comment is kind of deceiving.

Perhaps it is indeed better to add this comment in the starting gcode of your printer definition.

smartavionics: Sorry if this is me being stupid but if I declare it above that line it won't have file scope will it?

I understand what you are saying about having different layer heights per meshgroup but I don't think the Duet (reprap) firmware really knows about printing multiple objects sequentially so this is the best we can do.

If we added this comment to the starting code then it would have to be changed for each different layer height used, surely it is better to generate it automatically?

BagelOrb: I don't understand your comment about file scope.

There is not neccessarily one `layer_height` used throughout the whole gcode, so I don't like to have that variable as a global variable.

smartavionics: we can make it local to that file (static) and also change the name to something like gcode_comment_layer_height. How about that?

BagelOrb: Why not define it here where you use it?

smartavionics: Because meshgroup isn't in scope where the value of layer_height is used.

BagelOrb:
AH yes, of course.

Could you make it a member of the gcodeExport class and rename it to either
first_meshgroup_layer_height or last_... according to whichever it is?

Op 10 feb. 2017 18:24 schreef "Mark Burton" <notifications@github.com>:

smartavionics:
I have moved it into the class and it is called last_meshgroup_layer_height.

PR #430, opened 2017-01-02
Use different skin pattern on first layer

14bitVoid:
This pull req creates a setting to change the skin pattern for just the first layer. I like concentric skin for bed adhesion but not for anything else so this allows to set just the layer that does the bed adhesion to concentric.
I've used this for a while but now I decided to make a PR for it, it was tested on the Ultimaker 2 printer.

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -864,7 +864,9 @@ void FffGcodeWriter::addMeshPartToGCode(SliceDataStorage& storage, SliceMeshStor
         processSingleLayerInfill(gcode_layer, mesh, part, layer_nr, infill_line_distance, infill_overlap, infill_angle);
     }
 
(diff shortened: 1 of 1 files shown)


BagelOrb:
There should be an accompanying PR on the frontend. Did you also submit changes to Cura/resources/definitions/fdmprinter.def.json ?

BagelOrb:
Ah here it is: https://github.com/Ultimaker/Cura/pull/1290

BagelOrb:
Looking good. Just checking whether `"top_bottom_pattern"` occurs anywhere else in the code besides these places.

Ghostkeeper:
The only other place it's mentioned is here: https://github.com/Ultimaker/CuraEngine/blob/master/meta/15.04_to_2.0.txt#L174

Ghostkeeper:
Works nicely. Thanks!

PR #439, opened 2017-01-17
cubicsubdiv windows fix

mboerwinkle:
This fixes my cubicsubdiv when cross-compiled and run in wine. I'm not completely sure why this works, but it basically ends up passing and comparing smaller values. Should be good enough for now to get the beta build working again.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,15 @@ project(CuraEngine)
 
 cmake_minimum_required(VERSION 2.8.12)
 
+option (ENABLE_ARCUS
+    "Enable support for ARCUS" ON)
+
+if (ENABLE_ARCUS)
+    message(STATUS "Building with Arcus")
+    find_package(Arcus REQUIRED)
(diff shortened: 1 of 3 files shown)


BagelOrb:
Nice!

We're kind of busy with a lot of stuff at the moment, but we'll merge this ASAP.

mboerwinkle:
The problem seems to have been caused by different definitions of "long int". Specifying "int64_t" for sphere_slice_radius2 was all it needed. I have tested from Windows 10 and Wine. I have updated the pull request.

BagelOrb:
I didn't want the CMakeLists changes and all in the history, so I've fixed it myself: https://github.com/Ultimaker/CuraEngine/commit/113202cd3466f9db05f7125d12f8b9b1b3542993

`coord_t` evaluates to `long long singed int`.
An int is good enough for a layer_nr. There can't be that many layers.

BagelOrb:
Now I only have to check whether this actually solves the problem on our test machines.

PR #440, opened 2017-01-19
Remove areas of infill smaller than min_infill_area.

smartavionics:
When you have a small feature above an outer layer (e.g. raised text), the
area under the feature will be infill rather than skin and so the skin is
composed of multiple segments around the feature. Better surface quality is
obtained if the area under the raised feature is treated as skin thus allowing
more of the outer layer to be printed as a single segment. This setting
specifies the minimum area (in mm^2) of a filled region. Areas smaller than
this will be filled with skin rather than infill.

### DIFF
--- a/src/skin.cpp
+++ b/src/skin.cpp
@@ -31,7 +31,7 @@ void generateSkinAreas(int layer_nr, SliceMeshStorage& mesh, const int innermost
     {
         return;
     }
-    
+    int min_infill_area = mesh.getSettingInMillimeters("min_infill_area");
     for(unsigned int partNr = 0; partNr < layer.parts.size(); partNr++)
     {
         SliceLayerPart& part = layer.parts[partNr];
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Coincides with this pull request to add the setting to Cura's front-end: https://github.com/Ultimaker/Cura/pull/1359

Ghostkeeper:
I'm required to ask you if you're okay with this code becoming licensed under a commercial license in the future as well as the AGPL license it would get if it's merged now.

smartavionics:
Yes, as long as the AGPL license continues to apply, I don't mind it also being subject to a commercial license. This is true for all of the contributions I may make to Cura/CuraEngine.

BagelOrb:
There was some miscummunication internally. We will inform you shortly with more detailed information on the subject.

PR #446, opened 2017-01-30
Add 'No Walls' combing mode to suppress combing when destination is a wall.

smartavionics:
Print visual quality can suffer if a move to the start of an outer wall
line is combed because the excess "dribble" can reach the outer surface
(at least enough of it does to make a visual difference). This mod adds
the 'No Walls' combing mode which forces a retraction to be used when the
destination of the move is the first point in a wall.

I definitely see an improvement of print quality "downstream" of the z-seam when this mode is enabled. It's still rather experimental but I am creating the PR as it gives us something to discuss.

My intention was for this mode to only affect travel to outer walls but I don't yet understand the engine's code well enough to be sure that this is the case with this PR. Do you get a polygon for each of the walls of an area or just the outer wall?

### DIFF
--- a/src/gcodePlanner.cpp
+++ b/src/gcodePlanner.cpp
@@ -142,8 +142,8 @@ Polygons GCodePlanner::computeCombBoundaryInside(CombingMode combing_mode)
         for (SliceMeshStorage& mesh : storage.meshes)
         {
             SliceLayer& layer = mesh.layers[layer_nr];
-            const CombingMode combing_mode = mesh.getSettingAsCombingMode("retraction_combing");
-            if (combing_mode == CombingMode::NO_SKIN || combing_mode == CombingMode::NO_WALLS)
+            const CombingMode mesh_combing_mode = mesh.getSettingAsCombingMode("retraction_combing");
+            if (mesh_combing_mode == CombingMode::NO_SKIN || mesh_combing_mode == CombingMode::NO_WALLS)
             {
                 for (SliceLayerPart& part : layer.parts)
                 {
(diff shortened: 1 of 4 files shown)


smartavionics:
Furthermore, apart from reducing the possibility of excess filament arriving at the outer wall, this PR also helps by introducing a slight pause (while the prime occurs) after the travel to the outer wall and this may help to reduce the acceleration ripples that can occur after the rapid travel movement. On my Kossel XL delta, the extruder definitely bounces around after a rapid move. YMMV.

BagelOrb:
I think the intent of this PR is good, but I think it should be implemented in a totally different way.

The problem is that combing actually consists of two subfeatures, which aren't clearly distinct in the conceptualization of the feature:
- changing the path the nozzle takes to avoid crossing the boundary of the print
- doing no retraction for such a move

What you actually want is that the retraction is performed anyway when we're going to print the outer wall.
You should then perform the combing, but not set the `path.retract` to `true`.
In order to see where you should add the parameter you should walk up the call tree and see from where addPolygons is called for the outer walls. You should then pass along the setting to all functions in between `FffGcodeWriter::processInsets` and `addPolygon`. Just look for where `addPolygon` is called and repeat that for the function where it is called until you end up where the outer walls are processed.

I know it's really ugly because there are already some more settings which are passed along to those functions. Perhaps I should come up with a better solution than keep adding parameters to those functions, but for now you can just add another parameter with a default value.

Because combing should still be performed I think you should introduce a boolean setting called something like 'Retract Before Outer Wall' which is only visible if you have combing and retraction enabled.
I think this functionality shouldn't be part of the combing mode dropdown.

smartavionics:
Hi Tim, thanks for the feedback. I shall digest the detail in what you say and change the implementation.

smartavionics:
Closing this PR and creating a new one to replace it.

PR #447, opened 2017-01-30
Feature multithread omp rebased

BagelOrb:
Johan Kristensens changes rebased

Just to give myself a nice overview


BagelOrb:
Note to self: some of the latest commits of the original work are not in the rebased branch!

PR #449, opened 2017-01-31
Can now grow skin areas so that they become anchored by the infill above and below.

smartavionics:
If the anchor_skin_in_infill setting is true, skin areas are expanded by
an amount that is the infill line distance * 1.4. This should be sufficient
to ensure that the skin is anchored in the infill by at least the distance
between the infill lines.

Hi Tim,

This PR is what I really wanted to write when I did the minimum infill area PR a little time ago.

People (not only me) have been bleating at S3D to do something like this for a year or two. As far as I know, only slic3r currently behaves well in this regard. S3D (and Cura, for that matter) do not handle skins over sparse infill well. If you have a feature above a skin above sparse infill, it looks crap and is very weak. This PR should fix that.

Hope you like it (OK, if not the implementation, at least the idea!).

I realise that you are not using contributions at the moment due to some ongoing licensing issue but I would be grateful if you can consider this in the future.

### DIFF
--- a/src/skin.cpp
+++ b/src/skin.cpp
@@ -113,16 +113,27 @@ void generateSkinAreas(int layer_nr, SliceMeshStorage& mesh, const int innermost
             }
         }
 
-        int anchor_skin_distance = mesh.getSettingInMicrons("anchor_skin_distance");
+        int anchor_skin_expand_distance = mesh.getSettingInMicrons("anchor_skin_expand_distance");
         
-        if (anchor_skin_distance > 0 && mesh.getSettingBoolean("anchor_upper_skin_in_infill"))
+        if (anchor_skin_expand_distance > 0)
         {
-            upskin = upskin.offset(-innermost_wall_line_width/2).offset(anchor_skin_distance).intersection(virgin_upskin);
-        }
+            int anchor_skin_shrink_distance = mesh.getSettingInMicrons("anchor_skin_shrink_distance");
 
-        if (anchor_skin_distance > 0 && mesh.getSettingBoolean("anchor_lower_skin_in_infill"))
-        {
-            downskin = downskin.offset(-innermost_wall_line_width/2).offset(anchor_skin_distance).intersection(virgin_upskin);
+            // skin areas are to be enlarged by anchor_skin_distance but before they are expanded
+            // the skin areas are shrunk by shrink_distance so that very narrow regions of skin
+            // (often caused by the model's surface having a steep incline) are removed first
+
+            anchor_skin_expand_distance += anchor_skin_shrink_distance; // increase the expansion distance to compensate for the shrinkage
+
+            if (mesh.getSettingBoolean("anchor_upper_skin_in_infill"))
+            {
+                upskin = upskin.offset(-anchor_skin_shrink_distance).offset(anchor_skin_expand_distance).intersection(virgin_upskin);
+            }
+
+            if (mesh.getSettingBoolean("anchor_lower_skin_in_infill"))
+            {
+                downskin = downskin.offset(-anchor_skin_shrink_distance).offset(anchor_skin_expand_distance).intersection(virgin_upskin);
+            }
         }
 
         Polygons skin = upskin.unionPolygons(downskin);
--- a/src/skin.cpp
+++ b/src/skin.cpp
@@ -57,7 +57,8 @@ void generateSkinAreas(int layer_nr, SliceMeshStorage& mesh, const int innermost
         }
 
         Polygons upskin = part.insets.back().offset(-innermost_wall_line_width / 2);
-        Polygons virgin_upskin = Polygons(upskin);
+        // make a copy of the outline which we later intersect and union with the resized skins to ensure the resized skin isn't too large or removed completely.
+        Polygons original_outline = Polygons(upskin);
         Polygons downskin = (downSkinCount == 0) ? Polygons() : upskin;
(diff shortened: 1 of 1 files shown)


smartavionics:
Feature not enabled...
screenshot_2017-01-31_19-05-01
Feature enabled...
screenshot_2017-01-31_19-05-29

src/skin.cpp
```diff
@@ -111,9 +112,32 @@ void generateSkinAreas(int layer_nr, SliceMeshStorage& mesh, const int innermost
...
+
+            // skin areas are to be enlarged by expand_distance but before they are expanded
+            // the skin areas are shrunk by shrink_distance so that very narrow regions of skin
+            // (often caused by the model's surface having a steep incline) are removed first
+
+            expand_skins_expand_distance += expand_skins_shrink_distance; // increase the expansion distance to compensate for the shrinkage
+
+            if (mesh.getSettingBoolean("expand_upper_skins"))
+            {
+                upskin = upskin.offset(-expand_skins_shrink_distance).offset(expand_skins_expand_distance).intersection(virgin_upskin);
```
BagelOrb: I think this removes skin areas which are smaller than the shrink distance. Try a small cube and see whether it still gets top skin. I think you should do `upskin = upskin.union(...)`.

src/skin.cpp
```diff
@@ -42,6 +42,7 @@ void generateSkinAreas(int layer_nr, SliceMeshStorage& mesh, const int innermost
         }
 
         Polygons upskin = part.insets.back().offset(-innermost_wall_line_width / 2);
+        Polygons virgin_upskin = Polygons(upskin);
```
smartavionics: It's really just the original skin outline which gets used to initialize upskin (and downskin). I have now renamed virgin_upskin to original_outline.

src/skin.cpp
smartavionics: Good catch, done that.

BagelOrb:
Looking good. Ready to be component tested by us.

BagelOrb:
I am getting rather weird results on these prints:
[fox.stl.txt](https://github.com/Ultimaker/CuraEngine/files/816618/fox.stl.txt)
[Fennec_Fox.stl.txt](https://github.com/Ultimaker/CuraEngine/files/816619/Fennec_Fox.stl.txt)

I can't really tweak the settings to my liking here, but maybe that's me.

smartavionics:
With this scheme, models that have shallow slopes are always likely to produce areas of skin that aren't really useful. Increasing the value of min skin width for expansion will remove some of them but not all. I suspect that to do this optimally, requires quite a lot more computation than what it's doing at the moment.

So, yes, perhaps it's not so good for these kind of models but that doesn't detract from the benefit you get when you have features above a horizontal surface with low density infill below.

BagelOrb:
I'm refactoring how the settings work and their descriptions.
I'll send you a PR to include my changes in your PR when I'm ready ;)

On closer inspection I don't really understand the feature.
What do you mean by anchoring the skin in the infill?
Why do we do this for bottom surfaces?
I thought this was to have the infill of a small feature on top lie on top of skin fom the layer below...

Can you elaborate a bit on the feature?

smartavionics:
OK, the real point of this feature to ensure that skins always extend far enough under perimeter walls so that the skin is anchored to the infill below irrespective of how sparse the infill is (within reason, I don't expect it to work well with really low infill densities).

If you don't do this, and the skin stops immediately under the perimeter walls above, then you will tend to get gaps appearing where the skin meets the wall. Slic3r gets this right, S3D and Cura (without this PR) get this wrong.

So, yes, the skin is anchored in the infill because when the skins are expanded they will have infill both below and above them.

As for expanding the skins on bottom surfaces: imagine a part where you have, say, a vertical tube with a horizontal member joining the tube. If only the top skins of the horizontal member were extended into the vertical tube's infill, the part would be stronger in one orientation than another. By also anchoring the bottom skins of the horizontal member into the vertical tube you end up with a part that can cope equally with up/down forces on the horizontal member.

BagelOrb:
I still find the word 'anchoring' a bit weird. The goal is to not have skin hanging in the air, but have it end on infill lines, right? It seems to be more about having skin above infill, than having also infill above skin.

smartavionics:
Yes, it is more about having skin above infill than below but as I just mentioned, there is a case for extending lower skins you just don't need to do that very often.

As for "anchoring", I really don't mind if that is changed to something else that you feel is more appropriate. I'm interested in the capability more than the name!

BagelOrb:
Related frontend PR: https://github.com/Ultimaker/Cura/pull/1395

src/skin.cpp
smartavionics: Hi, adding the union stops small skin areas from being removed but that is exactly what some people are asking for (see https://ultimaker.com/en/community/39214-feature-add-angle-for-top-bottom-internal-layer-generation).

Perhaps we should add an option (something like Preserve Original Skin) to control whether we do the union or not?

BagelOrb: Those people only want to have small skin removed when it's next to infill. When the skin is enclosed directly by the walls, you would be removing the actual skin, which is a bad idea. You would then be able to see the infill when looking at your print.

I think the correct place to fix the issue you refer to is when the skins are generated in the first place. I'm not 100% sure it's possible there, but If it's possible we should do it there, because that would be less of an impact on performance.

PR #451, opened 2017-02-02
Fix spiralize args bug - untested but this just looked wrong to me.

smartavionics:


BagelOrb:
Not tested, but this code has to be wrong and your solution seems right when I look at the code.

Thanks a bunch!

JENkt4k:
Tested and works. When will the 2.4 Beta be updated? CuraEngine compiles and functions properly, Cura app from source currently throws GL exception on my system.

smartavionics:
Hi, glad that helped. As for when 2.4 Beta will be updated, I haven't a clue. The Cura source has just today had some new GL stuff merged so maybe that's the cause of your problem. Have you also updated/installed the Uranium source?

JENkt4k:
Thanks! "The Cura source has just today had some new GL stuff merged so maybe that's the cause of your problem." I checked out from master a few days ago so today's changes aren't likely to have caused the issue but if they were to resolve existing issues that may be my problem. I probably need to checkout from a specific branch (2.3 or 2.4). I tried setting the back-end in the config for Cura 2.3 to CuraEngine master but there are obviously some settings or incompatibilities. "cura-build" seems to be using some dated packages as well (Example: checking out/cloning version 5.4 vs 5.5 or 5.6 of PyQt) so getting a solid build has been a cumbersome process as well. I am running an older version of Ubuntu with Python 3.4 vs 3.6 (updated manually), but I really like the Cura 2.4 AppImage packaging. Made it a dream to test without wreaking 2.3 installed Cura.

smartavionics:
I am working on Ubuntu and I haven't been able to get cura-build to work so what I do is having built and installed libarcus and the protobuf stuff I then update CuraEngine, Uranium and Cura to whatever version I am interested in (mostly master + my own mods not yet integrated) and then in the build sub directory in Uranium and Cura I simply do "sudo cmake -P cmake_install.cmake" and that copies the changed files into /usr/local and in the CuraEngine directory I simply do make followed by "sudo cp CuraEngine /usr/local/bin".

One other thing is that I have a shell script to start cura so it can find what it needs:

#!/bin/sh
PYTHONPATH=/usr/local/lib/python3/dist-packages cura $*

Hope this helps.

JENkt4k:
Awesome, yeah I ended up doing something similar checking out each, and building them individually. Glad I'm not the only one.

probonopd:
> I really like the Cura 2.4 AppImage packaging. Made it a dream to test without wreaking 2.3 installed Cura.

Continuous builds (a new AppImage each time a commit is pushed to GitHub) would be a useful next step in this direction :-)

nallath:
We agree. This is something that we're looking into now.

CCS86:
Sweet!

JENkt4k:
Hmm I reverted the changes, tested just CuraEngine with "magic-spiralize=true" and got the same results. Not sure what, if any, effect the change had.

CCS86:
I'm still seeing an issue in 2.5 beta spiralize. Looks to be the same scar inducing gap. I have disabled all other settings that I could think of which would interfere (like coasting, gap filling, extra wall, etc).

image

CCS86:
@JENkt4k : make sure you set infill density to 0%, 1 perimeter wall, 0 top layers.

smartavionics:
If you have another gcode viewer such as S3D, take a look at the gcode with that because the scar may not actually be very visible at all. It is a known "feature" of the current Cura layer view that it shows the z-seam of a spiralized model.

Obviously, if the seam really is visible when the object is printed, then that is a bug in the spiralize code!

CCS86:
You are absolutely right. I skipped that step because the new layer view is so nice. Repetier shows no scar!

PR #454, opened 2017-02-03
Fix jerk setting to be compatible with RepRap firmware.

smartavionics:
M205 was being output which RepRap doesn't understand. RepRap firmware
uses M566 instead to set jerk speeds and the units are mm/min.

### DIFF
--- a/src/gcodeExport.cpp
+++ b/src/gcodeExport.cpp
@@ -889,15 +889,23 @@ void GCodeExport::writeJerk(double jerk)
 {
     if (current_jerk != jerk)
     {
-        if (getFlavor() == EGCodeFlavor::REPETIER)
+        if (getFlavor() == EGCodeFlavor::REPRAP)
         {
-            *output_stream << "M207 X";
+            const double jerkPerMin = jerk * 60;
+            *output_stream << "M566 X" << PrecisionedDouble{2, jerkPerMin} << " Y" << PrecisionedDouble{2, jerkPerMin} << new_line;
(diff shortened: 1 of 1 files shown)


src/gcodeExport.cpp
```diff
@@ -889,15 +889,23 @@ void GCodeExport::writeJerk(double jerk)
...
-            *output_stream << "M205 X";
+            if (getFlavor() == EGCodeFlavor::REPETIER)
+            {
+                *output_stream << "M207 X";
+            }
+            else
+            {
+                *output_stream << "M205 X";
+            }
+            *output_stream << PrecisionedDouble{2, jerk} << new_line;
```
Ghostkeeper: It might be more readable and efficient to use an actual switch statement here on getFlavor(), to be either REPRAP, REPETIER or default. Otherwise I think the code itself is a fine fix (but pull requests are on halt for a few weeks on CuraEngine due to internal talks going on).

BagelOrb:
The idea behind this PR is good, but unfortunately the problem is quite a bit bigger than this.

Currently we don't distinguish between RepRap and Marlin.
This means that we need to introduce a new flavor called MARLIN, add it to the setting definitions in fdmprinter.def.json and go over each use of `REPRAP` in the code to see whether it would only hold for Marlin, also for Repetier or for both.

smartavionics:
Well, I don't really care enough about this to do all that so I don't mind if this PR is held back until another time.

Issue #428, opened 2016-12-29
Missing checks when writing bed temperatur


Ocarthon:
In [gcodePlanner.cpp](https://github.com/Ultimaker/CuraEngine/blob/50df40c6c615f83f1e410aedcd22fddac5867b0e/src/gcodePlanner.cpp#L650) and [FffGcodeWriter.cpp](https://github.com/Ultimaker/CuraEngine/blob/2.4/src/FffGcodeWriter.cpp#L299) bed temperature commands get written to the GCode without checking if a heated bed exists like it's done here in [FffGcodeWriter.cpp](https://github.com/Ultimaker/CuraEngine/blob/50df40c6c615f83f1e410aedcd22fddac5867b0e/src/FffGcodeWriter.cpp#L242)

Ghostkeeper:
Indeed, nice find.

BagelOrb:
fixed

PR #455, opened 2017-02-07
Add checks to ensure that a heated bed exists

Ocarthon:
This PR adds missing checks to ensure that the bed is only heated if it exists and the temperature is not 0 degrees. This should fix https://github.com/Ultimaker/CuraEngine/issues/428. Sadly i cannot verify if it works as i cannot compile CuraEngine at the moment.


BagelOrb:
I'm sorry to say that our normal review process is delayed. Handling PRs
might take a week or two.

Op 7 feb. 2017 12:33 schreef "Philip Standt" <notifications@github.com>:

BagelOrb:
Whoops, I fixed this myself along the way.
This PR did the same as I did.

Issue #459, opened 2017-02-10
Repetier flavor - M204 used for setting acceleration instead of M201 and M202.


FlameRunner:
If I understand the code correctly, no matter which flavour is selected, acceleration-setting g-code is generated by [this line](https://github.com/Ultimaker/CuraEngine/blob/c7e621eeb5b56dce6fbccaf6d9c46628d3142c80/src/gcodeExport.cpp#L882)
`*output_stream << "M204 S" << PrecisionedDouble{0, acceleration} << new_line; // Print and Travel acceleration`

So, something like: M204 S2000 and that works fine with Marlin.

Problem is, Repetier firmware can't interpret that correctly.
In Repetier, M204 command sets PID parameters (fortunately, the Sxxx argument is omitted).
Setting acceleration in Repetier is done by:
M201 - Set max acceleration in units/s^2 for print moves (M201 X1000 Y1000)
M202 - Set max acceleration in units/s^2 for travel moves (M202 X1000 Y1000)
([source](https://github.com/repetier/Repetier-Firmware/wiki/G-codes) )

BagelOrb:
We accept pull requests with code which fixes the problem.

I think there is no reason for Ultimaker to invest in fixing such problems.

FlameRunner:
Unfortunately, I can't really code and my ability to use GitHub is also limited.

I guess something like this, although being rather crude way, would do the job:

```
if (getFlavor() == EGCodeFlavor::REPETIER)
{
	*output_stream << "M201 X" << PrecisionedDouble{0, acceleration} << " Y" << PrecisionedDouble{0, acceleration} << new_line; // Print acceleration
	*output_stream << "M202 X" << PrecisionedDouble{0, acceleration} << " Y" << PrecisionedDouble{0, acceleration} << new_line; // Travel acceleration
}
else
{
	*output_stream << "M204 S" << PrecisionedDouble{0, acceleration} << new_line; // Print and Travel acceleration
}

current_acceleration = acceleration;
estimateCalculator.setAcceleration(acceleration);
```

It would be much nicer if that code checked if the next move is going to be travel or print and output only one command, but I have absolutely no idea how to do this.

You are probably right that Ultimaker has no reason to fix this, but maybe someone else using Repetier would be willing to pick it up from here?

smartavionics:
Hopefully, someone who uses Repetier can test the PR and decide whether it works or not.

brickZA:
[...]

PR #461, opened 2017-02-11
Mb fix repetier acceleration

smartavionics:
Offered as a solution for https://github.com/Ultimaker/CuraEngine/issues/459

Compiles but untested as I don't have a machine that runs Repetier.

Happy to fix any problems it may have.

### DIFF
--- a/src/gcodeExport.h
+++ b/src/gcodeExport.h
@@ -96,6 +96,7 @@ private:
     Point3 currentPosition;
     double currentSpeed; //!< The current speed (F values / 60) in mm/s
     double current_acceleration; //!< The current acceleration in the XY direction (in mm/s^2)
+    double current_travel_acceleration; //!< The current acceleration in the XY direction used for travel moves if different from current_acceleration (in mm/s^2)
     double current_jerk; //!< The current jerk in the XY direction (in mm/s^3)
     double current_max_z_feedrate; //!< The current max z speed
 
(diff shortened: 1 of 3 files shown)


FlameRunner:
What if we try to print as follows:
1. Travel with acceleration=1500
2. Print with acceleration=2000
3. Travel with acceleration=2000
?
I believe that the acceleration-setting code won't be executed at third point ('new' acceleration is 2000 and current_acceleration set at second point is also 2000, so current_acceleration != acceleration condition is false), M202 X2000 Y2000 commend won't appear in the g-code and the printer will continue to use the 1500 value for travels.

smartavionics:
Yes, good point. I have added a new field, current_travel_acceleration so it will now track the last acceleration values set for both travel moves and non-travel moves.

src/gcodeExport.h
```diff
@@ -96,6 +96,7 @@ class GCodeExport : public NoCopy
     Point3 currentPosition;
     double currentSpeed; //!< The current speed (F values / 60) in mm/s
     double current_acceleration; //!< The current acceleration in the XY direction (in mm/s^2)
+    double current_travel_acceleration; //!< The current acceleration in the XY direction used for travel moves if different from current_acceleration (in mm/s^2)
```
BagelOrb: Would be nice to mention that this is only used by Repetier firmware at the moment.

Not blocking. This PR seems ready to be merged.

PR #465, opened 2017-02-19
Mb spiralize bug fix

smartavionics:
Hello again,

I have spent a few hours working on the spiralize bug and this PR does fix the problem for me. I confess that I am not 100% familiar with the code so my fix may not be optimal. Feel free to implement it differently if you wish.

I have also updated the earlier fix that was in the earlier PR because it wasn't actually doing the right thing.

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -1009,7 +1009,7 @@ void FffGcodeWriter::processInsets(GCodePlanner& gcode_layer, SliceMeshStorage* 
             { // on the last normal layer first make the outer wall normally and then start a second outer wall from the same hight, but gradually moving upward
                 WallOverlapComputation* wall_overlap_computation(nullptr);
                 int wall_0_wipe_dist(0);
-                gcode_layer.addPolygonsByOptimizer(part.insets[0], &mesh->insetX_config, wall_overlap_computation, EZSeamType::SHORTEST, z_seam_pos, wall_0_wipe_dist, spiralize);
+                gcode_layer.addPolygonsByOptimizer(part.insets[0], &mesh->inset0_config, wall_overlap_computation, z_seam_type, z_seam_pos, wall_0_wipe_dist, false);
             }
         }
         int processed_inset_number = -1;
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -1009,7 +1009,7 @@ void FffGcodeWriter::processInsets(GCodePlanner& gcode_layer, SliceMeshStorage* 
             { // on the last normal layer first make the outer wall normally and then start a second outer wall from the same hight, but gradually moving upward
                 WallOverlapComputation* wall_overlap_computation(nullptr);
                 int wall_0_wipe_dist(0);
-                gcode_layer.addPolygonsByOptimizer(part.insets[0], &mesh->inset0_config, wall_overlap_computation, z_seam_type, z_seam_pos, wall_0_wipe_dist, false);
+                gcode_layer.addPolygonsByOptimizer(part.insets[0], &mesh->inset0_config, wall_overlap_computation, EZSeamType::SHORTEST, z_seam_pos, wall_0_wipe_dist, false);
             }
         }
         int processed_inset_number = -1;
--- a/src/gcodePlanner.cpp
+++ b/src/gcodePlanner.cpp
@@ -409,23 +409,15 @@ void GCodePlanner::addPolygonsByOptimizer(Polygons& polygons, GCodePathConfig* c
     {
         return;
     }
-    if (spiralize)
+    PathOrderOptimizer orderOptimizer(lastPosition, z_seam_pos, z_seam_type);
+    for (unsigned int poly_idx = 0; poly_idx < polygons.size(); poly_idx++)
     {
-        // only interested in the outer perimeter
-        addPolygon(polygons[0], 0, config, wall_overlap_computation, wall_0_wipe_dist, spiralize);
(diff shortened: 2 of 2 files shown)


smartavionics:
Hmm, not sure the fix is good because although it worked perfectly for an object with rectangular sides, I just tried it with a circular object and it appears to leave gaps. Could of course be another bug showing itself now. More work required, I think.

smartavionics:
OK, I've been doing some more testing and it seems that the spiralize mode is sensitive in some way to the shape of the thing being spiralized. Let me explain, see this very simple test vase(!) ... 

screenshot_2017-02-19_19-14-53

When the wall is made from 90 segments or less, Cura spiralizes it just fine. If I increase the number of segments to 100 then it starts having problems with chunks getting left out. Like this...

screenshot_2017-02-19_19-21-36

So at the moment, with this PR, spiralize appears to work OK as long as the object is not too detailed.

src/FffGcodeWriter.cpp
```diff
@@ -1009,7 +1009,7 @@ void FffGcodeWriter::processInsets(GCodePlanner& gcode_layer, SliceMeshStorage*
             { // on the last normal layer first make the outer wall normally and then start a second outer wall from the same hight, but gradually moving upward
                 WallOverlapComputation* wall_overlap_computation(nullptr);
                 int wall_0_wipe_dist(0);
-                gcode_layer.addPolygonsByOptimizer(part.insets[0], &mesh->insetX_config, wall_overlap_computation, EZSeamType::SHORTEST, z_seam_pos, wall_0_wipe_dist, spiralize);
+                gcode_layer.addPolygonsByOptimizer(part.insets[0], &mesh->inset0_config, wall_overlap_computation, z_seam_type, z_seam_pos, wall_0_wipe_dist, false);
```
BagelOrb: Good changes, except for the `z_seam_type`. This is now a user configurable option. Let's keep it that way.

If you think a different seam type should be default please make it so in the frontend.

src/gcodePlanner.cpp
```diff
@@ -409,15 +409,23 @@ void GCodePlanner::addPolygonsByOptimizer(Polygons& polygons, GCodePathConfig* c
     {
         return;
     }
-    PathOrderOptimizer orderOptimizer(lastPosition, z_seam_pos, z_seam_type);
-    for (unsigned int poly_idx = 0; poly_idx < polygons.size(); poly_idx++)
+    if (spiralize)
     {
-        orderOptimizer.addPolygon(polygons[poly_idx]);
+        // only interested in the outer perimeter
+        addPolygon(polygons[0], 0, config, wall_overlap_computation, wall_0_wipe_dist, spiralize);
```
BagelOrb: While it is an interesting thing to take only the outer polygon in spiralize mode, this is not how spiralize is intended.
Think about a C shape which is gradually closing up into an O when going upward. Your change would entail that at some point suddenly half of the lines aren't printed any more.

Moreover, the first polygon in a Polygons object isn't guaranteed to be the outer polygon. That is only so for a PolygonsPart.

I would like you to revert this change.

src/FffGcodeWriter.cpp
smartavionics: Not sure what you're saying here but I will revert the z seam type to SHORTEST.

src/gcodePlanner.cpp
smartavionics: As you wish but that leaves spiralize completely broken as it will output both the outer and inner polygons which crash into each other for models such as I have above.

smartavionics:
I'm giving up with this one, I obviously don't know what I'm doing here.

src/gcodePlanner.cpp
BagelOrb: Spiralize is a heavily limited option. You shouldnt use it on models like the one you showed. The models should be solid.

I do like your thought, though. I was thinking of making a frontend setting out of that.

smartavionics: I'm even more confused now. I thought spiralize was used for objects like vases where you have a base and walls but it's open at the top.

src/FffGcodeWriter.cpp
BagelOrb: Oh I misread your change, I thought you changed it the other way around.

My mistake!

This code change is good.

BagelOrb:
Well I think you do know what you are doing.
Some of your changes definitely fix some of the bugs of spiralize mode.

Perhaps I have been a bit too grumpy in my reactions; I didn't mean t scare you off.

Perhaps it was my previous comment which you might have misunderstood. What I meant was that I like that you are working on this PR, because we don't get a lot of time from Ultimaker to work on spiralize mode.

src/gcodePlanner.cpp
BagelOrb: You can do `Polygon.splitIntoParts()` to get PolygonParts form which you can take the outer polygon only.

BagelOrb: A spiralize vase is meant to make a vase with an open top, but the model has to be solid.

Perhaps you are correct in throwing away this inner polygon. Let me discuss with colleagues.

smartavionics:
I have found a new problem. If compensate outer wall overlaps is enabled, it's possible to get this when using spiralize:

CuraEngine.git/src/gcodePlanner.h:106: void cura::ExtruderPlan::handleAllRemainingInserts(cura::GCodeExport&): Assertion `insert.path_idx == paths.size()' failed.

It fails around layer 208 of a 620 layer model.

As a short term fix, shall I disabled the outer wall overlap compensation when spiralize is true?

smartavionics:
Closing this as it has become redundant given the new spiralize implementation.

PR #468, opened 2017-02-23
Spiralize better

smartavionics:
This is an attempt to improve the performance of the spiralize feature. It's not perfect but it does at least make an excellent job of the watering can part (see https://github.com/Ultimaker/Cura/issues/1462). I can now print that with absolutely no visible seam whatsoever (it's really very smooth indeed).

It still needs more work and I would value some input to create a robust solution to the main problem which is reliably determining the right place to start the wall polygon for the current layer based on the last position in the last layer.

Also, this code assumes that the layers are processed sequentially with increasing layer number because it uses the previous layer outline when processing the current layer outline. With the move to multi-threaded processing I can see this breaking.

I hope we can at least use the basic idea in this PR to come up with a good solution.

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -126,11 +126,16 @@ void FffGcodeWriter::findLayerSeamsForSpiralize(SliceMeshStorage& mesh)
 
         if (layer_nr < mesh.layers.size())
         {
-            // use the vertex closest to Point(0, 0) for the first layer that has a part with insets
+            // use the vertex closest to the seam position for the first layer that has a part with insets
             // Doing this, we give the user some degree of control as to where the seam will be.
             // By moving the model on the bed they can move the start position of the seam and
             // this could be useful if the spiralization has a problem with a particular seam position
-            mesh.layers[layer_nr].seam_vertex_index = PolygonUtils::findClosest(Point(0, 0), mesh.layers[layer_nr].parts[0].insets[0][0]).point_idx;
+            Point seam_pos(0, 0);
+            if (mesh.getSettingAsZSeamType("z_seam_type") == EZSeamType::USER_SPECIFIED)
+            {
+                seam_pos = Point(mesh.getSettingInMicrons("z_seam_x"), mesh.getSettingInMicrons("z_seam_y"));
+            }
+            mesh.layers[layer_nr].seam_vertex_index = PolygonUtils::findClosest(seam_pos, mesh.layers[layer_nr].parts[0].insets[0][0]).point_idx;
             ++layer_nr;
         }
 
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -67,7 +67,7 @@ void FffGcodeWriter::writeGCode(SliceDataStorage& storage, TimeKeeper& time_keep
     for (SliceMeshStorage& mesh : storage.meshes)
     {
         total_layers = std::max(total_layers, mesh.layers.size());
-        if (getSettingBoolean("magic_spiralize"))
+        if (mesh.getSettingBoolean("magic_spiralize"))
         {
             findLayerSeamsForSpiralize(mesh);
         }
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -117,16 +117,24 @@ void FffGcodeWriter::findLayerSeamsForSpiralize(SliceMeshStorage& mesh)
 
     if (mesh.layers.size() > 0)
     {
-        if(mesh.layers[0].parts.size() > 0 && mesh.layers[0].parts[0].insets.size() > 0)
+        unsigned layer_nr = 0;
+        // set the seam vertex index to 0 for bottom layers that have no parts or insets
+        while (layer_nr < mesh.layers.size() && (mesh.layers[layer_nr].parts.size() == 0 || mesh.layers[layer_nr].parts[0].insets.size() == 0))
         {
-            // use the vertex closest to Point(0, 0) for the first layer
+            mesh.layers[layer_nr++].seam_vertex_index = 0;
+        }
+
+        if (layer_nr < mesh.layers.size())
+        {
+            // use the vertex closest to Point(0, 0) for the first layer that has a part with insets
             // Doing this, we give the user some degree of control as to where the seam will be.
             // By moving the model on the bed they can move the start position of the seam and
             // this could be useful if the spiralization has a problem with a particular seam position
-            mesh.layers[0].seam_vertex_index = PolygonUtils::findClosest(Point(0, 0), mesh.layers[0].parts[0].insets[0][0]).point_idx;
+            mesh.layers[layer_nr].seam_vertex_index = PolygonUtils::findClosest(Point(0, 0), mesh.layers[layer_nr].parts[0].insets[0][0]).point_idx;
+            ++layer_nr;
         }
 
-        for (unsigned layer_nr = 1; layer_nr < mesh.layers.size(); ++layer_nr)
+        for (; layer_nr < mesh.layers.size(); layer_nr++)
         {
             SliceLayer& layer = mesh.layers[layer_nr];
             const SliceLayer& last_layer = mesh.layers[layer_nr - 1];
@@ -1099,7 +1107,7 @@ void FffGcodeWriter::processInsets(GCodePlanner& gcode_layer, SliceMeshStorage* 
         if (spiralize && &wall_layer.parts[0] == &part)
         {
             std::vector<Polygons>& wall_insets = wall_layer.parts[0].insets;
-            if(wall_insets.size() == 0 || wall_insets[0].size() == 0)
+            if (wall_insets.size() == 0 || wall_insets[0].size() == 0)
             {
                 // wall doesn't have usable outline
                 return;
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -67,6 +67,10 @@ void FffGcodeWriter::writeGCode(SliceDataStorage& storage, TimeKeeper& time_keep
     for (SliceMeshStorage& mesh : storage.meshes)
     {
         total_layers = std::max(total_layers, mesh.layers.size());
+        if (getSettingBoolean("magic_spiralize"))
+        {
+            findLayerSeamsForSpiralize(mesh);
+        }
     }
     
     gcode.writeLayerCountComment(total_layers);
@@ -105,6 +109,77 @@ void FffGcodeWriter::writeGCode(SliceDataStorage& storage, TimeKeeper& time_keep
 
     constexpr bool force = true;
     gcode.writeRetraction(storage.retraction_config_per_extruder[gcode.getExtruderNr()], force); // retract after finishing each meshgroup
+}
+
+void FffGcodeWriter::findLayerSeamsForSpiralize(SliceMeshStorage& mesh)
+{
+    // The spiral has to continue on in an anti-clockwise direction from where the last layer finished, it can't jump backwards
+
+    if (mesh.layers.size() > 0)
+    {
+        if(mesh.layers[0].parts.size() > 0 && mesh.layers[0].parts[0].insets.size() > 0)
+        {
+            // use the vertex closest to Point(0, 0) for the first layer
+            // Doing this, we give the user some degree of control as to where the seam will be.
+            // By moving the model on the bed they can move the start position of the seam and
+            // this could be useful if the spiralization has a problem with a particular seam position
+            mesh.layers[0].seam_vertex_index = PolygonUtils::findClosest(Point(0, 0), mesh.layers[0].parts[0].insets[0][0]).point_idx;
+        }
+
+        for (unsigned layer_nr = 1; layer_nr < mesh.layers.size(); ++layer_nr)
+        {
+            SliceLayer& layer = mesh.layers[layer_nr];
+            const SliceLayer& last_layer = mesh.layers[layer_nr - 1];
+
+            if (layer.parts.size() < 1 || layer.parts[0].insets.size() < 1 || last_layer.parts.size() < 1 || last_layer.parts[0].insets.size() < 1)
+            {
+                // either this layer or the last has no parts (or a part has no insets) so just duplicate the index calculated for the last layer
+                layer.seam_vertex_index = last_layer.seam_vertex_index;
+                continue;
+            }
+            PolygonRef last_wall = last_layer.parts[0].insets[0][0];
+            PolygonRef wall = layer.parts[0].insets[0][0];
+            const int n_points = wall.size();
+            Point last_wall_seam_vertex = last_wall[last_layer.seam_vertex_index];
+
+            // seam_vertex_idx is going to be the index of the seam vertex in the current wall polygon
+            // initially we choose the vertex that is closest to the seam vertex in the last spiralized layer processed
+
+            int seam_vertex_idx = PolygonUtils::findClosest(last_wall_seam_vertex, wall).point_idx;
+
+            // now we check that the vertex following the seam vertex is not to the right of the seam vertex in the last layer
+            // and if it is we move forward
+
+            // get the inward normal of the last layer seam vertex
+            Point last_wall_seam_vertex_inward_normal = PolygonUtils::getVertexInwardNormal(last_wall, last_layer.seam_vertex_index);
+
+            // create a vector from the normal so that we can then test the vertex following the candidate seam vertex to make sure it is on the correct side
+            Point last_wall_seam_vertex_vector = last_wall_seam_vertex + last_wall_seam_vertex_inward_normal;
+
+            // now test the vertex following the candidate seam vertex and if it lies to the left or on the vector, it's good to use
+            const int first_seam_vertex_idx = seam_vertex_idx;
+            float a = LinearAlg2D::getAngleLeft(last_wall_seam_vertex_vector, last_wall_seam_vertex, wall[(seam_vertex_idx + 1) % n_points]);
+            //std::cerr << layer_nr << ": n_points= " << n_points << ", angle = " << a * 180/M_PI << "\n";
+
+            while (a > M_PI)
+            {
+                // the vertex was on the right of the vector so move the seam vertex on
+                seam_vertex_idx = (seam_vertex_idx + 1) % n_points;
+                a = LinearAlg2D::getAngleLeft(last_wall_seam_vertex_vector, last_wall_seam_vertex, wall[(seam_vertex_idx + 1) % n_points]);
+                //std::cerr << "new angle = " << a * 180/M_PI << "\n";
+
+                if (seam_vertex_idx == first_seam_vertex_idx)
+                {
+                    // this shouldn't happen!
+                    break;
+                }
+            }
+
+            // store result for use when spiralizing
+            mesh.layers[layer_nr].seam_vertex_index = seam_vertex_idx;
+            //std::cerr << "layer " << layer_nr << ": " << seam_vertex_idx << "\n";
+        }
+    }
 }
 
 void FffGcodeWriter::setConfigFanSpeedLayerTime(SliceDataStorage& storage)
@@ -998,7 +1073,6 @@ void FffGcodeWriter::processInsets(GCodePlanner& gcode_layer, SliceMeshStorage* 
     bool compensate_overlap_x = mesh->getSettingBoolean("travel_compensate_overlapping_walls_x_enabled");
     if (mesh->getSettingAsCount("wall_line_count") > 0)
     {
-        static int last_wall_seam_vertex_idx = -1;
         bool spiralize = false;
         if (mesh->getSettingBoolean("magic_spiralize"))
         {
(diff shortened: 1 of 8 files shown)


smartavionics:
This is now working well on the few test objects I have tried. The implementation requires some static storage and is dependent on the layers being processed in order of increasing layer number so I am sure it will all have to be changed in the future but in the meantime it would be good if other developers tried it out and if they have models that don't spiralize satisfactorily provide me with feedback.

smartavionics:
Must be working pretty well, my watering can is completely watertight, doesn't leak at all!

BagelOrb:
Here's a nice design for a spiralize vase which you might not have thought about.
Note that there's two layers which are noting alike.

PS: I didn't come up with this design, but just copied it from memory. I don't recall where I saw this in the first place.
[round_tri_vase.stl.txt](https://github.com/Ultimaker/CuraEngine/files/810236/round_tri_vase.stl.txt)

BagelOrb:
Because we are working on making the engine multithreaded, you cannot assume that the layers are processed sequentially.
However, `FffGcodeWriter` has access to all layers at once at the `SliceDataStorage` level.
Note that `processInsets` will soon be a `const` function.

src/gcodePlanner.cpp
```diff
@@ -448,6 +448,80 @@ void GCodePlanner::addLinesByOptimizer(Polygons& polygons, GCodePathConfig* conf
...
+            }
+        }
+    }
+
+    Polygons last_wall_polygons;
+    last_wall_polygons.add(last_wall);
+    const int max_dist = MM2INT(config->getLineWidth() * 5);
+    // extrude to the points following the seam vertex
+    // the last point is the seam vertex as the polygon is a loop
+    for (int i = 1; i <= n_points; ++i)
```
BagelOrb: Please use informative variable names. I would call this one 'point_idx'.

src/gcodePlanner.cpp
```diff
@@ -448,6 +448,80 @@ void GCodePlanner::addLinesByOptimizer(Polygons& polygons, GCodePathConfig* conf
...
+    {
+        // p is a point from the current wall polygon
+        const Point& p = wall[(seam_vertex_idx + i) % n_points];
```
BagelOrb: Please don't use abbreviations in variable names.
I myself only abbreviate 'index' to 'idx' and 'polygon' to 'poly', because these are used everywhere in the code.
Other variables are better written out fully, to make the code more reabible.

smartavionics:
Hi Tim, thanks for the vase, I will try that soon.

I now believe that for the spiralize to work well the critical thing is knowing the previous (aka layer below) poly outline and its start vertex. If we could have a loop that calculates all of those from the layer data before the gcode generation then the actual spiralizing can be done multithreaded.

To summarize, to spiralize a layer it needs access to the current layer outline poly, the previous outline layer poly and the start vertex of the previous outline poly.

src/gcodePlanner.cpp
```diff
@@ -448,6 +448,80 @@ void GCodePlanner::addLinesByOptimizer(Polygons& polygons, GCodePathConfig* conf
...
+        {
+            PolygonUtils::moveInside(last_wall_polygons, last_p, 0, max_dist);
+        }
+        // now interpolate between last_p and p depending on how far we have progressed along wall
+#if 0
+        if(i == n_points)
+            addTravel_simple(p);
+        else
+#endif
+        addExtrusionMove(Point(last_p + (p - last_p) * static_cast<double>(i) / n_points), config, SpaceFillType::Polygons, 1.0, true);
```
BagelOrb: You don't need to cast to `double`. `X * i / n` won't suffer from integer rounding because it's parsed as `(X * i) / n`.

BagelOrb: This logic causes problems for my round_tri_vase model.

BagelOrb:
The problem of my round_tri_vase is a principle problem of your approach. Perhaps it's better to have an option to revert to something like old functionality, which doesn't interpolate between layer outlines.

Or maybe put a maximum distance on the points between which you shift.

BagelOrb:
You could indeed calculate the polygon order before processing all layers.

Another approach would be to fully separate processing of spiralize mode from normal processing, so that spiralize never gets multithreaded. I don't know how much code duplication that would result in though, but I think the `processLayer` code would be a lot more readable when it doesn't deal with spiralize.

It's up to you, though. Whichever you think is best for you.

smartavionics:
I tried the tri-vase and it looks OK to me. I sliced it with Cura and it's being displayed in S3D. What's the problem?

screenshot_2017-03-01_10-11-25

screenshot_2017-03-01_10-14-26

smartavionics:
I'll make the style changes you requested and then I think I will not do any major reworking until the multi-threading stuff has made it into master and then I can see what I'm up against!

BagelOrb:
I suspect that under some rotations, the layer where it suddenly changes
shape would get weird paths, but I'm not sure.

I thought the idea of this PR was to gradually shift from the poly of the
previous layer to the poly of the next. Can you tell me the general idea of
this PR before I dive into the code?

Op 1 mrt. 2017 11:15 schreef "Mark Burton" <notifications@github.com>:

smartavionics:
Yes, you have the right idea. What it does is interpolate in X/Y between the points in the previous layer polygon and the current layer polygon. At the start of the new layer, it still using the previous layer outline, at the end of the new layer its using the new layer outline and half way around the new layer it's using points that lie half way between the previous and current layer. So it's a smooth progression.

BagelOrb:
Okay, but then I would expect that the straight pieces on the new layer would get bent a bit toward the round piece of the previous layer. I think the reason that is not happening is because the straight pieces are direct pieces and you only shift the vertices.
I can imagine that at least on some models this would fail.
I think there should be an option to turn off the interpolation between polygons, or to limit when it happens.

Here's some more interesting vases I concocted.
[cube_cylinder_vase.stl.txt](https://github.com/Ultimaker/CuraEngine/files/810425/cube_cylinder_vase.stl.txt)
[kartel_vase.stl.txt](https://github.com/Ultimaker/CuraEngine/files/810424/kartel_vase.stl.txt)

The cube_cylinder_vase is supposed to show that you shouldn't interpolate based on the vertex index, but based on the line distances. I haven't really tested them yet, though.

smartavionics:
So, let's see how we do...

screenshot_2017-03-01_13-30-07

The first layer after the transition has straight lines where there should be curves but it's a moot point because you can't print a curved line with no support.

As you can see, the layers above have the right curve...

screenshot_2017-03-01_13-32-35

The other one looks good to me...

screenshot_2017-03-01_13-28-49

I don't mind having an option to disable the interpolation so that people can choose between either having the more accurate outline with the possibility of a visable seam or the interpolated outline with no visible seam (and waterproof!)

BagelOrb:
I thought about it some more and perhaps you're right. The spiralize vase is a heuristic bunch of code anyway, so maybe we shouldn't put much effort in weird edge cases.

About multithreading: it's about to be merged somewhere this week. I really like to try your changes and have them included in Cura 2.5, which is going to be feature frozen somewhere next week. Try and see if you can get your changes to work if you merge the branch called `feature_omp_writeGcode`.

smartavionics:
OK, I will try that but what I really want to progress is the expand skins into infill PR and the infill/skin angles PR. For me, both of those are far more important than improving spiralize (which I am only working on because some users were having problems with the existing implementation).

src/gcodePlanner.cpp
BagelOrb: Please introduce a variable `coord_t polygon_length = wall.length()` and a variable `polygon_length_traversed`which you update on every iteration of this for loop, so that instead of doing `* i / n_points` you do `* polygon_length_traversed / polygon_length`.

BagelOrb:
> OK, I will try that but what I really want to progress is the expand skins into infill PR and the infill/skin angles PR.

I am trying to get all your PRs in the 2.5 release before it's feature frozen next week.
Indeed this PR has the highest chance of not making it in the 2.5 release because there are quite some code changes and because spiralize mode has a lower priority.

BagelOrb:
I'm going to do a full review of this PR instead of the multithreaded one because the other PR includes all my changes, so that would be hard to review.

There was some change of plans internally, which mean I have to merge all your PRs on Monday. I hope you still have the time to do all the fixes I requested today and/or in the weekend.

smartavionics:
Now interpolates based on distance rather than vertices consumed.

src/gcodePlanner.cpp
```diff
@@ -448,6 +448,80 @@ void GCodePlanner::addLinesByOptimizer(Polygons& polygons, GCodePathConfig* conf
...
+        while (a > M_PI)
+        {
+            // the vertex was on the right of the vector so move the seam vertex on
+            seam_vertex_idx = (seam_vertex_idx + 1) % n_points;
+            a = LinearAlg2D::getAngleLeft(last_wall_seam_vertex_vector, last_wall_seam_vertex, wall[(seam_vertex_idx + 1) % n_points]);
+            //std::cerr << "new angle = " << a * 180/M_PI << "\n";
+
+            if (seam_vertex_idx == first_seam_vertex_idx)
+            {
+                // this shouldn't happen!
```
BagelOrb: Note that the first point to the left of the vector might be pretty far away if the geometry of the next layer is rather different from the last.
Perhaps you should also check for the distance between the last wall seam and the seam vertex, compared to the distance between the last wall seam and the first seam vertex.
If that distance is relatively large, you should revert to using the closest point.

src/gcodePlanner.cpp
```diff
@@ -448,6 +448,80 @@ void GCodePlanner::addLinesByOptimizer(Polygons& polygons, GCodePathConfig* conf
...
+    const int max_dist = MM2INT(config->getLineWidth() * 5);
+    // extrude to the points following the seam vertex
+    // the last point is the seam vertex as the polygon is a loop
+    for (int i = 1; i <= n_points; ++i)
+    {
+        // p is a point from the current wall polygon
+        const Point& p = wall[(seam_vertex_idx + i) % n_points];
+        // last_p is p shifted onto last_wall
+        Point last_p(p);
+        if (last_wall.inside(last_p, false))
```
BagelOrb: You should replace these checks by using `PolygonUtils::findClosest(p, last_wall)` and the nseeing whether the distance is larger than the max_dist.
I think that's more verbose and more direct for what you are trying to do.

src/gcodePlanner.cpp
```diff
@@ -448,6 +448,80 @@ void GCodePlanner::addLinesByOptimizer(Polygons& polygons, GCodePathConfig* conf
...
+        {
+            PolygonUtils::moveOutside(last_wall_polygons, last_p, 0, max_dist);
+        }
+        else if(!last_wall.inside(last_p, true))
+        {
+            PolygonUtils::moveInside(last_wall_polygons, last_p, 0, max_dist);
+        }
+        // now interpolate between last_p and p depending on how far we have progressed along wall
+#if 0
```
BagelOrb: please remove debug code

src/gcodePlanner.cpp
```diff
@@ -448,6 +448,80 @@ void GCodePlanner::addLinesByOptimizer(Polygons& polygons, GCodePathConfig* conf
...
+            {
+                // this shouldn't happen!
+                break;
+            }
+        }
+    }
+
+    Polygons last_wall_polygons;
+    last_wall_polygons.add(last_wall);
+    const int max_dist = MM2INT(config->getLineWidth() * 5);
```
BagelOrb: I think the max_dist should be smaller.
If the shifting distance is more than the line width, the line wouldn't sit on top of the previous layer.

Also the `getLineWidth` function already returns coord_t, rather than mm.

So this line should read:
```
const int max_dist = config->getLineWidth();
```

src/gcodePlanner.cpp
```diff
@@ -448,6 +448,80 @@ void GCodePlanner::addLinesByOptimizer(Polygons& polygons, GCodePathConfig* conf
...
+    // extrude to the points following the seam vertex
+    // the last point is the seam vertex as the polygon is a loop
+    for (int i = 1; i <= n_points; ++i)
+    {
+        // p is a point from the current wall polygon
+        const Point& p = wall[(seam_vertex_idx + i) % n_points];
+        // last_p is p shifted onto last_wall
+        Point last_p(p);
```
BagelOrb: Suggested variable name: `closest_point_last_wall`
Suggested comment: `\\ or p if the closest point is too far`

src/gcodePlanner.cpp
```diff
@@ -515,11 +528,11 @@ int GCodePlanner::spiralizeWallSlice(GCodePathConfig* config, PolygonRef wall, P
         }
         // now interpolate between last_p and p depending on how far we have progressed along wall
 #if 0
-        if(i == n_points)
+        if(index == n_points)
             addTravel_simple(p);
         else
 #endif
-        addExtrusionMove(Point(last_p + (p - last_p) * static_cast<double>(i) / n_points), config, SpaceFillType::Polygons, 1.0, true);
+        addExtrusionMove(last_p + (p - last_p) * (wall_length / total_length), config, SpaceFillType::Polygons, 1.0, true);
```
BagelOrb: Again, you could have measured `wall_length` and `total_length` in integer logic and omitted the brackets on this line to achieve the same result.
The way you did it is also good.

src/gcodePlanner.cpp
```diff
@@ -448,6 +448,96 @@ void GCodePlanner::addLinesByOptimizer(Polygons& polygons, GCodePathConfig* conf
...
 }
 
+int GCodePlanner::spiralizeWallSlice(GCodePathConfig* config, PolygonRef wall, PolygonRef last_wall, int last_wall_seam_vertex_idx)
+{
+    // The spiral has to continue on in an anti-clockwise direction from where the last layer finished, it can't jump backwards
+
+    const int n_points = wall.size();
+
+    // origin is where we start from
+    const Point origin = (last_wall_seam_vertex_idx < 0) ? lastPosition : last_wall[last_wall_seam_vertex_idx];
```
BagelOrb: Note that in a multithreading context, `lastPosition` is *not* the last position in gcode, but the location of the z-seam you set in the frontend.

You really need to compute all start positions before `FffGcodeWriter::processLayer` is called if you want a nice z seam in spiralize vase mode.

Didn't you do this already in the other branch?

src/FffGcodeWriter.cpp
```diff
@@ -1002,52 +1075,88 @@ void FffGcodeWriter::processInsets(GCodePlanner& gcode_layer, SliceMeshStorage*
...
+            const PolygonRef& outer_wall = wall_insets[0][0]; // current layer outer wall outline
+            const PolygonRef& last_outer_wall = (*last_wall_insets)[0][0]; // last layer outer wall outline
+            // output a wall slice that is interpolated between the last and current walls
+            const int seam_vertex_idx = wall_layer.seam_vertex_index; // current layer seam vertex index
+            const int last_seam_vertex_idx = (layer_nr == 0) ? -1 : mesh->layers[layer_nr - 1].seam_vertex_index; // last layer seam vertex index
+            gcode_layer.spiralizeWallSlice(&mesh->inset0_config, outer_wall, last_outer_wall, seam_vertex_idx, last_seam_vertex_idx);
+        }
+        else
+        {
+            int processed_inset_number = -1;
```
BagelOrb: Note to self: all code of this function below is *still(?)* simply the code on the left with an extra indent. No changes at all.

src/gcodePlanner.cpp
```diff
@@ -448,6 +448,56 @@ void GCodePlanner::addLinesByOptimizer(Polygons& polygons, GCodePathConfig* conf
...
+    {
+        // nothing to do
+        return;
+    }
+
+    // extrude to the points following the seam vertex
+    // the last point is the seam vertex as the polygon is a loop
+    double wall_length = 0.0;
+    p0 = origin;
+    for (int index = 1; index <= n_points; ++index)
```
BagelOrb: I still find the variable name 'index' non-descriptive.
A better one would be 'wall_point_idx'.
This is not blocking merging this PR, though.

BagelOrb: Why do we start the point after the seam vertex index? Why do we start at index one rather than zero?

It seems to me we should start at zero and so maybe we should store the `seam_vertex_idx` as one lower than we currently do.

src/FffGcodeWriter.cpp
```diff
@@ -107,6 +111,75 @@ void FffGcodeWriter::writeGCode(SliceDataStorage& storage, TimeKeeper& time_keep
...
+            // now we check that the vertex following the seam vertex is not to the right of the seam vertex in the last layer
+            // and if it is we move forward
+
+            // get the inward normal of the last layer seam vertex
+            Point last_wall_seam_vertex_inward_normal = PolygonUtils::getVertexInwardNormal(last_wall, last_layer.seam_vertex_index);
+
+            // create a vector from the normal so that we can then test the vertex following the candidate seam vertex to make sure it is on the correct side
+            Point last_wall_seam_vertex_vector = last_wall_seam_vertex + last_wall_seam_vertex_inward_normal;
+
+            // now test the vertex following the candidate seam vertex and if it lies to the left or on the vector, it's good to use
```
BagelOrb: Note that the first point to the left of the vector might be pretty far away if the geometry of the next layer is rather different from the last.
Perhaps you should also check for the distance between the last wall seam and the seam vertex, compared to the distance between the last wall seam and the first seam vertex.
If that distance is relatively large, you should revert to using the closest point.

I'm not totally sure you should do this, though. This might be optimizing an edge case for which spiralize doesn't work so good in any case...

src/gcodePlanner.cpp
```diff
@@ -448,6 +448,56 @@ void GCodePlanner::addLinesByOptimizer(Polygons& polygons, GCodePathConfig* conf
...
+    {
+        // p is a point from the current wall polygon
+        const Point& p = wall[(seam_vertex_idx + index) % n_points];
+        wall_length += vSizeMM(p - p0);
+        p0 = p;
+
+        // last_p is p shifted onto last_wall
```
BagelOrb: Suggested variable name: `closest_point_last_wall`
Suggested comment: `\\ or p if the closest point is too far`

src/gcodePlanner.cpp
```diff
@@ -448,6 +448,56 @@ void GCodePlanner::addLinesByOptimizer(Polygons& polygons, GCodePathConfig* conf
...
+        Point last_p(p);
+        if (last_wall.inside(last_p, false))
```
BagelOrb: You should replace these checks by using `PolygonUtils::findClosest(p, last_wall)` and then seeing whether the distance is larger than the max_dist.
I think that's more verbose and more direct for what you are trying to do.

src/gcodePlanner.cpp
BagelOrb: The name `last_p` sounds like it is `p` from the last iteration, which it is not. That's why I am suggesting a different variable name - the current one is confusing.

src/gcodePlanner.cpp
BagelOrb: It's also less computationally expensive...

src/FffGcodeWriter.cpp
```diff
@@ -117,16 +117,24 @@ void FffGcodeWriter::findLayerSeamsForSpiralize(SliceMeshStorage& mesh)
...
+        }
+
+        if (layer_nr < mesh.layers.size())
+        {
+            // use the vertex closest to Point(0, 0) for the first layer that has a part with insets
             // Doing this, we give the user some degree of control as to where the seam will be.
             // By moving the model on the bed they can move the start position of the seam and
             // this could be useful if the spiralization has a problem with a particular seam position
-            mesh.layers[0].seam_vertex_index = PolygonUtils::findClosest(Point(0, 0), mesh.layers[0].parts[0].insets[0][0]).point_idx;
+            mesh.layers[layer_nr].seam_vertex_index = PolygonUtils::findClosest(Point(0, 0), mesh.layers[layer_nr].parts[0].insets[0][0]).point_idx;
```
BagelOrb: There's a frontend setting Z Seam X and Y, which you could use for this.

I'll add that to this PR.

BagelOrb:
I've made some additions here: https://github.com/smartavionics/CuraEngine/pull/1
Please merge that PR.

BagelOrb:
I am getting some rather bad results:
image

smartavionics:
That's odd, it works fine for me. What line width are you using?

BagelOrb:
[UM3_round_tri_vase.curaproject.3mf.txt](https://github.com/Ultimaker/CuraEngine/files/820687/UM3_round_tri_vase.curaproject.3mf.txt)

BagelOrb:
Hit File > Open project...

BagelOrb:
Line widths 0.35

smartavionics:
I loaded that project and, yes, I get similar errors. But for my Kossel XL settings, I get no errors so I am trying to work out what is different.

Unfortunately, when I load your project it overwrites my setting visibility and all the settings are hidden.

BagelOrb:
Oh dear. I think there will be a checkbox which lets you not modify the visibility, in the future...

I generally just use the search feature, which lets you edit settings which are not set to visible.

smartavionics:
OK, a little progress. I can get the errors simply by choosing UM3 machine profile. It appears to break the calculation of the seam vertices which makes the start point of a layer wildly different from the start point of the previous layer. Setting a layer height of 0.1 and line width of 0.35 in my kossel profile doesn't cause the same problem so there is something else that's causing this. Investigating...

smartavionics:
Should be fixed now.

BagelOrb:
testing

BagelOrb:
Ah yes, of course! Wow good catch!

I'm going to freeze all PRs in a couplde of hours. No more community contributions for Cura 2.5 after that.
First I'll test Grow Skin and Retract Before Outer Wall. I sure hope we find a solution for the problem in User Supplied Infill / Skin angles, so that I can properly test it.

BagelOrb:
Frontend visualization is off, but Repetier shows it's a perfect connected line:
image

smartavionics:
Yes, S3D gcode display gets it right too!

BagelOrb:
:1st_place_medal: 
I love this new spiralize! I wonder how other people like the vastly improved functionality.

smartavionics:
Well, there's sure to be models it will have problems with but until people try it, we won't know. But, yes, it's definitely a step forward.

filipgoc:
Nice going @smartavionics !

PR #472, opened 2017-02-27
Feature omp write gcode

BagelOrb:


tests/GcodeLayerThreaderTest.cpp
```diff
@@ -0,0 +1,123 @@
+//Copyright (c) 2015 Ultimaker B.V.
```
Ghostkeeper: We live in 2017. This file is copyrighted two years longer than this header would make it seem.

src/GcodeLayerThreader.h
```diff
@@ -0,0 +1,213 @@
...
+ * A layer_nr index is passed to the item producer in order to produce the different items.
+ * 
+ * Each thread does production and consumption, giving priority to consumption if some is available.
+ * 
+ * If there is only one thread, it consumes every time it has produced one item.
+ * 
+ * \warning This class is only adequate when the expected production time of an item is more than (n_threads - 1) times as much as the expected consumption time of an item
+ */
+template <typename T>
+class GcodeLayerThreader
```
Ghostkeeper: `GcodeLayerThreader` is a very implementation specific class name.

src/GcodeLayerThreader.h
```diff
@@ -0,0 +1,213 @@
...
+    : start_item_argument_index(start_item_argument_index)
+    , end_item_argument_index(end_item_argument_index)
+    , item_count(end_item_argument_index - start_item_argument_index)
+    , max_task_count(max_task_count)
+    , produce_item(produce_item)
+    , consume_item(consume_item)
+    , last_produced_argument_index(start_item_argument_index - 1)
+    {
+        produced.resize(item_count, nullptr);
+    }
```
Ghostkeeper: Implementation in header file.

src/GcodeLayerThreader.h
```diff
@@ -0,0 +1,213 @@
...
+            while (true)
+            {
+                if (finished())
+                {
+                    break;
+                }
+                act();
+            }
+        }
+    }
```
Ghostkeeper: Implementation in header file.

src/GcodeLayerThreader.h
```diff
@@ -0,0 +1,213 @@
...
+        #pragma omp critical
+        {
+            produced[item_idx] = produced_item;
+            if (item_idx == last_consumed_idx + 1 && last_consumed_idx + 1 < end_item_argument_index)
+            {
+                assert(!to_be_consumed_item_idx && "the just produced item shouldn't be consumable already!");
+                to_be_consumed_item_idx = item_idx;
+            }
+        }
+    }
```
Ghostkeeper: Implementation in header file.

src/GcodeLayerThreader.h
```diff
@@ -0,0 +1,213 @@
...
+            last_consumed_idx = item_idx;
+            if (produced[last_consumed_idx + 1] && last_consumed_idx + 1 < end_item_argument_index)
+            {
+                assert(!to_be_consumed_item_idx && "The next produced item shouldn't already be noted as being consumable because of the lock!");
+                to_be_consumed_item_idx = last_consumed_idx + 1;
+            }
+            active_task_count--;
+            assert(active_task_count >= 0);
+        }
+    }
```
Ghostkeeper: Implementation in header file.

src/GcodeLayerThreader.h
```diff
@@ -0,0 +1,213 @@
...
+            }
+            if (item_argument_index && *item_argument_index < end_item_argument_index)
+            {
+                produce(*item_argument_index);
+                return;
+            }
+        }
+        // thread is blocked by too many items being processed
+        std::this_thread::sleep_for(std::chrono::milliseconds(500));
+    }
```
Ghostkeeper: Implementation in header file.

src/GcodeLayerThreader.h
```diff
@@ -0,0 +1,213 @@
...
+    bool finished()
+    {
+        bool finished;
+        #pragma omp critical
+        {
+            finished = last_produced_argument_index >= end_item_argument_index - 1
+                && !to_be_consumed_item_idx;
+        }
+        return finished;
+    }
```
Ghostkeeper: Implementation in header file.

tests/GcodeLayerThreaderTest.h
```diff
@@ -0,0 +1,61 @@
...
+
+    /*!
+     * \brief Tears down the test suite when testing is done.
+     * 
+     * Since <em>StringTest</em> only has static functions, no instance
+     * exists that needs to be destroyed.
+     */
+    void tearDown();
+
+    //These are the actual test cases. The name of the function sort of describes what it tests but I refuse to document all of these, sorry.
```
Ghostkeeper: Hahaha, the name of the function describes what it tests?!?!

tests/GcodeLayerThreaderTest.h
```diff
@@ -0,0 +1,61 @@
+//Copyright (c) 2015 Ultimaker B.V.
```
Ghostkeeper: Wrong copyright year.

tests/GcodeLayerThreaderTest.cpp
```diff
@@ -0,0 +1,123 @@
...
+    }
+    
+    double time_diff = omp_get_wtime() - time_before;
+
+    for (unsigned int idx = 0; idx < times_produced.size(); idx++)
+    {
+        CPPUNIT_ASSERT_MESSAGE(std::string("Items are being produced more than once!"), times_produced[idx] == 1);
+        CPPUNIT_ASSERT_MESSAGE(std::string("Items are being consumed more than once!"), times_consumed[idx] == 1);
+    }
+    std::cerr << "computed in " << time_diff << "s.\n";
```
Ghostkeeper: Remove debug print.

tests/GcodeLayerThreaderTest.cpp
```diff
@@ -0,0 +1,123 @@
...
+    std::vector<int> times_consumed;
+    std::vector<int> times_produced;
+    times_consumed.resize(layer_count, 0);
+    times_produced.resize(layer_count, 0);
+    
+    const std::function<LayerPlan* (int)>& produce_item =
+        [&times_produced, &starting_layer_nr, &avg_computation_time](int layer_nr)
+        {
+            // simulate work:
+            std::this_thread::sleep_for(std::chrono::milliseconds(rand() % (avg_computation_time * 2 * 2 / 3)));
```
Ghostkeeper: Using sleep in a unit test is very bad practice! And `rand()` too.

This unit test is now nondeterministic. And many many times slower.

tests/GcodeLayerThreaderTest.cpp
```diff
@@ -0,0 +1,123 @@
...
+void GcodeLayerThreaderTest::test4()
+{
+    test(10000, 1000, 100);
+}
+
+
+
+void GcodeLayerThreaderTest::test(int avg_computation_time, int layer_count, int max_task_count)
+{
+//     srand(1);
```
Ghostkeeper: Remove commented code.

src/GcodeLayerThreader.h
```diff
@@ -0,0 +1,213 @@
...
+public:
+    /*!
+     * \param start_item_argument_index The first value with which to produce an item
+     * \param end_item_argument_index The last value with which to produce an item
+     * \param produce_item The function with which to produce an item
+     * \param consume_item The function with which to consume an item
+     * \param max_task_count The maximum number of items (being) produced without having been consumed
+     */
+    GcodeLayerThreader(
+        int start_item_argument_index,
```
Ghostkeeper: Won't we just always start at the 0th item? And why is this called `_argument_index`?

src/GcodeLayerThreader.h
```diff
@@ -0,0 +1,213 @@
...
+     * \param item_argument_index The parameter with which to call \ref GcodeLayerThreader::produce_item
+     */
+    void produce(int item_argument_index)
+    {
+        T* produced_item = produce_item(item_argument_index);
+        int item_idx = item_argument_index - start_item_argument_index;
+        #pragma omp critical
+        {
+            produced[item_idx] = produced_item;
+            if (item_idx == last_consumed_idx + 1 && last_consumed_idx + 1 < end_item_argument_index)
```
Ghostkeeper: It's more clear to use `item_idx` in the second check as well.

src/utils/polygon.h
```diff
@@ -36,81 +38,78 @@ typedef std::vector<ListPolygon> ListPolygons; //!< Polygons represented by a ve
...
-    : path(&polygon)
+    ConstPolygonRef(const ClipperLib::Path& polygon)
+    : path(const_cast<ClipperLib::Path*>(&polygon))
     {}
-    
+
+    bool operator==(ConstPolygonRef& other) const =delete;
+
+// needed in std::optional<ConstPolygonRef>
+//     ConstPolygonRef& operator=(const ConstPolygonRef& other) =delete;
```
Ghostkeeper: Remove commented code.

tests/LayerPlanTest.cpp
```diff
@@ -1,21 +1,22 @@
...
 
-CPPUNIT_TEST_SUITE_REGISTRATION(GCodePlannerTest);
+CPPUNIT_TEST_SUITE_REGISTRATION(LayerPlanTest);
 
-void GCodePlannerTest::setUp()
+void LayerPlanTest::setUp()
 {
     SettingsBase settings;
     settings.setSetting("machine_extruder_count", "1");
+//     settings.setSetting(""); speed, line width, acceleration, jerk, flow, layer height
```
Ghostkeeper: Remove debug code and nonsense text.

src/FffGcodeWriter.cpp
```diff
@@ -90,18 +63,54 @@ void FffGcodeWriter::writeGCode(SliceDataStorage& storage, TimeKeeper& time_keep
...
+        {
+            layer_plan_buffer.push(*gcode_layer);
+            LayerPlan* to_be_written = layer_plan_buffer.processBuffer();
+            if (to_be_written)
+            {
+                to_be_written->writeGCode(gcode);
+                delete to_be_written;
+            }
+        };
+    const unsigned int max_task_count = 20; // TODO: hardcoded value
```
Ghostkeeper: So... we going to do this then? We could at least make it a `#DEFINE` or something static.

src/FffGcodeWriter.cpp
```diff
@@ -90,18 +63,54 @@ void FffGcodeWriter::writeGCode(SliceDataStorage& storage, TimeKeeper& time_keep
...
+        // process filler layers to fill the airgap with helper object (support etc) so that they stick better to the raft.
+        process_layer_starting_layer_nr = -Raft::getFillerLayerCount(storage);
     }
-    
+
+
+    const std::function<LayerPlan* (int)>& produce_item =
+        [&storage, total_layers, this](int layer_nr)
+        {
+            LayerPlan& gcode_layer =processLayer(storage, layer_nr, total_layers);
```
Ghostkeeper: Code style: Space after binary operator.

src/main.cpp
```diff
@@ -98,6 +108,14 @@ void connect(int argc, char **argv)
                 case 'v':
                     cura::increaseVerboseLevel();
                     break;
+#ifdef _OPENMP
+                case 'm':
+                    str++;
+                    n_threads = *str - '0';
```
Ghostkeeper: If you just use `strtol` we can actually support more than 9 threads.

src/main.cpp
```diff
@@ -179,6 +201,14 @@ void slice(int argc, char **argv)
                     case 'v':
                         cura::increaseVerboseLevel();
                         break;
+#ifdef _OPENMP
+                    case 'm':
+                        str++;
+                        n_threads = *str - '0';
```
Ghostkeeper: If you just use `strtol` we can actually support more than 9 threads.

src/main.cpp
```diff
@@ -179,6 +201,14 @@ void slice(int argc, char **argv)
...
+                        n_threads = std::max(1, n_threads);
+                        omp_set_num_threads(n_threads);
+                        break;
```
Ghostkeeper: I have a feeling we have duplicate code here. Shouldn't we lift this out of the command-line-options bit and the socket connection bit and into a separate place to parse these arguments?

CMakeLists.txt
```diff
@@ -49,6 +49,7 @@ if (ENABLE_OPENMP)
     FIND_PACKAGE( OpenMP )
     if( OPENMP_FOUND )
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}" )
+        add_definitions(-DENABLE_OPENMP)
```
Ghostkeeper: This definition is unused.

src/settings/PathConfigStorage.cpp
```diff
@@ -0,0 +1,235 @@
...
+            PrintFeatureType::SupportInterface
+            , support_interface_train->getSettingInMicrons("support_interface_line_width")
+            , layer_thickness
+            , support_interface_train->getSettingInPercentage("material_flow")
+            , GCodePathConfig::SpeedDerivatives{support_interface_train->getSettingInMillimetersPerSecond("speed_support_interface"), support_interface_train->getSettingInMillimetersPerSecond("acceleration_support_interface"), support_interface_train->getSettingInMillimetersPerSecond("jerk_support_interface")}
+        )
+{
+    int extruder_count = storage.meshgroup->getExtruderCount();
+    travel_config_per_extruder.reserve(extruder_count);
+    skirt_brim_config.reserve(extruder_count);
```
BagelOrb: done

src/settings/PathConfigStorage.cpp
```diff
@@ -0,0 +1,235 @@
...
+            );
+    }
+
+    mesh_configs.reserve(storage.meshes.size());
+    for (const SliceMeshStorage& mesh_storage : storage.meshes)
+    {
+        mesh_configs.emplace_back(mesh_storage, layer_thickness);
+    }
+
+    int initial_speedup_layer_count = storage.getSettingAsCount("speed_slowdown_layers");
```
BagelOrb: done

src/FffGcodeWriter.h
```diff
@@ -60,33 +60,37 @@ class FffGcodeWriter : public SettingsMessenger, NoCopy
...
+    std::vector<std::vector<unsigned int>> mesh_order_per_extruder; //!< For each extruder, the order of the meshes (first element is first mesh to be printed)
 
     /*!
-     * For each extruder whether priming has already been planned
+     * For each extruder on which layer the prime will be planned,
+     * or a large negative number if it's already planned outside of \ref FffGcodeWriter::processLayer
+     * 
+     * Depending on whether we need to prime on the first layer, or anywhere in the print,
+     * the layer numbers are all zero (or less in case of raft)
+     * or they are the first layer at which the extruder is needed
```
BagelOrb: The UM3 can only prime on layer 0, other printers may not have that restriction.
For those printers this value can be any layer number...

src/FffGcodeWriter.h
```diff
@@ -60,33 +60,37 @@ class FffGcodeWriter : public SettingsMessenger, NoCopy
...
     , layer_plan_buffer(this, gcode)
-    , extruder_prime_is_planned {} // initialize all values in array with [false]
-    , last_position_planned(no_point)
-    , current_extruder_planned(0) // changed somewhere early in FffGcodeWriter::writeGCode
-    , is_inside_mesh_layer_part(false)
     {
+        for (unsigned int extruder_nr = 0; extruder_nr < MAX_EXTRUDERS; extruder_nr++)
+        { // initialize all as max layer_nr, so that they get updated to the lowest layer on which they are used.
+            extruder_prime_layer_nr[extruder_nr] = std::numeric_limits<int>::max();
+        }
```
BagelOrb: done

src/FffGcodeWriter.cpp

src/FffGcodeWriter.cpp
BagelOrb: made `#define`

src/FffGcodeWriter.cpp
```diff
@@ -346,12 +304,20 @@ void FffGcodeWriter::processRaft(SliceDataStorage& storage, unsigned int total_l
...
     
     { // raft base layer
-        
         int layer_nr = initial_raft_layer_nr;
         int layer_height = train->getSettingInMicrons("raft_base_thickness");
         z += layer_height;
         int64_t comb_offset = train->getSettingInMicrons("raft_base_line_spacing");
-        GCodePlanner& gcode_layer = layer_plan_buffer.emplace_back(storage, layer_nr, z, layer_height, last_position_planned, current_extruder_planned, is_inside_mesh_layer_part, fan_speed_layer_time_settings_per_extruder, combing_mode, comb_offset, train->getSettingBoolean("travel_avoid_other_parts"), train->getSettingInMicrons("travel_avoid_distance"));
+
+        std::vector<FanSpeedLayerTimeSettings> fan_speed_layer_time_settings_per_extruder_raft_base = fan_speed_layer_time_settings_per_extruder; // copy so that we change only the local copy
```
BagelOrb: Because the raft fan speed shouldn't influence the fan speed of other layers...

Only the raft use its own fan speeds; the rest use the fan speeds determined by the fan_speed_layer_time settings

src/FffGcodeWriter.cpp
```diff
@@ -346,12 +304,20 @@ void FffGcodeWriter::processRaft(SliceDataStorage& storage, unsigned int total_l
...
+        for (FanSpeedLayerTimeSettings& fsml : fan_speed_layer_time_settings_per_extruder_raft_base)
```
BagelOrb: I was naughty.
`fan_speed_minimal_layer_time_settings` is so long!

src/FffGcodeWriter.cpp
```diff
@@ -401,16 +372,17 @@ void FffGcodeWriter::processRaft(SliceDataStorage& storage, unsigned int total_l
...
 
-        gcode_layer.processFanSpeedAndMinimalLayerTime();
-        gcode_layer.overrideFanSpeeds(train->getSettingInPercentage("raft_interface_fan_speed"));
+        layer_plan_buffer.push(gcode_layer);
+        LayerPlan* to_be_written = layer_plan_buffer.processBuffer();
+        if (to_be_written)
+        {
+            to_be_written->writeGCode(gcode);
+            delete to_be_written;
+        }
```

src/FffGcodeWriter.cpp
```diff
@@ -496,15 +478,20 @@ void FffGcodeWriter::processLayer(SliceDataStorage& storage, int layer_nr, unsig
...
         max_inner_wall_width = std::max(max_inner_wall_width, mesh_settings.getSettingInMicrons((mesh_settings.getSettingAsCount("wall_line_count") > 1) ? "wall_line_width_x" : "wall_line_width_0")); 
     }
     int64_t comb_offset_from_outlines = max_inner_wall_width * 2;
 
+    assert(extruder_order_per_layer_negative_layers.size() + layer_nr >= 0 && "Layer numbers shouldn't get more negative than there are raft/filler layers");
+    const std::vector<unsigned int>& extruder_order =
+        (layer_nr < 0)?
+        extruder_order_per_layer_negative_layers[extruder_order_per_layer_negative_layers.size() + layer_nr]
+        :
+        extruder_order_per_layer[layer_nr];
```
BagelOrb: I found this better, because now we use the same index for the `extruder_order_per_layer` as the actual layer_nr. Otherwise it would get confusing and we could easily make mistakes.

src/FffGcodeWriter.cpp
```diff
@@ -535,52 +521,24 @@ void FffGcodeWriter::processLayer(SliceDataStorage& storage, int layer_nr, unsig
 
         if (layer_nr >= 0)
         {
-            std::vector<unsigned int>& mesh_order = mesh_order_per_extruder[extruder_nr];
-            unsigned int mesh_order_idx_starting_mesh = 0;
-            { // calculate mesh_order_idx_starting_mesh
-                Point layer_start_position = last_position_planned;
-                if (storage.getSettingBoolean("start_layers_at_same_position"))
```
BagelOrb: Exactly, that was the frontend commit I forgot to push ;)
This is *the* negative side effect of this PR.

Good review job!

src/FffGcodeWriter.cpp
```diff
@@ -677,52 +617,116 @@ void FffGcodeWriter::processDraftShield(SliceDataStorage& storage, GCodePlanner&
         }
     }
 
-    gcode_layer.addPolygonsByOptimizer(storage.draft_protection_shield, &storage.skirt_brim_config[0]); //TODO: Skirt and brim configuration index should correspond to draft shield extruder number.
+    gcode_layer.addPolygonsByOptimizer(storage.draft_protection_shield, &gcode_layer.configs_storage.skirt_brim_config[0]); //TODO: Skirt and brim configuration index should correspond to draft shield extruder number.
 }
 
-std::vector<int> FffGcodeWriter::calculateExtruderOrder(SliceDataStorage& storage, int current_extruder)
+void FffGcodeWriter::calculateExtruderOrderPerLayer(const SliceDataStorage& storage)
```
BagelOrb: No it's parsed as 'calculate extruder_order_per_layer`.
Changing calculateExtruderOrder to calculateLayerExtruderOrder

src/FffGcodeWriter.cpp
```diff
@@ -677,52 +617,116 @@ void FffGcodeWriter::processDraftShield(SliceDataStorage& storage, GCodePlanner&
...
         {
-            Mesh& mesh_data = storage.meshgroup->meshes[mesh_idx];
-            Point3 middle = (mesh_data.getAABB().min + mesh_data.getAABB().max) / 2;
+            const Mesh& mesh_data = storage.meshgroup->meshes[mesh_idx];
+            const Point3 middle = (mesh_data.getAABB().min + mesh_data.getAABB().max) / 2;
             mesh_idx_order_optimizer.addItem(Point(middle.x, middle.y), mesh_idx);
         }
     }
     std::list<unsigned int> mesh_indices_order = mesh_idx_order_optimizer.optimize();
+    std::list<unsigned int>::iterator starting_mesh_it = mesh_indices_order.end();
```

src/FffGcodeWriter.cpp
```diff
@@ -677,52 +617,116 @@ void FffGcodeWriter::processDraftShield(SliceDataStorage& storage, GCodePlanner&
...
+            const Point3 middle3 = mesh.getAABB().getMiddle();
+            const Point middle(middle3.x, middle3.y);
+            const coord_t dist2 = vSize2(middle - layer_start_position);
+            if (dist2 < best_dist2)
+            {
+                best_dist2 = dist2;
+                starting_mesh_it = mesh_it;
+            }
+        }
+    }
```
BagelOrb: Might be. Having separate functions is not a goal in itself.
I don't think this function is getting too long to understand without pulling this block into a new function.

Or do you think this function is less readable now?

src/FffGcodeWriter.cpp
```diff
@@ -677,52 +617,116 @@ void FffGcodeWriter::processDraftShield(SliceDataStorage& storage, GCodePlanner&
...
     std::vector<unsigned int> ret;
-    ret.reserve(mesh_indices_order.size());
-    for (unsigned int mesh_order_idx : mesh_indices_order)
+    ret.resize(mesh_indices_order.size());
```
BagelOrb: If I do reserve, I need to use push_back.
I don't really see the problem here.

src/LayerPlanBuffer.h
```diff
@@ -54,38 +60,63 @@ class LayerPlanBuffer : SettingsMessenger
...
-     * Place a new layer plan (GcodePlanner) by constructing it with the given arguments.
-     * Pop back the oldest layer plan is it exceeds the buffer size and write it to gcode.
+     * Push a new layer plan into the buffer
      */
-    template<typename... Args>
-    GCodePlanner& emplace_back(Args&&... constructor_args)
+    void push(LayerPlan& layer_plan)
     {
+        buffer.push_back(&layer_plan);
+    }
```
BagelOrb: The buffer should be private. Made it private.

BagelOrb: Moved to cpp

src/LayerPlanBuffer.h
```diff
@@ -54,38 +60,63 @@ class LayerPlanBuffer : SettingsMessenger
...
+
+    /*!
+     * Process all layers in the buffer
+     * This inserts the temperature commands to start warming for a given layer in earlier layers
```
BagelOrb: done

The spirit is 'Process all layers in the buffer'. I could expand on that.....
I do want to be explicit about what it exactly does, though, because it might come as a surprise.
Fan speeds and travel moves are more logically located in LayerPlan, but are not possible there due to multithreading.

src/GCodePathConfig.h
```diff
@@ -13,97 +13,73 @@ namespace cura
  */
 class GCodePathConfig
 {
-    friend class GCodePlannerTest;
+    friend class LayerPlanTest;
```
BagelOrb: Again?! Still!

The friendship is the other way around than you think.
The test case is meddling with private data because the test case isn't using the normal flow of the engine...

CMakeLists.txt
BagelOrb: done

CMakeLists.txt
```diff
@@ -66,10 +67,9 @@ set(engine_SRCS # Except main.cpp.
     src/FffProcessor.cpp
     src/gcodeExport.cpp
     src/GCodePathConfig.cpp
-    src/gcodePlanner.cpp
     src/infill.cpp
-    src/WallsComputation.cpp
```
BagelOrb: No it's still in there. I think it was double for no reason.

src/LayerPlan.cpp
```diff
@@ -74,52 +74,66 @@ GCodePath* GCodePlanner::getLatestPathWithConfig(GCodePathConfig* config, SpaceF
...
-void GCodePlanner::forceNewPathStart()
+void LayerPlan::forceNewPathStart()
 {
     std::vector<GCodePath>& paths = extruder_plans.back().paths;
     if (paths.size() > 0)
         paths[paths.size()-1].done = true;
 }
 
-GCodePlanner::GCodePlanner(SliceDataStorage& storage, int layer_nr, int z, int layer_thickness, Point last_position, int current_extruder, bool is_inside_mesh, std::vector<FanSpeedLayerTimeSettings>& fan_speed_layer_time_settings_per_extruder, CombingMode combing_mode, int64_t comb_boundary_offset, bool travel_avoid_other_parts, int64_t travel_avoid_distance)
+LayerPlan::LayerPlan(const SliceDataStorage& storage, int layer_nr, int z, int layer_thickness, unsigned int start_extruder, const std::vector<FanSpeedLayerTimeSettings>& fan_speed_layer_time_settings_per_extruder, CombingMode combing_mode, int64_t comb_boundary_offset, bool travel_avoid_other_parts, int64_t travel_avoid_distance)
```
BagelOrb: There are a couple of places where factories could/should be used.
There are a couple of functions which have multiple parameters with defaults, where you would have to reiterate the defaults if you only want a special value for the 4th parameter with default.

I don't think this is one of them, though.
The constructor is basically only used once in processLayer and there all parameters need to be filled in.
It's then also used in processRaft, but I don't think we should introduce default values for when it is called for processing raft, because the main purpose of the constructor is not for usage in raft.

I think this constructor is only used 4 times and there should be no default values, so I don't think there's enough reason to begin a factory.

In other places using factories is long overdue ':{

src/LayerPlan.cpp
```diff
@@ -74,52 +74,66 @@ GCodePath* GCodePlanner::getLatestPathWithConfig(GCodePathConfig* config, SpaceF
...
+, last_planned_extruder_setting_base(storage.meshgroup->getExtruderTrain(start_extruder))
 , comb_boundary_inside(computeCombBoundaryInside(combing_mode))
 , fan_speed_layer_time_settings_per_extruder(fan_speed_layer_time_settings_per_extruder)
 {
-    extruder_plans.reserve(storage.meshgroup->getExtruderCount());
-    extruder_plans.emplace_back(current_extruder, start_position, layer_nr, is_initial_layer, layer_thickness, fan_speed_layer_time_settings_per_extruder[current_extruder], storage.retraction_config_per_extruder[current_extruder]);
+    int current_extruder = start_extruder;
     comb = nullptr;
-    was_inside = is_inside_mesh; 
+    was_inside = true; // not used, because the first travel move is boguous
```
BagelOrb: Boguous is the """official""" spelling. :P

src/LayerPlan.cpp
```diff
@@ -622,8 +635,9 @@ TimeMaterialEstimates GCodePlanner::computeNaiveTimeEstimates()
     return ret;
 }
 
-void GCodePlanner::processFanSpeedAndMinimalLayerTime()
+void LayerPlan::processFanSpeedAndMinimalLayerTime(Point starting_position)
 {
+    extruder_plans[0].start_position = starting_position;
```
BagelOrb: Fixed. the `start_position` wasn't actually used. It seemed to only be used by the functions which are called from this function, so I just gave all of those functions an extra parameter.

BagelOrb: In the end I delted a lot of code because of hits. Including LayerPlan::computeNaiveTimeEstimates and LayerPlanTest unit test....

src/LayerPlanBuffer.cpp
```diff
@@ -17,16 +17,55 @@ void LayerPlanBuffer::flush()
...
-    
 }
 
+void LayerPlanBuffer::addConnectingTravelMove()
+{
+    auto newest_layer_it = --buffer.end();
+    auto prev_layer_it = newest_layer_it;
+    prev_layer_it--;
+    Point destination;
+    {
```
BagelOrb: To show that it doesn't interact with the rest of the code in that function.
Seems a bit redundant indeed.

Want me to refactor it?

src/LayerPlanBuffer.cpp
```diff
@@ -17,16 +17,55 @@ void LayerPlanBuffer::flush()
...
+        const LayerPlan* newest_layer = *newest_layer_it;
+        assert(newest_layer->extruder_plans.front().extruder == (*prev_layer_it)->extruder_plans.back().extruder);
+        const ExtruderPlan& extruder_plan_above = newest_layer->extruder_plans.front();
```
BagelOrb: renamed : `extruder_plan_above_prev_layer`

src/LayerPlanBuffer.cpp
```diff
@@ -17,16 +17,55 @@ void LayerPlanBuffer::flush()
...
+        if (extruder_plan_above.paths.size() == 0)
+        {
+            logWarning("There are empty layers (or layers with empty extruder plans) in the print! Temperature control and cross layer travel moves might suffer.\n");
+            return;
+        }
+        const GCodePath& path_above = extruder_plan_above.paths[0];
+        assert(path_above.points.size() == 1 && "The first move planned should have been a boguous direct travel move");
```
BagelOrb: In this case it's not. It's boguous spelling.

BagelOrb: I could try to leave out the travel move at the start of a layer, and instead have something like `extrusion_start_position` as a field in the layer plan.
However, this wouldn't work so well, because the bogus travel move is only inserted for the first extruder plan in a layer.
When switching extruder we first need to move to the switching bay, then start the new extruder plan with a combing travel move to the first extrusion position. If we add that combing move to the last extruder plan we would comb toward the object before we actually switch the extruder.

I do agree with your comment, but I don't really see a nice way to solve this. I'm not sure we should spend that much time on it, given that the 2.5 beta release is around the corner.

src/LayerPlanBuffer.cpp
```diff
@@ -17,16 +17,55 @@ void LayerPlanBuffer::flush()
...
+        Point first_location_new_layer = path_above.points[0];
+        destination = first_location_new_layer;
+    }
+    LayerPlan* prev_layer = *prev_layer_it;
+    prev_layer->addTravel(destination);
```
BagelOrb: done

src/LayerPlanBuffer.cpp
BagelOrb: did it any way/

src/LayerPlanBuffer.cpp
```diff
@@ -17,16 +17,55 @@ void LayerPlanBuffer::flush()
...
+    auto newest_layer_it = --buffer.end();
+    Point starting_position(0, 0);
+    if (buffer.size() >= 2)
+    {
+        auto prev_layer_it = newest_layer_it;
+        prev_layer_it--;
+        const LayerPlan* prev_layer = *prev_layer_it;
+        starting_position = prev_layer->getLastPosition();
+    }
+    LayerPlan* newest_layer = *newest_layer_it;
```
BagelOrb: refactored function such that its not possible any more
(though the pointer in the buffer is still assumed to point to an actual layer)

src/LayerPlanBuffer.cpp
```diff
@@ -17,16 +17,55 @@ void LayerPlanBuffer::flush()
...
+    newest_layer->processFanSpeedAndMinimalLayerTime(starting_position);
```
BagelOrb: Size 0 handled explicitly.
Size 1 described in new comments.

tests/GcodeLayerThreaderTest.cpp
BagelOrb: done

tests/GcodeLayerThreaderTest.cpp
BagelOrb: Ehh are you sure?
You were telling me that the test case shouldn't be random, so shouldn't I uncomment it then?

tests/GcodeLayerThreaderTest.cpp
BagelOrb: removing test altogether

tests/GcodeLayerThreaderTest.cpp
BagelOrb: removing test altogether

src/GcodeLayerThreader.h
BagelOrb: I did that on purpose. The functionality is very specific to the problem of multithreading gcode writing.

This class works for every pipeline where the production is ordered and each item produced can be identified by an integer in a consecutive list and the production can be performed in parallel, while the consumption cannot be performed in parallel. This class also minimizes the amount of items which are being / have been produced but haven't been consumed yet.
All this is so specific to our problem at hand that I found it weird to give the class a general name.

src/GcodeLayerThreader.h
BagelOrb: When we have a raft, the first time produce is called, it's called with -2 as an argument.
This class calls `produce` with a range of numbers, which can be negative.
It then puts the produced items in a vector, so the first item in the vector has index zero, but the argument used was start_item_argument_index.

src/GcodeLayerThreader.h
BagelOrb: This is a template class. The implementation is going to be in the header file any way, but I could move it outside the class definition to beneath it inside the same header file... Would that be better?

BagelOrb: done that

src/GcodeLayerThreader.h
BagelOrb: done

src/GcodeLayerThreader.h
BagelOrb: done

src/GcodeLayerThreader.h
BagelOrb: done

src/GcodeLayerThreader.h
BagelOrb: done

src/GcodeLayerThreader.h
BagelOrb: done

src/GcodeLayerThreader.h
BagelOrb: done

tests/GcodeLayerThreaderTest.h
BagelOrb: file removed

tests/GcodeLayerThreaderTest.h
BagelOrb: :P

src/utils/polygon.h
BagelOrb: did that along with the rest of the file ;)

tests/LayerPlanTest.cpp
BagelOrb: v

tests/GcodeLayerThreaderTest.cpp
Ghostkeeper: I was proposing to remove it. Not uncomment it.

src/main.cpp
BagelOrb: done

src/main.cpp
BagelOrb: I can't really lift the `case` and the `break` and it would be kind of weird to have a function to change the local variable `str`, so then we would introduce a new function to remove the code duplication of 3 lines.
Seems a bit useless.

src/utils/polygon.h
BagelOrb: Ah no oops. ok now I removed it.

src/FffGcodeWriter.h
Ghostkeeper: Previously this was a boolean and (as far as I know) therefore not possible to prime on a different layer than the first. Is this a new feature that was added at the same time? Or was this something that was unlocked by the changes you made to process the layers in parallel?

src/FffGcodeWriter.cpp
Ghostkeeper: No, well, it's a little long. But semantically putting it in a different function wouldn't make it better really. So let's leave it for now.

src/FffGcodeWriter.cpp
Ghostkeeper: I don't see the problem here either. `push_back` is fine, right?

src/main.cpp
Ghostkeeper: It's not just 3 lines. It's all of the parameter parsing really. But I understand that it's a lot of work. Just to say that I think a refactor is eventually needed.

src/FffGcodeWriter.h
BagelOrb: There used to be a boolean has_already_primed which was set on the first layer for which `processLayer` was called which needed to prime. Now the layers aren't processed in order, so we can't rely on that strategy. There also was a function named something like `extruderNeedsPrimeOnFirstLayer(.)`

BagelOrb: So it was possible to prime on a different layer.

src/FffGcodeWriter.cpp
BagelOrb: Eh ok
done

BagelOrb:
:cheers:

smartavionics:
Hi Tim, I'm seeing some bad gcode appearing after this was merged.

Gcode before omp merge...

screenshot_2017-03-13_11-59-48


Gcode after omp merge

screenshot_2017-03-13_11-50-00

smartavionics:
Hi Tim,

I've just discovered a problem with the spiralize code that has arrived since the omp gcode branch was merged. At the end of each layer it's now lowering the nozzle briefly which it didn't do before, here's some sample gcode...

G1 X-32.926 Y17.406 E10864.19571
G1 X-32.963 Y17.172 Z120.05 E10864.2095
G0 F1800 X-32.963 Y17.172 Z119.85
;TIME_ELAPSED:10282.831496
;LAYER:599
G0 X-32.963 Y17.172 Z120.05
;TYPE:WALL-OUTER
G1 F1004.6 X-31.751 Y5.492 Z120.059 E10864.89315
G1 X-27.688 Y-6.512 Z120.069 E10865.63095
G1 X-27.72 Y-6.688 E10865.64136

Notice that it's only a vertical move back to the initial layer height (x and y don't change).
Do you know what's doing that?

smartavionics:
I've created a new PR that has a fix for this problem - not yet tested much but appears to work OK.

PR #474, opened 2017-03-06
Spaghetti Infill

Ghostkeeper:
This implements a modifier on infill that produces a spaghetti-like infill structure. It can be combined with various infill patterns. The spaghetti infill drops the infill from a larger height so that it doesn't attach to the bottom and becomes chaotic.

Implements CURA-3238.


src/infill/SpaghettiInfill.cpp
```diff
@@ -0,0 +1,139 @@
...
+    coord_t line_width = mesh.getSettingInMicrons("infill_line_width");
+    double layer_height_mm = INT2MM(layer_height);
+    int spaghetti_max_layer_count = std::max(1, static_cast<int>(mesh.getSettingInMicrons("spaghetti_max_height") / layer_height));
+    // TODO: account for the initial layer height
+
+    coord_t filling_area_inset = mesh.getSettingInMicrons("spaghetti_inset");
+    
+    if (mesh.getSettingInAngleDegrees("spaghetti_max_infill_angle") >= 90)
+    {
+        return; // infill cannot be combined into pillars
```
Ghostkeeper: This is something that should be limited from the GUI as well. We now allow an angle of 90 in the GUI.

src/infill/SpaghettiInfill.h
```diff
@@ -0,0 +1,95 @@
...
+class SpaghettiInfill
+{
+public:
+    /*!
+     * Generate the filling areas and corresponding volume to extrude over such areas for spaghetti infill.
+     */
+    static void generateSpaghettiInfill(SliceMeshStorage& mesh);
+
+protected:
+    struct InfillPillar
```
Ghostkeeper: Why is this a struct? It's more than just a POD container for data.

thopiekar:
Images?

src/infill/SpaghettiInfill.cpp
```diff
@@ -0,0 +1,139 @@
...
+                SpaghettiInfill::InfillPillar& pillar = addPartToPillarBase(infill_part, pillar_base, connection_inset_dist);
+                pillar.top_slice_layer_part = &slice_layer_part;
+                pillar.last_layer_added = layer_idx;
+                pillar.layer_count++;
+            }
+        }
+
+
+        // handle finished pillars
+        for (auto it = pillar_base.begin(); it != pillar_base.end();)
```
Ghostkeeper: Bad variable name: `it`. "It" is a clown from a horror series and not applicable to pillars of infill in any way.

src/infill/SpaghettiInfill.cpp
```diff
@@ -0,0 +1,139 @@
...
+    assert(volume > 0.0);
+
+    // get filling area
+    Polygons filling_area = top_part.offset(-filling_area_inset);
+    assert(top_part.size() > 0 && top_part[0].size() > 0 && "the top part must be a non-zero area!");
+    if (filling_area.size() == 0)
+    {
+        AABB aabb(top_part);
+        Point inside = (aabb.min + aabb.max) / 2;
+        if (!top_part.inside(inside))
```
Ghostkeeper: You can use the `moveInside` functionality here instead.

src/infill/SpaghettiInfill.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+SpaghettiInfill::InfillPillar& SpaghettiInfill::addPartToPillarBase(const PolygonsPart& infill_part, std::list<SpaghettiInfill::InfillPillar>& pillar_base, coord_t connection_inset_dist)
+{
+    std::list<SpaghettiInfill::InfillPillar>::iterator ret = pillar_base.end();
+    for (auto it = pillar_base.begin(); it != pillar_base.end(); ++it)
+    {
+        InfillPillar& pillar = *it;
+        if (pillar.isConnected(infill_part))
+        {
+            pillar.total_area_mm2 += INT2MM(INT2MM(infill_part.area()));
```
Ghostkeeper: You later take this area and multiply it by the layer height. If taking initial layer height into account, this property should probably hold the volume rather than the area.

src/sliceDataStorage.cpp
```diff
@@ -22,6 +22,39 @@ Polygons& SliceLayerPart::getOwnInfillArea()
...
+            for (const Polygons& area : infill_area_per_combine)
+            {
+                if (area.size() > 0)
+                {
+                    return true;
+                }
+            }
+        }
+    }
+    return false;
```
Ghostkeeper: It might be worth considering if turning these if-statements around could make the code simpler: E.g. when infill density is 0, return `false`, if spaghetti is enabled but the volumes are empty, return `false`, etc. Perhaps that makes it read a bit more easily.

(Not blocking. And I don't know if that would make it better.)

src/utils/polygon.cpp
```diff
@@ -190,6 +190,22 @@ unsigned int Polygons::findInside(Point p, bool border_result)
     return ret;
 }
 
+int64_t Polygons::polyLineLength() const
+{
+    int64_t length = 0;
```
Ghostkeeper: Unsigned?

(Not blocking.)

src/FffGcodeWriter.cpp
```diff
@@ -898,7 +895,74 @@ void FffGcodeWriter::addMeshPartToGCode(SliceDataStorage& storage, SliceMeshStor
...
+void FffGcodeWriter::processInfill(GCodePlanner& gcode_layer, SliceMeshStorage* mesh, SliceLayerPart& part, unsigned int layer_nr, int infill_line_distance, int infill_overlap, int infill_angle)
+{
+    if (mesh->getSettingBoolean("spaghetti_infill_enabled"))
+    {
+        processSpaghettiInfill(gcode_layer, mesh,  part, layer_nr, infill_line_distance, infill_overlap, infill_angle);
+    }
+    else
+    {
+        processMultiLayerInfill(gcode_layer, mesh,  part, layer_nr, infill_line_distance, infill_overlap, infill_angle);
+        processSingleLayerInfill(gcode_layer, mesh,  part, layer_nr, infill_line_distance, infill_overlap, infill_angle);
```

BagelOrb:
> I see no handling of infill meshes. Is that working all right?

Yes, that is what we are testing with.

> Why do we need to move the nozzle while filling volumes with spaghetti?

That works better if we need to fill a whole volume with spaghetti/material. Otherwise there will be cravices which may not be reached.

> At several places there are now moves of 0.01 or 0.02mm. Isn't a normal extrusion move more appropriate?

There is no infrastructure in the engine to provide an extrusion move without traveling.
I could add an extra field to a `GcodePath` named `purge_amount`.
Do you want me to do that? Or do you think that will clutter the main workings of the engine too much with this feature?

src/FffGcodeWriter.cpp
```diff
@@ -898,7 +895,74 @@ void FffGcodeWriter::addMeshPartToGCode(SliceDataStorage& storage, SliceMeshStor
...
+    const unsigned int infill_line_width = config.getLineWidth();
+    const int64_t z = layer_nr * getSettingInMicrons("layer_height");
+    const int64_t infill_shift = 0;
+    const int64_t outline_offset = 0;
+
+    for (std::pair<PolygonsPart, double>& filling_area : part.spaghetti_infill_volumes)
+    {
+        Polygons infill_lines;
+        Polygons infill_polygons;
+
```

src/FffGcodeWriter.cpp
```diff
@@ -898,7 +895,74 @@ void FffGcodeWriter::addMeshPartToGCode(SliceDataStorage& storage, SliceMeshStor
...
+        const PolygonsPart& area = filling_area.first;
+        const double total_volume = filling_area.second * getSettingAsRatio("spaghetti_flow");
```
BagelOrb: Volume is 3D, area is 2D.
The volume is recorded in mm^3. It is already multiplied by the layer height in SpaghettiInfill.cpp::75

Yes you understood. adding more comments.

src/FffGcodeWriter.cpp
```diff
@@ -898,7 +895,74 @@ void FffGcodeWriter::addMeshPartToGCode(SliceDataStorage& storage, SliceMeshStor
...
+    }
+    else
+    {
+        processMultiLayerInfill(gcode_layer, mesh,  part, layer_nr, infill_line_distance, infill_overlap, infill_angle);
+        processSingleLayerInfill(gcode_layer, mesh,  part, layer_nr, infill_line_distance, infill_overlap, infill_angle);
+    }
+}
+
+void FffGcodeWriter::processSpaghettiInfill(GCodePlanner& gcode_layer, SliceMeshStorage* mesh, SliceLayerPart& part, unsigned int layer_nr, int infill_line_distance, int infill_overlap, int infill_angle)
+{
```
BagelOrb: *not* done (yet)

src/FffGcodeWriter.cpp
```diff
@@ -898,7 +895,74 @@ void FffGcodeWriter::addMeshPartToGCode(SliceDataStorage& storage, SliceMeshStor
...
+        Polygons* perimeter_gaps_output = nullptr;
+        const bool connected_zigzags = true;
+        const bool use_endpieces = false;
+        Infill infill_comp(pattern, area, outline_offset, infill_line_width, infill_line_distance, infill_overlap, infill_angle, z, infill_shift, perimeter_gaps_output, connected_zigzags, use_endpieces);
+        infill_comp.generate(infill_polygons, infill_lines, mesh);
+
+        const coord_t total_length = infill_polygons.polygonLength() + infill_lines.polyLineLength();
+        if (total_length > 0)
+        {
+            const double normal_volume = INT2MM(INT2MM(total_length * infill_line_width)) * mesh->getSettingInMillimeters("layer_height");
```

I don't see why we should convert before multiplying. The multiplication is within bounds of coord_t

src/FffGcodeWriter.cpp
```diff
@@ -898,7 +895,74 @@ void FffGcodeWriter::addMeshPartToGCode(SliceDataStorage& storage, SliceMeshStor
...
+            const float flow_ratio = total_volume / normal_volume;
```
BagelOrb: No it doesn't really make sense to have both these features.

Shall I edit the frontend settings such that when you set the one to true, the other becomes disabled?

BagelOrb: done that

src/FffGcodeWriter.cpp
```diff
@@ -898,7 +895,74 @@ void FffGcodeWriter::addMeshPartToGCode(SliceDataStorage& storage, SliceMeshStor
...
+            assert(flow_ratio / getSettingAsRatio("spaghetti_flow") >= 0.9);
 
+            gcode_layer.addPolygonsByOptimizer(infill_polygons, &config, nullptr, EZSeamType::SHORTEST, Point(0, 0), 0, false, flow_ratio);
+            if (pattern == EFillMethod::GRID || pattern == EFillMethod::LINES || pattern == EFillMethod::TRIANGLES)
+            {
+                gcode_layer.addLinesByOptimizer(infill_lines, &config, SpaceFillType::Lines, mesh->getSettingInMicrons("infill_wipe_dist"), flow_ratio);
+            }
+            else
+            {
+                gcode_layer.addLinesByOptimizer(infill_lines, &config, (pattern == EFillMethod::ZIG_ZAG)? SpaceFillType::PolyLines : SpaceFillType::Lines, 0, flow_ratio);
```
BagelOrb: This is copied from `FffGcodeWriter::processSingleLayerInfill`.
Concentric types and zigzag shouldn't use a wipe; only linear infill types should.

I don't know whether it would work for cubic subdivision, but it should definitely work for tetrahedral and cubic.
Shall I fix this in this issue, or make a new issue for this?

src/FffGcodeWriter.cpp
```diff
@@ -898,7 +895,74 @@ void FffGcodeWriter::addMeshPartToGCode(SliceDataStorage& storage, SliceMeshStor
...
+            }
+        }
+        else
+        {
+            Point middle = const_cast<PolygonsPart&>(area).outerPolygon().centerOfMass();
+            if (!area.inside(middle))
+            {
+                PolygonUtils::ensureInsideOrOutside(area, middle, infill_line_width / 2);
+            }
+            const double normal_volume = INT2MM(INT2MM(10 * infill_line_width)) * mesh->getSettingInMillimeters("layer_height");
```

BagelOrb: Because a length (10) times another lengths makes an area which is in mm^2 rather than in mm.
The length of 10 is now a named variable, to make it more clear.

src/FffGcodeWriter.cpp
```diff
@@ -898,7 +895,74 @@ void FffGcodeWriter::addMeshPartToGCode(SliceDataStorage& storage, SliceMeshStor
...
+            const float flow_ratio = total_volume / normal_volume;
+            gcode_layer.addTravel(middle);
+            gcode_layer.addExtrusionMove(middle + Point(0,10), &config, SpaceFillType::Lines, flow_ratio);
```

src/FffGcodeWriter.h
```diff
@@ -336,7 +336,33 @@ class FffGcodeWriter : public SettingsMessenger, NoCopy
      * 
      */
     void addMeshPartToGCode(SliceDataStorage& storage, SliceMeshStorage* mesh, SliceLayerPart& part, GCodePlanner& gcode_layer, int layer_nr);
-    
+
+    /*!
+     * Add sparse infill for a given part in a layer plan.
```
BagelOrb: removed 'sparse'

src/FffPolygonGenerator.cpp
```diff
@@ -499,18 +500,27 @@ void FffPolygonGenerator::processInfillMesh(SliceDataStorage& storage, unsigned
...
-    //SubDivCube Pre-compute Octree
-    if (mesh.getSettingAsFillMethod("infill_pattern") == EFillMethod::CUBICSUBDIV)
+    // generate spaghetti infill filling areas and volumes
+    if (mesh.getSettingBoolean("spaghetti_infill_enabled"))
     {
-        SubDivCube::precomputeOctree(mesh);
+        SpaghettiInfill::generateSpaghettiInfill(mesh);
     }
+    else
+    {
```
BagelOrb: Ah no gradual infill also doesn't really make sense in this context.

src/FffGcodeWriter.cpp
BagelOrb: fixed in frontend. too high sparse infill errors when you have spaghetti infill inabled

src/infill/SpaghettiInfill.cpp
```diff
@@ -0,0 +1,139 @@
...
+namespace cura {
+
+
+void SpaghettiInfill::generateSpaghettiInfill(SliceMeshStorage& mesh)
+{
+    coord_t layer_height = mesh.getSettingInMicrons("layer_height");
+    coord_t line_width = mesh.getSettingInMicrons("infill_line_width");
+    double layer_height_mm = INT2MM(layer_height);
+    int spaghetti_max_layer_count = std::max(1, static_cast<int>(mesh.getSettingInMicrons("spaghetti_max_height") / layer_height));
+    // TODO: account for the initial layer height
```
BagelOrb: took quite some refactoring, but I did it.

pillars now record volume rather than area and bottom z coord rather than layer_count

src/infill/SpaghettiInfill.cpp
BagelOrb: Well it's an edge case.
I wouldn't know why, but you could want spaghetti infill on every layer for some reason.
This is one way to enforce the spaghetti infill to be generated on each layer..

src/infill/SpaghettiInfill.cpp
BagelOrb: The clown is hiding behind the pillars...........................

BagelOrb: done

src/infill/SpaghettiInfill.cpp
BagelOrb: done

src/infill/SpaghettiInfill.cpp
BagelOrb: yup, that's also what I figured.
Already done it

src/infill/SpaghettiInfill.h
BagelOrb: done

src/sliceDataStorage.cpp
BagelOrb: meh. not done

src/utils/polygon.cpp
BagelOrb: It makes sense to make this coord_t, but coord_t is long long and I don't know whether this is int64_t rather than coord_t for a reason....

BagelOrb: Oh a long long is at least 64 bit, so I'm changing this to coord_t.
I'm not going to introduce an unsigned_coord_t because of this comment.....

src/FffGcodeWriter.cpp
BagelOrb: Discussed this with Jaime.
We concluded that it would be better to not have the FffPolygonGenerator part in the same class as the FffGcodeWriter part.
The thing is that I am using the class as an encapsulated environment. That environment is not usable in the GcodeWriter context.

Do you want me to create a new class `SpaghettiInfillPathGenerator` ?

Ghostkeeper: I think that would be better, yes.

src/FffGcodeWriter.cpp
Ghostkeeper: Better fix it. Being bugless is in the definition of done, after all.

src/infill/SpaghettiInfill.cpp
Ghostkeeper: Right, but that doesn't happen right now because it returns here if `spaghetti_max_infill_angle == 90`.

Ghostkeeper:
> I could add an extra field to a `GcodePath` named `purge_amount`. Do you want me to do that? Or do you think that will clutter the main workings of the engine too much with this feature?

I don't think it would clutter the main workings of the engine too much. But I do agree that it's kind of out of scope for this feature.

src/gcodeExport.cpp
```diff
@@ -556,6 +556,19 @@ void GCodeExport::writeMove(int x, int y, int z, double speed, double extrusion_
...
+        logError("Extrusion rate is infinite!");
+        assert(false && "Infinite extrusion move!");
+        std::exit(1);
+    }
+    if (isnan(extrusion_mm3_per_mm))
+    {
+        logError("Extrusion rate is not a number!");
+        assert(false && "NaN extrusion move!");
+        std::exit(1);
+    }
```
Ghostkeeper: You should put all of these checks in the ASSERT_INSANE_OUTPUT block, right?

src/FffGcodeWriter.cpp
BagelOrb: Yeah, but because the bug is also present in the normal infill code the question is whether I should fix it in this issue, or make a new issue which will be tested separately?

src/gcodeExport.cpp
BagelOrb: Nah. These checks are for *invalid* output, not for highly unlikely output.

src/FffGcodeWriter.cpp

src/FffGcodeWriter.cpp
BagelOrb: Made https://jira.ultimaker.com:8443/browse/CURA-3558

src/infill/SpaghettiInfill.cpp
BagelOrb: I was wrong. This is not a way to enforce spaghetti infill on each layer; it's a way to enforce that the angle doesn't play a role in checking whether parts are connected.

Fixed

PR #475, opened 2017-03-06
Multi-threaded support generation

Ghostkeeper:
This allows CuraEngine to use multiple processes for generating support, reducing computation time if support is enabled.

Implements CURA-3416.

### DIFF
--- a/src/support.cpp
+++ b/src/support.cpp
@@ -244,16 +244,29 @@ void AreaSupport::generateSupportAreas(SliceDataStorage& storage, unsigned int m
         AreaSupport::detectOverhangPoints(storage, mesh, overhang_points, layer_count, supportMinAreaSqrt);
     }
 
-    std::vector<Polygons> basic_overhang_per_layer;
+    std::vector<Polygons> xy_disallowed_per_layer;
     std::vector<Polygons> full_overhang_per_layer;
-    basic_overhang_per_layer.resize(support_layer_count);
+    xy_disallowed_per_layer.resize(support_layer_count);
     full_overhang_per_layer.resize(support_layer_count);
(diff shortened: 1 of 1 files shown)


src/support.cpp
```diff
@@ -240,10 +244,16 @@ void AreaSupport::generateSupportAreas(SliceDataStorage& storage, unsigned int m
...
+    std::vector<Polygons> full_overhang_per_layer;
+    basic_overhang_per_layer.resize(support_layer_count);
+    full_overhang_per_layer.resize(support_layer_count);
+#pragma omp parallel for default(none) shared(basic_overhang_per_layer, full_overhang_per_layer, support_layer_count, storage, mesh, max_dist_from_lower_layer) schedule(dynamic)
+    for (unsigned int layer_idx = 0; layer_idx < support_layer_count; layer_idx++)
     {
-        basic_and_full_overhang_above.push_front(computeBasicAndFullOverhang(storage, mesh, layer_idx, max_dist_from_lower_layer));
+        std::pair<Polygons, Polygons> basic_and_full_overhang = computeBasicAndFullOverhang(storage, mesh, layer_idx, max_dist_from_lower_layer);
+        basic_overhang_per_layer[layer_idx] = basic_and_full_overhang.first;
+        full_overhang_per_layer[layer_idx] = basic_and_full_overhang.second;
```
BagelOrb: You seriously find a vector of pairs better than two vectors?! Why?

In my eyes a better question would be: why was it a queue of pairs in the first place? Because we threw away old data from both at the same time.

Now we don't throw any data away, so that we can compute it all at the same time (at the cost of extra memory usage)

PR #476, opened 2017-03-07
Add only_grow_skin setting

smartavionics:
This is an additional option to the expand/shrink skins code to control whether the shrunk+expanded skin is unioned with the original skin shape or not. Doing the union means that the skin never ends up smaller than it's original size. Not doing the union means that the skin can end up smaller and even get removed completely. Some people could find the ability to remove the skin useful.

### DIFF
--- a/src/skin.cpp
+++ b/src/skin.cpp
@@ -143,12 +143,28 @@ void generateSkinAreas(int layer_nr, SliceMeshStorage& mesh, const int innermost
 
             if (mesh.getSettingBoolean("expand_upper_skins"))
             {
-                upskin = upskin.offset(-min_skin_width_for_expansion).offset(expand_skins_expand_distance).unionPolygons(upskin).intersection(original_outline);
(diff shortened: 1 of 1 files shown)


BagelOrb:
This can't make it into the Cura 2.5 release, I'm afraid. The code base is feature frozen now.

I don't really think this feature is helpful. It can remove skin areas which are actually top-most or bottom-most skin areas, meaning the infill will become visible from the outside. I think the feature is too dangerous to include and I think the added value is too small.

smartavionics:
OK, I understand that it can't go in 2.5 but I don't (yet) agree with you regarding the danger and lack of need for this feature. I will try again when you start working on 2.6.

BagelOrb:
Copy of what I said on your comment on the other PR:

Those people only want to have small skin removed when it's next to infill. When the skin is enclosed directly by the walls, you would be removing the actual skin, which is a bad idea. You would then be able to see the infill when looking at your print.

I think the correct place to fix the issue you refer to is when the skins are generated in the first place. I'm not 100% sure it's possible there, but If it's possible we should do it there, because that would be less of an impact on performance.

PR #479, opened 2017-03-14
Mould

Ghostkeeper:
Adds an option to print a mould for an object rather than the object itself.

### DIFF
--- a/src/Mold.cpp
+++ b/src/Mold.cpp
@@ -23,6 +23,7 @@ void Mold::process(Slicer& slicer, coord_t layer_height, double angle, coord_t w
             mold_outline_above = mold_outline_above.offset(-inset).unionPolygons(model_outlines.offset(width));
             layer.polygons = mold_outline_above.difference(model_outlines);
         }
+        layer.openPolylines.clear();
     }
 
 }
--- a/src/Mold.cpp
+++ b/src/Mold.cpp
@@ -1,0 +1,31 @@
+/** Copyright (C) 2017 Ultimaker - Released under terms of the AGPLv3 License */
+#include "Mold.h"
+#include "utils/intpoint.h"
+
+namespace cura
+{
+
+void Mold::process(Slicer& slicer, coord_t layer_height, double angle, coord_t width, coord_t open_polyline_width)
+{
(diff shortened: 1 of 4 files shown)


src/Mold.cpp
```diff
@@ -0,0 +1,31 @@
...
+#include "utils/intpoint.h"
+
+namespace cura
+{
+
+void Mold::process(Slicer& slicer, coord_t layer_height, double angle, coord_t width, coord_t open_polyline_width)
+{
+    Polygons mold_outline_above; // the outside of the mold
+
+    coord_t inset = tan(angle / 180 * M_PI) * layer_height;
```
Ghostkeeper: This is undefined if `angle == 90`.

src/Mold.h
```diff
@@ -0,0 +1,23 @@
...
+namespace cura
+{
+
+/*!
+ * Class for transforming the outlines of a model into a mold into which the model can be cast.
+ */
+class Mold
+{
+public:
+    static void process(Slicer& slicer, coord_t layer_height, double angle, coord_t width, coord_t open_polyline_width);
```
Ghostkeeper: Missing documentation. I can't guess what `open_polyline_width` does.

BagelOrb:
I prefer Brittish English as well, but in American English it's 'mold', so that's what it will be in Cura then.
http://english.stackexchange.com/questions/27542/mold-vs-mould

Ghostkeeper:
I never complained about your using American in the code ;)

daid:
The "Mal-function", Marvin will be proud ;-)

BagelOrb:
Nono, the voice in the back of my head complained ;)

src/Mold.cpp
BagelOrb: I'm getting a really large number, rather than NaN. M_PI is not exactly pi of course...
I don't think this can fail.
Maybe I should do a `std::max(0, ...)` though

BagelOrb: Ah no. Negative angles are possible if you want a really wide base.
Changed frontend such that the setting is limited to -89 to 90

src/Mold.h
BagelOrb: done

PR #480, opened 2017-03-16
Feature improvement - make retract when moving to outer wall heed min travel distance

smartavionics:
When the retract when moving to outer wall feature was originally implemented, it didn't force the retract for small distance moves. Sometime later, that was added but I now think that it's better not to force a retraction when moving to an outer wall that is less than the retraction min travel distance.

Without this PR, it retracts even when moving between two outer walls that lie next to each other (i.e. the move distance is a single outer wall width) which is plainly ridiculous. I actually see a noticeable reduction in print quality when the retraction is forced for such a small move.

### DIFF
--- a/src/LayerPlan.cpp
+++ b/src/LayerPlan.cpp
@@ -256,8 +256,9 @@ GCodePath& LayerPlan::addTravel(Point p, bool always_retract)
     const RetractionConfig& retraction_config = storage.retraction_config_per_extruder[getExtruder()];
 
     GCodePath* path = getLatestPathWithConfig(&travel_config, SpaceFillType::None);
-    if (always_retract)
+    if (always_retract && (!last_planned_position || !shorterThen(*last_planned_position - p, retraction_config.retraction_min_travel_distance))
     {
+        // path is not shorter than min travel distance, force a retraction
         path->retract = true;
     }
 
@@ -280,7 +281,7 @@ GCodePath& LayerPlan::addTravel(Point p, bool always_retract)
         combed = comb->calc(*last_planned_position, p, combPaths, was_inside, is_inside, retraction_config.retraction_min_travel_distance, via_outside_makes_combing_fail, fail_on_unavoidable_obstacles);
         if (combed)
         {
-            bool retract = always_retract || combPaths.size() > 1;
+            bool retract = path->retract || combPaths.size() > 1;
             if (!retract)
             { // check whether we want to retract
                 if (combPaths.throughAir)
@@ -326,8 +327,8 @@ GCodePath& LayerPlan::addTravel(Point p, bool always_retract)
     }
     
     if (!combed && last_planned_position) {
-        // no combing? always retract!
-        if (always_retract || !shorterThen(*last_planned_position - p, retraction_config.retraction_min_travel_distance))
+        // no combing? retract only when path is not shorter than minimum travel distance
+        if (!shorterThen(*last_planned_position - p, retraction_config.retraction_min_travel_distance))
         {
             if (was_inside) // when the previous location was from printing something which is considered inside (not support or prime tower etc)
             {               // then move inside the printed part, so that we don't ooze on the outer wall while retraction, but on the inside of the print.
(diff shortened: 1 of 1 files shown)


smartavionics:
I am happy that this is now working OK. The problem I was seeing was that the force retraction when moving to outer walls actually reduced print quality in the situation where the move is very short. So if you have a wall whose width is just 2 lines, it was retracting when moving from one line to the other which is obviously nonsense. Instead of introducing yet another setting, it made sense to use the min travel distance setting to provide a lower limit on travel distance before retraction is enforced.

src/LayerPlan.cpp
```diff
@@ -344,18 +345,16 @@ GCodePath& LayerPlan::addTravel(Point p, bool always_retract)
...
-        {
-            if (was_inside) // when the previous location was from printing something which is considered inside (not support or prime tower etc)
-            {               // then move inside the printed part, so that we don't ooze on the outer wall while retraction, but on the inside of the print.
-                assert (extr != nullptr);
-                moveInsideCombBoundary(extr->getSettingInMicrons((extr->getSettingAsCount("wall_line_count") > 1) ? "wall_line_width_x" : "wall_line_width_0") * 1);
-            }
-            path->retract = true;
-            path->perform_z_hop = perform_z_hops;
+    // no combing? retract only when path is not shorter than minimum travel distance
+    if (!combed && !is_first_travel_of_layer && !shorterThen(*last_planned_position - p, retraction_config.retraction_min_travel_distance))
```
BagelOrb: You removed the check whether `last_planned_position` is initialized. `*last_planned_position` might crash the engine.

src/LayerPlan.cpp
```diff
@@ -281,13 +277,18 @@ GCodePath& LayerPlan::addTravel(Point p, bool always_retract)
...
 
-    bool is_first_travel_of_layer = !static_cast<bool>(last_planned_position);
+    const bool is_first_travel_of_layer = !static_cast<bool>(last_planned_position);
     if (is_first_travel_of_layer)
     {
         bypass_combing = true; // first travel move is boguous; it is added after this and the previous layer have been planned in LayerPlanBuffer::addConnectingTravelMove
         first_travel_destination = p;
         first_travel_destination_is_inside = is_inside;
     }
+    else if (always_retract && !shorterThen(*last_planned_position - p, retraction_config.retraction_min_travel_distance))
```
BagelOrb: This code seems a bit misplaced. Why is it not below with the other check for `retraction_min_travel_distance`?

smartavionics: because when always_retract is true you only want to set path->retract if the distance is above the min allowed whether combing is used or not. The test below is only used when not combing.

src/LayerPlan.cpp
smartavionics: But the test does look at is_first_travel_of_layer which is the result of testing that last_planned_position has a value.

BagelOrb: Ah yes. I still think it's safer to double check this, given that some time in the future the value for `is_first_travel_of_layer` might be changed...

It's not required for merging this PR, though.

BagelOrb:
I though you said you would just revert one of my commits. Why didn't you? Did it give too many conflicts? Or is this just a revert?

smartavionics:
So, Tim, are you happy with how this is now? Can it be merged?

BagelOrb:
You didn't answer my question.

I though you said you would just revert one of my commits. Why didn't you? Did it give too many conflicts? Or is this just a revert?

smartavionics:
Sorry, I didn't think I could just revert your change because other stuff had changed in LayerPlan. As it stands now, I am happy with the code.

BagelOrb:
Ok I'll probably merge it on Monday.

Op 24 mrt. 2017 10:28 schreef "Mark Burton" <notifications@github.com>:

Sorry, I didn't think I could just revert your change because other stuff
had changed in LayerPlan. As it stands now, I am happy with the code.

—
You are receiving this because you commented.

<https://github.com/Ultimaker/CuraEngine/pull/480#issuecomment-288973660>,
or mute the thread

PR #481, opened 2017-03-17
Don't add a connecting travel move that doesn't actually move in x/y.

smartavionics:
LayerPlanBuffer::addConnectingTravelMove() was always adding a move from the last planned
position in the current layer to the first position in the next layer. In spiralize mode,
the move isn't required because the next layer starts at the same x/y as the current layer
but, worse than that, the move was causing the nozzle to bob down to the start height of
the current layer even though no x/y movement is required.

### DIFF
--- a/src/LayerPlanBuffer.cpp
+++ b/src/LayerPlanBuffer.cpp
@@ -91,7 +91,7 @@ void LayerPlanBuffer::addConnectingTravelMove(LayerPlan* prev_layer, const Layer
     Point first_location_new_layer = path_above.points[0];
 
     // if the last planned position in the previous layer isn't the same as the first location of the new layer, travel to the new location
-    if(!prev_layer->last_planned_position || *prev_layer->last_planned_position != first_location_new_layer)
+    if (!prev_layer->last_planned_position || *prev_layer->last_planned_position != first_location_new_layer)
     {
         prev_layer->addTravel(first_location_new_layer);
     }
--- a/src/LayerPlanBuffer.cpp
+++ b/src/LayerPlanBuffer.cpp
@@ -90,7 +90,11 @@ void LayerPlanBuffer::addConnectingTravelMove(LayerPlan* prev_layer, const Layer
 
     Point first_location_new_layer = path_above.points[0];
 
-    prev_layer->addTravel(first_location_new_layer);
+    // if the last planned position in the previous layer isn't the same as the first location of the new layer, travel to the new location
+    if(!prev_layer->last_planned_position || *prev_layer->last_planned_position != first_location_new_layer)
+    {
+        prev_layer->addTravel(first_location_new_layer);
+    }
 }
 
 void LayerPlanBuffer::processFanSpeedLayerTime()


BagelOrb:
This seems like a good bug fix, but the main problem is still not solved - if you ask me.
Calling `addTravel` shouldn't let the print head move to a different Z location in the first place.

I've traced the problem down to `LayerPlan::writeGcode::805` which calls `gcodeExport::writeZhopEnd` which moves the nozzle down to the current layer height.
I'm not sure how to fix this. Maybe we should just set the layer height to a new value with every extrusion move, when in spiralize?

Still your code fixes some problems, so I'm merging this.

smartavionics:
OK, well thanks for using this fix for now. I'll leave it up to you whether you want to make further changes.

BagelOrb:
Take a look at my latest commit. I think that should do it.

Op 17 mrt. 2017 10:56 schreef "Mark Burton" <notifications@github.com>:

smartavionics:
Hi Tim, just tried the latest master with spiralize...

2017-03-17 16:09:57,606 - DEBUG - UM.Backend.Backend._backendLog [90]: [Backend] CuraEngine: /home/markb/3dp/CuraEngine.git/src/pathPlanning/../utils/optional.h:129: constexpr T& std::optional<T>::operator*() const & [with T = ClipperLib::IntPoint]: Assertion `instance && "Instance should be instantiated!"' failed.
2017-03-17 16:09:57,829 - DEBUG - CuraEngineBackend.CuraEngineBackend._onBackendQuit [545]: Backend quit with return code -6. Resetting process and socket.

smartavionics:
Actually, spiralize doesn't make any difference, it just crashes.
Commit a38f58da is OK, it's the one after that causes the breakage.

BagelOrb:
Oh darn! That means i will have to bring cake on Monday.

Let's hope i can fix it in the weekend without my computer.

Op 17 mrt. 2017 17:19 schreef "Mark Burton" <notifications@github.com>:

Actually, spiralize doesn't make any difference, it just crashes.
Commit a38f58d
<https://github.com/Ultimaker/CuraEngine/commit/a38f58da3fe3d8c34bfc9b571c77e866957b5dfc>
is OK, it's the one after that causes the breakage.

—
You are receiving this because you modified the open/close state.

<https://github.com/Ultimaker/CuraEngine/pull/481#issuecomment-287400615>,
or mute the thread

BagelOrb:
Ah I found the problem.
https://github.com/Ultimaker/CuraEngine/pull/483

Thank for the heads up!

smartavionics:
Good, it doesn't crash now on the model I am trying (randomvase_V2) but when spiralize is enabled I get hundreds of these messages appearing (one per layer?)

DEBUG - UM.Backend.Backend._backendLog [90]: [Backend] [WARNING] There are empty layers (or layers with empty extruder plans) in the print! Temperature control and cross layer travel moves might suffer.

If spiralize is not enabled, non of these messages appear.

BagelOrb:
That's probably because the spiralized layers have no travel moves in them, so the LayerPlanBuffer doesn't know how to connect two layers.
A spiralized layer has no information on the starting point of the first extrusion move, so the LayerPlanBuffer cannot add a travel move from the end point of the previous layer to the starting point of the next.

We could simply ignore the warning - it shouldn't affect the gcode.

Still my code assumes that each layer starts with a travel move to the starting point of the first extrusion move. If a spiralized layer then has a travel move somewhere in the middle of the layer, the code assumes that that travel move is the travel move to the first extrusion in that layer.

It would be better if a spiralized layer starts with a travel move to the same location as the previous layer should have ended up in.

smartavionics:
Hi Tim,

Look how the spiral layer starts...

void LayerPlan::spiralizeWallSlice(const GCodePathConfig* config, ConstPolygonRef wall, ConstPolygonRef last_wall, const int seam_vertex_idx, const int last_seam_vertex_idx)
{
    const Point origin = (last_seam_vertex_idx >= 0) ? last_wall[last_seam_vertex_idx] : wall[seam_vertex_idx];
    addTravel_simple(origin);

So it does start with a travel move (after the omp gcode merge it required that adding in to stop an assertion)

BagelOrb:
Ah I see. I put my code to recognize the first travel move in addTravel, but not in addTravel_simple.
I will update the code tomorrow, so that spiralize doesn't receive all those warnings any more.

PR #484, opened 2017-03-22
Feature support top bottom speed

BagelOrb:
separate support roofs and bottoms into separate features with different speeds, line widths etc.


src/support.cpp
```diff
@@ -578,62 +589,75 @@ void AreaSupport::handleWallStruts(
...
+        generateSupportInterfaceLayer(support_layers[layer_idx].supportAreas, mesh_outlines, roof_line_width, roofs);
+        support_layers[layer_idx].support_roof.add(roofs);
     }
 }
 
-
+void AreaSupport::generateSupportInterfaceLayer(Polygons& support_areas, const std::vector<Polygons>& colliding_mesh_outlines, const coord_t safety_offset, Polygons& interface)
+{
+    Polygons model;
+    for (const Polygons outline : colliding_mesh_outlines)
```
BagelOrb: I think it's easier and more efficient to just add all outline polygons to one `Polygons` object and call `all_outlines.unionPolygons()` on it once.

Not a 100% sure, but I think that's how clipper works.

src/support.cpp
```diff
@@ -578,62 +589,75 @@ void AreaSupport::handleWallStruts(
...
-
-        const unsigned int top_layer_idx_above = layer_idx + roof_layer_count + z_distance_top;
         const unsigned int bottom_layer_idx_below = std::max(0, int(layer_idx) - int(bottom_layer_count) - int(z_distance_bottom));
-        if (top_layer_idx_above >= supportLayers.size())
-        {
-            continue;
-        }
-        Polygons roofs;
-        if (roof_layer_count > 0)
+        std::vector<Polygons> mesh_outlines;
```
BagelOrb: `Polygons` is already a vector. I don't think you need to keep the outlines of each layer separate. (See also comment below)

src/support.h
```diff
@@ -29,16 +32,44 @@ class AreaSupport {
      */
     static void generateSupportAreas(SliceDataStorage& storage, unsigned int mesh_idx, unsigned int layer_count, std::vector<Polygons>& supportAreas);
 
+    /*!
+     * Generate support bottom areas for a given mesh.
+     *
```
BagelOrb: The descriptions should say what these functions do to the support infill areas.

BagelOrb: These functions also subtract the interface from the support areas, right?

BagelOrb: It's not clear from the description what happens when you call generateBottoms before generateRoofs or the other way around.

It should be explicitly noted where bottoms are generated and how : above the model and within the support areas.

src/support.h
```diff
@@ -29,16 +32,44 @@ class AreaSupport {
...
+     * interface polygons, and this is subtracted from the support areas.
+     *
+     * \param support_areas The areas where support infill is going to be
+     * printed.
+     * \param mesh_outlines The outlines of the mesh above or below the layer
+     * we're generating interface for. These layers determine what areas are
+     * going to be filled with the interface.
+     * \param safety_offset An offset applied to the result to make sure
+     * everything can be printed.
+     * \param interface An output variable to put the resulting polygons in.
```
BagelOrb: You should supply it as output parameter like so: `\param[out] description`
It's also possible to mark a parameter like `\param[in,out]`.
I normally leave out the `\param[in]` because that is the default behaviour of a function; there's no need to mark it as (als0) being an output variable.

src/utils/polygon.h
```diff
@@ -454,6 +457,13 @@ class Polygons
         return paths.size();
     }
 
+    /*!
+     * Convenience function to check if the polygon has no points.
+     *
+     * \return `true` if the polygon has no points, or `false` if it does.
```
BagelOrb: Correct description.

I tend to just not document functions like this. It's just too simple for me :P

(Just a comment, no fix requested)

src/FffGcodeWriter.cpp
```diff
@@ -608,7 +610,6 @@ LayerPlan& FffGcodeWriter::processLayer(const SliceDataStorage& storage, int lay
     }
     int64_t comb_offset_from_outlines = max_inner_wall_width * 2;
 
-    assert(extruder_order_per_layer_negative_layers.size() + layer_nr >= 0 && "Layer numbers shouldn't get more negative than there are raft/filler layers");
```
Ghostkeeper: Because `extruder_order_per_layer_negative_layers.size()` and `layer_nr` are both unsigned, so this assertion is caught by the compiler.

src/FffGcodeWriter.cpp
```diff
@@ -1487,38 +1492,107 @@ bool FffGcodeWriter::addSupportRoofsToGCode(const SliceDataStorage& storage, Lay
...
+        setExtruder_addPrime(storage, gcode_layer, layer_nr, roof_extruder_nr);
+        gcode_layer.addPolygonsByOptimizer(roof_polygons, &gcode_layer.configs_storage.support_roof_config);
+        gcode_layer.addLinesByOptimizer(roof_lines, &gcode_layer.configs_storage.support_roof_config, (pattern == EFillMethod::ZIG_ZAG) ? SpaceFillType::PolyLines : SpaceFillType::Lines);
+        added = true;
+    }
+
+    return added;
+}
+
+bool FffGcodeWriter::addSupportBottomsToGCode(const SliceDataStorage& storage, LayerPlan& gcode_layer, int layer_nr) const
```
Ghostkeeper: I removed some duplication and made the code more compact so that there are now no two consecutive statements that aren't specific to roof/bottom.

src/support.cpp
Ghostkeeper: Indeed, Clipper and std::vector seem to have overloads to deal with this.

src/support.h
Ghostkeeper: > These functions also subtract the interface from the support areas, right?

Yes, both the support roof and support bottom subtract their respective parts from the support areas. I've documented this now. This happens in `generateSupportInterfaceLayer`.

I've worded it differently so that it's now clear that the support roof and bottom are separated from the normal support areas.

src/FffGcodeWriter.cpp
BagelOrb: layer_nr is not unsigned. I've fixed this compiler warning in the mean time on master, by casting the first term to int

src/FffGcodeWriter.cpp
```diff
@@ -1592,6 +1542,31 @@ bool FffGcodeWriter::addSupportBottomsToGCode(const SliceDataStorage& storage, L
...
+        return 0; //Concentric has no rotation.
+    }
+    if (pattern == EFillMethod::TRIANGLES)
+    {
+        return 90; //Triangular support interface shouldn't alternate every layer.
+    }
+
+    for (const SliceMeshStorage& mesh : storage.meshes)
+    {
+        if (mesh.getSettingInMicrons("support_roof_height") >= 2 * getSettingInMicrons("layer_height"))
```
BagelOrb: I don't think this logic makes sense for bottom interface angles.

This function is called for both top and bottom interface.

src/FffGcodeWriter.cpp
```diff
@@ -1460,66 +1466,105 @@ bool FffGcodeWriter::addSupportRoofsToGCode(const SliceDataStorage& storage, Lay
...
-    infill_comp.generate(support_polygons, support_lines);
 
-    if (support_lines.size() > 0 || support_polygons.size() > 0)
+    for (const SliceMeshStorage& mesh : storage.meshes)
     {
-        setExtruder_addPrime(storage, gcode_layer, layer_nr, skin_extruder_nr);
-        gcode_layer.addPolygonsByOptimizer(support_polygons, &gcode_layer.configs_storage.support_interface_config);
-        gcode_layer.addLinesByOptimizer(support_lines, &gcode_layer.configs_storage.support_interface_config, (pattern == EFillMethod::ZIG_ZAG)? SpaceFillType::PolyLines : SpaceFillType::Lines);
-        added = true;
+        if (mesh.getSettingInMicrons("support_roof_height") >= 2 * getSettingInMicrons("layer_height"))
```
BagelOrb: I don't think this logic makes sense for bottom interface angles.

This function is called for both top and bottom interface.

I don't know whether you should include logic for both top and bottom interface here or separate out the settigns from this cuntion and call the function with either top or bottom settings. Haven't really thought about it.

BagelOrb:
Please put the assert back to what it is on master.
See https://github.com/Ultimaker/CuraEngine/blob/master/src/FffGcodeWriter.cpp#L645

src/FffGcodeWriter.cpp
Ghostkeeper: I figured it would make sense to make sure that top and bottom infill fall on top of each other, so they must have the same direction.

BagelOrb: But currently it bases whether it should alternate the direction only on the amount of roof top layers. 
Using multiple bottom layers and a single top layer now results in the bottom layers all using the same direction.

I think the bottom layers should alternate as well is there are enough layers.

Ghostkeeper: This was always the case: https://github.com/Ultimaker/CuraEngine/blame/master/src/FffGcodeWriter.cpp#L1535

PR #485, opened 2017-03-26
Always make support towers

BagelOrb:
Make Support Towers (thicker support towers with a slanted roof up to the to-be-supported part) when support generates thin support.

Also fixes some cases where towers are generated for top surfaces, but not all :(

### DIFF
--- a/src/support.cpp
+++ b/src/support.cpp
@@ -170,7 +170,7 @@ void AreaSupport::generateSupportAreas(SliceDataStorage& storage, unsigned int m
     const int supportMinAreaSqrt = mesh.getSettingInMicrons("support_minimal_diameter");
     const double supportTowerRoofAngle = mesh.getSettingInAngleRadians("support_tower_roof_angle");
     const bool use_towers = mesh.getSettingBoolean("support_use_towers") && supportMinAreaSqrt > 0;
-    const coord_t tower_top_layer_count = 6; // number of layers after which to conclude that a tiny support area needs a tower
+    const unsigned int tower_top_layer_count = 6; // number of layers after which to conclude that a tiny support area needs a tower
 
     const int layerThickness = storage.getSettingInMicrons("layer_height");
     const int supportXYDistance = mesh.getSettingInMicrons("support_xy_distance");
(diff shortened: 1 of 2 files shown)


src/support.cpp
```diff
@@ -428,13 +427,16 @@ std::pair<Polygons, Polygons> AreaSupport::computeBasicAndFullOverhang(const Sli
 void AreaSupport::detectOverhangPoints(
     SliceDataStorage& storage,
     SliceMeshStorage& mesh, 
-    std::vector<std::pair<int, std::vector<Polygons>>>& overhang_points, // stores overhang_points along with the layer index at which the overhang point occurs)
+    std::vector<std::vector<Polygons>>& overhang_points, // stores overhang_points along with the layer index at which the overhang point occurs)
```
Ghostkeeper: Documentation is now outdated.

src/support.cpp
```diff
@@ -170,6 +170,7 @@ void AreaSupport::generateSupportAreas(SliceDataStorage& storage, unsigned int m
     const int supportMinAreaSqrt = mesh.getSettingInMicrons("support_minimal_diameter");
     const double supportTowerRoofAngle = mesh.getSettingInAngleRadians("support_tower_roof_angle");
     const bool use_towers = mesh.getSettingBoolean("support_use_towers") && supportMinAreaSqrt > 0;
+    const coord_t tower_top_layer_count = 6; // number of layers after which to conclude that a tiny support area needs a tower
```
Ghostkeeper: This documentation is incorrect, isn't it? It seems to behave like the support roof height, rather than a simple threshold for tower height of tiny towers.

src/support.cpp
```diff
@@ -294,6 +295,27 @@ void AreaSupport::generateSupportAreas(SliceDataStorage& storage, unsigned int m
...
+                if (poly.area() < supportMinAreaSqrt * supportMinAreaSqrt)
+                {
+                    if (layer_idx < support_layer_count - tower_top_layer_count && layer_idx >= 1)
+                    {
+                        const Polygons& support_layer_above = supportAreas[layer_idx + tower_top_layer_count];
+                        Polygons here;
+                        here.add(poly);
+                        if (here.intersection(support_layer_above).size() > 0)
+                        {
+                            towerRoofs.push_back(here);
```
Ghostkeeper: This just adds an already-existing support area to the tower roofs? What will that do?

(Not blocking, just asking for clarification.)

src/support.cpp
BagelOrb: Fixed 1st issue.

doc is correct. The idea is to start making tower roofs after a tiny support tower has remained tiny for the given amount of layers.

src/support.cpp
BagelOrb: The roofs are expanded until each tower is as big as required.
`towerRoofs` records all polygons which still need to be expanded before finally being added to the support.

src/support.cpp
BagelOrb: fixed

PR #488, opened 2017-03-31
Fix spiralization bugs that bite multi-material models

smartavionics:
The current implementation assumes that the previous layer is in the same mesh as the current layer but that is not true when crossing the boundaries between materials. The main changes in this PR are:

1 - the seam vertex index for each layer is no longer stored in the layer's SliceLayer object. Instead, it is stored in a vector in SliceDataStorage. Another vector stores the wall outline polygons for each layer.

2 - findLayerSeamsForSpiralize() is no longer called for each mesh, it's called once and it processes all the meshes together and stores the seam vertex indices and wall outline polygons for each layer into the SliceDataStorage object.

3 - processInsets() is now passed the SliceDataStorage object and when spiralizing it retrieves the seam vertex index and wall outline polygons for the previous layer from there.

The code appears to work and single extruder prints look the same as before. Testing of multi-extruder prints is ongoing but is hindered by the fact that I don't have access to a multi-extruder machine.
Issues noticed so far are:

1 - When moving away from the spiral at an material change, if combing is enabled, it first moves backwards to a position on the spiral and from there it goes back to the prime position. No z-hop occurs. This tends to leave some extrusion at the point it moves back to. Turning off combing stops it doing that. Instead, it does a z-hop and then moves directly to the prime positions and looks to avoid contacting the spiral again like it does when combing is enabled.
2 - We're seeing some blobs occurring at the point where the material changes. They may change  when the combing is disabled, my tester has yet to provide feedback on that.

To summarise, I'm pretty happy with this code now but testing continues...

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -59,15 +59,12 @@ void FffGcodeWriter::writeGCode(SliceDataStorage& storage, TimeKeeper& time_keep
         processNextMeshGroupCode(storage);
     }
 
-    bool doing_spiralize = false;
     size_t total_layers = 0;
     for (SliceMeshStorage& mesh : storage.meshes)
     {
         total_layers = std::max(total_layers, mesh.layers.size());
 
         setInfillAndSkinAngles(mesh);
-
-        doing_spiralize = doing_spiralize || mesh.getSettingBoolean("magic_spiralize");
     }
     
     gcode.writeLayerCountComment(total_layers);
@@ -82,7 +79,7 @@ void FffGcodeWriter::writeGCode(SliceDataStorage& storage, TimeKeeper& time_keep
     }
     calculateExtruderOrderPerLayer(storage);
 
-    if (doing_spiralize)
+    if (getSettingBoolean("magic_spiralize"))
     {
         findLayerSeamsForSpiralize(storage, total_layers);
     }
@@ -909,7 +906,7 @@ void FffGcodeWriter::addMeshLayerToGCode_meshSurfaceMode(const SliceDataStorage&
 
     EZSeamType z_seam_type = mesh->getSettingAsZSeamType("z_seam_type");
     Point z_seam_pos(mesh->getSettingInMicrons("z_seam_x"), mesh->getSettingInMicrons("z_seam_y"));
-    gcode_layer.addPolygonsByOptimizer(polygons, &mesh_config.inset0_config, nullptr, z_seam_type, z_seam_pos, mesh->getSettingInMicrons("wall_0_wipe_dist"), mesh->getSettingBoolean("magic_spiralize"));
+    gcode_layer.addPolygonsByOptimizer(polygons, &mesh_config.inset0_config, nullptr, z_seam_type, z_seam_pos, mesh->getSettingInMicrons("wall_0_wipe_dist"), getSettingBoolean("magic_spiralize"));
 
     addMeshOpenPolyLinesToGCode(storage, mesh, mesh_config, gcode_layer, layer_nr);
 }
@@ -1040,7 +1037,7 @@ void FffGcodeWriter::addMeshPartToGCode(const SliceDataStorage& storage, const S
     processSkinAndPerimeterGaps(gcode_layer, mesh, mesh_config, part, layer_nr, skin_overlap, skin_angle);
 
     //After a layer part, make sure the nozzle is inside the comb boundary, so we do not retract on the perimeter.
-    if (!mesh->getSettingBoolean("magic_spiralize") || static_cast<int>(layer_nr) < mesh->getSettingAsCount("bottom_layers"))
+    if (!getSettingBoolean("magic_spiralize") || static_cast<int>(layer_nr) < mesh->getSettingAsCount("bottom_layers"))
     {
         gcode_layer.moveInsideCombBoundary(mesh->getSettingInMicrons((mesh->getSettingAsCount("wall_line_count") > 1) ? "wall_line_width_x" : "wall_line_width_0") * 1);
     }
@@ -1151,7 +1148,7 @@ void FffGcodeWriter::processInsets(const SliceDataStorage& storage, LayerPlan& g
     if (mesh->getSettingAsCount("wall_line_count") > 0)
     {
         bool spiralize = false;
-        if (mesh->getSettingBoolean("magic_spiralize"))
+        if (getSettingBoolean("magic_spiralize"))
         {
             if (part.insets.size() == 0)
             {
@@ -1248,7 +1245,7 @@ void FffGcodeWriter::processSkinAndPerimeterGaps(LayerPlan& gcode_layer, const S
 
     constexpr int perimeter_gaps_extra_offset = 15; // extra offset so that the perimeter gaps aren't created everywhere due to rounding errors
     bool fill_perimeter_gaps = mesh->getSettingAsFillPerimeterGapMode("fill_perimeter_gaps") != FillPerimeterGapMode::NOWHERE
-                            && !mesh->getSettingBoolean("magic_spiralize");
+                            && !getSettingBoolean("magic_spiralize");
 
     Point z_seam_pos(0, 0); // not used
     PathOrderOptimizer part_order_optimizer(gcode_layer.getLastPosition(), z_seam_pos, EZSeamType::SHORTEST);
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -180,10 +180,11 @@ void FffGcodeWriter::findLayerSeamsForSpiralize(SliceDataStorage& storage, size_
                 if (!done_this_layer && mesh.layers.size() > layer_nr)
                 {
                     SliceLayer& layer = mesh.layers[layer_nr];
-                    // last_layer_nr will be < 0 until we have processed the first non-empty layer
-                    if (last_layer_nr < 0)
+                    // if the first part in the layer (if any) has insets, process it
+                    if (layer.parts.size() != 0 && layer.parts[0].insets.size() != 0)
                     {
-                        if (layer.parts.size() != 0 && layer.parts[0].insets.size() != 0)
+                        // last_layer_nr will be < 0 until we have processed the first non-empty layer
+                        if (last_layer_nr < 0)
                         {
                             // If the user has specified a z-seam location, use the vertex closest to that location for the seam vertex
                             // in the first layer that has a part with insets. This allows the user to alter the seam start location which
@@ -193,17 +194,13 @@ void FffGcodeWriter::findLayerSeamsForSpiralize(SliceDataStorage& storage, size_
                             {
                                 seam_pos = Point(mesh.getSettingInMicrons("z_seam_x"), mesh.getSettingInMicrons("z_seam_y"));
                             }
-                            storage.spiralize_wall_outlines[layer_nr] = &layer.parts[0].insets[0];
                             storage.spiralize_seam_vertex_indices[layer_nr] = PolygonUtils::findClosest(seam_pos, layer.parts[0].insets[0][0]).point_idx;
-                            last_layer_nr = layer_nr;
-                            // ignore any further meshes/extruders for this layer
-                            done_this_layer = true;
                         }
-                    }
-                    else
-                    {
-                        if (layer.parts.size() != 0 && layer.parts[0].insets.size() != 0)
+                        else
                         {
+                            // note that the code below doesn't assume that last_layer_nr is one less than layer_nr but the print is going
+                            // to come out pretty weird if that isn't true as it implies that there are empty layers
+
                             ConstPolygonRef last_wall = (*storage.spiralize_wall_outlines[last_layer_nr])[0];
                             ConstPolygonRef wall = layer.parts[0].insets[0][0];
                             const int n_points = wall.size();
@@ -236,18 +233,18 @@ void FffGcodeWriter::findLayerSeamsForSpiralize(SliceDataStorage& storage, size_
                                 if (seam_vertex_idx == first_seam_vertex_idx)
                                 {
                                     logWarning("WARNING: findLayerSeamsForSpiralize() failed to find a suitable seam vertex on layer %d\n", layer_nr);
-                                    // this shouldn't happen!
+                                    // this shouldn't happen very often - I have seen it occur when the seam moves into a very sharp corner
                                     break;
                                 }
                             }
 
-                            // store results for use when spiralizing
-                            storage.spiralize_wall_outlines[layer_nr] = &layer.parts[0].insets[0];
                             storage.spiralize_seam_vertex_indices[layer_nr] = seam_vertex_idx;
-                            last_layer_nr = layer_nr;
-                            // ignore any further meshes/extruders for this layer
-                            done_this_layer = true;
                         }
+                        // save the wall outline for this layer so it can be used in the spiralize interpolation calculation
+                        storage.spiralize_wall_outlines[layer_nr] = &layer.parts[0].insets[0];
+                        last_layer_nr = layer_nr;
+                        // ignore any further meshes/extruders for this layer
+                        done_this_layer = true;
                     }
                 }
             }
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -164,7 +164,7 @@ void FffGcodeWriter::findLayerSeamsForSpiralize(SliceDataStorage& storage, size_
         storage.spiralize_seam_vertex_indices[layer_nr] = 0;
 
         // iterate through extruders until we find a mesh that has a part with insets
-        const std::vector<unsigned int> &extruder_order = extruder_order_per_layer[layer_nr];
+        const std::vector<unsigned int>& extruder_order = extruder_order_per_layer[layer_nr];
         for (unsigned int extruder_idx = 0; !done_this_layer && extruder_idx < extruder_order.size(); ++extruder_idx)
         {
             const unsigned int extruder_nr = extruder_order[extruder_idx];
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -59,12 +59,15 @@ void FffGcodeWriter::writeGCode(SliceDataStorage& storage, TimeKeeper& time_keep
         processNextMeshGroupCode(storage);
     }
 
+    bool doing_spiralize = false;
     size_t total_layers = 0;
     for (SliceMeshStorage& mesh : storage.meshes)
     {
         total_layers = std::max(total_layers, mesh.layers.size());
 
         setInfillAndSkinAngles(mesh);
+
+        doing_spiralize = doing_spiralize || mesh.getSettingBoolean("magic_spiralize");
     }
     
     gcode.writeLayerCountComment(total_layers);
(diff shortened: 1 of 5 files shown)


smartavionics:
Hi Tim,
I have just created a new issue (https://github.com/Ultimaker/CuraEngine/issues/489) which I think is actually impossible to achieve due to the intractable problem of the print head smashing into the model when it tries to print a low prime tower. Maybe with some printers and models it could be done but, generally, I think it is a loser.

So, with that in mind, I still think that this PR is worth having in that it does fix some problems with multi-extruder spiralized prints. It just doesn't solve the big prime blob problem. So please consider this PR for merging.

smartavionics:
I am also testing a way of avoiding the big prime blob that occurs at the interface between materials. Because a prime tower is not being used, it just currently moves the new extruder back to the print and does the prime there which produces a huge blob. The change I am working on does the prime extrude into thin air before it moves back to the print. I am hoping that will be easier to clean up than the big blob. I await feedback from my tester.

smartavionics:
Actually, I would label these changes as bug fixes rather than enhancement as it just didn't work right for multi-material prints.

BagelOrb:
The Ultimaker 3 needs to do the prime poop on the very first layer, so pooping mid print shouldn't be a problem from that machine - unless there's a bug.
There's a function `FffGcodeWriter::getExtrudersNeedPrimeDuringFirstLayer` for that.
What machine are you using?

smartavionics:
It wants to do the prime after each extruder change and it was occurring on the print which caused a big blob so my change forces the prime to occur at the extruder change position.

BagelOrb:
We should really make the prime occur on the first layer. That's what the prime poop gcode G280 was made for. It wasn't made to extrude in mid air.

I'll review your code later.

src/FffGcodeWriter.cpp
```diff
@@ -59,16 +59,15 @@ void FffGcodeWriter::writeGCode(SliceDataStorage& storage, TimeKeeper& time_keep
...
     {
         total_layers = std::max(total_layers, mesh.layers.size());
-        if (mesh.getSettingBoolean("magic_spiralize"))
-        {
-            findLayerSeamsForSpiralize(mesh);
-        }
 
         setInfillAndSkinAngles(mesh);
+
+        doing_spiralize = doing_spiralize || mesh.getSettingBoolean("magic_spiralize");
```
BagelOrb: I thin kthat with the new spiralize implementation it doesn't really make sense any more to be able to set spiralize per object.
You can just do `storage.getSettingBoolean("magic_spiralize")` instead of looping over all meshes.

fdmprinter.def.json should then get
```
                    "settable_per_mesh": false,
                    "settable_per_extruder": false
```

src/FffGcodeWriter.cpp
```diff
@@ -143,83 +147,107 @@ void FffGcodeWriter::writeGCode(SliceDataStorage& storage, TimeKeeper& time_keep
...
+    for (unsigned layer_nr = 0; layer_nr < total_layers; ++layer_nr)
+    {
+        bool done_this_layer = false;
 
-            // now test the vertex following the candidate seam vertex and if it lies to the left or on the vector, it's good to use
-            const int first_seam_vertex_idx = seam_vertex_idx;
-            float a = LinearAlg2D::getAngleLeft(last_wall_seam_vertex_vector, last_wall_seam_vertex, wall[(seam_vertex_idx + 1) % n_points]);
+        // default is no information available
+        storage.spiralize_wall_outlines[layer_nr] = nullptr;
+        storage.spiralize_seam_vertex_indices[layer_nr] = 0;
```
BagelOrb: You could also have done `storage.spiralize_wall_outlines.resize(layer_count, nullptr);` instead of reserving and setting separately...

It's not required for you to change this, though.
Just saying...

src/FffGcodeWriter.cpp
```diff
@@ -143,83 +147,107 @@ void FffGcodeWriter::writeGCode(SliceDataStorage& storage, TimeKeeper& time_keep
...
 
-            while (a > M_PI)
+        // iterate through extruders until we find a mesh that has a part with insets
+        const std::vector<unsigned int> &extruder_order = extruder_order_per_layer[layer_nr];
```
BagelOrb: Code style: we view the ampersand as part of the type, so the space should be after the `&`, not before.

src/FffGcodeWriter.cpp
```diff
@@ -143,83 +147,107 @@ void FffGcodeWriter::writeGCode(SliceDataStorage& storage, TimeKeeper& time_keep
...
+                            // in the first layer that has a part with insets. This allows the user to alter the seam start location which
+                            // could be useful if the spiralization has a problem with a particular seam path.
+                            Point seam_pos(0, 0);
+                            if (mesh.getSettingAsZSeamType("z_seam_type") == EZSeamType::USER_SPECIFIED)
+                            {
+                                seam_pos = Point(mesh.getSettingInMicrons("z_seam_x"), mesh.getSettingInMicrons("z_seam_y"));
+                            }
+                            storage.spiralize_seam_vertex_indices[layer_nr] = PolygonUtils::findClosest(seam_pos, layer.parts[0].insets[0][0]).point_idx;
+                        }
+                        else
```
BagelOrb: This whole function is becoming quite large.
Could you please factor a piece of this code into a new function?
Perhaps something like `findSpiralizeLayerStart(...)`

src/LayerPlan.cpp
```diff
@@ -774,6 +774,31 @@ void LayerPlan::writeGCode(GCodeExport& gcode)
                 }
                 gcode.writeTemperatureCommand(prev_extruder, prev_extruder_temp, wait);
             }
+
+            if (!getPrimeTowerIsPlanned())
```
BagelOrb: This code should be removed and the prime poop should instead be performed on the first layer.
There seems to be a bug there, cause the prime poop should already always be planned on the first layer...

src/sliceDataStorage.h
```diff
@@ -195,6 +194,9 @@ class SliceDataStorage : public SettingsMessenger, NoCopy
     std::vector<int> max_print_height_per_extruder; //!< For each extruder the highest layer number at which it is used.
     std::vector<size_t> max_print_height_order; //!< Ordered indices into max_print_height_per_extruder: back() will return the extruder number with the highest print height.
 
+    std::vector<int> spiralize_seam_vertex_indices; //!< the index of the seam vertex for each layer
+    std::vector<Polygons *> spiralize_wall_outlines; //!< the wall outline polygons for each layer
```
BagelOrb: Code style: no space between `Polygons` and `*`

src/FffGcodeWriter.cpp
```diff
@@ -1143,31 +1170,32 @@ void FffGcodeWriter::processInsets(LayerPlan& gcode_layer, const SliceMeshStorag
                 gcode_layer.addPolygonsByOptimizer(part.insets[0], &mesh_config.inset0_config, wall_overlap_computation, EZSeamType::SHORTEST, z_seam_pos, wall_0_wipe_dist);
             }
         }
-        // only spiralize the first part in the mesh, other parts won't be printed
-        if (spiralize && &wall_layer.parts[0] == &part)
+        // Only spiralize the first part in the mesh, any other parts will be printed using the normal, non-spiralize codepath.
+        // This sounds weird but actually does the right thing when you have a model that has multiple parts at the bottom that merge into
+        // one part higher up. Once all the parts have merged, layers above that level will be spiralized
+        if (spiralize && &mesh->layers[layer_nr].parts[0] == &part)
```
BagelOrb: Please move the spiralize functionality here into a new function with a name something like `processSpiralizedWall(...)`

src/LayerPlan.cpp
```diff
@@ -774,6 +774,31 @@ void LayerPlan::writeGCode(GCodeExport& gcode)
...
+                    if (!paths[path_idx].config->isTravelPath())
+                    {
+                        first_extrusion_move_is_spiralized = paths[path_idx].spiralize;
+                        break;
+                    }
+                }
+                if (first_extrusion_move_is_spiralized)
+                {
+                    // set the extrusion temperature and wait for it to be achieved
+                    gcode.writeTemperatureCommand(extruder, extruder_plan.extrusion_temperature.value_or(extruder_plan.required_start_temperature), true);
```
BagelOrb: Temperatures are already handled in a different part of the code.
There is no need to handle them here.

Or is there a bug in the handling of temperature commands?

src/FffGcodeWriter.cpp
smartavionics: But that only initialises any added elements, not existing elements.

src/LayerPlan.cpp
smartavionics: But what's happening is that after each extruder change it was moving back to the print and extruding 16mm of filament. Some of that would be the normal post-retract prime but around 10mm was extra. The code I have added makes that big extrude occur before moving back to the print.

src/LayerPlan.cpp
smartavionics: It was moving back to the print and then waiting to get from the required_start_temp back to the extrusion_temp. I want it to be up to temperature before moving back to the print.

smartavionics:
Tim, I am unsure as to whether the addition in LayerPlan::writeGCode() to force the prime to be done before moving back to the print is a good idea or not and maybe it should be removed again. The big problem is I can't test this myself as I don't have a dual extruder printer. I think the other changes are all basically sound.

src/LayerPlan.cpp
BagelOrb: No, there is a special extruder-change retract which is 16mm.
If you look carefully, each time we switch from T0 to T1 we retract 16mm in total compared to the last extrusion E value and that is also the amount primed before the first extrusion move when we switch back to that extruder.

I don't see a problem when looking at the gcode you posted named 'UM3_SpiralMultiMaterial_5.gcode`...

If there still is a bug with unretraction or with the prime poop then that problem should be solved properly by fixing the unretraction and/or prime poop implementation, not by introducing a hotfix to solve it only for spiralize mode.

src/LayerPlan.cpp
BagelOrb: I think you misread the gcode you generated.
What I see in your generated gcode is that it requires the initial_printing_temp right after switching the extruder and then does an M104 to start heating to the printing_temp without waiting for it to be reached.

Again: if there is a problem with temperature control then that problem should be fixed in the temperature control code. It shouldn't be fixed by introducing a hotfix for spiralize only.

src/LayerPlan.cpp
smartavionics: If you look at the photo(s) posted on the forum thread by the person who is kindly testing the code, it does look rather messy when doing the prime before the move. But is that worse than leaving a big blob on the print?

src/FffGcodeWriter.h
```diff
@@ -472,6 +482,15 @@ class FffGcodeWriter : public SettingsMessenger, NoCopy
...
      */
     void findLayerSeamsForSpiralize(SliceDataStorage& storage, size_t total_layers);
+
+    /*!
+     * Calculate the index of the vertex that is considered to be the seam for the given layer
+     * \param storage where the slice data is stored.
+     * \param mesh the mesh containing the layer of interest
+     * \param layer_nr layer number of the layer whose seam verted index is required
+     * \param last_layer_nr layer number of the previous layer
+     */
```
BagelOrb: Please document the return value, as such:
```
* \return [explanation here]
```

Why is it a signed int, rather than an unsigned int?
Does it mean something when it returns a value below zero? If so, please document that.

src/LayerPlan.cpp
smartavionics: OK, but I was trying to minimize the changes to existing code. So do you think that the temperature change immediately after the extruder swap should be altered to what I have added below? i.e :

`
// set the extrusion temperature and wait for it to be achieved 
gcode.writeTemperatureCommand(extruder, extruder_plan.extrusion_temperature.value_or(extruder_plan.required_start_temperature), true);
`

src/FffGcodeWriter.h
```diff
@@ -398,6 +398,16 @@ class FffGcodeWriter : public SettingsMessenger, NoCopy
      */
     void processInsets(const SliceDataStorage& storage, LayerPlan& gcodeLayer, const SliceMeshStorage* mesh, const PathConfigStorage::MeshPathConfigs& mesh_config, const SliceLayerPart& part, unsigned int layer_nr, EZSeamType z_seam_type, Point z_seam_pos) const;
     
+    /*!
+     * Generate the a spiralized wall for a given layer part.
+     * \param[in] storage where the slice data is stored.
+     * \param gcodeLayer The initial planning of the gcode of the layer.
```
BagelOrb: We generally add the specification if a function parameter is an output value.
Please make this line read
```
     * \param[out] gcodeLayer The initial planning of the gcode of the layer.
```

smartavionics:
OK,  I made the changes you requested. Don't forget there's a little Cura PR to make the spiralize setting non settable for meshes (or extruders).

BagelOrb:
https://github.com/Ultimaker/Cura/pull/1650

BagelOrb:
I merged this to master for 2.6. Sadly we're too far into the process of releasing 2.5, so this cannot be in the 2.5 release.

BagelOrb:
I think we need to fix the upgrade system, so that people who have spiralize enabled on all extruders will have spiralize enabled globally when updating to 2.6..

smartavionics:
OK, that's all fine. Perhaps by the time 2.6 arrives, the issue with the post extruder change prime blobs will have a solution?

BagelOrb:
Maybe, maybe not.
After the second extruder has been inactive for a while, some material has leaked toward the tip of the nozzle. When unretracting, this leaked material exists the nozzle too early and causes the blob.

I think this is a general problem which doesn't affect spiralize alone, but also the normal printing process.

smartavionics:
Tim, the guy who has been testing the multi-material spiralization has posted another thread on the forum about printing objects that are part spiral part non-spiral. It's an interesting idea. Have we done the right thing in stopping spiral mode being settable per mesh?

BagelOrb:
Combining spiralize with normal prints can lead to some weird defects. You
can't satisfy everyone. I think it's okay like it is now.

On the other hand: if it would work, there's really no reason to forbid it.
Be aware that where you switch from nonspirwlize to spiralize, the
spiralized segment is printed twice: one to get from the previous layer
height to the current and one for the next. It's something like that..

So all in all, in takes quite some effort to get it to work properly with
mixed spiralize and nonspirwlize, while the feature was never really
i.plemented for that.

Op 7 apr. 2017 09:48 schreef "Mark Burton" <notifications@github.com>:

BagelOrb:
We're getting some weird results at the moment.
How are layers supposed to be handled which have multiple parts?

I thought you enforced that every layer only contains one single polygon and the rest would simply not be printed...

We are getting results like this:
spiralise
img_5169

The engine also says:
```
2017-04-07 09:39:23,115 - DEBUG - UM.Backend.Backend._backendLog [90]: [Backend] [WARNING] WARNING: findLayerSeamsForSpiralize() failed to find a suitable seam vertex on layer 1
2017-04-07 09:39:23,118 - DEBUG - UM.Backend.Backend._backendLog [90]: [Backend] [WARNING] WARNING: findLayerSeamsForSpiralize() failed to find a suitable seam vertex on layer 9
2017-04-07 09:39:23,120 - DEBUG - UM.Backend.Backend._backendLog [90]: [Backend] [WARNING] WARNING: findLayerSeamsForSpiralize() failed to find a suitable seam vertex on layer 116
2017-04-07 09:39:23,120 - DEBUG - UM.Backend.Backend._backendLog [90]: [Backend] [WARNING] WARNING: findLayerSeamsForSpiralize() failed to find a suitable seam vertex on layer 136
2017-04-07 09:39:23,120 - DEBUG - UM.Backend.Backend._backendLog [90]: [Backend] [WARNING] WARNING: findLayerSeamsForSpiralize() failed to find a suitable seam vertex on layer 178
2017-04-07 09:39:23,135 - DEBUG - UM.Backend.Backend._backendLog [90]: [Backend] [WARNING] WARNING: findLayerSeamsForSpiralize() failed to find a suitable seam vertex on layer 204
```

BagelOrb:
Note: I know that using spiralize on models with multiple parts doesn't really make sense.. Still it would be nice if the engine would either throw away the second part or print it normally without spiralizing it - if that's at all possible.

smartavionics:
No, what happens at the moment is that the outline of the first part is spiralized and any further parts are printed as outlines but not spiralized. On the master branch, processInsets() contains this comment:

        // Only spiralize the first part in the mesh, any other parts will be printed using the normal, non-spiralize codepath.
        // This sounds weird but actually does the right thing when you have a model that has multiple parts at the bottom that merge into
        // one part higher up. Once all the parts have merged, layers above that level will be spiralized

BagelOrb:
What about the print moves between the two models? Shouldn't those be travel moves?

smartavionics:
Ah, I hadn't grokked that it was extruding between models, there. Well, yes, something isn't right. I don't know why it's doing that. I can see that each model is using a different extruder.

Tell me, in the above print, will the extruder order be the same for each layer (i.e. LRLRLR...) or will it be LLRRLLRRLLRR... If the former, the start vertex for each layer should always be on the same print (white or red), if the latter, I can imagine that the start point for each layer will alternate between the white and red models. In that case, I guess it's possible that it's starting the first extrude in the wrong position and extruding to the next vertex which is actually on the other model?

BagelOrb:
The order is LRRLLRRLL etc.

Note that it only failed on layers 1, 9, 116, 136, 178, 204.

smartavionics:
Yes, but it extruded between models on all layers?

BTW, I am not using the Cura master branch at the moment because when I move to it, all of my machines and profiles are no longer visible. So I am stuck using child branches of 2.5 at the moment.

BagelOrb:
There is probably no upgrade functionality yet to move your profiles from 2.5 to master/2.6...
The 2.6 release puts all its profiles and machines etc in a separate folder so that it doesn't affect your 2.5 profiles. You can safely check out master without being scared that that will f up your profiles and everything. That way you can still test something on master, except that you are not easily able to test it on a custom machine.

This PR still introduces problems of extrusion moves between models, as far as I know. If it's not fixed I'm afraid I'll have to revert the merge :(

I'm not sure whether this was also a problem before this PR, though...

BagelOrb:
I actually think the problem is larger than this PR. I think that when you go from a non-spiralized part on the previous layer to a spiralized part on the next layer, you will create an extrusion move toward the first point in the spiralized part. When the previous point was also from a spiralized model this is nice, because it doesn't introduce a travel move. However, when you also have non-spiralized parts on a layer you get really long extrusion moves between models :/

Perhaps we have to accept this as a limitation of spiralize mode, but it sure isn't graceful, behaviour.
We have to think about how to solve this...

smartavionics:
Hi Tim, profile stuff understood, thanks.
Here's what I think about this PR. We should keep this PR (or a variant of it) so that it helps with the multi-extruder spiralized single-model print (we know that priming blobs still need sorting). I don't think we should be concerned with supporting multi-object spiralized prints like your example which has the extrusion moves. Spiralizing doesn't make any sense to me for multiple models whether they use one or two extruders.

BagelOrb:
I agree. That's why I think it would be better to just throw away the second part - given that having two parts only results in bugs at the moment.

I'll leave this as a known limitation of spiralize mode for the moment. Let's hope there aren't a lot of users who try spiralize on such models.

PR #491, opened 2017-04-02
Truncated Octadehron Infill

BagelOrb:
From lulzbot Cura


src/infill.cpp
```diff
@@ -48,12 +52,127 @@ void Infill::generate(Polygons& result_polygons, Polygons& result_lines)
...
+        int x = boundary.min.X-(offset/2);
+        if(it == 1){
+          x = (lineCountX * (lineSpacing)) + boundary.min.X + lineSpacing / 2 - (offset/2);
+        }
+        p.add(Point(x,y));
+        for(unsigned int lx=0; lx < lineCountX; lx++){
+          if(it == 1){
+            side = (2*((lx + ly + it + rtMod + lineCountX) % 2) - 1);
+            y = (ly * lineSpacing) + boundary.min.Y + lineSpacing / 2 + (offset/2 * side);
+            x = ((lineCountX - lx - 1) * lineSpacing) + boundary.min.X + lineSpacing / 2;
```
Ghostkeeper: These updates to `x` and `y` are not trivial and should be documented. You'd expect something to get updated with `lineSpacing` rather than `lineSpacing / 2`.

src/infill.cpp
```diff
@@ -48,12 +52,127 @@ void Infill::generate(Polygons& result_polygons, Polygons& result_lines)
...
+        int y = (ly * lineSpacing) + boundary.min.Y + lineSpacing / 2 - (offset/2 * side);
+        int x = boundary.min.X-(offset/2);
+        if(it == 1){
+          x = (lineCountX * (lineSpacing)) + boundary.min.X + lineSpacing / 2 - (offset/2);
+        }
+        p.add(Point(x,y));
+        for(unsigned int lx=0; lx < lineCountX; lx++){
+          if(it == 1){
+            side = (2*((lx + ly + it + rtMod + lineCountX) % 2) - 1);
+            y = (ly * lineSpacing) + boundary.min.Y + lineSpacing / 2 + (offset/2 * side);
```
Ghostkeeper: Why does the `y` get updated with `offset / 2` but the `x` not?

src/infill.cpp
```diff
@@ -48,12 +52,127 @@ void Infill::generate(Polygons& result_polygons, Polygons& result_lines)
...
+    for(unsigned int ly=0; ly < lineCountY;){
+      for(size_t it = 0; it < 2; ly++, it++){
+        int side = (2*((ly + it + rtMod) % 2) - 1);
+        int y = (ly * lineSpacing) + boundary.min.Y + lineSpacing / 2 - (offset/2 * side);
+        int x = boundary.min.X-(offset/2);
+        if(it == 1){
+          x = (lineCountX * (lineSpacing)) + boundary.min.X + lineSpacing / 2 - (offset/2);
+        }
+        p.add(Point(x,y));
+        for(unsigned int lx=0; lx < lineCountX; lx++){
```
Ghostkeeper: This loop seems to do its very best to prevent some minor code duplication but as an end result there's so much magic going on that simply listing the 8 sides of the polygon in order would be much simpler to read and easier to maintain.

src/infill.cpp
```diff
@@ -48,12 +52,127 @@ void Infill::generate(Polygons& result_polygons, Polygons& result_lines)
...
+        y = (ly * lineSpacing) + boundary.min.Y + lineSpacing / 2 - (offset/2 * side);
+        p.add(Point(x,y));
+      }
+    }
+    // Generate tops / bottoms of octohedrons
+    if(abs((abs(offset) - Zscale/4)) < (extrusionWidth/2)){
+      uint64_t startLine = (offset < 0) ? 0 : 1;
+      uint64_t coverWidth = OCTSLEN(lineSpacing);
+      std::vector<Point> points;
+      for(size_t xi = 0; xi < (lineCountX+1); xi++){
```
Ghostkeeper: `xi`? Write these variables out completely.

src/infill.cpp
```diff
@@ -48,12 +52,127 @@ void Infill::generate(Polygons& result_polygons, Polygons& result_lines)
...
+        for(size_t yi = 0; yi < (lineCountY); yi += 2){
+          points.push_back(Point(boundary.min.X + OCTDLEN(lineSpacing)
+                                 + (xi - startLine + rtMod) * lineSpacing,
+                                 boundary.min.Y + OCTDLEN(lineSpacing)
+                                 + (yi + (xi%2)) * lineSpacing
+                                 + extrusionWidth/2));
+        }
+      }
+      uint64_t order = 0;
```
Ghostkeeper: Seems like a boolean would be enough for a variable that can only be 0 or 1.

Ghostkeeper: Order is a bit of a weird variable name for a variable that alternates between 0 and 1. In this case, `go_left_to_right` seems more appropriate.

src/infill.cpp
```diff
@@ -48,12 +52,127 @@ void Infill::generate(Polygons& result_polygons, Polygons& result_lines)
...
+      for(Point pp : points){
+        PolygonRef p = po.newPoly();
+        for(size_t yi = 0; yi <= coverWidth; yi += extrusionWidth) {
+          if(order == 0){
+            p.add(Point(pp.X, pp.Y + yi));
+            p.add(Point(pp.X + coverWidth + extrusionWidth, pp.Y + yi));
+          } else {
+            p.add(Point(pp.X + coverWidth + extrusionWidth, pp.Y + yi));
+            p.add(Point(pp.X, pp.Y + yi));
```
Ghostkeeper: You're adding horizontal lines here. Diagonal lines generally produce better results due to being able to accelerate on two axes.

src/infill.cpp
```diff
@@ -4,6 +4,10 @@
 #include "utils/polygonUtils.h"
 #include "utils/logoutput.h"
 
+#define SQRT2MUL(x) ((30547*(x))/21600)
+#define OCTSLEN(x) ((43200*(x))/73747)
```
BagelOrb: sqrt(2) / (1 + sqrt(2))
WolframAlpha FTW!

Ghostkeeper: Or maybe it's `2 - sqrt(2)`? Perhaps `1 - (1 / (1 + sqrt(2)))`?

BagelOrb: Yes. Inclusive.
Or it's `1/sqrt(3)` which is the ratio between the side and the diagonal of a cube, but with severe rounding errors :P

Appesteijn:
Too much issues, no visible change.

BagelOrb:
I don't know whether this PR needs to be open for us to get email notifications..

Ghostkeeper:
No we get e-mail notifications regardless of whether an issue is open or closed.

PR #495, opened 2017-04-08
SpaghettiInfill: Replacing isnan with std::isnan

thopiekar:
Currently using 'isnan' gives an compiling error in one of my PPA builds.
It seems that the function has only been renamed in more recent compiler versions, so nothing special to say here.

Signed-off-by: Thomas Karl Pietrowski <thopiekar@gmail.com> (github: thopiekar)


BagelOrb:
Shouldn't the same hold for `isnan`?

BagelOrb:
`isinf` I mean..

thopiekar:
Could be, haven't tested my PR here locally. In that case I'll get another broken build tomorrow :sweat_smile:

Ghostkeeper:
I don't know what problems Thopiekar encountered on his PPA, but on Windows builds, `isinf` indeed has the same problem as `isnan`. Both will need `std::`.

thopiekar:
Already done.

PR #498, opened 2017-04-13
Feature tests

jackha:
Added a bunch of unit tests:
- Settings
- AABB
- AABB3D

### DIFF
--- a/tests/utils/AABB3DTest.h
+++ b/tests/utils/AABB3DTest.h
@@ -1,0 +1,59 @@
+//Copyright (c) 2017 Ultimaker B.V.
+//CuraEngine is released under the terms of the AGPLv3 or higher.
+//Use this file as a template for testing new classes
+
+#ifndef TEMPLATE_TEST_H
+#define TEMPLATE_TEST_H
+
+#include <cppunit/TestFixture.h>
+#include <cppunit/extensions/HelperMacros.h>
+
+#include <../src/utils/AABB3D.h>
+
+namespace cura
+{
+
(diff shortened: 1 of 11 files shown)


tests/utils/AABB3DTest.h
```diff
@@ -0,0 +1,58 @@
...
+    void includeZTest();
+    void offsetTest();
+    void offsetTest2();
+    void expandXYTest();
+
+private:
+    /*!
+     * \brief The maximum allowed error in floats.
+     */
+    static constexpr double epsilon = 0.000001f;
```
BagelOrb: epsilon should be of coord_t rather than floating point.
You are only comparing it to values of coord_t...

jackha: Can you show me how? If I change it some assertions fail and I do not understand why...

Ghostkeeper: `coord_t` is an integer-type so an epsilon of 0.000001 wouldn't make sense. It just gets rounded down to 0. If it's changed to `coord_t` it has a different function.

In this particular case, `epsilon` is being used both to compare the values of transformation matrices (in which case 0.000001f makes sense) and to compare the values of (integer) coordinates in microns (in which case an epsilon of 10 or 15 makes more sense).

BagelOrb:
Made a JRIA ticket: CURA-3830

ChrisTerBeke:
This was closed in Jira because of no activity. It also has conflicts. If we really still want this we can re-open.

PR #499, opened 2017-04-18
CURA-3688: Fix compiler warnings

LipuFei:
Still one warning remains, but that's caused by the generated code of protobuf:
```
CuraEngine/build/Cura.pb.cc:6234:48: warning: unused parameter 'output' [-Wunused-parameter]
     ::google::protobuf::io::CodedOutputStream* output) const {
                                                ^
```

The callback functions in `Wireframe2gcode.cpp` can be bound to be member functions, which requires a self-reference at the first parameter, so to suppress those warnings, I removed the name of the parameter.

### DIFF
--- a/src/ExtruderTrain.h
+++ b/src/ExtruderTrain.h
@@ -10,7 +10,6 @@ namespace cura
 class ExtruderTrain : public SettingsBase
 {
     int extruder_nr;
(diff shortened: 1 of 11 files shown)


Appesteijn:
👍 No more warnings on MacOS

PR #501, opened 2017-04-19
Fix crash when definition has more extruder trains than extruder count

fieldOfView:
CuraEngine crashes when a definition defines more extruder trains in its def.json file metadata than specified in the machine_extruder_count setting. This PR fixes that crash, ignoring settings for the extra extruder_trains.

See https://github.com/Ultimaker/Cura/pull/1521 for a usecase of having a lower machine_extruder_count than defined extruder_trains: more easily defining printers with a variable number of extruder trains (eg custom dual- or quad-extrusion printers)

### DIFF
--- a/src/commandSocket.cpp
+++ b/src/commandSocket.cpp
@@ -429,7 +429,7 @@ void CommandSocket::handleObjectList(cura::proto::ObjectList* list, const google
             int extruder_nr = extruder.id();
             if (extruder_nr >= extruder_count)
             {
-                logWarning("Definition has more extruder trains than extruder count suggests, ignoring extra extruder trains.");
+                logWarning("Definition has more extruder trains than extruder count suggests, ignoring extra extruder trains.\n");
                 break;
             }
             ExtruderTrain* train = meshgroup->getExtruderTrain(extruder_nr);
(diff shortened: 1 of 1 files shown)


LipuFei:
@fieldOfView just curious, so there's a concurrency issue with `FffProcessor::getInstance()->getSettingAsCount("machine_extruder_count")`?

LipuFei:
Ah, sorry... I was thinking about something else...

BagelOrb:
I can't imagine this fixes all problems for such a use case, but the code looks good this way...

Ghostkeeper:
Seems to work fine, actually. I added a newline after the warning though because the logWarning doesn't do that by itself.

fieldOfView:
@bagelorb: It took me quite a while of hunting for this one place where CuraEngine relies on the number of extruder definitions matches the machine_extruder_count. CuraEngine uses machine_extruder_count very consistently, kudos for that ;-)

BagelOrb:
@fieldOfView are you kidding? Are you sarcastic? It couldn't possibly be the case that something is consistent in the engine! :D

fieldOfView:
(by now I have found a way to fix the issue from Cura, but it never hurts to be defensive and fix it on both ends)

PR #503, opened 2017-04-23
Fix bug in vertex position test in findLayerSeamsForSpiralize() (2.5 branch only)

smartavionics:
When testing to ensure that the vertex that follows the seam vertex for the current layer
did not fall "behind" the position of the last vertex in the previous layer, it was testing
to see if the point lay to the left of a vector or on that vector. It should have only
succeeded if the point was to the left of the vector and not actually on the vector.

A similar fix needs to be created for the master branch.

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -191,8 +191,8 @@ void FffGcodeWriter::findLayerSeamsForSpiralize(SliceMeshStorage& mesh)
 
             int seam_vertex_idx = PolygonUtils::findClosest(last_wall_seam_vertex, wall).point_idx;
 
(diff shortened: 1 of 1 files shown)


smartavionics:
This fixes some of the problems reported in https://github.com/Ultimaker/Cura/issues/1728

Ghostkeeper:
This is not going to be merged into 2.5, since 2.5 is already released and we don't plan on making a 2.5.1. Could you aim it at Master instead?

smartavionics:
Already done,  https://github.com/Ultimaker/CuraEngine/pull/504
I produced PRs for both branches because I thought maybe you would issue 2.5.1... if you had any other defects you wanted to correct ASAP.

Ghostkeeper:
No, I'm sorry. I'll close this one up then. I hadn't realised they were the same fix.

PR #504, opened 2017-04-23
Fix bug in vertex position test in findSpiralizedLayerSeamVertexIndex().

smartavionics:
When testing to ensure that the vertex that follows the seam vertex for the current layer
did not fall "behind" the position of the last vertex in the previous layer, it was testing
to see if the point lay to the left of a vector or on that vector. It should have only
succeeded if the point was to the left of the vector and not actually on the vector.
Fixes one of the problems reported in https://github.com/Ultimaker/Cura/issues/1728.


Ghostkeeper:
Devs, see CURA-3727.

PR #505, opened 2017-04-25
CURA-3634 Make prime blob optional

LipuFei:
Linked to: https://github.com/Ultimaker/Cura/pull/1713

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -712,40 +712,29 @@ LayerPlan& FffGcodeWriter::processLayer(const SliceDataStorage& storage, int lay
     return gcode_layer;
 }
 
-bool FffGcodeWriter::getAnyExtruderNeedPrimePoopDuringFirstLayer(const SliceDataStorage& storage) const
+bool FffGcodeWriter::getExtruderNeedPrimeBlobDuringFirstLayer(const SliceDataStorage& storage, uint32_t extruder_nr) const
 {
-    bool need_prime = false;
+    bool need_prime_blob = false;
     switch (gcode.getFlavor())
     {
         case EGCodeFlavor::GRIFFIN:
-            need_prime = true;
+            need_prime_blob = true;
             break;
         default:
-            need_prime = false; // TODO: change this once priming for other firmware types is implemented
+            need_prime_blob = false; // TODO: change this once priming for other firmware types is implemented
             break;
     }
 
-    // check the settings if the prime poop is disabled
-    if (need_prime)
+    // check the settings if the prime blob is disabled
+    if (need_prime_blob)
     {
-        bool has_extruder_that_needs_prime = false;
-        const std::vector<bool> extruders_is_used_overall = storage.getExtrudersUsed();
+        const bool is_extruder_used_overall = storage.getExtrudersUsed()[extruder_nr];
+        const bool extruder_prime_blob_enabled = storage.getExtruderPrimeBlobEnabled(extruder_nr);
 
-        for (uint32_t extruder_nr = 0; extruder_nr < extruders_is_used_overall.size(); ++extruder_nr)
-        {
-            const bool extruder_prime_poop_enabled = storage.getExtruderPrimePoopEnabled(extruder_nr);
-            has_extruder_that_needs_prime = has_extruder_that_needs_prime
-                || (extruders_is_used_overall[extruder_nr] && extruder_prime_poop_enabled);
-            if (has_extruder_that_needs_prime)
-            {
-                break;
-            }
-        }
-
-        need_prime = has_extruder_that_needs_prime;
+        need_prime_blob = is_extruder_used_overall && extruder_prime_blob_enabled;
     }
 
-    return need_prime;
+    return need_prime_blob;
 }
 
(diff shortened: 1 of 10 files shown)


src/FffGcodeWriter.h
```diff
@@ -225,11 +225,11 @@ class FffGcodeWriter : public SettingsMessenger, NoCopy
     LayerPlan& processLayer(const SliceDataStorage& storage, int layer_nr, unsigned int total_layers) const;
 
     /*!
-     * Whether the extruders need to be primed separately just before they are used.
+     * Whether any extruder needs to be primed separately just before they are used.
```
BagelOrb: This description should note that this functions checks whether priming should happen on the first layer.

Actually an extruder should always be primed before it starts, but that priming may or may not include a poop.

src/FffGcodeWriter.cpp
```diff
@@ -712,15 +712,39 @@ LayerPlan& FffGcodeWriter::processLayer(const SliceDataStorage& storage, int lay
...
+
+    // check the settings if the prime poop is disabled
+    if (need_prime)
+    {
+        bool has_extruder_that_needs_prime = false;
+        const std::vector<bool> extruders_is_used_overall = storage.getExtrudersUsed();
+
+        for (uint32_t extruder_nr = 0; extruder_nr < extruders_is_used_overall.size(); ++extruder_nr)
+        {
+            const bool extruder_prime_enabled = storage.getExtruderPrimeEnabled(extruder_nr);
```
BagelOrb: This variable name is not correct. It should be extruder_prime_poop_enabled

The word poop is important here. An extruder should always be primed before it starts.
Such priming may include extruder start gcode, the G280 gcode or it may even be empty.

src/FffGcodeWriter.cpp
```diff
@@ -712,15 +712,39 @@ LayerPlan& FffGcodeWriter::processLayer(const SliceDataStorage& storage, int lay
     return gcode_layer;
 }
 
-bool FffGcodeWriter::getExtrudersNeedPrimeDuringFirstLayer() const
+bool FffGcodeWriter::getAnyExtruderNeedPrimeDuringFirstLayer(const SliceDataStorage& storage) const
```
BagelOrb: Why not make a function `getExtruderNeedsPrimeDuringFirstLayer` which works per extruder?

Nothing depends on whether any extruder needs a prime; stuff only depends per extruder on whether a prime poop is needed.

src/gcodeExport.cpp
```diff
@@ -90,6 +90,7 @@ void GCodeExport::preSetup(const MeshGroup* meshgroup)
         extruder_attr[extruder_nr].prime_pos = Point3(train->getSettingInMicrons("extruder_prime_pos_x"), train->getSettingInMicrons("extruder_prime_pos_y"), train->getSettingInMicrons("extruder_prime_pos_z"));
         extruder_attr[extruder_nr].prime_pos_is_abs = train->getSettingBoolean("extruder_prime_pos_abs");
         extruder_attr[extruder_nr].use_temp = train->getSettingBoolean("machine_nozzle_temp_enabled");
+        extruder_attr[extruder_nr].is_prime_enabled = train->getSettingBoolean("prime_enable");
```
BagelOrb: Please rename the setting here and in fdmprinter.def.json.
It should be called `prime_poop_enable`.

src/gcodeExport.h
```diff
@@ -42,6 +42,7 @@ class GCodeExport : public NoCopy
         bool prime_pos_is_abs; //!< Whether the prime position is absolute, rather than relative to the last given position
         bool is_primed; //!< Whether this extruder has currently already been primed in this print
         bool use_temp; //!< Whether to insert temperature commands for this extruder
+        bool is_prime_enabled; //! < Whether the priming poop is enabled
```
BagelOrb: Please rename to `is_prime_poop_enabled`

src/sliceDataStorage.cpp
```diff
@@ -390,30 +390,15 @@ std::vector<bool> SliceDataStorage::getExtrudersUsed(int layer_nr) const
     return ret;
 }
 
+bool SliceDataStorage::getExtruderPrimeEnabled(int extruder_nr) const
```
BagelOrb: Please rename to `getExtruderPrimePoopEnabled`

LipuFei:
@BagelOrb I agree. I was trying to figure out a "better" word than "poop". If you think that's the right word then I'm happy to use it :)

LipuFei:
@BagelOrb All fixed.

src/FffGcodeWriter.cpp
```diff
@@ -712,15 +712,40 @@ LayerPlan& FffGcodeWriter::processLayer(const SliceDataStorage& storage, int lay
     return gcode_layer;
 }
 
-bool FffGcodeWriter::getExtrudersNeedPrimeDuringFirstLayer() const
+bool FffGcodeWriter::getAnyExtruderNeedPrimePoopDuringFirstLayer(const SliceDataStorage& storage) const
```
BagelOrb: Why not make a function getExtruderNeedsPrimePoopDuringFirstLayer which works per extruder instead of the `getAny...` construction?

Nothing depends on whether any extruder needs a prime; stuff only depends per extruder on whether a prime poop is needed.

LipuFei:
@BagelOrb Refactored the `getAll...()` function and changed all names to "blob"

PR #508, opened 2017-05-01
This change fills the gaps that occur in the spiralized layer view.

smartavionics:
It's not a perfect solution because it will cause the z-seam to be visible on those parts
of the model that feature layers that differ greatly in outline. For smooth models, it
works well. This does not change the printed model, just the model in the layer view.


Ghostkeeper:
Devs, see CURA-3760.

jackha:
@smartavionics Thanks!

PR #509, opened 2017-05-02
Fix loading extruders when there are more extruders than machine_extruder_count 

fieldOfView:
This PR is a followup to https://github.com/Ultimaker/CuraEngine/pull/501. That PR fixed a crash, but did not introduce the correct behavior for cases when there are more extruder definitions than machine_extruder_count.

The previous fix ignored any extruder that came in after an extruder_nr that was too high was encountered. It turns out the extruders may not come in in a specific order, so this would often ignore too many extruders.


jackha:
Good find! Thanks

PR #512, opened 2017-05-05
Time estimate per print feature

14bitVoid:
This PR lets curaengine send a time estimate for each of the print features, like support, infill, skin, etc. That lets you know where the printer spends most of its time, so you can adjust the speeds more effectively. For example you could see that the print takes 9h but 7h of that is skin, then changing the skin speed is better than changing the outer wall speed.

The normal time estimate looks the same but it only lists the details of each print feature in the tooltip of the normal time estimate. I can't make a screenshot because it's a tooltip.

Uranium: https://github.com/Ultimaker/Uranium/pull/247
Cura: https://github.com/Ultimaker/Cura/pull/1794

### DIFF
--- a/src/PrintFeature.h
+++ b/src/PrintFeature.h
@@ -16,7 +16,8 @@ enum class PrintFeatureType: unsigned char
     SupportInfill,
     MoveCombing,
     MoveRetraction,
-    SupportInterface
+    SupportInterface,
+    NumPrintFeatureTypes
 };
 
 
--- a/src/PrintFeature.h
+++ b/src/PrintFeature.h
@@ -6,18 +6,18 @@ namespace cura
 
 enum class PrintFeatureType: unsigned char
 {
-    NoneType, // used to mark unspecified jumps in polygons. libArcus depends on it
-    OuterWall,
-    InnerWall,
-    Skin,
-    Support,
-    SkirtBrim,
-    Infill,
-    SupportInfill,
-    MoveCombing,
-    MoveRetraction,
-    SupportInterface,
-    NumPrintFeatureTypes
+    NoneType = 0, // used to mark unspecified jumps in polygons. libArcus depends on it
+    OuterWall = 1,
+    InnerWall = 2,
+    Skin = 3,
+    Support = 4,
+    SkirtBrim = 5,
+    Infill = 6,
+    SupportInfill = 7,
+    MoveCombing = 8,
+    MoveRetraction = 9,
+    SupportInterface = 10,
+    NumPrintFeatureTypes = 11 // used to find the number of feature types in this enum
 };
 
 
--- a/src/commandSocket.cpp
+++ b/src/commandSocket.cpp
@@ -629,18 +629,18 @@ void CommandSocket::sendPrintTimeMaterialEstimates()
     logDebug("Sending print time and material estimates.\n");
     auto message = std::make_shared<cura::proto::PrintTimeMaterialEstimates>();
 
-    std::unordered_map<PrintFeatureType, double> time_estimates = FffProcessor::getInstance()->getTotalPrintTimes();
-    message->set_time_infill(time_estimates[PrintFeatureType::Infill]);
-    message->set_time_inset_0(time_estimates[PrintFeatureType::OuterWall]);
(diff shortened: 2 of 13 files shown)


Ghostkeeper:
Devs, see CURA-3773.

src/gcodeExport.cpp
```diff
@@ -602,10 +630,10 @@ void GCodeExport::writeTravel(int x, int y, int z, double speed)
     CommandSocket::sendLineTo(travel_move_type, Point(x, y), display_width);
 
     *output_stream << "G0";
-    writeFXYZE(speed, x, y, z, current_e_value);
+    writeFXYZE(speed, x, y, z, current_e_value, PrintFeatureType::MoveCombing);
```
BagelOrb: I would rather see the travel time as a part of the last print feature, unless when moving between different print features.

Because infill patterns can severely impact the travel time I think it would be best to count the travel between infill lines as infill time.

src/commandSocket.cpp
```diff
@@ -629,7 +629,18 @@ void CommandSocket::sendPrintTimeMaterialEstimates()
     logDebug("Sending print time and material estimates.\n");
     auto message = std::make_shared<cura::proto::PrintTimeMaterialEstimates>();
 
-    message->set_time(FffProcessor::getInstance()->getTotalPrintTime());
+    std::unordered_map<PrintFeatureType, double> time_estimates = FffProcessor::getInstance()->getTotalPrintTimes();
```
BagelOrb: Because enums evaluate to integer types it's better to just make this a vector.
Do make sure the enum constants are the right integer values by defining them.

E.g.:
````
enum class PrintFeature
{
   INFILL = 0,
   SKIN = 1,
    etc.
}
time_estimates[PrintFeature::INFILL] = ...;
````

**Comment by BagelOrb** [comment]
Linked PRs: https://github.com/Ultimaker/Cura/pull/1794
https://github.com/Ultimaker/Uranium/pull/247

File: src/gcodeExport.cpp (hunk 1 above)
Comment (nallath) [inline src/gcodeExport.cpp]: I'd rather see travel time as a global thing; It's good to know that for a given print, Cura spent X time doing travels.

File: src/commandSocket.cpp (hunk 2 above)
Comment (14bitVoid) [inline src/commandSocket.cpp]: Okay because its an enum class you have to cast it to unsigned char first though.

File: src/gcodeExport.cpp (hunk 1 above)
Comment (14bitVoid) [inline src/gcodeExport.cpp]: I can probably change it if you want, just takes some figuring out where the travels to new print features are made :-)

I meant it like a global thing tho. You can see that travel time goes down 50% when you change infill pattern to zigzag. You see how these settings influence each other.

Comment (BagelOrb) [inline src/gcodeExport.cpp]: Okay let's leave it like this then.

File: src/PrintFeature.h [hunk 3]
```diff
@@ -16,7 +16,8 @@ enum class PrintFeatureType: unsigned char
     SupportInfill,
     MoveCombing,
     MoveRetraction,
-    SupportInterface
+    SupportInterface,
+    NumPrintFeatureTypes
```
BagelOrb: Please put a little comment here to explain what this value is used for.

You can explicitly set the type of underlying integer to use for the enum. I'm unsure, but maybe you should also explicitly set the values of the enum constants like`SupportInterface = 6,`.

14bitVoid:
Okay I've added explicit values for the enum constants and a comment.

src/gcodeExport.cpp
```diff
@@ -377,14 +377,27 @@ double GCodeExport::getTotalFilamentUsed(int extruder_nr)
     return extruder_attr[extruder_nr].totalFilament;
 }
 
-double GCodeExport::getTotalPrintTime()
+std::vector<double> GCodeExport::getTotalPrintTimes()
```
BagelOrb: The CuraEngine side of this PR looks good to me!

Just one little comment: I'd like to rename this function to `getTotalPrintTimePerFeature` so as to be more specific.

14bitVoid: Okay I renamed all of them because there were several with this name and functionality.

PR #520, opened 2017-05-22
CURA-3848: Fix priming on the first layer when using raft with dual extrusion

LipuFei:


src/FffGcodeWriter.cpp
```diff
@@ -514,6 +514,20 @@ void FffGcodeWriter::processRaft(const SliceDataStorage& storage)
...
+        std::vector<uint32_t> extruder_order = calculateLayerExtruderOrder(storage, extruder_nr, layer_nr);
+        bool has_extruders_primed = false;
+        for (uint32_t each_extruder_need_prime : extruder_order)
+        {
+            setExtruder_addPrime(storage, gcode_layer, layer_nr, each_extruder_need_prime);
+            has_extruders_primed = true;
+        }
+        if (has_extruders_primed)
+        {
+            gcode_layer.setExtruder(extruder_nr);
```
BagelOrb: I don't think you should switch back.
I layer shouldn't use one extruder twice, so I think the switching back should happen on the next layer.

src/FffGcodeWriter.cpp
```diff
@@ -838,7 +852,7 @@ std::vector<unsigned int> FffGcodeWriter::calculateLayerExtruderOrder(const Slic
     std::vector<bool> extruder_is_used_on_this_layer = storage.getExtrudersUsed(layer_nr);
     
     // check if we are on the first layer
-    if ((getSettingAsPlatformAdhesion("adhesion_type") == EPlatformAdhesion::RAFT && layer_nr == -Raft::getFillerLayerCount(storage))
+    if ((getSettingAsPlatformAdhesion("adhesion_type") == EPlatformAdhesion::RAFT && layer_nr == -Raft::getTotalExtraLayers(storage))
```
BagelOrb: Is this first check still needed? I thought you said priming during the raft didn't make use of what this function does...

src/FffGcodeWriter.cpp
```diff
@@ -514,6 +514,20 @@ void FffGcodeWriter::processRaft(const SliceDataStorage& storage)
         infill_comp.generate(raft_polygons, raftLines);
         gcode_layer.addLinesByOptimizer(raftLines, &gcode_layer.configs_storage.raft_base_config, SpaceFillType::Lines);
 
+        // When we use raft, we need to make sure that all used extruders for this print will get primed on the first raft layer,
+        // and then switch back to the original extruder.
+        std::vector<uint32_t> extruder_order = calculateLayerExtruderOrder(storage, extruder_nr, layer_nr);
```
BagelOrb: -> unsigned int

src/FffGcodeWriter.cpp
LipuFei: This is still needed because the process raft function now uses it to get the extruders that need to be primed, so this layer number check is required.

src/FffGcodeWriter.cpp
LipuFei: This is the last part of the first layer. Without the change, it will just print the first layer with an extruder and go on. In case we have dual extrusion, we do an extra extruder switch, and I think it makes sense to make sure that it switches back to the original extruder if there is any extruder switches here.

BagelOrb: I'm not saying you should remove it, I'm saying you should move the switch to the second layer.
I think it makes sense to set the extruder for each raft layer..

src/FffGcodeWriter.cpp
BagelOrb: Ah nevermind.

LipuFei:
@BagelOrb all fixed.

src/FffGcodeWriter.cpp
```diff
@@ -555,6 +563,9 @@ void FffGcodeWriter::processRaft(const SliceDataStorage& storage)
 
     for (int raftSurfaceLayer = 1; raftSurfaceLayer <= train->getSettingAsCount("raft_surface_layers"); raftSurfaceLayer++)
     { // raft surface layers
+        // make sure that we are using the correct extruder to print raft
```
BagelOrb: You also need to set the extruder for the interface layer.
The raft would then be like so:
raft base, make prime blob for each extruder, raft interface, raft surface layers

BagelOrb: Oh it was already set!

src/FffGcodeWriter.cpp
```diff
@@ -555,6 +563,9 @@ void FffGcodeWriter::processRaft(const SliceDataStorage& storage)
...
+        gcode_layer.setExtruder(extruder_nr);
```
BagelOrb: This variable is undefined at this point.

src/FffGcodeWriter.cpp
```diff
@@ -568,6 +576,9 @@ void FffGcodeWriter::processRaft(const SliceDataStorage& storage)
         }
 
         LayerPlan& gcode_layer = *new LayerPlan(storage, layer_nr, z, layer_height, extruder_nr, fan_speed_layer_time_settings_per_extruder_raft_surface, combing_mode, comb_offset, train->getSettingBoolean("travel_avoid_other_parts"), train->getSettingInMicrons("travel_avoid_distance"));
+        // make sure that we are using the correct extruder to print raft
+        gcode_layer.setExtruder(extruder_nr);
```
BagelOrb: You set the extruder before you set the isinside. This causes a combing bug because the next travel move will assume it started from inside, while the print head is actually at the extruder bay.

src/FffGcodeWriter.cpp
```diff
@@ -520,6 +522,7 @@ void FffGcodeWriter::processRaft(const SliceDataStorage& storage)
         for (unsigned int each_extruder_need_prime : extruder_order)
         {
             setExtruder_addPrime(storage, gcode_layer, layer_nr, each_extruder_need_prime);
+            last_extruder_on_initial_raft_layer = each_extruder_need_prime;
```
BagelOrb: The variable `each_extruder_need_prime` is not clear to me. 
I think it should be something like `to_be_primed_extruder_nr`

BagelOrb: I think it's better to make a variable `current_extruder_nr` which you use for all layers.

You can set that using `gcode_layer.getExtruder()` at the end of planning the layer.

LipuFei: I will use `each_to_be_primed_extruder_nr`, the word `current` occurs too often...

LipuFei:
@BagelOrb Fixed

src/FffGcodeWriter.cpp
```diff
@@ -519,10 +519,10 @@ void FffGcodeWriter::processRaft(const SliceDataStorage& storage)
         // When we use raft, we need to make sure that all used extruders for this print will get primed on the first raft layer,
         // and then switch back to the original extruder.
         std::vector<unsigned int> extruder_order = getUsedExtrudersOnLayerExcludingStartingExtruder(storage, extruder_nr, layer_nr);
-        for (unsigned int each_extruder_need_prime : extruder_order)
+        for (unsigned int each_to_be_primed_extruder_nr : extruder_order)
```
BagelOrb: This is not each number!
This is a for-each loop, but that doesn't mean you should put 'each' in the variable name! That's not relevant to this variable.

src/FffGcodeWriter.cpp
```diff
@@ -519,10 +519,10 @@ void FffGcodeWriter::processRaft(const SliceDataStorage& storage)
...
         {
-            setExtruder_addPrime(storage, gcode_layer, layer_nr, each_extruder_need_prime);
-            last_extruder_on_initial_raft_layer = each_extruder_need_prime;
+            setExtruder_addPrime(storage, gcode_layer, layer_nr, each_to_be_primed_extruder_nr);
+            last_extruder_on_initial_raft_layer = each_to_be_primed_extruder_nr;
```
BagelOrb: You didn't address my other comment.



I think it's better to make a variable `current_extruder_nr` which you *use for all layers*.

You can set that using `gcode_layer.getExtruder()` at the end of planning the layer.

Issue #225, opened 2015-07-23
Draw all perimeters one feature at a time.


nallenscott:
Moved from daid/Cura#1319:

Cura is drawing one or two perimeters on a feature, then leaving to draw one or two on another feature and so on, until it eventually gets all the perimeters drawn for the entire layer. The resulting perimeters are pretty sloppy due to the starting / stopping (especially on smaller features). Is it possible to force Cura to draw all of the perimeters one feature at a time?

Response to @BagelOrb:

To clarify, let's say we have a part with a few through holes, and a shell thickness (3x) the nozzle diameter so that Cura will draw (3) perimeter lines around the various features of the part. When Cura draws perimeters for a layer, it jumps around between features drawing one or two perimeter lines on each one until all the perimeters for that layer are finished.

This probably isn't a big deal if you're printing figurines. But if you need functional parts, then having features with clean / contiguous shells could be very important. I have another printer with proprietary software that draws all of the perimeter lines for a layer one feature at a time, and draws them in such a way that results in a nice solid shell around each feature.

Drawing perimeters for a layer one feature at a time ( i.e. finish all perimeter lines for a feature before moving to the next feature ) would improve the consistency and contiguity of individual feature shells, and therefore the mechanics and aesthetic of the entire part. Slic3r does this, and I use it for certain parts, but I would prefer to see this behavior in Cura.

CCS86:
I totally agree

BagelOrb:
Ah now I understand.

This problem is more difficult than you might think, because while the first perimeter might consist of multiple parts, the second can consist of a single polygon. It would require a bit more computation time to group perimeters which belong together.

JustAnother1:
+1
Slicing speed is not as crucial as quality. Especially for functional parts. Having this enabled as optional feature could be a solution that fits all users.

BagelOrb:
It is on the agenda...

Perhaps you guys can help me think of a good algorithm?

I was thinking something along the following lines:
- Perform an inset/outset on each polygon separately
- For each combination of two polygons that might intersect (based on bounding boxes for example):
  - If they intersect, group* these polygons 

printing:
- for each group*
  - for each inset number (wall 2, wall 3, outer wall)
     - print all insets of the given number for all polygons in the group

---

[...]

PR #524, opened 2017-05-23
Add the ability to process inset walls as groups

smartavionics:
First attempt at grouping insets together so that it prints all the insets for a hole before moving onto the next. It appears to work pretty well but it's early days. At the moment, it ignores the overlap compensation setting but as that currently appears to be broken anyway, I haven't rushed to implement that.
Really just looking for feedback at this time.
I will post the matching Cura PR for the UI checkbox to enable this feature. It's https://github.com/Ultimaker/Cura/pull/1865.


smartavionics:
Working pretty well now if anyone wants to try it out.

nallath:
Devs see CURA-3862

BagelOrb:
Related issues:
https://github.com/Ultimaker/Cura/issues/1380
https://github.com/Ultimaker/CuraEngine/issues/225

Please read through the discussions there to learn about a couple of the ins and outs of this feature.

smartavionics:
I've looked at those issues you cited. It's interesting that what I have done is quite close to what is suggested but I haven't come up with any fancy data structure to reduce the time taken to find the intersecting insets. Performance doesn't seem to be an issue even with dozens of holes (I haven't tried hundreds of holes). The end result looks pretty good, big reductions in distance travelled and retractions. I think I will do a test now with hundreds of holes and compare the time taken to slice with and without optimization.

BagelOrb:
I don't think you should test for hundreds of holes, but for a high poly model with holes which have a lot of vertices.

smartavionics:
I did a simple test which was just 11 x 11 grid of circular holes. When the holes had 10 vertices, the time taken to slice when optimizing was often less than the time taken when not optimizing! With 20 vertices per hole, it takes a bit longer but really it's hard to tell unless you look at the time reported by the back end. With 4 walls, and 25 layers, the print time was being reduced from 4h11m to 3h18m!

BagelOrb:
That's a pretty big difference!!

I was wondering more what would happen when you have under 5 holes but they each have more than 200 vertices... High poly models can easily get there.

smartavionics:
OK, just tested 9 holes each with 300 vertices and the slicing time with/without optimization is pretty much the same. It varies from run to run anyway. Sometimes one is quicker than the other and vice versa.

BagelOrb:
Yeah I always test by running the engine from the command line and having no other program open. I then slice with both versions alternatingly to reduce the noise as much as possible.

If I do such tests I normally do them about 5 times each.
It takes a bit of time, but it's a good excuse to get some coffee and relax for a bit :)

src/FffGcodeWriter.cpp
```diff
@@ -1211,6 +1211,356 @@ void FffGcodeWriter::processSpiralizedWall(const SliceDataStorage& storage, Laye
...
+            if (LinearAlg2D::getDist2FromLineSegment(lp0, inner_poly_point, lp1) < max_gap2)
+            {
+                return true;
+            }
+        }
+    }
+    return false;
+}
+
+static int findAdjacentEnclosingPoly(const ConstPolygonRef& enclosed_inset, const std::vector<ConstPolygonRef>& possible_enclosing_polys, const coord_t max_gap)
```
BagelOrb: These functions should be moved to utils/polygonUtils.cpp

Also you might want to have a look at the functions in that file, because there is some similar functionality in there.

src/FffGcodeWriter.cpp
```diff
@@ -1211,6 +1211,356 @@ void FffGcodeWriter::processSpiralizedWall(const SliceDataStorage& storage, Laye
...
+    {
+        if (polyOutlinesAdjacent(inset, possible_adjacent_polys[poly_idx], max_gap))
+        {
+            return poly_idx;
+        }
+    }
+    return -1;
+}
+
+static void processInsetsWithOptimizedOrdering(const SliceDataStorage& storage, LayerPlan& gcode_layer, const SliceMeshStorage* mesh, const PathConfigStorage::MeshPathConfigs& mesh_config, const SliceLayerPart& part, unsigned int layer_nr, EZSeamType z_seam_type, Point z_seam_pos)
```
BagelOrb: It would be nice to have a general function which returns the order in which to print insets, so we can use it in other places as well.

Something like `std::vector<std::pair<ConstPolygonRef, GCodePathConfig*>> optimizeInsetOrder(std::vector<std::pair<Polygons, GCodePathConfig*>>, bool outside_first);`

Or instead of returning a reference to a polygon, return an index, just like the path order optimizer.

In fact, it might be a good idea to change the path order optimizer itself, rather than implementing all this functionality in FffGcodeWriter.

src/FffGcodeWriter.cpp
```diff
@@ -1211,6 +1211,356 @@ void FffGcodeWriter::processSpiralizedWall(const SliceDataStorage& storage, Laye
...
+{
+    const coord_t wall_line_width_0 = mesh->getSettingInMicrons("wall_line_width_0");
```
BagelOrb: You can obtain these settings from the config. Then these settings don't have to be parsed over again.

Moreover, some settings might be different on the initial layer, so getting the settings from the GCodePathConfig probably avoids some bugs.

src/FffGcodeWriter.cpp
```diff
@@ -1211,6 +1211,356 @@ void FffGcodeWriter::processSpiralizedWall(const SliceDataStorage& storage, Laye
...
+            if (adjacent_poly_idx < 0 && !outer_inset_first)
+            {
+                // we didn't find a level 1 inset that encloses this hole so now look to see if there is a level 1 inset that simply touches
+                // this hole and use that instead - however, as the level 1 inset will touch other holes and/or the outer wall we don't want
+                // to do this when printing the outer walls first
+                adjacent_poly_idx = findAdjacentPoly(hole_outer_wall[0], inset_polys[1], std::max(wall_line_width_0, wall_line_width_x) * 1.5f);
+            }
+        }
+        // now test for the special case where we are printing the outer wall first and we have two or more holes so close together that they share a level 1 inset
+        // in this situation we don't want to output the level 1 inset until after all the holes' outer walls have been printed
```
BagelOrb: The inverse case is also problematic.
You can have a big 8-shaped outline which has two second walls for the outside part, but you can also have a big part with two round holes close to each other, such that the second wall around those two holes forms an 8-shape.

In the first case we have a problem when we need to process the inner inset first, while in the second case we have a problem when we need to process the outer inset first.

src/FffGcodeWriter.cpp
```diff
@@ -1211,6 +1211,356 @@ void FffGcodeWriter::processSpiralizedWall(const SliceDataStorage& storage, Laye
...
+                gcode_layer.addPolygonsByOptimizer(hole_outer_wall, &mesh_config.inset0_config, &wall_overlap_computation, z_seam_type, z_seam_pos, wall_0_wipe_dist, spiralize, flow, retract_before_outer_wall);
+            }
+            else
+            {
+                WallOverlapComputation* wall_overlap_computation(nullptr);
+                gcode_layer.addPolygonsByOptimizer(hole_outer_wall, &mesh_config.inset0_config, wall_overlap_computation, z_seam_type, z_seam_pos, wall_0_wipe_dist, spiralize, flow, retract_before_outer_wall);
+            }
+        }
+    }
+    // now process the part's outer wall and the level 1 insets that it surrounds
```
BagelOrb: This function is becoming very long. Please try to split it up into smaller parts to make the code more readable.

BagelOrb:
Maybe we should talk a bit about how you want to tackle this problem, before you run off and implement everything. I think this PR has quite some architecture implementations.

smartavionics:
Why don't you just try it out and see if you can find situations where it does a bad job. I think it's working pretty well and it isn't a CPU hog.

BagelOrb:
Having a working algorithm is one thing, but having clean code is another.
What I mean is that we should maybe talk about where the algorithms should be implemented.

Given your CPU tests I believe your algorithms are sufficiently fast.

However, we should think about where to implement this: in the pathOrderOptimizer? In a new class which extends the pathOrderOptimizer? In a totally new class? 

Should we make a function which optimizes the order of the walls only, or one which is applicable to any set of polygons with their accompanying GCodePathConfig?

How should we split the function up so that the code becomes readable?

These are all important questions. We need code which is easily adaptable in the future when we want to expand on the functionality.

src/FffGcodeWriter.cpp
smartavionics: BTW the original processInsets() function still gets wall widths using getSettingInMicrons().

smartavionics:
I've done a bit of refactoring to split the big function into 3.

Personally, I don't think it needs a new class. It's so specific to the job at hand that I can't imagine how it could be used for any other purpose. Perhaps in the future it will become obvious that what it does could be applicable to other situations or that we need to do a similar thing elsewhere. If that happens, it could be carved out into a separate class and made more general then.

What it needs more than anything else right now is testing against as many models as possible to see where it could be improved/fixed.

smartavionics:
Hi Tim, I see that lines like the following have appeared:

       added_something = added_something | func();

Is it really good style to use bitwise OR with boolean values? I have used the following instead:

added_something = func() || added_something;

smartavionics:
Also, IMHO, the added_something code just adds a bunch of unnecessary complexity. All you needed to do was have a flag in LayerPlan that gets set true if something was added to the plan and then clear the flag in addMeshPartToGCode() before processing the infill/walls, etc and then test the flag at the end to decide whether to add the move inside the boundary. That would have been a lot simpler!

BagelOrb:
Hey Mark,

just letting you know that I'll be on vacation for two weeks. The rest of the team is working on their own research project. I don't think any PR will be picked up during the next two weeks.

smartavionics:
Hi Tim, no problem, I'm in no hurry. Have a great vacation. Cheers, Mark

Ghostkeeper:
I would've gone for `added_something |= func()` myself.

smartavionics:
Hi, well as I mentioned above, I think the whole added_something mechanism that has been recently introduced is a really ugly way of achieving the desired goal. It would have been so much neater to have a flag in LayerPlan that gets set when a move or extrusion is added and then that could just be tested after the insets/infill, etc. have been processed and if the flag is set, the extra move is added. Perhaps there's more to it than I have understood.

Ghostkeeper:
I agree it's a bit chunky. I'd have tried to detect changes without a boolean flag myself, if that were possible. But that typically requires a more object-oriented set-up than we have now...

smartavionics:
Well, LayerPlan is an object, it could easily hold the flag to say that an extrusion or move has been added by one of its methods. Perhaps when Tim gets back from holiday he will consider reworking this?

BagelOrb:
Hey Mark,

I've been a bit scared of reviewing this, because it has 60 (!!!) commits and there will probably be loads of comments from me.

I'm also scared this is going to give huge merge conflicts once https://github.com/Ultimaker/CuraEngine/pull/548 is merged.

Sorry for dragging this PR on for so long. I'll really get into it now.

smartavionics:
Why not get #548 merged first and let me worry about merging this one. I'm in no particular hurry!

BagelOrb:
Before I really dive into the code I'd like to make clear what you've actually implemented.

Suppose we have these walls:
hole_perimeter_connections


I've given them tags based on the inset count

Now in which order they are allowed to be processed depends on the following setting:
- Outer Before Inner Walls (outer_inset_first)

We have the following partial ordering on the walls (where higher is closer to air)
```
1 > 2 > 3
or in tree form:
1.1  1.2  1.3
 \     \  /
  2.1   2.2
    \   / 
     3.1
```

If outer_inset_first then what we currently do (without this PR) could be:
1.1, 1.2, 1.3, 2.1, 2.2, 3.1 (current)
An optimized order (disregarding distance) would be
1.1, 2.1, 1.2, 1.3, 2.2, 3.1 (optimal)
An order which is not allowed by the ordering constraints is:
1.2, 2.2, 3.1, 1.3, 1.1, 2.1 (bad!)

My question is the following: did you ensure to never violate the ordering constraint?

smartavionics:
Can you please share that stl so I can check with exactly the same model?

smartavionics:
I have created a similar model and get this sequence when outer wall first is active...

First 1.2 and 1.3
screenshot_2017-07-04_14-17-40

Then 1.1, 2.1 and 3.1 merged with 3.3
screenshot_2017-07-04_14-18-00

Then 2.3 and 2.3 which are merged
screenshot_2017-07-04_14-18-13

Then the remaining level 3s
screenshot_2017-07-04_14-18-28

screenshot_2017-07-04_14-18-41

And finally it starts on the skin...

screenshot_2017-07-04_14-18-56

I think that looks OK?

BagelOrb:
No, it seems to me you are violating the required order when you are doing 3.3 before you do 2.3.

At those places you first do the outer wall, then the innermost wall and then a wall in the middle.
I think this can result in defects visible on the outside.
There may be small deviations from the actual line width. If you print the walls from the outside inward then the excess material will be deposited on the inside.

The other way around: if you have `outer_inset_first == false` you want to print the innermost perimeter first. Then each next perimeter has the previous perimeter to attach to, so that you can print overhangs more reliably. Printing the 3rd wall and then the 1st wall makes the outer wall less attached to any printed lines, which makes overhangs worse.

BagelOrb:
Implementation-wise I'd ideally like to see something like an `OrderedInsetOptimizer` which returns the order in which the polygons should be printed.
Take a look at how PathOrderOptimizer does it, but don't look too much at that file, cause it's not coded very clearly ;)

`OrderedInsetOptimizer::optimize()` could return something like a `vector<InsetPolygonIndex>` where
```
struct InsetPolygonIndex
{
    unsigned int inset_idx;
    unsigned int polygon_idx;
};
```

That way we don't infect so much code in FffGcodeWriter with this change...

smartavionics:
OK, I will look into refactoring it. I am away next week and it is unlikely that it will get done before then but we shall see...

smartavionics:
I've moved all of the functions into their own class but haven't changed how it works.
To be honest, I am not sure how to do it at the moment because it's not just a matter of working out the order in which to output the insets. It also generates travel moves before writing some of the polygon groups so to influence the ordering of the insets and the choice of start vertices.

BagelOrb:
Hey Mark,

The architecture is already a bit better with this refactor.
However, let's leave the architecture for what it is for a while.

The most important problem is the order of insets we discussed. (See https://github.com/Ultimaker/CuraEngine/pull/524#issuecomment-313352808 and a couple of comments above)
I feel that I cannot merge this code while that bug is present.

I don't know whether your current approach for implementing this feature allows for that bug to be solved. It might be the case that you need to rethink the whole implementation, because I think it's quite a fundamental problem in general.

I hope you can figure it out!

I'm here if you need help thinking about the problem.

smartavionics:
> The most important problem is the order of insets we discussed. (See #524 (comment) and a couple of comments above)
> I feel that I cannot merge this code while that bug is present.

Hi Tim, I have already made a change which I think fixes the ordering issue you mention.

smartavionics:
So now we get this sequence...
screenshot_2017-07-14_10-06-30
screenshot_2017-07-14_10-06-43
screenshot_2017-07-14_10-06-57
screenshot_2017-07-14_10-07-24
screenshot_2017-07-14_10-07-40
screenshot_2017-07-14_10-07-52
screenshot_2017-07-14_10-08-13

BagelOrb:
Still getting wrong order when the setting Outer Before Inner Walls is disabled: 2-1-3
image

[hole_perimeter_connections.stl.txt](https://github.com/Ultimaker/CuraEngine/files/1148193/hole_perimeter_connections.stl.txt)

smartavionics:
OK, so you are referring to the fact that it fills in some of the level 3 insets after it has done the level 2 and level 1 insets (in the correct order). Well, personally, I don't consider that a problem. Why is it a problem?

BagelOrb:
The setting was meant to determine whether the walls should be processed inside outward or outside inward. I don't know what the impact is if the 3rd wall doesn't satisfy those conditions, but not knowing the problem doesn't mean there is none. I'm not sure yet whether this causes print problems.

I don't understand why it works like this, though. Perhaps the fact that the 3rd wall doesn't satisfy the order means that in some other cases the 2nd wall doesn't satisfy the order either - but I haven't found any.
Could you explain why the algorithm enforces the order between the 1st and 2nd wall but not between the 2nd and 3rd wall? Did you ahrdcode that it should enforce the order between the 1st and 2nd wall, but didn't code anything for the other walls?

BagelOrb:
Also: try this lovely model:
[hole_dist_forrest.stl.txt](https://github.com/Ultimaker/CuraEngine/files/1148497/hole_dist_forrest.stl.txt)

It's nice to see how your algorithm solves this piece.

BagelOrb:
This order doesn't really seem optimized.
Outer Before Inner Walls was enabled.
(Besides the fact that the 3rd wall came before the 1st wall.)
A lot of 1st walls are done before any of the 2nd walls are done.
It looks like it would be more efficient to do the 2nd walls whenever all enclosed 1st walls are done.
That's the whole goal of this PR ;)

smartavionics:
> (Besides the fact that the 3rd wall came before the 1st wall.)
A lot of 1st walls are done before any of the 2nd walls are done.
It looks like it would be more efficient to do the 2nd walls whenever all enclosed 1st walls are done.
That's the whole goal of this PR ;)

Hi Tim, I've made some changes and it makes a better job of the hole forrest when outer walls are printed first.

smartavionics:
Your simpler test now looks like this when outer wall first is used...
screenshot_2017-07-14_16-56-51
screenshot_2017-07-14_16-57-03
screenshot_2017-07-14_16-57-17
screenshot_2017-07-14_16-57-37
screenshot_2017-07-14_16-57-49

BagelOrb:
I'm still getting wrong results. I did have to update quite some code, though, to even get it to compile.
I just merged some code which undoes quite some code changes from another PR which was already merged... Sorry about that!

Here's what I'm getting now:
image

smartavionics:
What I get is this sequence...

screenshot_2017-07-17_18-31-41
screenshot_2017-07-17_18-31-59
screenshot_2017-07-17_18-32-19
screenshot_2017-07-17_18-32-44
screenshot_2017-07-17_18-33-01
screenshot_2017-07-17_18-33-31
screenshot_2017-07-17_18-33-54
screenshot_2017-07-17_18-34-10
screenshot_2017-07-17_18-34-23

What's the problem with that?

BagelOrb:
I've just talked to our printing expert team and they said the following:
If you do infill -> 2 -> 1 -> 1 -> 3 then that means you are doing a second wall after infill.
This causes a little blob on the second wall, because the infill is usually printed with a higher flow, which in Bowden extruders makes the hysteresis effect dump a little more material when it first starts the second wall.
This bump on the second wall is then perpetuated a bit into the first wall, making a visual defect.
That's we some of our profiles have 3 walls; then the bump gets smoothed out over 2 walls, which makes it almost disappear.

To conclude: we should always keep the same order 3 > 2 > 1 when outer before inner walls is disabled

smartavionics:
Hi Tim, OK, I will disable the optimisation when printing infill before walls.

smartavionics:
Tim, I do think this PR would benefit by being tested by the Cura users. That way, we will find out if it has any real issues but also we will get some feedback as to the benefits. If you're worried about it breaking peoples prints, why don't we just move the setting into the experimental category and not enable the optimisation by default?

Ghostkeeper:
BagelOrb requested me to give a neutral opinion, but I haven't had the time for it yet. Sorry!

smartavionics:
My view is that the optimisation works well and it is ready for some serious testing. I think the only way that's going to happen is if we let the whole community have access to it so I am suggesting that we move the setting into the experimental category and disable it by default so people have to explicitly turn it on to try it. Then we will get some decent feedback as to whether it's working OK or not.

smartavionics:
OK, this has been brought up to date WRT the master branch. How about making it available as an experimental feature?

ChrisRiddell:
I'd like to see this as experimental if the devs feel its not ready for prime time, Pretty keen in wanting to test this out.

smartavionics:
So, you haven't pulled this into 2.7, I have to say I'm greatly disappointed. Cura users are regularly requesting fewer retractions and travel and this goes some way towards solving that. I know that the real solution also involves optimising the infill but even so, this is still a useful PR. Why won't you let people give it a go?

BagelOrb:
Hi Mark.

I'm sorry for the delay.

I have withdrawn myself from this pull request. I find it very difficult to decide what to do with this.
On the one hand this is one of the most requested features, but on the other hand there are some limitations to your algorithm and I'm not sure about the code quality.

I have assigned @Ghostkeeper to review this PR.
He will be taking over my job soon.
I will become a researcher in September, working both for Ultimaker and the Technical University of Delft. That means that I won't be the one working on CuraEngine any more, although I hope to make pull requests with new research developments every so often.

I'm sorry for causing so much delay in handling this PR! I hope @Ghostkeeper will have time to review this soon.

smartavionics:
Hi Tim, good luck with your new role.

Ghostkeeper:
Sorry man, due to resources being pulled away from Cura and the busy release process for 2.7, we aren't keeping up with the pull request fast enough. This afternoon I did have time and just as I was reading through the comments and evaluating the differences between what Smartavionics seems to see and what BagelOrb seems to see, you closed the PR.

I'd like to still add this, if you agree. I think this could improve the print quality in all sorts of ways: Better bed adhesion by not dragging over loose lines on the first layer, less blips by not requiring a retraction, better speed, etc.
But the chance that it makes it to 2.7 is getting very slim now. Still, would 2.8 be nice?

smartavionics:
Look, it's sat around for weeks going nowhere. Tim's parting comment was "but on the other hand there are some limitations to your algorithm and I'm not sure about the code quality." Well, code quality can alway be improved (although I don't actually know what he's referring to as I'm not a mind reader) and, yes, there are limitations to the algorithm but, really, the only way it will get a decent test is if the community can try it out and provide feedback. i'm quite pissed off that it's not been given a chance to be widely tested in the next release given how long it's been stable. However, if it makes 2.8 in some form or other, that will be something. Right now, I'm not feeling very cooperative.

Ghostkeeper:
Thanks. Then perhaps if there are small things that need to be improved such as code quality, I can do that in my free time to make sure you don't really need to put a lot more work in. You don't need to cooperate much, aside from giving permission for us to work further on it. I'm typically a bit more lax in code quality when it comes from an external contributor or a new developer, if they haven't really had the opportunity to read and feel our code standards. If I'd find large obstructions then they'd be with the architecture but BagelOrb also said that the architecture is fine now.

I've mostly left long discussions about the engine such as the one above to BagelOrb, but I really appreciate the sheer amount of effort & enthusiasm that people such as Smartavionics put into Cura. If there's anything I can do to make that easier... I just really needed more time.

BagelOrb:
> "I'm not sure about the code quality"
I said this because I haven't digged deep enough into your code. I'm not sure because I haven't really looked ;)

src/utils/polygonUtils.cpp
```diff
@@ -994,5 +994,44 @@ bool PolygonUtils::polygonCollidesWithLineSegment(const Polygons& polys, const P
...
+{
+    const coord_t max_gap2 = max_gap * max_gap;
+    const unsigned outer_poly_size = outer_poly.size();
+    for (unsigned line_index = 0; line_index < outer_poly_size; ++line_index)
+    {
+        const Point lp0 = outer_poly[line_index];
+        const Point lp1 = outer_poly[(line_index + 1) % outer_poly_size];
+        for (Point inner_poly_point : inner_poly)
+        {
+            if (LinearAlg2D::getDist2FromLineSegment(lp0, inner_poly_point, lp1) < max_gap2)
```
Ghostkeeper: That's quadratic! Quite expensive, especially when combined with the additional quadratic factor in `processInsets()`. I can think of one optimisation that would save a significant amount of time here in the random case: Add a heuristic by first checking whether the AABB's of the polygons are adjacent.

Off the top of my head I can also think of an `O(n log n)` algorithm: Find the MST of the combined set of vertices (`O(n log n)`), then check each of those edges (`O(n)`) for the smallest connection of a point from one subset to the other. But perhaps that is too much work and in many small-scale slices it would perform slower. Perhaps the heuristic combined with brute-force search is the best we can achieve.

Ghostkeeper:
See https://github.com/smartavionics/CuraEngine/pull/2

smartavionics:
Thanks for the review!

I've merged your changes and removed a, now, redundant declaration of part_outer_wall as it was being defined more locally later where it was used.

I've given it a basic test and behaviour appears to be unchanged.

>  I think that actually adding the polygons is not the task of the InsetOrderOptimizer

Yes, I'd agree but the practical problem is that it isn't just adding a sequence of polygons, it can also add travel moves. It does that so that the starting position is favourable for the correct ordering of the polygons that follow. I didn't try to introduce a new means of achieving that so I added the moves.

Ghostkeeper:
Ah, yes. I replaced that `part_outer_wall` variable and then forgot to remove it... Thanks for reviewing :grimacing:

Ghostkeeper:
I'd like to leave this in experimental for a while to see what the performance impact is in the wild. I've tested quite a few models and noticed no significant performance impact, but you never know what kind of complex models people come up with sometimes.

smartavionics:
OK, I think that's a good idea.

smartavionics:
A plus point of the optimisation is that it fixes the recently introduced regression of inner walls not starting near their matching outer walls when the z-seam position is specified by the user. Compare these two images, first without optimisation enabled, the second with. Although the total print time only came down by a couple of minutes, it does save printer wear.
screenshot_2017-08-03_19-00-42
screenshot_2017-08-03_19-01-10

JoshMcCullough:
So what's this setting called, assuming it's included as an experimental feature still, and in Cura v3.6.0...?

PR #525, opened 2017-05-23
M204 S argument deprecated

notnyt:
RRF does not support S as an option and it has been deprecated in Marlin.  This uses the newer P and T options.

https://github.com/MarlinFirmware/Marlin/blob/1.1.x/Marlin/Marlin_main.cpp#L7581

http://marlinfw.org/docs/gcode/M204.html

### DIFF
--- a/src/gcodeExport.cpp
+++ b/src/gcodeExport.cpp
@@ -1015,7 +1015,7 @@ void GCodeExport::writeAcceleration(double acceleration, bool for_travel_moves)
     {
         if (current_acceleration != acceleration)
         {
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Then we should make this dependent on the g-code flavour. Since all Ultimaker printers still use S, I believe.

notnyt:
This was just a quick patch to get going... It can probably be further optimized since this covers travel and printing acceleration, yet the code currently changes acceleration before moves of each type.

Ghostkeeper:
You'll understand: We can't merge this because it breaks the acceleration tweaking for printers that still only understand `S`.

PR #529, opened 2017-06-06
Avoid printing skin as multiple segments by starting at a more sensible position.

smartavionics:
In FffGcodeWriter::processSkinAndPerimeterGaps() skin areas are printed starting from the
current location which often causes a skin area that could be printed in one segment to be
split into two segments. This can produce a visible seam on the skin surface where the two
segments meet. However, by moving the start position of the skin area, it may be possible
to print as a single segment hence avoiding an ugly seam.

Imagine that the skin area is a simple shape like a rectangle. If we start the skin area
in one of the two corners where the skin lines are shortest, we should be able to print
the whole skin as a single segment. If we start the area at any other location, it will
require two segments. What this commit does is find locations on the outline of the
skin area that only have skin lines on one side of the location (they are "corner like")
and it adds a move to the nearest of those locations.


Ghostkeeper:
Devs, see CURA-3922.

Ghostkeeper:
Too bad it doesn't work with the new ironing feature. I really would've loved that! Maybe I'll so something similar there...

PR #532, opened 2017-06-09
Fix bug in GCodeExport::writeExtrusion() - don't include z hop height in extrusion length.

smartavionics:


BagelOrb:
Oh wow! That's a pretty big bug!

Thanks for finding it!

I think we need to merge this to 2.6 as well..

smartavionics:
Yes, for very short segments it could have made quite a difference, would have increased the chance of a blob occurring at the start of the path.

smartavionics:
BTW, aren't you meant to be on holiday?!

BagelOrb:
@Devs see CURA-3920

BagelOrb:
Today is packing day ;)

smartavionics:
Err, hold your fire with this one. I think it's not actually going to make any difference because when z hopped, the z hop height isn't assigned to currentPosition.z so that should remain at the layer z height. Sorry, false alarm!

smartavionics:
Take a look, if you agree, please close this PR.

BagelOrb:
Ah yes. But the code change still makes sense.

The writeZHopEnd does a separate move before the actual extrusion move is written, so any code relating to the eventual extrusion move should be placed after the initial unhop.

I think this code should be merged nonetheless.

Ghostkeeper:
Tested by using multiple cubes (so it would retract and z-hop) then looking at the E values of the moves afterwards, in the g-code.

The first extrusion move afterwards had a delta E of 0.05925. The second extrusion move, which should be equally long (two walls of a cube), had a delta E of 0.05924. So it looks like there is no incorrect extrusion going on after the fix.

The gcode-diff tool indicates no change before or after, so it was likely not a problem in any case. I agree that this is more robust though.

PR #533, opened 2017-06-10
Fix bug whereby support was allowed to touch the model on the bottom layer.

smartavionics:
It wasn't shrinking the support polygons by the xy distance on the bottom layer, does now.


smartavionics:
Compare and contrast...

Before PR
screenshot_2017-06-10_16-18-46

After PR
screenshot_2017-06-10_16-17-33

Ghostkeeper:
I seem to recall something about that this could've been intended. Not sure though, and I can't think of a reason why.

Ghostkeeper:
Devs, see CURA-3927.

smartavionics:
> I seem to recall something about that this could've been intended. Not sure though, and I can't think of a reason why.

If that's the case, they should have added a comment explaining what their reasoning was so that someone didn't come along and try and change it later!

Anyway, for the print shown above, it's not the right thing to do.

Ghostkeeper:
Yeah, I agree. Perhaps it was just a mistake because we handle that initial layer a bit differently.

BagelOrb:
The bug was probably caused by the fact that that for loop used to be used
just to compute the overhang at each layer.

Doesn't the new code segfault when it tries to compute overhang on layer 0?
Does it still produce an empty Polygons object?

Op 19 jun. 2017 09:09 schreef "Tim Kuipers" <t.kuipers@ultimaker.com>:

smartavionics:
Hi Tim, yes, I can see there's a potential problem, how about something like this instead...

diff --git a/src/support.cpp b/src/support.cpp
index e522685a..ab0fdd2b 100644
--- a/src/support.cpp
+++ b/src/support.cpp
@@ -331,6 +331,7 @@ void AreaSupport::generateSupportAreas(SliceDataStorage& storage, const Settings
     std::vector<Polygons> full_overhang_per_layer;
     xy_disallowed_per_layer.resize(support_layer_count);
     full_overhang_per_layer.resize(support_layer_count);
+    xy_disallowed_per_layer[0] = storage.getLayerOutlines(0, false).offset(supportXYDistance);
     #pragma omp parallel for default(none) shared(xy_disallowed_per_layer, full_overhang_per_layer, support_layer_count, storage, mesh, max_dist_from_lower_layer, tanAngle) schedule(dynamic)
     for (unsigned int layer_idx = 1; layer_idx < support_layer_count; layer_idx++)
     {

BagelOrb:
The code looks okay.

The code within the `if (use_support_xy_distance_overhang)` isn't executed, but then again no overhang is computed, so that's okay.

It might be good to comment why this line needs separate handling, so that we don't fall into this pitfall in the future.

BagelOrb:
👍

PR #534, opened 2017-06-15
Horizontal Ironing

Ghostkeeper:
This implements the basic ironing technique: Just horizontal ironing.

Ironing causes the nozzle to go over the top surface once again while extruding only very little material. This makes the top surface very smooth.

See also the accompanying pull request in Cura: https://github.com/Ultimaker/Cura/pull/1978

Implements CURA-3903.


src/TopSurface.h
```diff
@@ -0,0 +1,61 @@
...
+     * \brief The areas of top surface, for each layer.
+     */
+    Polygons areas;
+
+    /*!
+     * \brief Generate paths for sanding over the top surface.
+     *
+     * This generates an infill pattern over the top surface that is supposed to
+     * strike the surface smooth by melting it with the hot nozzle, without
+     * extruding anything.
```
nallath: Oh? So where is the extrusion done then? We have a setting for it right?

src/sliceDataStorage.h
```diff
@@ -43,6 +46,15 @@ class SliceLayerPart
...
     Polygons perimeter_gaps; //!< The gaps betwee nconsecutive walls and between the inner wall and outer skin inset
     std::vector<SkinPart> skin_parts;     //!< The skin parts which are filled for 100% with lines and/or insets.
+
+    /*!
+     * \brief The parts of the model that are exposed at the very top of the
+     * model.
+     *
+     * This is filled only when the top surface is needed.
+     */
+    TopSurface* top_surface;
```
nallath: Shouldn't this be initialised to a nullptr somewhere?

src/FffGcodeWriter.cpp
```diff
@@ -1528,6 +1529,16 @@ bool FffGcodeWriter::processSkinPart(const SliceDataStorage& storage, LayerPlan&
     return added_something;
 }
 
+bool FffGcodeWriter::processSanding(const SliceMeshStorage& mesh, const SliceLayerPart& part, const GCodePathConfig& line_config, LayerPlan& gcode_layer) const
```
LipuFei: Since you have commented here, let's use the work "ironing" then.

src/FffGcodeWriter.cpp
```diff
@@ -1025,69 +1025,70 @@ void FffGcodeWriter::addMeshLayerToGCode(const SliceDataStorage& storage, const
...
-        skin_angle = mesh->skin_angles.at(layer_nr % mesh->skin_angles.size());
+        skin_angle = mesh.skin_angles.at(layer_nr % mesh.skin_angles.size());
     }
     if (skin_alternate_rotation && ( layer_nr / 2 ) & 1)
         skin_angle -= 45;
 
-    int64_t skin_overlap = mesh->getSettingInMicrons("skin_overlap_mm");
+    int64_t skin_overlap = mesh.getSettingInMicrons("skin_overlap_mm");
     added_something = added_something | processSkinAndPerimeterGaps(storage, gcode_layer, mesh, extruder_nr, mesh_config, part, layer_nr, skin_overlap, skin_angle);
+    added_something |= processSanding(mesh, part, mesh_config.sanding_config, gcode_layer);
```
LipuFei: "Ironing"

src/FffGcodeWriter.cpp
```diff
@@ -1528,6 +1529,16 @@ bool FffGcodeWriter::processSkinPart(const SliceDataStorage& storage, LayerPlan&
     return added_something;
 }
 
+bool FffGcodeWriter::processSanding(const SliceMeshStorage& mesh, const SliceLayerPart& part, const GCodePathConfig& line_config, LayerPlan& gcode_layer) const
+{
+    bool added_something = false;
+    if (part.top_surface)
+    {
+        added_something |= part.top_surface->sand(mesh, line_config, gcode_layer);
```
LipuFei: iron

src/sliceDataStorage.h
```diff
@@ -13,10 +13,13 @@
 #include "mesh.h"
 #include "MeshGroup.h"
 #include "PrimeTower.h"
+#include "TopSurface.h"
 #include "gcodeExport.h" // CoastingConfig
 
 namespace cura 
 {
+class TopSurface;
```
LipuFei: `TopSurface.h` is included, I guess this can go.

src/FffGcodeWriter.cpp
Ghostkeeper: Yes, originally it was called sanding. At the end I renamed it to ironing but only renamed the user-visible strings.

I've now renamed the implementation under water as well, for consistency: https://github.com/Ultimaker/CuraEngine/commit/d927f7317a2601eb90f650e936377d3f7aa5b360

src/FffGcodeWriter.h
```diff
@@ -478,7 +478,22 @@ class FffGcodeWriter : public SettingsMessenger, NoCopy
      * \param infill_angle The angle in the XY plane at which the infill is generated.
      * \return Whether this function added anything to the layer plan
      */
-    bool processSkinPart(const SliceDataStorage& storage, LayerPlan& gcode_layer, const SliceMeshStorage* mesh, const int extruder_nr, const PathConfigStorage::MeshPathConfigs& mesh_config, const SkinPart& skin_part, unsigned int layer_nr, int skin_overlap, int infill_angle) const;
+    bool processSkinPart(const SliceDataStorage& storage, LayerPlan& gcode_layer, const SliceMeshStorage& mesh, const int extruder_nr, const PathConfigStorage::MeshPathConfigs& mesh_config, const SkinPart& skin_part, unsigned int layer_nr, int skin_overlap, int infill_angle) const;
+
+    /*!
+     * Add the g-code for sanding the top surface.
+     *
+     * This produces travel moves that cover the entire top surface.
```
Ghostkeeper: Indeed, there is a setting that insets the area. I've now edited the documentation to no longer say "entire", and in general made it a bit more clear: c4901204b85f6ec386b26904819e577b5df27525.

src/TopSurface.cpp
```diff
@@ -0,0 +1,64 @@
...
+}
+
+bool TopSurface::sand(const SliceMeshStorage& mesh, const GCodePathConfig& line_config, LayerPlan& layer)
+{
+    if (areas.empty())
+    {
+        return false; //Nothing to do.
+    }
+    //Generate the lines to cover the surface.
+    const EFillMethod pattern = mesh.getSettingAsFillMethod("sanding_pattern");
```
Ghostkeeper: Yes. Most of our patterns leave holes open. We have 3 patterns that don't: Lines, ZigZag and Concentric. Other patterns, like triangles, would leave holes open at 100% infill (line distance equal to line width) and that's not expected.

As the ironing was supposed to make travel moves, there would be no real difference between lines and zigzag. Later on I added a flow setting which didn't make it travel moves any more, so there would be a difference, but these travel moves are between adjacent lines and since the default line distance is at 0.1mm, that travel move is always very small anyway, so I just went for ZigZag.

src/TopSurface.h
Ghostkeeper: Indeed, this was an outdated documentation. Updated it here: 995969a8f2d450dc7a3ed480fa688225443f323c

src/sliceDataStorage.h
Ghostkeeper: Good point. I let it initialise here: 17ee75ed515ac6ba1dde729bd465ae7869d4544e

src/FffGcodeWriter.cpp
Ghostkeeper: Renamed it here: d927f7317a2601eb90f650e936377d3f7aa5b360

src/FffGcodeWriter.cpp
Ghostkeeper: Renamed it here: d927f7317a2601eb90f650e936377d3f7aa5b360

src/sliceDataStorage.h
Ghostkeeper: Indeed. Removed it here: 501f0a113e045d61f23875521f5d0f6bb8ddc463

Dim3nsioneer:
Cool that this feature also comes to Cura. Maybe you want to credit Neotko for it (https://ultimaker.com/en/community/22649-neosanding-automatic-layer-sanding) in the release notes?

Ghostkeeper:
We're planning to put something like that in the change log, that it was based on that idea.

src/FffPolygonGenerator.cpp
```diff
@@ -680,6 +681,12 @@ void FffPolygonGenerator::processSkinsAndInfill(SliceMeshStorage& mesh, unsigned
         }
         generateInfill(layer_nr, mesh, innermost_wall_line_width, infill_skin_overlap, wall_line_count);
     }
+
+    if (mesh.getSettingBoolean("ironing_enabled"))
+    {
+        TopSurface* top_surface = new TopSurface(mesh, layer_nr); //Generate the top surface to iron over.
```
BagelOrb: You initialize this on the heap, but don't delete it.

Please revert https://github.com/Ultimaker/CuraEngine/pull/534/commits/6d1020f37536d037a45fcea7824a4ee37fa04daf

src/TopSurface.h
```diff
@@ -0,0 +1,59 @@
...
+
+#include "utils/polygon.h" //For the polygon areas.
+#include "sliceDataStorage.h" //For the input mesh.
+
+namespace cura
+{
+
+class SliceMeshStorage;
+
+class TopSurface
```
BagelOrb: class documentation is missing here..

BagelOrb: This architecture is rather different from how features are implemented elsewhere. Generally the polygon generation phase is separated from the path generation phase. This is because classes in the engine are generally wrappers for computation.

I've made the same mistake when implementing the prime tower. There I used the same class for the polygon generation phase as the path planning phase as well. The problem was that class members can then only be used for either phase. The prime tower class has private members used in the polygon generation phase, but all parameters of the computation of the path generation phase had to be passed via arguments.

I think your architecture might be better than what we had, but it does introduce a new type of architecture in the engine, which is already chaotic in its architecture. I think we should discuss the architecture and decide together whether the architecture of this class is the type of architecture we want all print features to have eventually or whether this print feature should conform to existing architecture.

src/TopSurface.h
```diff
@@ -0,0 +1,59 @@
...
+public:
+    /*!
+     * Create an empty top surface area.
+     */
+    TopSurface();
+
+    /*!
+     * \brief Generate new top surface for a specific layer.
+     *
+     * The surface will be generated by subtracting the layer above from the
```
BagelOrb: This way of computing the surface generally leads to top surface areas everywhere where theres a slanted surface.
If you want to remove these small areas you'd need to perform a close operation.

If you want to avoid the double offset required for the close operation you can just use e.g. second wall instead of computing an inset from the outline.

Currently you have an Ironing Inset setting with a default value of .35. You should change it such that the top surface area is computed by subtracting the next layer from the top_bottom_area and performing an offset equal to the Ironing Offset which would have a default value of `wall_line_width_0 + wall_line_width_x * (wall_line_count - 1) - .35`

src/TopSurface.h
```diff
@@ -0,0 +1,59 @@
...
+     *
+     * \param mesh The mesh to generate the top surface area for.
+     * \param layer_number The layer to generate the top surface area for.
+     */
+    TopSurface(SliceMeshStorage& mesh, size_t layer_number);
+
+    /*!
+     * \brief The areas of top surface, for each layer.
+     */
+    Polygons areas;
```
BagelOrb: These top_surface areas are specific to ironing. I think this class should be renamed to `Ironing`.

When we would like to separate out a new feature type called Upper Top Skin, we will not use the areas generated here. I don't see why you have conceptually distinguished between TopSurface area and ironing area in this class.

src/TopSurface.cpp
```diff
@@ -0,0 +1,63 @@
...
+    //Do nothing. Areas stays empty.
+}
+
+TopSurface::TopSurface(SliceMeshStorage& mesh, size_t layer_number)
+{
+    //The top surface is all parts of the mesh where there's no mesh above it, so find the layer above it first.
+    Polygons mesh_above;
+    if (layer_number < mesh.layers.size() - 1)
+    {
+        mesh_above = mesh.layers[layer_number + 1].getOutlines();
```
BagelOrb: You can optimize out this copy quite easily.

```
if (...)
{
areas = outlines.difference(layer_above);
}
else
{
areas = outlines;
}
```

src/pathPlanning/GCodePath.cpp
```diff
@@ -6,6 +6,20 @@
...
+GCodePath::GCodePath(const GCodePathConfig* config, SpaceFillType space_fill_type, float flow, bool spiralize) :
+config(config),
+space_fill_type(space_fill_type),
+flow(flow),
+spiralize(spiralize)
+{
+    retract = false;
+    perform_z_hop = false;
+    perform_prime = false;
+    points = std::vector<Point>();
```
BagelOrb: `points` was already initialized to an empty vector. There's no need in setting it to a new empty vector again.

src/pathPlanning/GCodePath.cpp
```diff
@@ -6,6 +6,20 @@
...
+    done = false;
+    estimates = TimeMaterialEstimates();
```
BagelOrb: No need to set again.

src/pathPlanning/GCodePath.cpp
```diff
@@ -6,6 +6,20 @@
 namespace cura
 {
 
+GCodePath::GCodePath(const GCodePathConfig* config, SpaceFillType space_fill_type, float flow, bool spiralize) :
+config(config),
+space_fill_type(space_fill_type),
+flow(flow),
+spiralize(spiralize)
+{
+    retract = false;
```
BagelOrb: Why not initialize every member above, rather than in the body of the constructor?

src/pathPlanning/GCodePath.h
```diff
@@ -37,6 +39,18 @@ class GCodePath
...
+     * \brief Creates a new g-code path.
+     *
+     * \param config The line configuration to use when printing this path.
+     * \param space_fill_type The type of space filling of which this path is a
+     * part.
+     * \param flow The flow rate to print this path with.
+     * \param spiralize Gradually increment the z-coordinate while traversing
+     * this path.
+     */
+    GCodePath(const GCodePathConfig* config, SpaceFillType space_fill_type, float flow, bool spiralize);
```
BagelOrb: :+1:

src/settings/PathConfigStorage.cpp
```diff
@@ -49,6 +49,13 @@ PathConfigStorage::MeshPathConfigs::MeshPathConfigs(const SliceMeshStorage& mesh
     , mesh.getSettingInPercentage("material_flow")
     , GCodePathConfig::SpeedDerivatives{mesh.getSettingInMillimetersPerSecond("speed_topbottom"), mesh.getSettingInMillimetersPerSecond("acceleration_topbottom"), mesh.getSettingInMillimetersPerSecond("jerk_topbottom")}
 )
+, ironing_config(
+    PrintFeatureType::Skin
+    , mesh.getSettingInMicrons("skin_line_width")
+    , layer_thickness
+    , mesh.getSettingInPercentage("material_flow")
```
BagelOrb: Shouldn't this be `ironing_flow` ?!

src/settings/PathConfigStorage.cpp
```diff
@@ -49,6 +49,13 @@ PathConfigStorage::MeshPathConfigs::MeshPathConfigs(const SliceMeshStorage& mesh
     , mesh.getSettingInPercentage("material_flow")
     , GCodePathConfig::SpeedDerivatives{mesh.getSettingInMillimetersPerSecond("speed_topbottom"), mesh.getSettingInMillimetersPerSecond("acceleration_topbottom"), mesh.getSettingInMillimetersPerSecond("jerk_topbottom")}
 )
+, ironing_config(
+    PrintFeatureType::Skin
+    , mesh.getSettingInMicrons("skin_line_width")
```
BagelOrb: I think this should be `ironing_line_spacing`

src/sliceDataStorage.h
```diff
@@ -102,6 +105,14 @@ class SliceLayer
     Polygons openPolyLines; //!< A list of lines which were never hooked up into a 2D polygon. (Currently unused in normal operation)
 
     /*!
+     * \brief The parts of the model that are exposed at the very top of the
+     * model.
+     *
+     * This is filled only when the top surface is needed.
+     */
+    TopSurface* top_surface = nullptr;
```
BagelOrb: I thought this kind of syntax caused problems for Windows...
I thought it had to be `TopSurface* top_surface(nullptr);`

Not sure, though...

src/TopSurface.cpp
```diff
@@ -0,0 +1,63 @@
...
+//CuraEngine is released under the terms of the AGPLv3 or higher.
+
+#include "infill.h"
+#include "LayerPlan.h"
+#include "TopSurface.h"
+
+namespace cura
+{
+
+TopSurface::TopSurface()
```
BagelOrb: This constructor is unused. You already catch not doing ironing by setting top_surface to nullptr.

PR #535, opened 2017-06-16
CURA-3930 fix extruder train settings for single extrusion machines

LipuFei:


src/config.h
```diff
@@ -0,0 +1,23 @@
...
+    bool coasting_enable; //!< Whether coasting is enabled on the extruder to which this config is attached 
+    double coasting_volume; //!< The volume leeked when printing without feeding
+    double coasting_speed; //!< A modifier (0-1) on the last used travel speed to move slower during coasting
+    double coasting_min_volume;  //!< The minimal volume printed to build up enough pressure to leek the coasting_volume
+};
+
+
+} // cura
+
+#endif // CURAENGINE_CONFIGS_H
```
Ghostkeeper: Code style: The file name must equal the struct name (and each class/struct must have its own header file).

src/gcodeExport.cpp
```diff
@@ -57,7 +57,37 @@ void GCodeExport::preSetup(const MeshGroup* meshgroup)
...
+            settings_to_fetch_list.push_back("extruder_prime_pos_abs");
+            settings_to_fetch_list.push_back("machine_nozzle_temp_enabled");
+            settings_to_fetch_list.push_back("prime_blob_enable");
+            settings_to_fetch_list.push_back("machine_nozzle_size");
+            //settings_to_fetch_list.push_back("machine_nozzle_offset_x");  // only for multi-extrusion machines
+            //settings_to_fetch_list.push_back("machine_nozzle_offset_y");  // only for multi-extrusion machines
+            settings_to_fetch_list.push_back("material_guid");
+            //settings_to_fetch_list.push_back("machine_extruder_start_code");  // only for multi-extrusion machines
+            //settings_to_fetch_list.push_back("machine_extruder_end_code");  // only for multi-extrusion machines
+            settings_to_fetch_list.push_back("retraction_prime_speed");
```
Ghostkeeper: Shouldn't it need to just update ALL settings in the extruder train?

If we're debugging later for some bug in other pieces of code, we'll discover that somehow the `retraction_prime_speed` gets inherited from the global stack into the extruder stack, but its sister `retraction_retract_speed` doesn't. I'd expect the extruder train to contain all settings properly, not just the ones that happened to be necessary on line 103 of GCodeExport. And I'd expect the extruder train to contain all settings from the beginning of the slicing, not just from the Export stage onwards.

src/config.h
BagelOrb: Furthermore the header guard should be derived from the file name and location. Please look at how this is done in other header files..

src/gcodeExport.cpp
BagelOrb: The extruder trains already inherit from the global settings base.

BagelOrb:
This is not the right way to solve the problem.

Extruder Train settings bases already inherit from the global settins. There is no reason to copy over setting values to a different settings base.

Moreover, the fact that the frontend puts all extruder train settings in the global stack for single extrusion machines is a frontend implementation detail. That shouldn't have effect on how the backend handles settings.

LipuFei:
@BagelOrb This is correct for multi-extrusion machines but not single-extrusion ones. The problem is that, Cura won't have any extruder stacks for single-extrusion machines. There will only be a global stack. The ExtruderTrains initialized here are based on the fdmextruder definitions and the extruder stack values from Cura. In this case, if you set a 0.8mm nozzle for UM2, nozzle size in the global stack will be 0.8mm, but no extruder stack settings will be sent by Cura, so the engine will initialize from fdmextruder definition, by default giving the value 0.4mm, and that doesn't inherit from the global stack. So, eventually, if we use a single-extrusion machine, the engine uses the default values from the fdmextruder definition instead of what we set in the global stack.

BagelOrb:
I understand where the problem comes from, but I think we should first explore what the ideal solution would look like before we introduce hacks like the one you implemented.

I am not in the office, so maybe you already discussed how the problem should be solved ideally..

If you ask me the ideal solution is to also use an extruder stack in the frontend when it's a single extrusion machine.
Given that fdmprinter refers to fdmextruder it seems to me that the existence of an extruder is already acknowledged for all machines including single extrusion machines. The problem is that that fact is not represented in the setting stacks.
Moreover, conceptually each printer does have an extruder train; an extruder train is defined as the combination of an extruder, filament and nozzle. These are things a single extrusion machine does have.

Making single extrusion machines have multiple extruders might be too much work.
One way in which to introduce a hack, but keep the hack in the frontend is to copy all settings from the global stack to the extruder stack just before sending the settings to the engine...

src/config.h
LipuFei: @BagelOrb A prefix such as the application/project name for things like header file guards is commonly used to make it more secure. For example, if two separate libraries/applications both happen to have a `setting.h` in them both using the header guard `SETTING_H`, and one header includes the other, there will be a problem.

LipuFei:
@BagelOrb I agree this is a hack, and that was why I created this PR to have a discussion.

This problem is mainly cause by the inconsistency of global stack and extruder stacks between Cura and CuraEngine, so making them consistency is the most logically way.

Another thing I was thinking about is to prepare those settings first thing before starting a slice. Now the settings preparation is kinda scattered around multiple places (`FffGcodeWriter`, `gcodeExporter`, etc.), so I think it makes more sense to put centralize those settings preparation in one place in the beginning, and all other components can just go there to fetch the settings. But this is more like an extra refactoring work of course.

src/config.h
BagelOrb: Okay, but our current scheme is FILE_LOCATION_FILE_NAME_H. Either you have to omit the `CURAENGINE_` part or add it to all files.

BagelOrb:
I don't really understand what you mean by setting preparation.
The only type of setings preparation I can think of is loading in settings and loading in settings base inheritance configuration (limit_to_extruder).

Or do you mean reading in settings into temporary settings structs?

src/gcodeExport.cpp
```diff
@@ -57,7 +57,23 @@ void GCodeExport::preSetup(const MeshGroup* meshgroup)
...
+            for (auto setting_key : train->getAllLocalSettingKeys())
+            {
+                if (not storage.hasSettingKey(setting_key))
+                {
+                    continue;
+                }
+
+                train->setSetting(setting_key, storage.getSettingString(setting_key));
+            }
+        }
```
Ghostkeeper: This bit gets called rather late in the process. I would rather see that this is done in the settings registry upon creating the extruder train. Perhaps it doesn't need to be only the first extruder, either.

BagelOrb:
I'm strongly of the opinion that the problem should be fixed in the frontend.

If you ask me I would reject this hack.

src/gcodeExport.cpp
LipuFei: `MeshGroup.createExtruderTrain()` is a nicer place for this piece of code, but it's not an easy task to move it there. A major refactoring might be required.

LipuFei:
Changed this to base on 2.6

src/gcodeExport.cpp
BagelOrb: The CommandSocket would be a good candidate perhaps, though.

LipuFei:
We decided to use this approach: https://github.com/Ultimaker/Cura/pull/2015
Closing this PR.

PR #536, opened 2017-06-19
Make user specified z seam position immune to position of part on platform

smartavionics:
What this PR does is make the user specified z seam origin relative to the mesh's BB rather than being an
absolute position on the platform. The rational behind this is that without the PR, the z seam will change position depending on where the part is on the platform. With the PR, the z seam will not change position when the part is moved and if the part is multiplied, all the parts will have exactly the same z seams.

This appears to work OK but should we have a setting to control whether the z seam position is absolute (as it is now) or relative (as in this PR)?

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -160,7 +160,7 @@ unsigned int FffGcodeWriter::findSpiralizedLayerSeamVertexIndex(const SliceDataS
         Point seam_pos(0, 0);
         if (mesh.getSettingAsZSeamType("z_seam_type") == EZSeamType::USER_SPECIFIED)
         {
-            seam_pos = Point(mesh.getSettingInMicrons("z_seam_x"), mesh.getSettingInMicrons("z_seam_y"));
+            seam_pos = mesh.getZSeamOrigin();
         }
         return PolygonUtils::findClosest(seam_pos, layer.parts[0].insets[0][0]).point_idx;
     }
@@ -965,8 +965,7 @@ void FffGcodeWriter::addMeshLayerToGCode_meshSurfaceMode(const SliceDataStorage&
     }
 
     EZSeamType z_seam_type = mesh.getSettingAsZSeamType("z_seam_type");
-    Point z_seam_pos(mesh.getSettingInMicrons("z_seam_x"), mesh.getSettingInMicrons("z_seam_y"));
-    gcode_layer.addPolygonsByOptimizer(polygons, &mesh_config.inset0_config, nullptr, z_seam_type, z_seam_pos, mesh.getSettingInMicrons("wall_0_wipe_dist"), getSettingBoolean("magic_spiralize"));
+    gcode_layer.addPolygonsByOptimizer(polygons, &mesh_config.inset0_config, nullptr, z_seam_type, mesh.getZSeamOrigin(), mesh.getSettingInMicrons("wall_0_wipe_dist"), getSettingBoolean("magic_spiralize"));
 
     addMeshOpenPolyLinesToGCode(mesh, mesh_config, gcode_layer, layer_nr);
 }
@@ -1014,9 +1013,8 @@ void FffGcodeWriter::addMeshLayerToGCode(const SliceDataStorage& storage, const 
 
 
     EZSeamType z_seam_type = mesh.getSettingAsZSeamType("z_seam_type");
-    Point z_seam_pos(mesh.getSettingInMicrons("z_seam_x"), mesh.getSettingInMicrons("z_seam_y"));
     Point layer_start_position = Point(train->getSettingInMicrons("layer_start_x"), train->getSettingInMicrons("layer_start_y"));
-    PathOrderOptimizer part_order_optimizer(layer_start_position, z_seam_pos, z_seam_type);
+    PathOrderOptimizer part_order_optimizer(layer_start_position, mesh.getZSeamOrigin(), z_seam_type);
     for(unsigned int partNr = 0; partNr < layer.parts.size(); partNr++)
     {
         part_order_optimizer.addPolygon(layer.parts[partNr].insets[0][0]);
@@ -1057,8 +1055,7 @@ void FffGcodeWriter::addMeshPartToGCode(const SliceDataStorage& storage, const S
     }
 
     EZSeamType z_seam_type = mesh.getSettingAsZSeamType("z_seam_type");
-    Point z_seam_pos(mesh.getSettingInMicrons("z_seam_x"), mesh.getSettingInMicrons("z_seam_y"));
-    added_something = added_something | processInsets(storage, gcode_layer, mesh, extruder_nr, mesh_config, part, layer_nr, z_seam_type, z_seam_pos);
+    added_something = added_something | processInsets(storage, gcode_layer, mesh, extruder_nr, mesh_config, part, layer_nr, z_seam_type, mesh.getZSeamOrigin());
 
     if (!mesh.getSettingBoolean("infill_before_walls"))
     {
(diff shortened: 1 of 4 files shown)


smartavionics:
See https://github.com/Ultimaker/Cura/pull/1995.

BagelOrb:
I'm not really sure about this feature. I'm not sure whether people really want this extra setting.

The main reason not to merge this would be that adding another setting clutters the UI more than it already is.

I'll have to talk this over with other people.

smartavionics:
If people don't want to use it, they can leave the setting hidden. The default value keeps the current (IMHO broken) behaviour.

Ghostkeeper:
Devs, see CURA-3971.

BagelOrb:
Okay I guess there's different ways to look at this.
Let's merge this PR.

First let me do a review, though ;)

src/FffGcodeWriter.cpp
```diff
@@ -886,17 +886,20 @@ std::vector<unsigned int> FffGcodeWriter::getUsedExtrudersOnLayerExcludingStarti
...
 
     for (unsigned int mesh_idx = 0; mesh_idx < storage.meshes.size(); mesh_idx++)
     {
-        const SliceMeshStorage& mesh = storage.meshes[mesh_idx];
+        SliceMeshStorage& mesh = storage.meshes[mesh_idx];
         if (mesh.getExtruderIsUsed(extruder_nr))
         {
             const Mesh& mesh_data = storage.meshgroup->meshes[mesh_idx];
             const Point3 middle = (mesh_data.getAABB().min + mesh_data.getAABB().max) / 2;
+            mesh.middle.x = middle.x;
```
BagelOrb: I think this is a weird place to initialize that member variable.
It should be initialized as early as possible.
I think it can be initialized at the moment the SliceMeshStorage is constructed.

It might also be nice to just store the whole AABB, because then we can use the member for other things as well in the future.

Note: rather than computing the middle of the AABB, I should just have called `mesh_data.getAABB().getMiddle()`

src/FffGcodeWriter.h
```diff
@@ -312,7 +312,7 @@ class FffGcodeWriter : public SettingsMessenger, NoCopy
      * \param extruder_nr The extruder for which to determine the order
      * \return A vector of mesh indices ordered on print order for that extruder.
      */
-    std::vector<unsigned int> calculateMeshOrder(const SliceDataStorage& storage, int extruder_nr) const;
+    std::vector<unsigned int> calculateMeshOrder(SliceDataStorage& storage, int extruder_nr) const;
```
BagelOrb: When you initialize the middle (or the aabb when you save that instead) at an earlier stage, please revert this change.

src/sliceDataStorage.h
```diff
@@ -184,11 +184,13 @@ class SliceMeshStorage : public SettingsMessenger // passes on settings from a M
 
     std::vector<int> infill_angles; //!< a list of angle values (in degrees) which is cycled through to determine the infill angle of each layer
     std::vector<int> skin_angles; //!< a list of angle values (in degrees) which is cycled through to determine the skin angle of each layer
+    Point3 middle; //!< the middle of the mesh's AABB
```
BagelOrb: Proposed: save the AABB3D instead.
[not blocking]

src/sliceDataStorage.h
```diff
@@ -208,6 +210,18 @@ class SliceMeshStorage : public SettingsMessenger // passes on settings from a M
...
+
+    /*!
+     * \return the mesh's user specified z seam hint
+     */
+    Point getZSeamOrigin() const {
+        Point pos(getSettingInMicrons("z_seam_x"), getSettingInMicrons("z_seam_y"));
+        if(getSettingBoolean("z_seam_relative"))
+        {
+            pos += Point(middle.x, middle.y);
+        }
```
BagelOrb: Please put implementation in cpp file.

src/sliceDataStorage.h
```diff
@@ -208,6 +210,18 @@ class SliceMeshStorage : public SettingsMessenger // passes on settings from a M
      * \return whether a particular extruder is used by this mesh on a particular layer
      */
     bool getExtruderIsUsed(int extruder_nr, int layer_nr) const;
+
+    /*!
+     * \return the mesh's user specified z seam hint
+     */
+    Point getZSeamOrigin() const {
+        Point pos(getSettingInMicrons("z_seam_x"), getSettingInMicrons("z_seam_y"));
+        if(getSettingBoolean("z_seam_relative"))
```
BagelOrb: Code style: space after `if`: i.e. `if (getSettingBoolean("z_seam_relative"))`

src/sliceDataStorage.h
```diff
@@ -208,6 +210,18 @@ class SliceMeshStorage : public SettingsMessenger // passes on settings from a M
      * \return whether a particular extruder is used by this mesh on a particular layer
      */
     bool getExtruderIsUsed(int extruder_nr, int layer_nr) const;
+
+    /*!
+     * \return the mesh's user specified z seam hint
+     */
+    Point getZSeamOrigin() const {
```
BagelOrb: Code style: `{` on separate line.

I think the naming of this function is a bit weird. The word 'origin' doesn't seem the right word...
Maybe just `getZSeamLocationHint()` ?

src/sliceDataStorage.h
```diff
@@ -208,6 +210,18 @@ class SliceMeshStorage : public SettingsMessenger // passes on settings from a M
      * \return whether a particular extruder is used by this mesh on a particular layer
      */
     bool getExtruderIsUsed(int extruder_nr, int layer_nr) const;
+
+    /*!
+     * \return the mesh's user specified z seam hint
```
BagelOrb: I like that you use the word 'hint' here.
Makes it a lot more clear what this is! :+1:

smartavionics:
Thanks for the review, will work on it later today...

src/FffGcodeWriter.cpp

src/FffGcodeWriter.h
BagelOrb: fixed

src/sliceDataStorage.h
BagelOrb: fixed

src/sliceDataStorage.h
BagelOrb: fixed

src/sliceDataStorage.h
BagelOrb: fixed

src/sliceDataStorage.h
BagelOrb: fixed

BagelOrb:
There seems to be a problem:
image

As you can see the z seam isn't on the same location for both meshes.
(I've enabled coasting to make it easy to see where the seam is.)

However the same seam locations are generated without this PR, so I don't really know where the problem is.

In any case I cannot really think of a test which wouldn't fail on this, so I cannot conclude that this PR works.

smartavionics:
Hmm, how strange, can you please make the model available for me to test

BagelOrb:
Hmm wait.

BagelOrb:
The picture above is a bit misleading because the seam position is right on the middle of the left cone.

Here you see that relative Seam doesn't work (at least not on the UM3)
image

I'm getting the same results on the UM2, a single extrusion machine.

BagelOrb:
Wait

BagelOrb:
Branch name was too long and I ended up not pulling your branch...

Would be nice if you keep the branch names a bit shorter

BagelOrb:
All is good. Never mind my stupid mistakes :D

smartavionics:
I don't know what platform you develop on but if you use linux and have the git completion set up then you get branch name completion so you can say "git merge mb-TAB" and it will list all mb-branches (just like file name completion)

BagelOrb:
I don't know whether that also holds for external remotes...

smartavionics:
Ah, yes, you're probably right there, silly me.

PR #539, opened 2017-06-22
Feature prime tower purge

BagelOrb:


LipuFei:
@BagelOrb I have removed the new line type "PurgeMove" and changed it to use an "extrusion_speed_factor" in GCodePathConfig for doing the speed compensation.

src/LayerPlan.cpp
```diff
@@ -849,8 +849,13 @@ void LayerPlan::writeGCode(GCodeExport& gcode)
             if (path.config->isTravelPath())
                 speed *= extruder_plan.getTravelSpeedFactor();
             else
+            {
                 speed *= extruder_plan.getExtrudeSpeedFactor();
 
+                // for some movements such as prime tower purge, the speed may get changed by this factor
+                speed *= path.config->getExtrusionSpeedFactor();
```
BagelOrb: I think this edit is slightly in the wrong place.
It would be better 10 lines above.

Create a function `GCodePath::getSpeed()` which returns `config->getSpeed() * extrusion_speed_factor`

The line above should then just read `double speed = path.getSpeed();`

src/GCodePathConfig.h
```diff
@@ -29,9 +29,10 @@ class GCodePathConfig
     const int line_width; //!< width of the line extruded
     const int layer_thickness; //!< current layer height in micron
     const double flow; //!< extrusion flow modifier in %
+    const double extrusion_speed_factor; //! < extrusion speed factor (only used by purge moves for now)
```
BagelOrb: I don't see any reason to limit this factor to extrusion moves.

Because the factor can be set separately for any move (extrusion or travel) I think you should rename this to `speed_factor`.

src/PrimeTower.cpp
```diff
@@ -247,16 +249,118 @@ void PrimeTower::preWipe(const SliceDataStorage& storage, LayerPlan& gcode_layer
...
+#endif // NDEBUG
+
+                const double compensated_speed = purge_move_length_mm / min_time_needed_for_extrusion;
+                extrusion_speed_factor = compensated_speed / speed;
+            }
+
+            // As we need a plan, which can't have a stationary extrusion, we use an extrusion move to prime.
+            // This has the added benefit that it will evenly spread the primed material inside the tower.
+            // FIXME: this can cause out-of-memory issues because we are allocating memory blocks on the heap without releasing them
+            GCodePathConfig *prime_tower_purge_gcode_path_config = new GCodePathConfig(
```
BagelOrb: There's a FIXME here and I think this should be fixed.

Creating a new GCodePathConfig in the right way can be quite a hassle.

You can just reuse the existing prime tower config and adjust the speed and extrusion of the line segment for the purge move.

src/PrimeTower.cpp
```diff
@@ -247,16 +249,118 @@ void PrimeTower::preWipe(const SliceDataStorage& storage, LayerPlan& gcode_layer
...
+            // Normal move behavior to wipe start location.
+            GCodePath& toward_wipe_start = gcode_layer.addTravel_simple(prime_start);
+            toward_wipe_start.perform_z_hop = false;
+            toward_wipe_start.retract = true;
+        }
     }
     else
     {
-        gcode_layer.addTravel(start);
+        if (purge_volume > 0)
```
BagelOrb: There's a lot of code duplication here. Please merge these two blocks into a new function, which gets called twice.

src/GCodePathConfig.h
```diff
@@ -77,6 +80,12 @@ class GCodePathConfig
 
     double getFlowPercentage() const;
 
+    const SpeedDerivatives& getSpeedDerivatives() const;
```
BagelOrb: This function is unused when you follow on my comments below.
In that case remove this function please.

src/PrimeTower.cpp
```diff
@@ -247,16 +249,118 @@ void PrimeTower::preWipe(const SliceDataStorage& storage, LayerPlan& gcode_layer
...
+            const double max_possible_extursion_speed_mm3_per_sec = 3.0;
+
+            const double speed = current_gcode_path_config->getSpeed();
+            double extrusion_speed_factor = 1.0;
+
+            if (purge_extrusion_speed_mm3_per_sec > max_possible_extursion_speed_mm3_per_sec)
+            {
+                // compensate the travel speed for the large extrusion amount
+                const double min_time_needed_for_extrusion = purge_volume / max_possible_extursion_speed_mm3_per_sec;
+#ifndef NDEBUG
```
BagelOrb: What's this `NDEBUG`? I've never seen that macro before..

BagelOrb: Shouldn't it just be `DEBUG`?

src/GCodePathConfig.h
LipuFei: This is actually a travel speed factor, but this is only for the travel speed during extrusion, so pure travel moves should not be affected by this. That's why I named it as `extrusion_speed_factor`, but then this can still be confused with the speed of the extruder motor(s).

src/PrimeTower.cpp
LipuFei: `NDEBUG` is defined when we are building a `RELEASE` version. This is the macro which enables/disables the `assert()`s. In a `DEBUG` build, `NDEBUG` is *not* defined and `assert` will work; in a `RELEASE` build, things happen in the opposite way.

src/LayerPlan.cpp
LipuFei: In the scope of this feature, this speed factor should only affect extrusion travel speeds, not pure travel moves. It is possible to extend this factor to be used for both cases, but it is not intended at the moment. I think changing the name to speed factor to `speed_factor` makes it too general and implies that it does something to the travel move too, which it doesn't. I think we should leave this to when we want to extend this feature.

src/GCodePathConfig.h
LipuFei: Will be removed.

LipuFei: Wait... This is actually used, See `PrimeTower.cpp`

src/PrimeTower.cpp
```diff
@@ -268,5 +288,44 @@ void PrimeTower::subtractFromSupport(SliceDataStorage& storage)
...
+    const double max_possible_extursion_speed_mm3_per_sec = 3.0;
+
+    const double speed = current_gcode_path_config->getSpeed();
+    double speed_factor = 1.0;
+
+    if (purge_extrusion_speed_mm3_per_sec > max_possible_extursion_speed_mm3_per_sec)
+    {
+        // compensate the travel speed for the large extrusion amount
+        const double min_time_needed_for_extrusion = purge_volume / max_possible_extursion_speed_mm3_per_sec;
+#ifndef NDEBUG
```
BagelOrb: make it `#ifdef DEBUG`

src/LayerPlan.cpp
BagelOrb: We discussed this in person and I said we could have just as well made the factor work for travel moves as well, but it's not necessary for you to fix this now. We can also do this in the future.

src/GCodePathConfig.h
BagelOrb: Same as comment above. It's okay like this.

src/GCodePathConfig.h
BagelOrb: It's removed now..

src/PrimeTower.cpp
```diff
@@ -268,5 +288,44 @@ void PrimeTower::subtractFromSupport(SliceDataStorage& storage)
...
+    // Find out how much purging needs to be done.
+    const coord_t purge_move_length = vSize(start_pos - end_pos);
+    const unsigned int line_width = gcode_layer.configs_storage.prime_tower_config_per_extruder[extruder_nr].getLineWidth();
+    const double layer_height_mm = (layer_nr == 0) ? train->getSettingInMillimeters("layer_height_0") : train->getSettingInMillimeters("layer_height");
+    const double normal_volume = INT2MM(INT2MM(purge_move_length * line_width)) * layer_height_mm; // Volume extruded on the "normal" move
+    float purge_flow = purge_volume / normal_volume;
+
+    const GCodePathConfig *current_gcode_path_config = &gcode_layer.configs_storage.prime_tower_config_per_extruder[extruder_nr];
+
+    const coord_t purge_move_length_mm = INT2MM(purge_move_length);
```
BagelOrb: The type for millimeters is `double`, not `coord_t`. Please make this `double`

src/PrimeTower.cpp
```diff
@@ -268,5 +288,44 @@ void PrimeTower::subtractFromSupport(SliceDataStorage& storage)
...
+
+    if (purge_extrusion_speed_mm3_per_sec > max_possible_extursion_speed_mm3_per_sec)
+    {
+        // compensate the travel speed for the large extrusion amount
+        const double min_time_needed_for_extrusion = purge_volume / max_possible_extursion_speed_mm3_per_sec;
+#ifdef DEBUG
+        // sanity check to make sure that the newly calculated travel time is no less than the old travel time
+        // because we are trying to give time for the extrusion.
+        const double old_travel_time = purge_move_length_mm / speed;
+        assert(old_travel_time <= min_time_needed_for_extrusion);
```
BagelOrb: This could be wrong due to rounding errors...

src/PrimeTower.h
```diff
@@ -137,14 +137,33 @@ class PrimeTower
     void addToGcode_denseInfill(LayerPlan& gcode_layer, const int layer_nr, const int extruder) const;
 
     /*!
-     * Plan the moves for wiping the current nozzles oozed material before starting to print the prime tower.
+     * Plan the moves for wiping the current nozzles oozed material before starting to print the prime tower. 
+     * This generates wipe moves from inside the tower towards the outside if the tower is hollow. Otherwise
+     * it wipes from outside to inside.
```
BagelOrb: Please add that this function also adds a purge, right?

src/PrimeTower.cpp
```diff
@@ -225,7 +225,7 @@ void PrimeTower::generateWipeLocations(const SliceDataStorage& storage)
     PolygonUtils::spreadDots(segment_start, segment_end, number_of_pre_wipe_locations, pre_wipe_locations);
 }
 
-void PrimeTower::preWipe(const SliceDataStorage& storage, LayerPlan& gcode_layer, const int layer_nr, const int extruder_nr) const
+void PrimeTower::preWipeAndPurge(const SliceDataStorage& storage, LayerPlan& gcode_layer, const int layer_nr, const int extruder_nr) const
```
BagelOrb: would have been better in the chronological order, i.e. purgeAndPreWipe, but whatever..
[not blocking]

src/pathPlanning/GCodePath.cpp
```diff
@@ -6,11 +6,12 @@
...
 {
 
-GCodePath::GCodePath(const GCodePathConfig* config, SpaceFillType space_fill_type, float flow, bool spiralize) :
+GCodePath::GCodePath(const GCodePathConfig* config, SpaceFillType space_fill_type, float flow, bool spiralize, double speed_factor) :
 config(config),
 space_fill_type(space_fill_type),
 flow(flow),
-spiralize(spiralize)
+spiralize(spiralize),
+speed_factor(speed_factor)
```
BagelOrb: I'm getting a compile warning here:
"spiralize will be initialized after"

You should initialize the member variables in the same order as they are defined.

PR #540, opened 2017-06-26
Add setting to change horizontal expansion for layer 0

fieldOfView:
This PR adds functionality to change the Horizontal Expansion for the first layer only. This can be used to compensate for "elephant's foot".

See https://github.com/Ultimaker/Cura/issues/1573 for discussion
Also see Cura PR https://github.com/Ultimaker/Cura/pull/2031

### DIFF
--- a/src/slicer.cpp
+++ b/src/slicer.cpp
@@ -766,7 +766,9 @@ void SlicerLayer::makePolygons(const Mesh* mesh, bool keep_none_closed, bool ext
     for (PolygonRef polyline : open_polylines)
     {
         if (polyline.size() > 0)
+        {
             openPolylines.add(polyline);
+        }
     }
 
     //Remove all the tiny polygons, or polygons that are not closed. As they do not contribute to the actual print.
@@ -780,8 +782,10 @@ void SlicerLayer::makePolygons(const Mesh* mesh, bool keep_none_closed, bool ext
     polygons.removeDegenerateVerts(); // remove verts connected to overlapping line segments
 
     int xy_offset = mesh->getSettingInMicrons("xy_offset");
-    if(is_initial_layer)
+    if (is_initial_layer)
+    {
         xy_offset = mesh->getSettingInMicrons("xy_offset_layer_0");
+    }
 
     if (xy_offset != 0)
     {
--- a/src/slicer.cpp
+++ b/src/slicer.cpp
@@ -734,7 +734,7 @@ ClosePolygonResult SlicerLayer::findPolygonPointClosestTo(Point input)
     return ret;
 }
 
-void SlicerLayer::makePolygons(const Mesh* mesh, bool keep_none_closed, bool extensive_stitching)
+void SlicerLayer::makePolygons(const Mesh* mesh, bool keep_none_closed, bool extensive_stitching, bool is_initial_layer)
 {
     Polygons open_polylines;
 
@@ -780,6 +780,9 @@ void SlicerLayer::makePolygons(const Mesh* mesh, bool keep_none_closed, bool ext
     polygons.removeDegenerateVerts(); // remove verts connected to overlapping line segments
 
     int xy_offset = mesh->getSettingInMicrons("xy_offset");
+    if(is_initial_layer)
+        xy_offset = mesh->getSettingInMicrons("xy_offset_layer_0");
+
     if (xy_offset != 0)
     {
         polygons = polygons.offset(xy_offset);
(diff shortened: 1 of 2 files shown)


BagelOrb:
Jaaa hooooor! Komt ie mij ook ff lastig vallen! :P

src/slicer.cpp
```diff
@@ -780,6 +780,9 @@ void SlicerLayer::makePolygons(const Mesh* mesh, bool keep_none_closed, bool ext
     polygons.removeDegenerateVerts(); // remove verts connected to overlapping line segments
 
     int xy_offset = mesh->getSettingInMicrons("xy_offset");
+    if(is_initial_layer)
```
BagelOrb: Code style: `if (condition)`

src/slicer.cpp
```diff
@@ -780,6 +780,9 @@ void SlicerLayer::makePolygons(const Mesh* mesh, bool keep_none_closed, bool ext
...
+        xy_offset = mesh->getSettingInMicrons("xy_offset_layer_0");
```
BagelOrb: Code style: always use brackets on separate lines in C++

nallath:
CURA-3968

src/slicer.cpp
```diff
@@ -780,6 +780,9 @@ void SlicerLayer::makePolygons(const Mesh* mesh, bool keep_none_closed, bool ext
...
+
```
BagelOrb: I expected the initial layer XY offset to add onto the normal xy offset.

The initial layer XY offset is to get rid of elephant feat.

The normal XY offset is to compensate for hole dimensions for example.

If you do a large negative xy_offset it would be weird that you get the elephant feat back again because you didn't also change the xy_offset_layer_0...

src/slicer.cpp
BagelOrb: i.e.
```
if (cond)
{
    doStuff();
}
```

src/slicer.cpp
fieldOfView: Fixed

src/slicer.cpp
fieldOfView: Fixed. Also fixed one in the existing code.

src/slicer.cpp
nallath: It should override it right? It's up to the front end to inherit the value of the xy_offset_layer_0 from xy_offset

fieldOfView: I implemented it in a way that it is in line with other "Initial Layer" settings. "Initial Layer Height" is not a delta to "Layer Height", and neither is "Printing Temperature Initial Layer". So in my opinion it would be counter-intuitive if this one new setting were.

BagelOrb:
I disagree with your sentiment about relative settings.

You set a prime tower size instead of a prime tower end point position.

The raft margin is added onto the horizontal expansion, the skirt margin is added onto the horizontal expansion and even the brim starts at the edge of the offsetted model, not the original outlines.

If you look at it like this then the general rule is that horizontal expansion is applied before anything else. It changes the model and anything along with it. The elephant feet compensation shouldn't be any different.

Moreover, these settings control different things: dimensional accuracy of holes and elephant feet.
You wouldn't want it if changing your dimensional accuracy would lead to different elephant feet.

Perhaps if you call it something like "Initial Layer Shrinkage" you don't conceptually link it to Horizontal Expansion, which is effectively used for something totally different.

fieldOfView:
"Layer Height" and "Initial Layer Height" also control different things: quality and strength of the print vs bed adhesion. Yet initial layer height is not specified as an offset to layer height.

As @nallath says, normally the frontend inheritance settings will keep the Initial Layer Horizontal Expansion and the Horizontal Expansion together.

I remember a lunch-lecture on design patterns (in code). UX design has also adopted the concept of design patterns, but in UX these patterns are often "deduced" from behavior in the application; you can also see it as "consistent behavior". I don't want to break the pattern between a main setting and a specialized setting for the first layer.

I see two options:
* I rename the setting to something entirely different, eg "Elephant Foot Compensation" or "Initial Layer Squish Compensation", which behaves as you describe, or
* I keep "Horizontal Expansion" and "Initial Layer Horizontal Expansion" and the latter overrides the former like other "initial layer" settings.

I don't mind either way.

nallath:
I agree with FieldOfView, but I've been unable to convince BagelOrb that this is the "right" way.

BagelOrb:
The thing is that the initial layer height depends on the bed, so it should stay the same while the layer height changes over the profiles.
Moreover, both settings control the height of a layer, while the horizontal expansion is in itself already a relative setting to the actual outline of a layer.

Perhaps 'Initial Layer Inset' ?

Your argument about consistency doesn't hold when the new setting is conceptualized in such a way that it doesn't seem related to Horizontal Expansion. I think that you see the connection between those settings because you know how such things are implemented in the same way.

Now that that argument is out of the way, the only thing that remains is the usability of the setting. If a user has its own values for Horizontal Expansion and for Initial Layer Inset, having the Initial Layer Inset work independently of Horizontal Expansion makes it easier to tweak the value of Horizontal Expansion.

I hope that I have convinced you now. I've already put quite some effort into this. If you still don't agree with me than just leave it as it is...

fieldOfView:
> If you still don't agree with me than just leave it as it is...

Given the choice, I would keep the current implementation. Not to be lazy, but because I feel that it is better than the alternatives I have come up with sofar.

Note that I just implemented a feature request because it sounded like something trivial to add while fixing an issue that I am also dealing with from time to time with prints. I did not mean to take up (much) of your time (or mine).

BagelOrb:
I didn't mean to blame you for the time. Ihose to invest time..

You have my hesitant
[approved]

Op 28 jun. 2017 11:18 schreef "Aldo Hoeben" <notifications@github.com>:

nallath:
Thanks for the contribution!

PR #541, opened 2017-06-26
Now outputs "M205 Xval Yval" rather than "M205 Xval".

smartavionics:
The newer versions of Marlin now expect the X and Y axis jerk values to be specified
separately. Before, Xval would set both the X and Y axis to use the same jerk val.

I have looked at the early Marlin source and it doesn't look like it will barf on the extra
Y value so this change should be backwards compatible.

### DIFF
--- a/src/gcodeExport.cpp
+++ b/src/gcodeExport.cpp
@@ -1014,7 +1014,7 @@ void GCodeExport::writeJerk(double jerk)
         }
         else
         {
-            *output_stream << "M205 X";
+            *output_stream << "M205 X" << PrecisionedDouble{2, jerk} << " Y";
         }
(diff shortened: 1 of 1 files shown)


src/gcodeExport.cpp
```diff
@@ -1014,7 +1014,7 @@ void GCodeExport::writeJerk(double jerk)
         }
         else
         {
-            *output_stream << "M205 X";
+            *output_stream << "M205 X" << PrecisionedDouble{2, jerk} << " Y";
```
BagelOrb: :+1:

BagelOrb:
Easy fix.
Thanks!

PR #545, opened 2017-06-30
Spaghetti infill total volume (CURA-3861)

BagelOrb:


src/infill/SpaghettiInfill.cpp
```diff
@@ -4,8 +4,71 @@
...
 
+void SpaghettiInfill::generateTotalSpaghettiInfill(SliceMeshStorage& mesh)
+{
+    double total_volume_mm3 = 0.0;
+
+    const int max_layer = mesh.layers.size() - 1 - mesh.getSettingAsCount("top_layers");
+    for (int layer_idx = 0; layer_idx <= max_layer; layer_idx++)
+    {
+        const coord_t layer_height = (layer_idx == 0)? mesh.getSettingInMicrons("layer_height_0") : mesh.getSettingInMicrons("layer_height");
+        if (static_cast<int>(layer_idx) < mesh.getSettingAsCount("bottom_layers"))
```
LipuFei: `layer_idx` is already `int`, no need to cast.

src/infill/SpaghettiInfill.cpp
```diff
@@ -4,8 +4,71 @@
...
+        SliceLayer& layer = mesh.layers[layer_idx];
+        for (SliceLayerPart& part : layer.parts)
+        {
+            if (part.getOwnInfillArea().size() > 0)
+            { // WARNING: we just use the very first part we encounter; there's no way for the user to choose which filling area will be used
+                top_filling_layer_part = &part;
+                break;
+            }
+        }
+        if (top_filling_layer_part)
```
LipuFei: This is not necessary. The `if` above already ensures this.

LipuFei:
@BagelOrb  Ah, never mind, I will fix it.

src/infill/SpaghettiInfill.cpp
LipuFei: Ah, wait... you are right.

src/infill/SpaghettiInfill.cpp
LipuFei: Fixed in 0d403493d3030cb6ed6b496bef8715c835a915a7

PR #548, opened 2017-07-01
Skin infill refactor (CURA-3989)

BagelOrb:


src/FffGcodeWriter.cpp
```diff
@@ -1592,6 +1547,62 @@ void FffGcodeWriter::processSkinInsets(const SliceDataStorage& storage, LayerPla
...
+
+        if (pattern == EFillMethod::GRID || pattern == EFillMethod::LINES || pattern == EFillMethod::TRIANGLES || pattern == EFillMethod::CUBIC || pattern == EFillMethod::TETRAHEDRAL || pattern == EFillMethod::CUBICSUBDIV)
+        {
+            gcode_layer.addLinesByOptimizer(skin_lines, &mesh_config.skin_config, SpaceFillType::Lines, mesh.getSettingInMicrons("infill_wipe_dist"));
+        }
+        else
+        {
+            gcode_layer.addLinesByOptimizer(skin_lines, &mesh_config.skin_config, (pattern == EFillMethod::ZIG_ZAG)? SpaceFillType::PolyLines : SpaceFillType::Lines);
+        }
+    }
```
Ghostkeeper: The generating of lines for the perimeter gaps isn't in this function. Shouldn't that be close to the generating of polygons?

src/FffGcodeWriter.cpp
```diff
@@ -1496,25 +1498,6 @@ bool FffGcodeWriter::processSkinPart(const SliceDataStorage& storage, LayerPlan&
         skin_angle = bridge;
     }
 
-    // add insets
-    if (pattern != EFillMethod::CONCENTRIC)
```
Ghostkeeper: This condition got lost in the refactor.

src/FffGcodeWriter.cpp
```diff
@@ -1562,6 +1573,7 @@ bool FffGcodeWriter::processSkinPart(const SliceDataStorage& storage, LayerPlan&
         Polygons gap_polygons; // will remain empty
         Polygons gap_lines;
         int offset = 0;
+        int extra_infill_shift = 0;
```
Ghostkeeper: These two can be `constexpr`?

src/FffGcodeWriter.cpp
```diff
@@ -1517,15 +1515,28 @@ bool FffGcodeWriter::processSkinPart(const SliceDataStorage& storage, LayerPlan&
...
+    const bool generate_perimeter_gaps = fill_concentric_perimeter_gaps && pattern == EFillMethod::CONCENTRIC; // whether we need to generate perimeter gaps for concentric infill of the skinfill area
+
+    // generate skin_polygons and skin_lines (and concentric_perimeter_gaps if needed)
+    if (process_skin_infill || generate_perimeter_gaps)
+    {
+        // only generate infill if we're going to print it with this extruder
+        // or when we need to compute the perimeter gaps
+        // TODO: only compute this once when the infill is concentric and the perimeter gpas are printed with a different extruder
+        int extra_infill_shift = 0;
+        coord_t offset_from_inner_skin_infill = 0;
```
Ghostkeeper: These can be `constexpr`?

src/FffGcodeWriter.cpp
```diff
@@ -1517,15 +1515,28 @@ bool FffGcodeWriter::processSkinPart(const SliceDataStorage& storage, LayerPlan&
...
-    int extra_infill_shift = 0;
-    coord_t offset_from_inner_skin_infill = 0;
+    Polygons skin_polygons;
+    Polygons skin_lines;
     Polygons concentric_perimeter_gaps; // the perimeter gaps of the insets of concentric skin pattern of this skin part
-    Polygons* perimeter_gaps_output = (fill_concentric_perimeter_gaps)? &concentric_perimeter_gaps : nullptr;
-    Infill infill_comp(pattern, skin_part.inner_infill, offset_from_inner_skin_infill, skin_line_width, skin_line_width, skin_overlap, skin_angle, z, extra_infill_shift, perimeter_gaps_output);
-    infill_comp.generate(skin_polygons, skin_lines);
+
+    const bool process_skin_infill = extruder_nr == top_bottom_extruder_nr;
```
Ghostkeeper: If the "processing skin infill" and the "generating perimeter gaps" are in separate functions, this indirection won't be necessary.

src/skin.cpp
```diff
@@ -245,16 +245,11 @@ void SkinInfillAreaComputation::applySkinExpansion(const Polygons& original_outl
  *
  * this function may only read/write the skin and infill from the *current* layer.
  */
-void SkinInfillAreaComputation::generateSkinInsets(SliceLayerPart* part)
+void SkinInfillAreaComputation::generateSkinInsetsAndInnerSkinInfill(SliceLayerPart* part)
```
Ghostkeeper: Rule of thumb: If a function has the word "and" in its name, it does two things! I think the generating of inner skin infill can quite easily be separated even here.

src/FffGcodeWriter.cpp
```diff
@@ -1517,24 +1515,13 @@ bool FffGcodeWriter::processSkinPart(const SliceDataStorage& storage, LayerPlan&
...
-        }
-    }
-
-    if (inner_skin_outline == nullptr)
-    {
-        inner_skin_outline = &skin_part.outline;
     }
 
     int extra_infill_shift = 0;
+    coord_t offset_from_inner_skin_infill = 0;
```
Ghostkeeper: I think a variable name like `offset` could be more clear. You'd just see "oh, there is no offset" because it says `offset = 0` right above it.

(Not blocking.)

src/skin.cpp
```diff
@@ -49,6 +49,29 @@ Polygons SkinInfillAreaComputation::getInsidePolygons(const SliceLayerPart& part
  * This function is executed in a parallel region based on layer_nr.
  * When modifying make sure any changes does not introduce data races.
  *
+ * this function may only read/write the skin and infill from the *current* layer.
+ */
+Polygons SkinInfillAreaComputation::getOutlines(const SliceLayerPart& part_here, int layer2_nr)
```
Ghostkeeper: This function is never used. It also doesn't seem to coincide with what the commit message says happened here.

src/FffGcodeWriter.cpp
```diff
@@ -1475,41 +1475,39 @@ bool FffGcodeWriter::processSkinPart(const SliceDataStorage& storage, LayerPlan&
         mesh.getSettingAsFillPerimeterGapMode("fill_perimeter_gaps") != FillPerimeterGapMode::NOWHERE
         && !getSettingBoolean("magic_spiralize")
         && extruder_nr == wall_0_extruder_nr;
-    const bool fill_concentric_perimeter_gaps =
+
+    const bool fill_top_bottom_concentric_perimeter_gaps =
```
BagelOrb: That's in order to distinguish from the future variable `fill_top_most_`...

src/FffGcodeWriter.cpp
BagelOrb: That was actually a bug in the original code. I tested it and saw that the skin perimeters were missing when the pattern was concentric.

src/FffGcodeWriter.cpp
BagelOrb: I don't understand what you mean.
How is this boolean an indirection?
Please comment in the final PR; your comments are already outdated.

src/FffGcodeWriter.cpp
BagelOrb: meh

src/FffGcodeWriter.cpp
BagelOrb: I think so. Fixing..

BagelOrb: Actually, I have no idea. I can't find this variable any more.
Please comment on final PR, this comment is outdated already.

BagelOrb: Fixed all occurrences of extra_infill_shift.

src/FffGcodeWriter.cpp

src/FffGcodeWriter.cpp
```diff
@@ -1491,53 +1490,9 @@ bool FffGcodeWriter::processSkinPart(const SliceDataStorage& storage, LayerPlan&
...
-        if (pattern == EFillMethod::GRID || pattern == EFillMethod::LINES || pattern == EFillMethod::TRIANGLES || pattern == EFillMethod::CUBIC || pattern == EFillMethod::TETRAHEDRAL || pattern == EFillMethod::CUBICSUBDIV)
-        {
-            gcode_layer.addLinesByOptimizer(skin_lines, &mesh_config.skin_config, SpaceFillType::Lines, mesh.getSettingInMicrons("infill_wipe_dist"));
-        }
-        else
-        {
-            gcode_layer.addLinesByOptimizer(skin_lines, &mesh_config.skin_config, (pattern == EFillMethod::ZIG_ZAG)? SpaceFillType::PolyLines : SpaceFillType::Lines);
-        }
-    }
+    processSkinPartInfillGeneratePerimeterGaps(storage, gcode_layer, mesh, extruder_nr, mesh_config, skin_part, layer_nr, skin_overlap, skin_angle, pattern, z, generate_perimeter_gaps, concentric_perimeter_gaps, added_something);
```
BagelOrb: Good points!

I've never had a clear view on whether we should move to functions which don't depend on the actual settings base, so that they can be called with different settings in a different place or whether we should move to the least amount of parameters in order to keep the code clean.

The function is called `processSkinPartInfillGeneratePerimeterGaps` because the perimeter gaps are also processed *if they should be processed*; not because they will always be processed.

added_something is still needed because the gaps may be so small that no single line will be generated..

I did added_something as a return parameter so that we don't have to do the ugly construction any more which looks like `added_something = added_something | processX(.)`

Could we please move this feedback to CURA-3990?
Otherwise we would get all kinds of horrible merge conflicts...

src/FffGcodeWriter.cpp
BagelOrb: No. All perimeter_gaps are printed at the same time, the skin part only adds extra perimeter gaps to the perimeter gaps already computed which are lying outside of the skin part.

Added doc to explain that in e6d2df4fb805e81a05042ab79d4c5aaa93a09610

src/FffGcodeWriter.h
```diff
@@ -483,13 +483,23 @@ class FffGcodeWriter : public SettingsMessenger, NoCopy
...
+     */
+    void processSkinInsets(const SliceDataStorage& storage, LayerPlan& gcode_layer, const SliceMeshStorage& mesh, const int extruder_nr, const PathConfigStorage::MeshPathConfigs& mesh_config, const SkinPart& skin_part, unsigned int layer_nr, bool& added_something) const;
+
+    /*!
+     * Add the normal skinfill which is the area inside the innermost skin inset
      * 
-     * Note that the normal perimeter gaps are printed with the outer wall extruder,
-     * while newly generated perimeter gaps between consecutive insets of a concentric top/bottom pattern
-     * are printed with the top bottom extruder.
+     * Perimeter gaps are generated when the pattern is concentric
```
BagelOrb: No. The Infill.generate processes areas into lines/polygons and generates new areas for the gaps created by the infill pattern when it's concentric. These gaps aren't processed here. They are generated here. They are processed in `processSkinPart`.

This will make a bit more sense in CURA-3940, because then we add the topmost_skin which also generated gaps which are handled in processSkinPart.

src/skin.cpp
BagelOrb: It was added accidentally. It's going to be used in CURA-3940, though. Removing it from this PR is only going to give more merge conflicts...

Was it too messy to separate this PR from CURA-3940 ?
Maybe that wasn't such a good idea after all.

src/skin.cpp
BagelOrb: Well maybe it's just the wording.

The general working of the polygon processor is as follows:
- we get an area which needs to be filled in some way
- we split that area up into two parts which need to be filled differently from each other
- recurse until finished ^^

This is exactly what happens here: we split up the outline of the skin area and produce both the insets and the infill.

BagelOrb: Do you think I should change the wording?
I can't really come up with something better in the couple of seconds I tried... :P

src/FffGcodeWriter.cpp
```diff
@@ -1588,6 +1526,81 @@ bool FffGcodeWriter::processSkinPart(const SliceDataStorage& storage, LayerPlan&
...
+
+void FffGcodeWriter::processSkinPartInfillGeneratePerimeterGaps(const SliceDataStorage& storage, LayerPlan& gcode_layer, const SliceMeshStorage& mesh, const int extruder_nr, const PathConfigStorage::MeshPathConfigs& mesh_config, const SkinPart& skin_part, unsigned int layer_nr, int skin_overlap, int skin_angle, EFillMethod pattern, const coord_t z, const bool generate_perimeter_gaps, Polygons& concentric_perimeter_gaps, bool& added_something) const
+{
+    const coord_t skin_line_width = mesh_config.skin_config.getLineWidth();
+    const int top_bottom_extruder_nr = mesh.getSettingAsExtruderNr("top_bottom_extruder_nr");
+
+    Polygons skin_polygons;
+    Polygons skin_lines;
+
+    const bool process_skin_infill = extruder_nr == top_bottom_extruder_nr;
```
Ghostkeeper: If the "processing skin infill" and the "generating perimeter gaps" are in separate functions, this indirection won't be necessary.

It won't be necessary because in a function called "processing skin infill" it is obvious that it will be processing skin infill.

It is an indirection because down in the condition below I don't see "if this then that", but I see "if that-needs-to-happen then that". It makes it hard to find out what an if condition does if the actual condition is placed elsewhere.

src/FffGcodeWriter.cpp
Ghostkeeper: I think moving to functions which don't depend on the settings base is a good plan. What's previously stopped me from doing that is the horrible architectural design to actually put the slice data in the same instance as the settings, because a function will often require a few settings as well as the slice data, but the slice data is specific to one mesh.

Splitting the function into `processSkinPartInfill` and `generatePerimeterGaps` allows you to call the specific functions only if they should be processed. Makes testing easier and makes the functions more semantic.

The construction `added_something |= processX(...)` isn't so ugly.

src/skin.cpp
Ghostkeeper: What this function really does is:
* Generate insets for skin (lines 269 to 288). The result gets stored by reference in `skin_part`.
* Generate the skin infill area by taking the last inset in the `skin_part` and insetting it a bit more (lines 289 and 290).

There's an edge case at the top (lines 264 through 268) that could also be done in the second part, generating the skin infill areas.

BagelOrb: done

src/FffGcodeWriter.cpp
BagelOrb: I think `added_something |= processX(...)` is uglier than passing `added_something` as an output parameter.

The thing is that `Infill::generate()` already produces the interperimeter_gaps. That's logican because it depends on the infill pattern and all specificities to infill patterns are handled by that class. If you want me to split up this function there's a whole lot of other stuff which needs to be refactored.

Again: please let's handle these comments in CURA-3990, because otherwise we will be getting a lot of merge conflicts if we fix it here. It's going to be horrible.

Ghostkeeper: > I think added_something |= processX(...) is uglier than passing added_something as an output parameter.

Let's categorise that as opinion/code style then and leave it as-is.

I get your point about the infill generator also producing the perimeter gaps. It's a painful thorn leftover from legacy, so I think we must see the perimeter gaps as a side product of the infill generation, and just update the data storage as soon as possible. In the future this can be solved by letting the infill pattern generator update the perimeter gaps in storage by itself.

6 of the [7 original points](https://github.com/Ultimaker/CuraEngine/pull/548#discussion_r125499793) in this discussion still hold though.

BagelOrb: Yes. So let's postpone these until CURA-3990 - as I said.

src/FffGcodeWriter.h
```diff
@@ -498,9 +508,15 @@ class FffGcodeWriter : public SettingsMessenger, NoCopy
      * \param mesh_config the line config with which to print a print feature
      * \param skin_part The skin part for which to create gcode
      * \param layer_nr The current layer number.
+     * \param skin_overlap The distance by which the skin overlaps with the wall insets and the distance by which the perimeter gaps overlap with adjacent print features.
+     * \param skin_angle The angle in the XY plane at which the infill is generated.
+     * \param pattern The pattern with which we'd like to fill skin areas (unless we bridge, then we force the lines pattern)
+     * \param z The (approximate) height at which we are adding gcode plans
```
BagelOrb: We are adding to the layer plans. How would you call it? What are we adding to the layer plans if not 'gcode plans'?

src/FffGcodeWriter.cpp
BagelOrb: So you mean to say that I should inline this line?

I don't agree. I made this line so that you see what this check is used for.
Whether we need to process the skin-infill here (currently) only depends on the extruder_nr.

Ghostkeeper: The same could be said for any if condition, but the prevailing consensus is:

    if (extruder_nr == top_bottom_extruder_nr)
    {
        processSkinInfill();
    }

rather than

    const bool process_skin_infill = extruder_nr == top_bottom_extruder_nr;
    if (process_skin_infill)
    {
        processSkinInfill();
    }

The first is much easier to read, right? It's also much more normal.

src/FffGcodeWriter.cpp
Ghostkeeper: Well no, let's not put these ugly 14-parameter functions in our code.

src/skin.cpp
Ghostkeeper: It wouldn't give a merge conflict if it gets removed here and added to the branch where it's necessary. Unless that branch then proceeds to modify the function, in which case you'd have to do a rebase to move the commit to the start of the branch.

src/FffGcodeWriter.cpp
BagelOrb: Did point 1 - 5 in the new branch

PR #549, opened 2017-07-01
Top most skin feature (CURA-3940)

BagelOrb:


LipuFei:
@BagelOrb Sadly, you need to resolve some merge conflicts...

src/FffGcodeWriter.cpp
```diff
@@ -1261,15 +1250,15 @@ bool FffGcodeWriter::processInsets(const SliceDataStorage& storage, LayerPlan& g
                 // nothing to do
                 return false;
             }
-            const unsigned bottom_layers = mesh.getSettingAsCount("bottom_layers");
-            if (layer_nr >= bottom_layers)
+            const int bottom_layers = mesh.getSettingAsCount("bottom_layers");
```
BagelOrb: So the value cannot be negative. This change was just because it is compared with unsigned integers and it gave compiler warnings.

src/FffGcodeWriter.cpp
```diff
@@ -1383,22 +1372,33 @@ std::optional<Point> FffGcodeWriter::getSeamAvoidingLocation(const Polygons& fil
...
 }
 
-bool FffGcodeWriter::processSkinAndPerimeterGaps(const SliceDataStorage& storage, LayerPlan& gcode_layer, const SliceMeshStorage& mesh, const int extruder_nr, const PathConfigStorage::MeshPathConfigs& mesh_config, const SliceLayerPart& part, unsigned int layer_nr, int skin_overlap, int skin_angle) const
+bool FffGcodeWriter::processSkinAndPerimeterGaps(const SliceDataStorage& storage, LayerPlan& gcode_layer, const SliceMeshStorage& mesh, const int extruder_nr, const PathConfigStorage::MeshPathConfigs& mesh_config, const SliceLayerPart& part) const
 {
     int top_bottom_extruder_nr = mesh.getSettingAsExtruderNr("top_bottom_extruder_nr");
+    int topmost_skin_extruder_nr = mesh.getSettingAsExtruderNr("topmost_skin_extruder_nr");
     int wall_0_extruder_nr = mesh.getSettingAsExtruderNr("wall_0_extruder_nr");
-    if (extruder_nr != top_bottom_extruder_nr && extruder_nr != wall_0_extruder_nr)
+    if (extruder_nr != top_bottom_extruder_nr && extruder_nr != wall_0_extruder_nr && extruder_nr != topmost_skin_extruder_nr)
```

src/FffGcodeWriter.cpp
```diff
@@ -1383,22 +1372,33 @@ std::optional<Point> FffGcodeWriter::getSeamAvoidingLocation(const Polygons& fil
...
     bool added_something = false;
-    int64_t z = layer_nr * getSettingInMicrons("layer_height");
     const unsigned int perimeter_gaps_line_width = mesh_config.perimeter_gap_config.getLineWidth();
 
     bool fill_perimeter_gaps = mesh.getSettingAsFillPerimeterGapMode("fill_perimeter_gaps") != FillPerimeterGapMode::NOWHERE
                             && !getSettingBoolean("magic_spiralize")
                             && extruder_nr == wall_0_extruder_nr;
 
+    int skin_angle = 45;
+    bool skin_alternate_rotation = mesh.getSettingBoolean("skin_alternate_rotation") && (mesh.getSettingAsCount("top_layers") >= 4 || mesh.getSettingAsCount("bottom_layers") >= 4 );
```
BagelOrb: I don't know. That's only relevant to the Skin Alternate Rotation feature. This code was just copied to a different place...

src/FffGcodeWriter.cpp
```diff
@@ -1425,155 +1425,237 @@ bool FffGcodeWriter::processSkinAndPerimeterGaps(const SliceDataStorage& storage
...
-        if (skin_part.insets.size() > 0)
-        {
-            inner_skin_outline = &skin_part.insets.back();
-            offset_from_inner_skin_outline = -mesh_config.insetX_config.getLineWidth() / 2;
-        }
     }
+}
 
-    if (inner_skin_outline == nullptr)
+void FffGcodeWriter::processSkinPartTopMostInfillGeneratePerimeterGaps(const SliceDataStorage& storage, LayerPlan& gcode_layer, const SliceMeshStorage& mesh, const int extruder_nr, const PathConfigStorage::MeshPathConfigs& mesh_config, const SkinPart& skin_part, int skin_angle, const bool generate_perimeter_gaps, Polygons& concentric_perimeter_gaps, bool& added_something) const
```
BagelOrb: What you are asking for is a major refactor.
I don't think it's a good idea to make this issue bigger than it already is.

I know this architecture is rather suboptimal, but we shouldn't want to fix all the architecture touched by this single issue. The issue might then explode because each time we fix some architecture, some new piece of code will be touched of which the architecture should be changed.

src/FffGcodeWriter.cpp
```diff
@@ -1425,155 +1425,237 @@ bool FffGcodeWriter::processSkinAndPerimeterGaps(const SliceDataStorage& storage
...
+    coord_t skin_overlap = 0; // skinfill already expanded over the topmost skin; don't overlap with perimeters
+    EFillMethod pattern = mesh.getSettingAsFillMethod("topmost_skin_pattern");
+
+    const coord_t skin_line_width = mesh_config.topmost_skin_config.getLineWidth();
+    const int topmost_skin_extruder_nr = mesh.getSettingAsExtruderNr("topmost_skin_extruder_nr");
+
+    Polygons skin_polygons;
+    Polygons skin_lines;
+
+    const bool process_fill = extruder_nr == topmost_skin_extruder_nr;
```
BagelOrb: The if-statement is more complex than this simple statement, so I think it does make sense.

src/FffGcodeWriter.cpp
```diff
@@ -1425,155 +1425,237 @@ bool FffGcodeWriter::processSkinAndPerimeterGaps(const SliceDataStorage& storage
...
+
+    // generate skin_polygons and skin_lines (and concentric_perimeter_gaps if needed)
+    if (process_fill || generate_perimeter_gaps)
```
BagelOrb: If that's true the compiler will optimize it out. This code is to conceptually make clear that there are two cases in which the code below should be called.

src/FffGcodeWriter.cpp
```diff
@@ -1425,155 +1425,237 @@ bool FffGcodeWriter::processSkinAndPerimeterGaps(const SliceDataStorage& storage
...
 
-    int extra_infill_shift = 0;
-    Polygons concentric_perimeter_gaps; // the perimeter gaps of the insets of concentric skin pattern of this skin part
-    Polygons* perimeter_gaps_output = (fill_concentric_perimeter_gaps)? &concentric_perimeter_gaps : nullptr;
-    Infill infill_comp(pattern, *inner_skin_outline, offset_from_inner_skin_outline, skin_line_width, skin_line_width, skin_overlap, skin_angle, z, extra_infill_shift, perimeter_gaps_output);
-    infill_comp.generate(skin_polygons, skin_lines);
+        // calculate polygons and lines
+        constexpr int extra_infill_shift = 0;
+        constexpr coord_t offset_from_inner_skin_infill = 0;
+        Polygons* perimeter_gaps_output = (generate_perimeter_gaps)? &concentric_perimeter_gaps : nullptr;
```
BagelOrb: fixed by introducing topmost_skin_angles.

src/FffGcodeWriter.cpp
```diff
@@ -1425,155 +1425,237 @@ bool FffGcodeWriter::processSkinAndPerimeterGaps(const SliceDataStorage& storage
...
+        Infill infill_comp(pattern, skin_part.top_most_skinfill, offset_from_inner_skin_infill, skin_line_width, skin_line_width, skin_overlap, skin_angle, gcode_layer.z, extra_infill_shift, perimeter_gaps_output);
+        infill_comp.generate(skin_polygons, skin_lines);
+    }
 
     // add skin itself!
-    if ((skin_polygons.size() > 0 || skin_lines.size() > 0) && extruder_nr == top_bottom_extruder_nr)
+    if (process_fill && (skin_polygons.size() > 0 || skin_lines.size() > 0))
```
BagelOrb: also fixed.

src/FffGcodeWriter.cpp
```diff
@@ -1455,168 +1456,275 @@ bool FffGcodeWriter::processSkinAndPerimeterGaps(const SliceDataStorage& storage
...
+        skin_angle -= 45;
+    }
+
+    Polygons skin_polygons;
+    Polygons skin_lines;
+
+    const bool process_skin_infill = extruder_nr == top_bottom_extruder_nr;
+
+    // generate skin_polygons and skin_lines (and concentric_perimeter_gaps if needed)
+    if (process_skin_infill || generate_perimeter_gaps)
```
Ghostkeeper: `generate_perimeter_gaps` implies `process_skin_infill`, so this is equivalent to `if (process_skin_infill)` and consequently just `if (extruder_nr == top_bottom_extruder_nr)`.

src/FffGcodeWriter.cpp
```diff
@@ -1455,168 +1456,275 @@ bool FffGcodeWriter::processSkinAndPerimeterGaps(const SliceDataStorage& storage
...
+        constexpr int extra_infill_shift = 0;
+        constexpr coord_t offset_from_inner_skin_infill = 0;
+        const coord_t skin_overlap = mesh.getSettingInMicrons("skin_overlap_mm");
+        Polygons* perimeter_gaps_output = (generate_perimeter_gaps)? &concentric_perimeter_gaps : nullptr;
+        Infill infill_comp(pattern, skin_part.inner_infill, offset_from_inner_skin_infill, skin_line_width, skin_line_width, skin_overlap, skin_angle, gcode_layer.z, extra_infill_shift, perimeter_gaps_output);
+        infill_comp.generate(skin_polygons, skin_lines);
+    }
+
+    // add skin itself!
+    if (process_skin_infill && (skin_polygons.size() > 0 || skin_lines.size() > 0))
```
Ghostkeeper: `if` clause can be nested in previous `if` clause.

src/FffGcodeWriter.cpp
```diff
@@ -1455,168 +1456,275 @@ bool FffGcodeWriter::processSkinAndPerimeterGaps(const SliceDataStorage& storage
...
-        if (gap_lines.size() > 0)
+        else
         {
-            added_something = true;
-            setExtruder_addPrime(storage, gcode_layer, layer_nr, extruder_nr);
-            gcode_layer.setIsInside(true); // going to print stuff inside print object
-            gcode_layer.addLinesByOptimizer(gap_lines, mesh_config.perimeter_gap_config, SpaceFillType::Lines);
+            gcode_layer.addLinesByOptimizer(skin_lines, mesh_config.skin_config, (pattern == EFillMethod::ZIG_ZAG)? SpaceFillType::PolyLines : SpaceFillType::Lines);
         }
     }
```
Ghostkeeper: Lots of this code is duplicate from the `processSkinPartRoofingInfillGeneratePerimeterGaps` function. Once you've broken these functions up, please remember to re-use the parts where possible.

src/FffGcodeWriter.h
```diff
@@ -480,17 +462,67 @@ class FffGcodeWriter : public SettingsMessenger, NoCopy
...
-    bool processSkinPart(const SliceDataStorage& storage, LayerPlan& gcode_layer, const SliceMeshStorage& mesh, const int extruder_nr, const PathConfigStorage::MeshPathConfigs& mesh_config, const SkinPart& skin_part, unsigned int layer_nr, int skin_overlap, int infill_angle) const;
+    bool processSkinPart(const SliceDataStorage& storage, LayerPlan& gcode_layer, const SliceMeshStorage& mesh, const int extruder_nr, const PathConfigStorage::MeshPathConfigs& mesh_config, const SkinPart& skin_part) const;
+
+    /*!
+     * Add the extra skin walls
+     * 
+     * \param[in] storage where the slice data is stored.
+     * \param gcode_layer The initial planning of the gcode of the layer.
+     * \param mesh The mesh for which to add to the layer plan \p gcode_layer.
+     * \param extruder_nr The extruder for which to print all features of the mesh which should be printed with this extruder
```
Ghostkeeper: Isn't that just "the extruder with which the skin walls must be printed"?

src/FffGcodeWriter.h
```diff
@@ -480,17 +462,67 @@ class FffGcodeWriter : public SettingsMessenger, NoCopy
...
+     * \param mesh_config the line config with which to print a print feature
```
Ghostkeeper: Which print feature?

src/FffGcodeWriter.h
```diff
@@ -480,17 +462,67 @@ class FffGcodeWriter : public SettingsMessenger, NoCopy
...
+     * \param skin_part The skin part for which to create gcode
+     * \param generate_perimeter_gaps Whether we need to generate perimeter gaps for concentric infill of the skinfill area
+     * \param[out] concentric_perimeter_gaps The perimeter gaps output which are generated when the pattern is concentric
+     * \param[out] added_something Whether this function added anything to the layer plan
+     */
+    void processSkinPartRoofingInfillGeneratePerimeterGaps(const SliceDataStorage& storage, LayerPlan& gcode_layer, const SliceMeshStorage& mesh, const int extruder_nr, const PathConfigStorage::MeshPathConfigs& mesh_config, const SkinPart& skin_part, const bool generate_perimeter_gaps, Polygons& concentric_perimeter_gaps, bool& added_something) const;
+
+    /*!
+     * Add the normal skinfill which is the area inside the innermost skin inset which doesn't have air directly above it
```
Ghostkeeper: Not entirely accurate since the roofing can be multiple layers.

src/FffGcodeWriter.h
```diff
@@ -480,17 +462,67 @@ class FffGcodeWriter : public SettingsMessenger, NoCopy
...
+     * \param mesh The mesh for which to add to the layer plan \p gcode_layer.
+     * \param extruder_nr The extruder for which to print all features of the mesh which should be printed with this extruder
+     * \param mesh_config the line config with which to print a print feature
+     * \param skin_part The skin part for which to create gcode
+     * \param[out] added_something Whether this function added anything to the layer plan
+     */
+    void processSkinInsets(const SliceDataStorage& storage, LayerPlan& gcode_layer, const SliceMeshStorage& mesh, const int extruder_nr, const PathConfigStorage::MeshPathConfigs& mesh_config, const SkinPart& skin_part, bool& added_something) const;
+
+    /*!
+     * Add the roofing which is the area inside the innermost skin inset which has air directly above
```
Ghostkeeper: Not entirely accurate since the roofing can be multiple layers.

src/skin.cpp
```diff
@@ -52,6 +52,36 @@ Polygons SkinInfillAreaComputation::getInsidePolygons(const SliceLayerPart& part
  * This function is executed in a parallel region based on layer_nr.
  * When modifying make sure any changes does not introduce data races.
  *
+ * this function may only read/write the skin and infill from the *current* layer.
+ */
+Polygons SkinInfillAreaComputation::getWalls(const SliceLayerPart& part_here, int layer2_nr, unsigned int wall_idx)
+{
+    Polygons result;
+    if (layer2_nr < static_cast<int>(mesh.layers.size()))
```
Ghostkeeper: This is an input check. You could save an indentation by flipping this if-statement around:

    if (layer2_nr >= static_cast<int>(mesh.layers.size()))
    {
        return Polygons(); //The other layer is above the model, so it must be empty.
    }
    const SliceLayer& layer2...

(Not blocking.)

src/skin.h
```diff
@@ -113,21 +113,37 @@ class SkinInfillAreaComputation
     void generateInfill(SliceLayerPart& part, const Polygons& skin);
 
     /*!
-     * Generate the skin insets.
+     * Calculate the areas which are directly under air,
```
Ghostkeeper: Not directly, really. There may be other roofing above it.

src/sliceDataStorage.cpp
```diff
@@ -115,6 +115,10 @@ bool SliceMeshStorage::getExtruderIsUsed(int extruder_nr) const
     {
         return true;
     }
+    if ((getSettingAsCount("top_layers") > 0 || getSettingAsCount("bottom_layers") > 0) && getSettingAsCount("roofing_layer_count") > 0 && getSettingAsExtruderNr("roofing_extruder_nr") == extruder_nr)
```
Ghostkeeper: It might be an idea to nest this together with the previous `if` statement. Sort of subjective style though.

(Not blocking.)

src/FffGcodeWriter.cpp
```diff
@@ -362,6 +362,17 @@ void FffGcodeWriter::setInfillAndSkinAngles(SliceMeshStorage& mesh)
...
+    if (mesh.roofing_angles.size() == 0)
+    {
+        mesh.roofing_angles = mesh.getSettingAsIntegerList("roofing_angles");
+        if (mesh.roofing_angles.size() == 0)
+        {
+            // user has not specified any infill angles so use defaults
+            mesh.roofing_angles.push_back(45);
+            mesh.roofing_angles.push_back(135);
+        }
+    }
```
BagelOrb: I don't want to make Roofing a child setting of anything, because then we have to introduce a parent setting for each place we introduced a new Roofing setting.

BagelOrb: We could let the settings inherit only via the value function. Let's discuss this live.

src/FffGcodeWriter.cpp
BagelOrb: The frontend says "minimum_value": "0".

The frontend does all user input checking.
I don't think we should change all that for this feature.

src/FffGcodeWriter.cpp
```diff
@@ -1455,168 +1456,275 @@ bool FffGcodeWriter::processSkinAndPerimeterGaps(const SliceDataStorage& storage
...
-        bridge = bridgeAngle(skin_part.outline, &mesh.layers[layer_nr-1]);
-    }
-    if (bridge > -1)
+    const bool generate_top_bottom_perimeter_gaps = fill_top_bottom_concentric_perimeter_gaps && top_bottom_pattern == EFillMethod::CONCENTRIC; // whether we need to generate perimeter gaps for concentric infill of the skinfill area
+
+    Polygons top_bottom_concentric_perimeter_gaps; // the perimeter gaps of the insets of concentric skin pattern of this skin part
+    processSkinPartInfillGeneratePerimeterGaps(storage, gcode_layer, mesh, extruder_nr, mesh_config, skin_part, generate_top_bottom_perimeter_gaps, top_bottom_concentric_perimeter_gaps, added_something);
+
+
+    // handle perimeter_gaps of concentric skin
```

src/FffGcodeWriter.cpp
```diff
@@ -1455,168 +1456,275 @@ bool FffGcodeWriter::processSkinAndPerimeterGaps(const SliceDataStorage& storage
...
+    if (fill_perimeter_gaps || generate_top_bottom_perimeter_gaps)
```
BagelOrb: There's definitely a bug here.

I need to find out how the perimeter gaps should be printed. Either all with the outer wall extruder, or always after the perimeters have been generated between which there are gaps. You can't have both, or can you?

src/FffGcodeWriter.cpp
```diff
@@ -1455,168 +1456,275 @@ bool FffGcodeWriter::processSkinAndPerimeterGaps(const SliceDataStorage& storage
...
+        else if (extruder_nr == wall_0_extruder_nr)
+        {
+            perimeter_gaps = &skin_part.perimeter_gaps;
+        }
+        else
+        {
+            perimeter_gaps = &top_bottom_concentric_perimeter_gaps;
+        }
+        assert(mesh.roofing_angles.size() > 0);
+        int perimeter_gaps_angle = mesh.roofing_angles[gcode_layer.getLayerNr() % mesh.roofing_angles.size()]; // use roofing angles for perimeter gaps
```
BagelOrb: That's a bug. The perimeter gaps which are generated should also be printed here.

src/FffGcodeWriter.cpp
BagelOrb: No. The previous clause has different conditions, i.e. `|| generate_perimeter_gaps`

src/FffGcodeWriter.cpp
```diff
@@ -1455,168 +1456,275 @@ bool FffGcodeWriter::processSkinAndPerimeterGaps(const SliceDataStorage& storage
...
-        // determine inner_skin_outline
-        if (skin_part.insets.size() > 0)
-        {
-            inner_skin_outline = &skin_part.insets.back();
-            offset_from_inner_skin_outline = -mesh_config.insetX_config.getLineWidth() / 2;
-        }
     }
+}
+
+void FffGcodeWriter::processSkinPartRoofingInfillGeneratePerimeterGaps(const SliceDataStorage& storage, LayerPlan& gcode_layer, const SliceMeshStorage& mesh, const int extruder_nr, const PathConfigStorage::MeshPathConfigs& mesh_config, const SkinPart& skin_part, const bool generate_perimeter_gaps, Polygons& concentric_perimeter_gaps, bool& added_something) const
```
BagelOrb: renamed to `processRoofing`. Please reevaluate.

src/FffGcodeWriter.cpp
```diff
@@ -1455,168 +1456,275 @@ bool FffGcodeWriter::processSkinAndPerimeterGaps(const SliceDataStorage& storage
...
-    Polygons* perimeter_gaps_output = (fill_concentric_perimeter_gaps)? &concentric_perimeter_gaps : nullptr;
-    Infill infill_comp(pattern, *inner_skin_outline, offset_from_inner_skin_outline, skin_line_width, skin_line_width, skin_overlap, skin_angle, z, extra_infill_shift, perimeter_gaps_output);
-    infill_comp.generate(skin_polygons, skin_lines);
+    Polygons skin_polygons;
+    Polygons skin_lines;
+
+    const bool process_fill = extruder_nr == roofing_extruder_nr;
+
+    // generate skin_polygons and skin_lines (and concentric_perimeter_gaps if needed)
+    if (process_fill || generate_perimeter_gaps)
```
BagelOrb: did a big refactor

src/FffGcodeWriter.cpp
```diff
@@ -1455,168 +1456,275 @@ bool FffGcodeWriter::processSkinAndPerimeterGaps(const SliceDataStorage& storage
...
         }
     }
+}
 
-    if (fill_perimeter_gaps || fill_concentric_perimeter_gaps)
-    { // handle perimeter_gaps of concentric skin
-        const Polygons* perimeter_gaps = nullptr;
-        Polygons perimeter_gaps_recomputed;
-        if (wall_0_extruder_nr == top_bottom_extruder_nr) // then also both are equalt to extruder_nr
+void FffGcodeWriter::processSkinPartInfillGeneratePerimeterGaps(const SliceDataStorage& storage, LayerPlan& gcode_layer, const SliceMeshStorage& mesh, const int extruder_nr, const PathConfigStorage::MeshPathConfigs& mesh_config, const SkinPart& skin_part, const bool generate_perimeter_gaps, Polygons& concentric_perimeter_gaps, bool& added_something) const
```
BagelOrb: renamed to `processTopBottom`

src/FffGcodeWriter.cpp
BagelOrb: refactored away

src/FffGcodeWriter.cpp
```diff
@@ -1455,168 +1456,275 @@ bool FffGcodeWriter::processSkinAndPerimeterGaps(const SliceDataStorage& storage
...
+    Polygons skin_lines;
+
+    const bool process_skin_infill = extruder_nr == top_bottom_extruder_nr;
+
+    // generate skin_polygons and skin_lines (and concentric_perimeter_gaps if needed)
+    if (process_skin_infill || generate_perimeter_gaps)
+    {
+        // only generate infill if we're going to print it with this extruder
+        // or when we need to compute the perimeter gaps
+        // TODO: only compute this once when the infill is concentric and the perimeter gpas are printed with a different extruder
```
BagelOrb: refactored away. The TODO doesn't make sense any more.
Perimeter gaps of a skin part are now always printed with the extruder with which the skin part feature was pritned. Not always with the outer wall extruder any more.

src/FffGcodeWriter.cpp
BagelOrb: refactored away

src/FffGcodeWriter.cpp

src/FffGcodeWriter.h
BagelOrb: No definitely not!
This is the extruder with which we are currently checking if we should print something.
It is not really the current extruder, because we might not yet have encountered any feature which caused us to actually switch to this extruder.

For each extruder we process each print feature and in each `processPrintFeatureX` function we check whether we should actually print something with the extruder.

src/FffGcodeWriter.h
BagelOrb: any. The mesh config storage contains the configs for all features.

src/FffGcodeWriter.h
BagelOrb: fixed

src/FffGcodeWriter.h
BagelOrb: This has been the name of this parameter everywhere where it is used in  functions in this class.

src/FffGcodeWriter.h
BagelOrb: fixed

src/skin.cpp
BagelOrb: done

src/skin.h
BagelOrb: fixed, but kind of a lame fix: `'directly' under`.
Any better wording?

src/sliceDataStorage.cpp
BagelOrb: Meh

src/FffGcodeWriter.cpp
BagelOrb: Chose the latter

Ghostkeeper: I indeed don't think you can have both, with the other constraints of which features will be printed first (e.g. inner before outer wall) and with the hard requirement that there may be at most `extruder_count - 1` extruder switches per layer.

src/FffGcodeWriter.h
```diff
@@ -490,7 +490,7 @@ class FffGcodeWriter : public SettingsMessenger, NoCopy
     void processSkinInsets(const SliceDataStorage& storage, LayerPlan& gcode_layer, const SliceMeshStorage& mesh, const int extruder_nr, const PathConfigStorage::MeshPathConfigs& mesh_config, const SkinPart& skin_part, bool& added_something) const;
 
     /*!
-     * Add the roofing which is the area inside the innermost skin inset which has air directly above
+     * Add the roofing which is the area inside the innermost skin inset which has air 'directly' above
```
Ghostkeeper: Hmm, yeah, well I can't think of anything better either...

LipuFei:
@BagelOrb conflicts again

smartavionics:
OK, just getting to grips with this feature. I understand the rational behind it. By default, it enables 1 top layer which caught me out on this print. Because the next layer looks like this:

screenshot_2017-07-27_15-07-16

What I got was this:

screenshot_2017-07-27_15-06-58

But what I wanted is this:

screenshot_2017-07-27_15-15-55

i.e. that layer is printed as  a single (basically triangular) region rather than as two regions (one triangular and another circular).
So I think that people are maybe going to get a bit confused with this change?
Also, in a similar way to how the skins can be grown into the infill, is there an argument here to shrink the non-top layer region so that the boundary moves inside the walls a bit more? Otherwise, I think you will get some artifacts that you would not otherwise see if the skin was printed as one region.

smartavionics:
Also! If the top-layer has the same settings as the non-top layer (speed, pattern, etc.), surely it is not actually needed and can be ignored. i.e. the triangular area and the circular area in my example above are printed with the same settings so the top layer region is really a no-op and so the whole region should be printed as one.

BagelOrb:
The overlap is the same as the skin overlap. It is also applied to make the top-layer skin overlap with the rest of the skin.

Yes in the case of your settings it doesn't make sense to print a top layer, or in other words: it doesn't make sense to print it with the same settings. You shouldn't ask for a top layer if you're going to print it with the same settings. The top layer is off by default, so if someone changes the number of top layers, he should also think about changing other settings. There's no way in Cura to automatically change that setting back to zero if the user changes it to one but doesn't change any of the other top skin settings.

BagelOrb:
New feature might be needed, such as expanding the top skin into the non-top skin or making non-top skin areas which are too small a part of the top skin.

smartavionics:
For me, top surface infill layers defaults to 1, not 0.

BagelOrb:
Whoops!
Fixed that!

Ghostkeeper:
> Also! If the top-layer has the same settings as the non-top layer (speed, pattern, etc.), surely it is not actually needed and can be ignored.

Sounds like an enhancement request right there! I agree, since the defaults are currently equal to the top/bottom settings. I guess that enhancement request won't have high priority though...

Issue #550, opened 2017-07-03
Question: why not use m201 instead of m204 to set acceleration?


smartavionics:
If you compare these two mcodes:

http://reprap.org/wiki/G-code#M201:_Set_max_printing_acceleration

http://reprap.org/wiki/G-code#M204:_Set_default_acceleration

They are both supported by most firmwares. Cura currently uses m204 to alter acceleration for non-repetier flavor gcode but the parameter (S) it uses is not supported by reprap firmware. I am suggesting that we change to using m201 with X, Y, Z parameters which is supported by both marlin and reprap.

The above referenced wiki says under m204: _Use M201 to set per-axis accelerations and extruder accelerations._

We have the choice of changing to m201 for the reprap (and marlin) gcode flavor and keeping m204 for the other flavors or we can change all of the non-repetier flavors to use m201. The code for repetier flavor would not need to change.

Thoughts, anyone?

smartavionics:
Maybe it would be easier to just fiddle with the M204 params so that it is acceptable to both reprap and marlin. I will try that.

PR #551, opened 2017-07-04
Use M201 rather than M204 to set accelerations for reprap (reprap/marlin/sprinter) flavor gcode.

smartavionics:
This does what is suggested in #550.


smartavionics:
Hmm, had a better idea...!

smartavionics:
OK, i've reopened this pull request. It's been tested on a Duet running reprap 1.18 and works OK, it would be good if it could be tested on Ultimaker marlin to see that it works OK.

smartavionics:
Nope, this is going again because I am going to integrate it into a new PR that introduces a new gcode flavor that really supports reprap firmware.

PR #552, opened 2017-07-04
Add P and T values to M204 when setting the acceleration.

smartavionics:
The M204 output to set the acceleration had just the S parameter which is understood by legacy Marlin to
set both the print and travel accelerations. It was not supported by RepRap.

I have added the P and T values (same actual acceleration values as for the S value) . P and T are understood by RepRap (1.18 onwards) to set the print and travel accelerations, respectively. Recent Marlin also understands the P and T values.

### DIFF
--- a/src/gcodeExport.cpp
+++ b/src/gcodeExport.cpp
@@ -1012,7 +1012,9 @@ void GCodeExport::writeAcceleration(double acceleration, bool for_travel_moves)
     {
         if (current_acceleration != acceleration)
         {
-            *output_stream << "M204 S" << PrecisionedDouble{0, acceleration} << new_line; // Print and Travel acceleration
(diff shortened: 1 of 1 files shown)


BagelOrb:
M204 Tx is interpreted by legacy Marlin to set the acceleration of filament-only moves.

I don't think this PR can be merged. :/

smartavionics:
Yes, but it gets the S value set the same, surely that together, that covers all cases?

smartavionics:
BTW, what version of Marlin (e.g. date, version number) are you using? I would like to check the source.

smartavionics:
Ahh, OK, I found your marlin source and understand now that T sets the extruder acceleration. Well, yes, I can see that this PR is not going to work so I shall close this and bring my previous PR back to life that changes the M204 into M201. Fuck, legacy marlin is such a pain!!!

PR #553, opened 2017-07-05
Add support for real reprap flavor gcode

smartavionics:
Rather than trying to code around the support for legacy Marlin, I think it's better to have a new gcode flavor that is compatible with the RepRap firmware. This PR adds the REPRAP_REPRAP gcode flavor and adds the ability to output acceleration, jerk and max z feedrate in gcode suitable for that firmware.

### DIFF
--- a/src/gcodeExport.cpp
+++ b/src/gcodeExport.cpp
@@ -1035,7 +1035,7 @@ void GCodeExport::writeJerk(double jerk)
         {
             *output_stream << "M207 X";
         }
-        else if(getFlavor() == EGCodeFlavor::REPRAP)
+        else if (getFlavor() == EGCodeFlavor::REPRAP)
         {
             *output_stream << "M566 X" << PrecisionedDouble{2, jerk * 60} << " Y" << PrecisionedDouble{2, jerk * 60} << new_line;
         }
--- a/src/gcodeExport.cpp
+++ b/src/gcodeExport.cpp
@@ -36,7 +36,7 @@ GCodeExport::GCodeExport()
     current_max_z_feedrate = -1;
 
     isZHopped = 0;
-    setFlavor(EGCodeFlavor::REPRAP);
+    setFlavor(EGCodeFlavor::MARLIN);
     initial_bed_temp = 0;
 
     extruder_count = 0;
@@ -254,7 +254,7 @@ void GCodeExport::setFlavor(EGCodeFlavor flavor)
     else
         for(int n=0; n<MAX_EXTRUDERS; n++)
             extruder_attr[n].extruderCharacter = 'E';
-    if (flavor == EGCodeFlavor::ULTIGCODE || flavor == EGCodeFlavor::REPRAP_VOLUMATRIC)
+    if (flavor == EGCodeFlavor::ULTIGCODE || flavor == EGCodeFlavor::MARLIN_VOLUMATRIC)
     {
         is_volumatric = true;
     }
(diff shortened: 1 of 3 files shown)


BagelOrb:
Nice! This was long overdue, but there wasn't really any gain for Ultimaker..

Is there also an accompanying frontend PR?

smartavionics:
Yes, it's https://github.com/Ultimaker/Cura/pull/2073

BagelOrb:
See accompanying frontend PR: https://github.com/Ultimaker/Cura/pull/2073

src/gcodeExport.cpp
```diff
@@ -1012,7 +1012,15 @@ void GCodeExport::writeAcceleration(double acceleration, bool for_travel_moves)
     {
         if (current_acceleration != acceleration)
         {
-            *output_stream << "M204 S" << PrecisionedDouble{0, acceleration} << new_line; // Print and Travel acceleration
+            // Print and Travel acceleration
+            if (getFlavor() == EGCodeFlavor::REPRAP)
+            {
+                *output_stream << "M201" << " X" << PrecisionedDouble{0, acceleration} << " Y" << PrecisionedDouble{0, acceleration} << new_line;
```
BagelOrb: This is actually a different firmware setting. You are now controlling the maximum acceleration per axis, while the actual moves are limited by acceleration settings along both axis as well as the acceleration along all axes at the same time.

You should change this to `M204 Px Tx`

src/gcodeExport.cpp
```diff
@@ -1027,6 +1035,10 @@ void GCodeExport::writeJerk(double jerk)
         {
             *output_stream << "M207 X";
         }
+        else if (getFlavor() == EGCodeFlavor::REPRAP)
```
BagelOrb: Maybe a `swtich(getFlavor())` is in place here.
[not blocking]

src/gcodeExport.cpp
```diff
@@ -1027,6 +1035,10 @@ void GCodeExport::writeJerk(double jerk)
...
+        {
+            *output_stream << "M566 X" << PrecisionedDouble{2, jerk * 60} << " Y" << PrecisionedDouble{2, jerk * 60} << new_line;
```
BagelOrb: You should refactor the lines below.
The lines 1046-1048 should be copied into the Repetier case and the else case, but not into the reprap case.

BagelOrb:
Looking good.
Testing after lunch.

BagelOrb:
`GCodeExport::getFileHeader(.)` had a check for `(flavor == EGCodeFlavor::REPRAP)` which isn't converted into `(flavor == EGCodeFlavor::REPRAP || flavor == Marlin)`

BagelOrb:
I'll add that one..

smartavionics:
Hold on, if you referring to where I think you are, the following lines are only used by the RepRap firmware? It's the comments for filament used and layer height?

smartavionics:
Also, I was going to differentiate between travel and print acceleration for reprap.

BagelOrb:
Whoops
too late!

smartavionics:
No matter, leave it as it is, there's no harm in having those comments in the marlin gcode

BagelOrb:
Leaving out those comments for the UMO isn't worth it. Some people may like those comments for the UMO.

BagelOrb:
Yeah

BagelOrb:
You've already had a PR for differentiating between travel and print acceleration some time ago
Maybe revisit that PR?

BagelOrb:
Anyway I don't think it's weird to make a new PR for that. You can view that as a separate issue any way.

smartavionics:
Yes, i'll do a new PR some time for that.

Ghostkeeper:
Devs, see CURA-4014.

PR #555, opened 2017-07-07
Mb travel accel mods

smartavionics:
This little PR fixes a small bug when generating repetier flavor gcode (the time estimate calculator was not getting updated as it should) and adds support for separate print and travel accelerations in reprap gcode.

### DIFF
--- a/src/gcodeExport.h
+++ b/src/gcodeExport.h
@@ -100,8 +100,8 @@ private:
     double current_e_value; //!< The last E value written to gcode (in mm or mm^3)
     Point3 currentPosition; //!< The last build plate coordinates written to gcode (which might be different from actually written gcode coordinates when the extruder offset is encoded in the gcode)
     double currentSpeed; //!< The current speed (F values / 60) in mm/s
-    double current_acceleration; //!< The current acceleration in the XY direction (in mm/s^2)
-    double current_travel_acceleration; //!< The current acceleration in the XY direction used for travel moves if different from current_acceleration (in mm/s^2) (Only used for Repetier flavor)
+    double current_print_acceleration; //!< The current acceleration (in mm/s^2) used for print moves (and also for travel moves if the gcode flavor doesn't have separate travel acceleration)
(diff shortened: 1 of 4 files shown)


BagelOrb:
If `writeAcceleration` doesn't output any gcode, then why would we cahnge the acceleration of the time estimator?

smartavionics:
Because if the estimator was last estimating a print move, and what comes next is a travel move, (or vice-versa), the estimator will need to be updated irrespective of whether an M code was output or not because the printer will use the appropriate acceleration (P or T) for the move.

BagelOrb:
Ah ok.

smartavionics:
But updating the estimator acceleration is very cheap anyway, it doesn't trigger any calculation, just sets some state.

BagelOrb:
Perhaps the code would look a lot more clean if we separate this function into two: `writeTravelAcceleration` and `writePrintAcceleration`.
For Marlin `writeTravelAcceleration` could then simply call `writePrintAcceleration`.

That way we can have more `switch` statements whih makes this easier to read.

[not blocking]

smartavionics:
OK, i'll do that now...

Ghostkeeper:
Devs, see CURA-4014.

src/gcodeExport.h
```diff
@@ -100,8 +100,8 @@ class GCodeExport : public NoCopy
     double current_e_value; //!< The last E value written to gcode (in mm or mm^3)
     Point3 currentPosition; //!< The last build plate coordinates written to gcode (which might be different from actually written gcode coordinates when the extruder offset is encoded in the gcode)
     double currentSpeed; //!< The current speed (F values / 60) in mm/s
-    double current_acceleration; //!< The current acceleration in the XY direction (in mm/s^2)
-    double current_travel_acceleration; //!< The current acceleration in the XY direction used for travel moves if different from current_acceleration (in mm/s^2) (Only used for Repetier flavor)
+    double current_print_acceleration; //!< The current acceleration (in mm/s^2) used for print moves (and also for travel moves if the gcode flavor doesn't have separate travel acceleration)
```
BagelOrb: 👍

BagelOrb:
Thanks!
Looks a lot better now! :)

PR #557, opened 2017-07-09
First layer line width Multiplier

14bitVoid:
My attempt at making the line width different on the first layer. This creates a setting for a multiplier to the line width only on the first layer of the print. You get less lines but thicker ones, this improves adhesion to the buildplate.
Fixes: https://github.com/Ultimaker/Cura/issues/1165

This was a bit more hard than I thought! So please check!

Some caveats:
It doesn't multiply brim, skirt and raft because those have their own line widths for the first layer.
It doesn't multiply spaghetti infill because that does some weird shit with the line width basically and I don't dare to change that. Please check!
It doesn't multiply ironing line width because that would be useless.
It doesn't multiply wire printing. I never tried actually that but I don't think it would make it better.
When support or infill touches the buildplate the infill density is wrong. This is the lesser of two evils, because changing the line spacing to correct the density will make lines misalign with higher layers.

Setting the multiplier to 120% makes my ABS stick a bit better. Maybe you want me to put that in the ABS profiles?


whosawhatsis:
Thank you, now I'll have to see if I can figure out how to build this for Mac...

Ghostkeeper:
Devs, see issue CURA-4027.

Ghostkeeper:
* The intent of spaghetti infill is to lay the filament loosely in your model, so I don't think bed adhesion is the issue there. This feature won't need to work on that.
* I agree that ironing is not necessary.
* I agree that applying this to wire printing would probably make it worse.
* It's good that you thought of the (support) infill density and making sure that the lines align.

I'm not sure about the brim and raft line widths, because those are actually intended for bed adhesion. But indeed they already have their own settings (Brim Line Width and Raft Base Line Width). It might be useful to make those settings scale automatically using inheritance formulas, or else just apply it in the engine as well to prevent the user having to do that manually.

src/settings/PathConfigStorage.cpp
```diff
@@ -163,6 +222,65 @@ PathConfigStorage::PathConfigStorage(const SliceDataStorage& storage, int layer_
...
+{
+    return (layer_nr == 0)? &inset0_config_layer0 : &inset0_config;
+}
+
+const GCodePathConfig *PathConfigStorage::MeshPathConfigs::getInsetXConfig(int layer_nr) const
+{
+    return (layer_nr == 0)? &insetX_config_layer0 : &insetX_config;
+}
+
+const GCodePathConfig *PathConfigStorage::MeshPathConfigs::getSkinConfig(int layer_nr) const
```
Ghostkeeper: These three have a `layer_nr` parameter that's not marked as `const` while it is marked as `const` in the header file.

src/settings/PathConfigStorage.cpp
```diff
@@ -163,6 +222,65 @@ PathConfigStorage::PathConfigStorage(const SliceDataStorage& storage, int layer_
...
+{
+    return (layer_nr == 0)? &support_infill_config_layer0 : &support_infill_config;
+}
+
+const GCodePathConfig *PathConfigStorage::getSupportRoofConfig(const int layer_nr) const
+{
+    return (layer_nr == 0)? &support_roof_config_layer0 : &support_roof_config;
+}
+
+void PathConfigStorage::MeshPathConfigs::smoothAllSpeeds(GCodePathConfig::SpeedDerivatives first_layer_config, int layer_nr, int max_speed_layer)
```
Ghostkeeper: Why did you put these in a separate function? It seemed logical to have them in `handleInitialLayerSpeedup`.

(Not blocking.)

src/settings/PathConfigStorage.h
```diff
@@ -24,30 +24,48 @@ class PathConfigStorage
...
         GCodePathConfig skin_config;
-        GCodePathConfig ironing_config;
+        GCodePathConfig skin_config_layer0;
         GCodePathConfig perimeter_gap_config;
+        GCodePathConfig perimeter_gap_config_layer0;
+        GCodePathConfig infill_config_layer0; // combined infill layers can't be printed on the first layer
         std::vector<GCodePathConfig> infill_config;
 
+    public:
+        GCodePathConfig ironing_config;
```
Ghostkeeper: I'd like to keep these configs consistent. Right now half of them are private with their own `get***Config` functions (the ones that are different for the first layer), and the other half are public without such a function. This makes it inconsistent since the programmer has to remember which configs are public and which aren't.

So please also make the rest of them private and provide a `get***Config` function. Perhaps you can just give it no parameters since it doesn't vary by layer number.

src/settings/PathConfigStorage.cpp
14bitVoid: I made the actual configs private to prevent them from being used accidentally and help me find all uses. PathConfigStorage needs to smooth the speeds, so smoothAllSpeeds will edit the private variables but you can call it from PathConfigStorage since it is public.

14bitVoid:
Done, thanks for the suggestions

Ghostkeeper:
Seems to work nicely. Thanks!

BagelOrb:
Hi Niklas,

We're having some problems with this PR after it was merged.
We were considering reverting the whole merge, but we thought it would be better to ask you to fix the problems instead.

These are the issues I encountered:
- You are introducing new configs for the initial layer while you can just change the config itself. The configs are already stored per layer now, so just changing the config would already do it. This would make the code a lot easier to read. Implementation hint: I suggest you introduce a private class member `std::vector<float> line_width_factor_per_extruder` to `PathConfigStorage` which is either 1 or the setting value based on the `layer_nr` with which the constructor of that class is called.
- This also means that the getXxConfig functions can be removed and the configs can be made public again. That would also make it a lot easier to merge some of the other branches we still have open which add extra configs or edit them.
- The setting is retrieved from a mesh level as well as the extruder level. You should retrieve the setting from the extruder level everywhere. For example, the outer wall line width should be retrieved from the outer wall extruder. That way the initial layer line width depends on the material with which you are printing even if you are printing one model with different extruders per print feature.

Using the regex `line_width[^ ]*"` I've also found the following places where you forgot to account for the initial line width multiplier:
- SkinInfillAreaComputation::generateInfill line 296. You forgot to assign the `line_width_factor`
- support: you forgot to apply the line width factor to support. For example: it's needed in `AreaSupport::splitGlobalSupportAreasIntoSupportInfillParts`. I think it's a good idea to take something like `min(line_width_layer_0, support_line_distance)`.

While you're editing code anyway, it might be nice to capture the `if (layer_nr == 0)` inside a separate function somewhere. I'm not sure where, though. Maybe a function `getSettingAsLineWidth(std::string key, int layer_nr)` in `settings.h/cpp`?


I was hoping to get this fixed yesterday myself, but I didn't have enough time for it after all... :(
I hope you can find the time to fix these things!

14bitVoid:
Ok sure, I'll try. I'm going to make a new PR for that.

Makes sense to do this per extruder then. I haven't really experimented a lot with setting the wall/infill extruders in master yet I use 2.6.2 by default (until I started on this PR).

The option you propose would still make merge conflicts. If we make line_width_factor_per_extruder private then something in PathConfigStorage has to access and use it. The getXxConfig getters could do that now, but then I can't remove those getters and you'll have merge conflicts. You can't just get the xx_config public members because you have to multiply it with the factor of the correct extruder. Alternatively if we make line_width_factor_per_extruder public and then remove getXxConfig getters, every place that uses any config has to manually multiply it with the correct extruder, so you'll still get merge conflicts.

I'm going to try passing line_width_factor_per_extruder as constructor parameter and multiplying the line widths in the constructor so we don't even need to save it in a member.

Sorry about the forgotten places. The line width factor was applied to support but that place where the inset for support with 1/2 line width was made have I forgotten.

I will try at layer == 0 too!

BagelOrb:
> The option you propose would still make merge conflicts. If we make line_width_factor_per_extruder private then something in PathConfigStorage has to access and use it. 

What I meant was that PathConfigStorage then uses that private member when initializing all the GcodePathConfig members of it's own class. It's private because it's only used during initialization of the whole object.
You can then just get `configs.support_config` for example, because the factor is already taken along when constructing `support_config`.

BagelOrb:
So in the constructor `PathConfigStorage::PathConfigStorage(SliceDataStorage storage, int layer_nr, int layer_thickness)` you construct for example the `support_roof_config` something like so:
```
support_roof_config(
    PrintFeatureType::Support
    , support_infill_train->getSettingInMicron("support_line_width") * line_width_factor
   , etc.
```
where the line width factor is initialized in PathConfigStorage either to 1.0 or to the value of the setting.

BagelOrb:
Nver mind. I'm fixing it myself.

14bitVoid:
Sorry I was on vacation and flying back. 2 days Flight from Brisbane to Frankfurt stopping at Bangkok! I made this on the airplane! But now you made it too!

You could've said you were going to fix it yourself....

BagelOrb:
Oh wow, so you spend the night in Bangkok...

I wasn't planning on fixing it, but then I had some free time, so i did it
after all ;)


Op 19 jul. 2017 16:39 schreef "Niklas" <notifications@github.com>:

Sorry I was on vacation and flying back. 2 days Flight from Brisbane to
Frankfurt stopping at Bangkok! I made this on the airplane! But now you
made it too!

You could've said you were going to fix it yourself....

—
You are receiving this because you commented.

<https://github.com/Ultimaker/CuraEngine/pull/557#issuecomment-316408708>,
or mute the thread

Ghostkeeper:
It's in the beta for 2.7, and will be in the stable release for 2.7 as well.

PR #561, opened 2017-07-17
Correct the infill mesh areas added to the comb boundary when combing mode is NO_SKIN.

### LINKED ISSUE #560: Bug - No-skin combing always retracts within infill mesh
I have a print which uses infill meshes to beef up certain areas and I've noticed that when combing is set to no-skin, the travels within the infill areas (that have skin and wall thicknesses set to 0) are all retracts. See this pic:

screenshot_2017-07-16_14-31-43

I've spent a while looking at the code but can't figure out what's wrong. As an experiment, I  stopped it adding the infill mesh's infill_area to the combing boundary in LayerPlan::computeCombBoundaryInside() and then the travels don't use retract but that's not a good solution because if the infill mesh has any skin (why would anyone give an infill mesh skin/walls anyway?) the travels also don't retract when printing the skin parts which is wrong when combing mode is no-skin. Perhaps someone who is familiar with this code and take a look as I'm wasting too much time on this!

smartavionics:
It was adding the area of the infill mesh's infill but this just makes a hole in the
comb boundary the shape of the infill mesh. Now, it adds the outlines of the skin areas
in the infill mesh (rather than the infill area).

Fixes #560.

### DIFF
--- a/src/LayerPlan.cpp
+++ b/src/LayerPlan.cpp
@@ -156,7 +156,7 @@ Polygons LayerPlan::computeCombBoundaryInside(CombingMode combing_mode)
                         // for infill mesh parts, add the outlines of its skin parts to the comb boundary
                         for (const SkinPart& skin_part : part.skin_parts)
                         {
-                            comb_boundary.add(skin_part.outline[0]);
+                            comb_boundary.add(skin_part.outline.outerPolygon());
                         }
                     }
                     else
--- a/src/LayerPlan.cpp
+++ b/src/LayerPlan.cpp
@@ -146,27 +146,17 @@ Polygons LayerPlan::computeCombBoundaryInside(CombingMode combing_mode)
         for (const SliceMeshStorage& mesh : storage.meshes)
         {
             const SliceLayer& layer = mesh.layers[layer_nr];
-            const bool is_infill_mesh = mesh.getSettingBoolean("infill_mesh");
+            if (mesh.getSettingBoolean("infill_mesh")) {
+                continue;
+            }
             if (mesh.getSettingAsCombingMode("retraction_combing") == CombingMode::NO_SKIN)
             {
                 for (const SliceLayerPart& part : layer.parts)
                 {
-                    if (is_infill_mesh)
-                    {
-                        // for infill mesh parts, add the outlines of its skin parts to the comb boundary
-                        for (const SkinPart& skin_part : part.skin_parts)
-                        {
-                            comb_boundary.add(skin_part.outline.outerPolygon());
-                        }
-                    }
-                    else
-                    {
-                        // for non-infill mesh parts, add the outline of its infill to the comb boundary
-                        comb_boundary.add(part.infill_area);
-                    }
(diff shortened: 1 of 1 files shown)


src/LayerPlan.cpp
```diff
@@ -146,19 +146,28 @@ Polygons LayerPlan::computeCombBoundaryInside(CombingMode combing_mode)
...
             {
                 for (const SliceLayerPart& part : layer.parts)
                 {
-                    comb_boundary.add(part.infill_area);
+                    if (is_infill_mesh)
+                    {
+                        // for infill mesh parts, add the outlines of its skin parts to the comb boundary
+                        for (const SkinPart& skin_part : part.skin_parts)
+                        {
+                            comb_boundary.add(skin_part.outline[0]);
```
BagelOrb: You should never do a `[0]` on polygons. The first polygon has no special meaning in general. Instead you should use `PolygonsPart::getOuterPolygon()`

BagelOrb: This forces you to be 100% sure you are working with a PoylgonsPart rather than just a Polygons.

src/LayerPlan.cpp
```diff
@@ -146,19 +146,28 @@ Polygons LayerPlan::computeCombBoundaryInside(CombingMode combing_mode)
         for (const SliceMeshStorage& mesh : storage.meshes)
         {
             const SliceLayer& layer = mesh.layers[layer_nr];
+            const bool is_infill_mesh = mesh.getSettingBoolean("infill_mesh");
```
BagelOrb: I think your solution isn't the right one.
I think the check which *was* on line 158 should be performed here.
So right below this line it should say:
```
if (is_infill_mesh)
{
    continue;
}
```

smartavionics: So you're saying ignore the NO_SKIN setting for skins of infill meshes?

BagelOrb:
Please change the PR to the simple check I proposed - if that indeed solves the problem.

Please understand the difference between the `infill_area` and the `infill_area_own`. The latter is the infill area without the areas which belong to infill areas, while the first is the area inside of the mesh, including the areass which belong to infill meshes.

src/LayerPlan.cpp
BagelOrb: Yes, that's the easiest.

You could also do something like:
```
if (NO_SKIN)
{
if (infill_mesh)
{
   comb_boundary.add(part.skin_outline); // make hole polygons in the surrounding infill_area of the mesh of which this is an infill mesh
}
else
{
    comb_boundary.add(part.infill_area);
}
}
```

src/LayerPlan.cpp
BagelOrb: This line doesn't make a difference between the skin and the infill. The outlines of the infill mesh encompass both the infill and the skin.
You wanted that combing should be performed on the infill. but not on the skin of infill meshes, but you aren't looking at either skin or infill areas of infill meshes.

BagelOrb:
Did you see my earlier comments?

PR #564, opened 2017-07-20
feat: quarter cubic infill

BagelOrb:
similar to tetrahedral (octet) infill, but doesn't have double lines crossing in two directions.

Also it probably has totally different structural properties...

See https://en.wikipedia.org/wiki/Quarter_cubic_honeycomb
The pattern is tilted on its side.
And compare with octet: https://en.wikipedia.org/wiki/Tetrahedral-octahedral_honeycomb

And compare with all lovely possibilities: https://en.wikipedia.org/wiki/Convex_uniform_honeycomb


BagelOrb:
Note: I was browsing for cool regular tilings of 3D space and while playing with the Quarter Cubic pattern I noticed it was extremely easy to implement.

BagelOrb:
image

[Quarter cubic honeycomb_angled.skp.txt](https://github.com/Ultimaker/CuraEngine/files/1163810/Quarter.cubic.honeycomb_angled.skp.txt)

BagelOrb:
Devs, see CURA-4071

PR #565, opened 2017-07-21
feat: force layer to start with given model

BagelOrb:
force a specific model to be the first to be printed with the current extruder on that layer


BagelOrb:
See also https://github.com/Ultimaker/Cura/pull/2142

BagelOrb:
@Devs see CURA-4073

rburema:
@BagelOrb Going through stale branches right now: can this one be deleted?

BagelOrb:
It's an unmerged feature and I think we decided back then we didn't want the feature after all.

Some people like the feature, though.

See the linked frontend PR.

rburema:
I'm only deleting things if there isn't any work in there that could be used, so I'll leave this for now.

PR #567, opened 2017-07-21
fix_first_layer_support_grid

BagelOrb:


BagelOrb:
According to materials the initial layer grid pattern helps a lot to make support stick to the build plate better.

PR #568, opened 2017-07-26
Remove some zags for zig zag infill

LipuFei:
CURA-3957


src/infill/ZigzagConnectorProcessorConnectedEndPieces.cpp
```diff
@@ -20,50 +20,54 @@ void ZigzagConnectorProcessorConnectedEndPieces::registerScanlineSegmentIntersec
...
 
 void ZigzagConnectorProcessorConnectedEndPieces::registerPolyFinished()
 {
-    // write end segment if needed (first half of start/end-crossing segment)
-    if (!last_scanline_is_even && !first_zigzag_connector_ends_in_even_scanline)
+    const bool is_last_piece_end_piece = last_scanline_is_even == first_zigzag_connector_ends_in_even_scanline;
+    const bool need_to_skip_this_piece = skip_some_zags && ++current_zag_count >= zag_skip_count;
+    const bool add_last_piece = is_last_piece_end_piece || (last_scanline_is_even && !need_to_skip_this_piece);
+
+    if (add_last_piece)
```
Ghostkeeper: I always find it easier to read when it's like this:

    if (<condition>)
    {
        <do stuff>
    }

Rather than:

    const bool should_do_stuff = <condition>;
    if (should_do_stuff)
    {
        <do stuff>
    }

It's sort of what if-clauses are made for.

src/infill/ZigzagConnectorProcessorDisconnectedEndPieces.cpp
```diff
@@ -19,54 +19,64 @@ void ZigzagConnectorProcessorDisconnectedEndPieces::registerScanlineSegmentInter
...
-        { // if we left from an even scanline, but not if this is the line segment connecting that zigzag_connector to an even scanline
-            addLine(last_connector_point, intersection);
-        }
-        else if (!previous_scanline_is_even && !this_scanline_is_even) // if we end an odd boundary in an odd segment
-        { // add whole oddBoundarySegment (including the just obtained point)
-            for (unsigned int point_idx = 1; point_idx < zigzag_connector.size(); point_idx++)
+        const bool is_end_piece = previous_scanline_is_even == this_scanline_is_even;
+        const bool add_connection = previous_scanline_is_even || is_end_piece;  // add connections on even segments or it is an endpiece
+
+        if (add_connection)
```
Ghostkeeper: I always find it easier to read when it's like this:

    if (<condition>)
    {
        <do stuff>
    }

Rather than:

    const bool should_do_stuff = <condition>;
    if (should_do_stuff)
    {
        <do stuff>
    }

It's sort of what if-clauses are made for.

src/infill/ZigzagConnectorProcessorNoEndPieces.cpp
```diff
@@ -33,13 +33,20 @@ void ZigzagConnectorProcessorNoEndPieces::registerScanlineSegmentIntersection(co
...
+            else
+            {   // add whole zigzag_connector (including the just obtained point)
+                for (unsigned int point_idx = 1; point_idx < zigzag_connector.size(); point_idx++)
+                {
+                    addLine(zigzag_connector[point_idx - 1], zigzag_connector[point_idx]);
+                }
+                addLine(zigzag_connector.back(), intersection);
             }
-            addLine(zigzag_connector.back(), intersection);
-            zigzag_connector.clear();
```
Ghostkeeper: Why was this `clear()` call removed?

src/infill/ZigzagConnectorProcessorEndPieces.cpp
```diff
@@ -11,11 +11,6 @@ void ZigzagConnectorProcessorEndPieces::registerVertex(const Point& vertex)
     {
         first_zigzag_connector.push_back(vertex);
     }
-    else if (last_scanline_is_even)
-    { // when a boundary segments starts in an even scanline it's either a normal zigzag connector or an endpiece to be included
-        // note that for ZigzagConnectorProcessorDisconnectedEndPieces only the last line segment from a boundary vertex to a scanline-boundary intersection is omitted
-        addLine(last_connector_point, vertex);
```
Ghostkeeper: The `last_connector_point` field is now unused.

src/infill/ZigzagConnectorProcessorNoEndPieces.cpp
LipuFei: Outside this big if block there is a `zigzag_connector.clear();` which will always be called, so this is kinda redundant.

src/infill/ZigzagConnectorProcessorDisconnectedEndPieces.cpp
```diff
@@ -19,54 +19,64 @@ void ZigzagConnectorProcessorDisconnectedEndPieces::registerScanlineSegmentInter
...
     {
-        if (previous_scanline_is_even && !this_scanline_is_even)
-        { // if we left from an even scanline, but not if this is the line segment connecting that zigzag_connector to an even scanline
-            addLine(last_connector_point, intersection);
-        }
-        else if (!previous_scanline_is_even && !this_scanline_is_even) // if we end an odd boundary in an odd segment
-        { // add whole oddBoundarySegment (including the just obtained point)
-            for (unsigned int point_idx = 1; point_idx < zigzag_connector.size(); point_idx++)
+        const bool is_end_piece = previous_scanline_is_even == this_scanline_is_even;
+        const bool add_connection = previous_scanline_is_even || is_end_piece;  // add connections on even segments or it is an endpiece
```
LipuFei: Fixed in 2e0ffbd

src/infill/ZigzagConnectorProcessorDisconnectedEndPieces.cpp
LipuFei: Fixed in 2e0ffbd

src/infill/ZigzagConnectorProcessorConnectedEndPieces.cpp
```diff
@@ -20,50 +20,54 @@ void ZigzagConnectorProcessorConnectedEndPieces::registerScanlineSegmentIntersec
...
-        { // when a boundary segment starts in an even scanline it is either a normal zigzag connector or an endpiece, so it should be included anyway
-            addLine(last_connector_point, intersection);
-        }
-        else if (!previous_scanline_is_even && !this_scanline_is_even) // if we end an odd boundary in an odd segment
-        { // add whole zigzag_connector (including the just obtained point)
-            for (unsigned int point_idx = 1; point_idx < zigzag_connector.size(); point_idx++)
+        bool add_connection = previous_scanline_is_even;  // add connections on even segments
+        add_connection |= previous_scanline_is_even == this_scanline_is_even;  // end piece
+
+        if (add_connection)
```
LipuFei: Fixed in 2e0ffbd

src/infill/ZigzagConnectorProcessorConnectedEndPieces.cpp
LipuFei: Fixed in 2e0ffbd

src/infill/ZigzagConnectorProcessorEndPieces.cpp
LipuFei: Fixed in 34be16e32c495e7e276bf17c3636f32419a42836 and 976dd9a1ceb69c3f00cc23437d2fa343e4277cab

src/infill/ZigzagConnectorProcessorConnectedEndPieces.cpp
```diff
@@ -20,50 +20,54 @@ void ZigzagConnectorProcessorConnectedEndPieces::registerScanlineSegmentIntersec
...
     last_connector_point = intersection;
 }
 
 
 void ZigzagConnectorProcessorConnectedEndPieces::registerPolyFinished()
 {
-    // write end segment if needed (first half of start/end-crossing segment)
-    if (!last_scanline_is_even && !first_zigzag_connector_ends_in_even_scanline)
+    const bool is_last_piece_end_piece = last_scanline_is_even == first_zigzag_connector_ends_in_even_scanline;
+    const bool need_to_skip_this_piece = skip_some_zags && ++current_zag_count >= zag_skip_count;
```
LipuFei: Fixed in d7bf8cce69f35f0eeb19bb3bec95e65d3853ad8a

LipuFei:
796fec9 is the new commit, some earlier commits are merged from master.

src/infill/ZigzagConnectorProcessorConnectedEndPieces.cpp
BagelOrb: In that case it would be better if there's a comment saying `should_do_stuff` because the variable name makes explicit  what the combination of conditions amount to.

src/infill/ZigzagConnectorProcessorDisconnectedEndPieces.cpp
BagelOrb: I'd then also have a comment like
```
if (previous_scanline_is_even || is_end_piece)
{ // add connection
  do_stuff
}
```

src/infill/ZigzagConnectorProcessor.cpp
```diff
@@ -0,0 +1,145 @@
...
+    {
+        //
+        // Skipping some zags is done in the following way:
+        //  > direction: negative number means from <right> -> <left>
+        //               positive number means from <left> -> <right>
+        //  > for a connection:
+        //    - if it comes from <left> -> <right> (direction >= 0), we skip connections that
+        //      lay in segments which mean the condition: "index mod skip_count = 0"
+        //    - for connections that come from the other direction, we skip on
+        //      "(index - 1) mod skip_count = 0"
```
BagelOrb: This comment should explain why these conditions are as such.
You should say that zag connections in even scan segments go from lower scanline index to higher, while connections in odd scansegments go from higher scanline index to lower. (Or is it the other way around???)

src/infill/ZigzagConnectorProcessor.h
```diff
@@ -94,59 +80,130 @@ namespace cura
...
+    const PointMatrix& rotation_matrix; //!< The rotation matrix used to enforce the infill angle
+    Polygons& result; //!< The result of the computation
+
+    bool use_endpieces; //!< Whether to include end pieces or not
+    bool connected_endpieces; //!< Whether the end pieces should be connected with the rest part of the infill
+    int skip_some_zags; //!< Whether to skip some zags
+    int zag_skip_count; //!< Skip 1 zag in every N zags
+
+    bool is_first_connector; //!< indicating whether we are still looking for the first connector or not
+    int first_connector_index; //!< scanline segment index of the first connector
```
BagelOrb: rename to `first_connector_end_scanline_idx`

src/infill/ZigzagConnectorProcessor.h
```diff
@@ -94,59 +80,130 @@ namespace cura
...
+    int first_connector_direction; //!< direction of the first connector
```
BagelOrb: This variable doesn't need to be saved. See comment further on about computing the direction from the start adn end scnaline index

src/infill/ZigzagConnectorProcessor.cpp
```diff
@@ -0,0 +1,145 @@
...
+        this->first_connector_index = scanline_index;
+        this->first_connector_direction = direction;
+        this->is_first_connector = false;
+    }
+    else
+    {
+        // add this connector if needed
+        if (this->shouldAddThisConnector(this->last_connector_index, scanline_index, direction))
+        {
+            for (unsigned int point_idx = 1; point_idx < this->current_connector.size(); ++point_idx)
```
BagelOrb: I think it's weird that this for loop starts from 1 and then in the loop we subtract 1 immediately, but this was in my original code, so I fixed this myself.

src/infill/ZigzagConnectorProcessor.cpp
```diff
@@ -0,0 +1,145 @@
...
+        || (!is_endpiece && this->shouldAddThisConnector(scanline_start_index, scanline_end_index, direction)))
+    {
+        // for convenience, put every point in one vector
+        for (const Point& point : this->first_connector)
+        {
+            this->current_connector.push_back(point);
+        }
+        this->first_connector.clear();
+
+        for (unsigned int point_idx = 1; point_idx < this->current_connector.size() - 1; ++point_idx)
```
BagelOrb: Please move the below loop and if-statement into a new function and remove the code duplication with the same code in `registerScanlineSegmentIntersection`.

Make a new function something like `void addCurrentZagConnector()`

src/infill/ZigzagConnectorProcessor.cpp
```diff
@@ -85,9 +85,9 @@ void ZigzagConnectorProcessor::registerScanlineSegmentIntersection(const Point&
         // add this connector if needed
         if (this->shouldAddThisConnector(this->last_connector_index, scanline_index, direction))
         {
-            for (unsigned int point_idx = 1; point_idx < this->current_connector.size(); ++point_idx)
+            for (unsigned int point_idx = 0; point_idx < this->current_connector.size() - 1; ++point_idx)
             {
-                addLine(this->current_connector[point_idx - 1], this->current_connector[point_idx]);
+                addLine(this->current_connector[point_idx], this->current_connector[point_idx]);
```
LipuFei: @BagelOrb This is adding the same points...

src/infill.cpp
```diff
@@ -4,9 +4,20 @@
...
+ * Function which returns the scanline_idx for a given x coordinate
+ *
+ * For negative \p x this is different from simple division.
+ *
+ * \warning \p line_width is assumed to be positive
+ *
+ * \param x the point to get the scansegment index for
+ * \param line_width the width of the scan segments
+ */
+static inline int computeScanSegmentIdx(int x, int line_width);
```
LipuFei: This function is only used in this C++ file, so the scope of this function is here only. Putting it in the header function will make other modules who import this header file aware of the function, which is not necessary. This is why I moved it into this file.

src/infill/ZigzagConnectorProcessor.h
```diff
@@ -94,59 +80,130 @@ namespace cura
...
+    this->first_connector_direction = 0;
+    this->last_connector_index = 0;
+    this->last_connector_direction = 0;
+    this->first_connector.clear();
+    this->current_connector.clear();
+}
+
+inline void ZigzagConnectorProcessor::addLine(Point from, Point to)
+{
+    assert(from != to);
```
LipuFei: I think it is caused by a small mistake... 4e04807818939470e0bb0974fc7ea79d3502d056

src/infill/ZigzagConnectorProcessor.cpp
```diff
@@ -0,0 +1,145 @@
...
+    else
+    { // it's yet unclear whether the line segment should be included, so we store it until we know
+        this->current_connector.push_back(vertex);
+    }
+}
+
+
+bool ZigzagConnectorProcessor::shouldAddThisConnector(int start_scanline_idx, int end_scanline_idx, int direction) const
+{
+    assert((direction != 1 || direction != -1) && "direction should be either +1 or -1");
```
LipuFei: Fixed in 17176b75

src/infill/ZigzagConnectorProcessor.h
LipuFei: Fixed in 136518ba09df1f9d0d09dc76f9e9c4436079f93a

src/infill/ZigzagConnectorProcessor.h
LipuFei: Fixed in f9db1ef85f3a446965de0a2fc5243dbecff73eeb

src/infill/ZigzagConnectorProcessor.h
LipuFei: Fixed in 5634a63b14e8c6d0251cac92f70e1e0c0cd03610

src/infill/ZigzagConnectorProcessor.h
```diff
@@ -94,59 +80,130 @@ namespace cura
...
+    void addLine(Point from, Point to);
+
+    /*!
+     * Checks whether this connector should be added or not.
+     *
+     * \param start_scanline_idx the start scanline index of this scanline segment
+     * \param end_scanline_idx The the end scanline index of this scanline segment
+     * \param direction The direction of this scanline segment
+     */
+    bool shouldAddThisConnector(int start_scanline_idx, int end_scanline_idx, int direction) const;
```
LipuFei: Fixed in 491cd8c779c5b768414bd67633baa04c61eebd95

src/infill/ZigzagConnectorProcessor.cpp
LipuFei: I fixed this in 136518b

src/infill/ZigzagConnectorProcessor.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+
+void ZigzagConnectorProcessor::registerVertex(const Point& vertex)
+{
+    if (this->is_first_connector)
+    {
+        this->first_connector.push_back(vertex);
+    }
+    else
+    { // it's yet unclear whether the line segment should be included, so we store it until we know
```
LipuFei: Fixed in 678aec1a

src/infill/NoZigZagConnectorProcessor.h
```diff
@@ -8,16 +8,19 @@
...
 
+/*!
+ * This processor adds no connection. This is for line infill pattern.
+ */
 class NoZigZagConnectorProcessor : public ZigzagConnectorProcessor
 {
 public:
     NoZigZagConnectorProcessor(const PointMatrix& rotation_matrix, Polygons& result)
-    : ZigzagConnectorProcessor(rotation_matrix, result)
+    : ZigzagConnectorProcessor(rotation_matrix, result, false, false)
```
LipuFei: Fixed in 678aec1a

src/infill/ZigzagConnectorProcessor.h
LipuFei: Also in a6703e4343c422a43213e06b53f082f251724e07

src/infill/ZigzagConnectorProcessor.cpp
```diff
@@ -0,0 +1,145 @@
+#include <cassert>
+#include "ZigZagConnectorProcessor.h"
```
LipuFei: Fixed by @BagelOrb  :)

src/infill/ZigzagConnectorProcessor.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+
+bool ZigzagConnectorProcessor::shouldAddThisConnector(int start_scanline_idx, int end_scanline_idx, int direction) const
+{
+    assert((direction != 1 || direction != -1) && "direction should be either +1 or -1");
+    //
+    // Decide whether we should add this connection or not.
+    // Add this zag connection is the following cases:
+    //  - if this zag lays in an even-numbered scanline segment
+    //  - if this zag is an end piece (check if the previous and the current scanlines are the same)
```
LipuFei: Fixed in 3b9b9358

src/infill/ZigzagConnectorProcessor.cpp
LipuFei: Fixed in 3b9b9358

src/infill/ZigzagConnectorProcessor.cpp
```diff
@@ -0,0 +1,145 @@
...
+}
+
+
+bool ZigzagConnectorProcessor::shouldAddThisConnector(int start_scanline_idx, int end_scanline_idx, int direction) const
+{
+    assert((direction != 1 || direction != -1) && "direction should be either +1 or -1");
+    //
+    // Decide whether we should add this connection or not.
+    // Add this zag connection is the following cases:
+    //  - if this zag lays in an even-numbered scanline segment
```
LipuFei: Fixed in 3b9b9358

src/infill/ZigzagConnectorProcessor.h
```diff
@@ -94,59 +80,130 @@ namespace cura
...
      * 
      * \param rotation_matrix The rotation matrix used to enforce the infill angle
      * \param result The resulting line segments (Each line segment is a Polygon with 2 points)
+     * \param use_endpieces Whether to include end pieces or not
+     * \param connected_endpieces Whether the end pieces should be connected with the rest part of the infill
+     * \param skip_some_zags; //!< Whether to skip some zags
+     * \param zag_skip_count; //!< Skip 1 zag in every N zags
      */
-    ZigzagConnectorProcessor(const PointMatrix& rotation_matrix, Polygons& result)
+    ZigzagConnectorProcessor(const PointMatrix& rotation_matrix, Polygons& result,
```
LipuFei: Fixed in 7f8f293c

src/infill/ZigzagConnectorProcessor.cpp
LipuFei: Fixed in 1934a7cc

src/infill/ZigzagConnectorProcessor.cpp
LipuFei: Fixed by @BagelOrb

src/infill.cpp
BagelOrb: I guess this is just a matter of taste...

If you put it in the header file with private scope then other modules won't be able to acces the function anyway.

src/infill/ZigzagConnectorProcessorConnectedEndPieces.cpp
```diff
@@ -22,7 +22,11 @@ void ZigzagConnectorProcessorConnectedEndPieces::registerScanlineSegmentIntersec
     {
         if (previous_scanline_is_even)
         { // when a boundary segment starts in an even scanline it is either a normal zigzag connector or an endpiece, so it should be included anyway
-            addLine(last_connector_point, intersection);
+            for (unsigned int point_idx = 1; point_idx < zigzag_connector.size(); point_idx++)
```
BagelOrb: (outdated)

src/infill/ZigzagConnectorProcessor.cpp
BagelOrb: You forgot the second half of this comment.

src/infill/ZigzagConnectorProcessor.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+    const bool should_add =
+        (is_this_connection_even && !is_this_endpiece && !should_skip_this_connection) // normal connections that should be added
+        || (this->use_endpieces && is_this_endpiece);  // end piece if it is enabled;
+
+    return should_add;
+}
+
+
+void ZigzagConnectorProcessor::registerScanlineSegmentIntersection(const Point& intersection, int scanline_index, int direction)
```
BagelOrb: Instead of passing `direction` all the way from infil.cpp you can simply do
```
int direction = end_scanline_idx - start_scanline_idx;
assert((direction == 1 || direction == -1 || direction == 0) && "if the current connection is not an endpiece it should have a direction");
```

src/infill/ZigzagConnectorProcessor.cpp
```diff
@@ -0,0 +1,156 @@
...
+        this->current_connector.push_back(vertex);
+    }
+}
+
+
+bool ZigzagConnectorProcessor::shouldAddCurrentConnector(int start_scanline_idx, int end_scanline_idx, int direction) const
+{
+    assert((direction == 1 || direction == -1) && "direction should be either +1 or -1");
+    //
+    // Decide whether we should add this connection or not.
```
BagelOrb: this -> the current

src/infill/ZigzagConnectorProcessor.cpp
```diff
@@ -0,0 +1,156 @@
...
+    // Add the current zag connection in the following cases:
+    //  - if the current zag starts on an even scanline and ends on an odd scanline
+    //  - if the current zag is an end piece (check if the previous and the current scanlines are the same)
+    //    and "use end piece" is enabled
+    // Don't add a zag if:
+    //  - the current zag is NOT an end piece, "skip some zags" is enabled, and the current zag lays in a
+    //    segment which needs to be skipped.
+    // Moreover:
+    //  - if a "connected end pieces" is not enabled and this is an end piece, the last line
```
BagelOrb: this -> the current

src/infill/ZigzagConnectorProcessor.cpp
```diff
@@ -0,0 +1,156 @@
...
+    {
+        // process as the first connector if we haven't found one yet
+        // this will be processed with the last remaining piece at the end (when the polygon finishes)
+        this->first_connector.push_back(intersection);
+        this->first_connector_end_scanline_index = scanline_index;
+        this->is_first_connector = false;
+    }
+    else
+    {
+        // add this connector if needed
```
BagelOrb: this -> the current

src/infill/ZigzagConnectorProcessor.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+
+void ZigzagConnectorProcessor::registerPolyFinished()
+{
+    int scanline_start_index = this->last_connector_index;
+    int scanline_end_index = this->first_connector_end_scanline_index;
+    int direction = scanline_end_index > scanline_start_index ? 1 : -1;
+    const bool is_endpiece = this->is_first_connector || (!this->is_first_connector && scanline_start_index == scanline_end_index);
+
+    // decides whether to add this zag according to the following rules
```
BagelOrb: this -> the current

src/infill/ZigzagConnectorProcessor.h
```diff
@@ -94,59 +80,136 @@ namespace cura
...
-    
+    void registerVertex(const Point& vertex);
+
     /*!
      * Handle the next intersection between a scanline and the outer boundary.
      * 
      * \param intersection The intersection
-     * \param scanline_is_even Whether the scanline was even
+     * \param scanline_index Index of the current scanline
+     * \param direction Direction of this line segment
```
BagelOrb: this -> the current

src/infill/ZigzagConnectorProcessor.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+        this->addZagConnector(this->current_connector, is_endpiece);
+    }
+
+    // reset member variables
+    this->reset();
+}
+
+
+void ZigzagConnectorProcessor::addZagConnector(const std::vector<Point>& points, bool is_endpiece)
```
BagelOrb: This function is always called with the same first argument, namely `this->current_connector`, so you could leave out the argument from the function signature.
If you do so you could rename the function to `addCurrentZagConnector`

src/infill/ZigzagConnectorProcessor.h
LipuFei: Fixed in b55e4bf8590b368bb18a3d7338a873616f711029

src/infill/ZigzagConnectorProcessor.cpp
LipuFei: Fixed in b55e4bf8590b368bb18a3d7338a873616f711029

src/infill/ZigzagConnectorProcessor.cpp
LipuFei: Fixed in b55e4bf8590b368bb18a3d7338a873616f711029

src/infill/ZigzagConnectorProcessor.cpp
LipuFei: Fixed in b55e4bf8590b368bb18a3d7338a873616f711029

src/infill/ZigzagConnectorProcessor.cpp
LipuFei: Fixed in b55e4bf8590b368bb18a3d7338a873616f711029

src/infill/ZigzagConnectorProcessor.cpp
LipuFei: Fixed in b55e4bf8590b368bb18a3d7338a873616f711029

src/infill/ZigzagConnectorProcessor.cpp
LipuFei: Fixed in b55e4bf8590b368bb18a3d7338a873616f711029

src/infill/ZigzagConnectorProcessor.cpp
LipuFei: I made it this way because it is more flexible. If there is another list of points need to be added, it can use this function directly.

src/infill.cpp
alekseisasin: In my opinion if the function is used only in this class then is should be private and not visible from outside. It makes code more understandable, because if after a few months some body will review/use this code he will may assume that this function is also used somewhere else.

src/infill/ZigzagConnectorProcessor.h
```diff
@@ -43,20 +43,6 @@ namespace cura
...
- *                                                                          /               \                                   .
- *                                                  ActualZigzagConnectorProcessor      NoZigZagConnectorProcessor
- *                                                         /                \             for lines infill                                      .
- *                                                        /                  \                                                  .
- *                          ZigzagConnectorProcessorEndPieces   ZigzagConnectorProcessorNoEndPieces
- *                                     /            \                 for zigzag infill (without end pieces)                                                          .
- *                                    /              \                                                                          .
- * ZigzagConnectorProcessorConnectedEndPieces     ZigzagConnectorProcessorDisconnectedEndPieces
- * for zigzag support with normal endpieces          for zigzag support with disconnected endpieces for more easy removability
- *
```
alekseisasin: You can leave some part of sketch. We still have ZigzagConnectorProcessor as parent class and NoZigZagConnectorProcessor as child class.

src/infill/ZigzagConnectorProcessor.h
```diff
@@ -94,59 +80,134 @@ namespace cura
...
      * Handle the end of a polygon and prepare for the next.
      * This function should reset all member variables.
      */
-    virtual void registerPolyFinished() = 0;
+    void registerPolyFinished();
+
+    /*!
+     * Reset the state so it can be used for processing another polygon.
+     */
+    void reset();
```
alekseisasin: This method also can be moved to the protected

src/infill/ZigzagConnectorProcessor.h
LipuFei: This sketch is to complain the complex relationships of those classes. Since now we have cleaned it up, I added a small comment on the NoZigzagConnector (in 81e93783)

src/infill/ZigzagConnectorProcessor.h
LipuFei: Fixed in 94dee5e576b96cfd5217897032292d87f07cda00

src/infill.cpp
BagelOrb: Not changed, but that's ok - as discussed.

src/infill/ZigzagConnectorProcessor.cpp
BagelOrb: I think you are optimizing code for some future use while there is not really a reason for thinking we would need this function to be used as such.
For now we don't need the extra flexibility and I don't think we will ever need it.

I don't consider this a blocking comment, though. I would also merge this without this being fixed.

smartavionics:
See #584

src/infill.cpp
Ghostkeeper: I think putting all function declarations in the header provides an overview of each function with its documentation, which is also valuable.

It may be a matter of taste, but in the interest of code consistency we've specified in our code conventions that function declarations and their documentation go in the header file: https://github.com/Ultimaker/Meta/blob/master/code_conventions.md#documentation

PR #572, opened 2017-07-29
Added Z offset feature - can now apply Z offset to layer 0 and taper offset over following layers.

smartavionics:
Engine component for https://github.com/Ultimaker/Cura/pull/2176


Ghostkeeper:
Devs, see issue CURA-4171.

alekseisasin:
I tested your solution and have next Z values:
Case #1: Settings: no
0 layer = z0,3
1 layer = z0,4
2 layer = z0,5
3 layer = z0,6
Case #2: Settings: (z_offset_taper_layers=3 -s z_offset_layer_0=0.1) 
0 layer = z0,4
1 layer = z0,467
2 layer = z0,534
3 layer = z0,6
Case #3: Settings: (z_offset_taper_layers=0 -s z_offset_layer_0=0.1) OR (z_offset_taper_layers=1 -s z_offset_layer_0=0.1)
0 layer = z0,4
1 layer = z0,5
2 layer = z0,6
3 layer = z0,7

From this test I see that case#3 with two different settings produces same result, in case#2 the distance between 0 and 3 layer will be small and material may stick to the nozzle, what is not good.

smartavionics:
Hi, thanks for your feedback. Obviously, we can do some checks to ensure that the gap between the layers is not too small. What is the smallest gap between layers that would be OK as a fraction of the layer height? So, for example, if it was decided that the gap between two layers cannot be less than 80% of the normal layer height, we would have to have a minimum of 5 for the taper layers. If we require the gap to be not less than 90% of the normal layer height, taper layers would have to be at least 10. For larger layer heights, perhaps it is not so critical.
Can you provide some numbers for this, please?

src/FffGcodeWriter.cpp
```diff
@@ -654,6 +654,19 @@ LayerPlan& FffGcodeWriter::processLayer(const SliceDataStorage& storage, int lay
             }
             layer_thickness = getSettingInMicrons("layer_height_0");
         }
+        int64_t z_offset = getSettingInMicrons("z_offset_layer_0");
```
Ghostkeeper: You should use `coord_t` here instead of `int64_t`. In practice it's the same, but the length of the coordinates could be changed later.

smartavionics: Yeah, well when you change z to be a coord_t, I will change z_offset!

Ghostkeeper: Yeah, I see that the original `z` variable is also `int64_t`, so that's a bug in the original code then so nevermind that :)

alekseisasin:
smartavionics , Adding a hard coded parameter(gap between the layers) will solve it maybe for now, but make it very dynamic will require more effort (maybe will require some extra settings or etc.). I think lets keep the current code.

From my side the code works and lets keep it simple. I will explain this solution to my colleagues. If everything will be fine and I commit it with Cura pull request.

smartavionics:
Hi, did you see that I had added some checks in the UI? @Ghostkeeper queried that, can you please discuss with him whether you want to keep those value limits and advise me if you want any changes.

alekseisasin:
smartavionics, Yes, I spoke with @Ghostkeeper about min values. We will do some additional testing. If everything is fine then we will merge.

LipuFei:
@smartavionics I have tested the branches. There is a problem that the engine still does z offset when raft is enabled. From the settings you have made, I suppose that those settings are not meant for raft. This produces an interesting result:
image

**How to reproduce:**
- enable brim
- set initial layer z offset and taper layer to some number
- enable raft
- see the screenshot, the engine still uses the z offset

smartavionics:
Good catch, fixed!

smartavionics:
Closing as it will be implemented differently.

PR #574, opened 2017-07-30
Added cool_fan_min_layer setting

smartavionics:
The fans are off until this layer is reached. Then the normal ramp up to max speed occurs.

When set to the default value of 0, the original behaviour is preserved.


Ghostkeeper:
Devs, see issue CURA-4172.

smartavionics:
This PR has been withdrawn.

PR #575, opened 2017-08-02
Rename GCodePath::getLineWidth() to GCodePath::getLineWidthForLayerView().

smartavionics:
This avoids confusion as there's another getLineWidth() in use.


Ghostkeeper:
Devs, see CURA-4173.

PR #576, opened 2017-08-02
Carry out overlap compensation on all level-0 or level-X walls together.

smartavionics:
The old un-optimised wall printing code would overlap compensate all walls that have the
same level, i.e. all level-0 walls would be compensated, all the level-1 walls and so on.

The optimised wall printing code would overlap compensate walls that had been grouped
together but as it was now grouping walls by hole rather than by inset level that meant
that some overlaps were being missed.

This new code should catch all overlaps between all inner walls and all overlaps between
all outer walls but not overlaps between inner and outer walls (they should not
occur anyway?) It's part of the wall order optimizer so only applicable when optimization is enabled.

Appears to work OK but it's still WIP rather than a merge candidate at this time.


smartavionics:
This has now received quite a lot of testing and doesn't appear to have any problems. A beneficial side effect of this PR is that it simplifies the wall order optimising code as the repetitive checks for whether overlap compensation is in use or not are now all removed.

Ghostkeeper:
The outer wall can overlap with the inner walls due to the Outer Wall Inset setting, but you wouldn't want to compensate lines for that then.

Ghostkeeper:
Looks useful.

Devs, see CURA-4183.

smartavionics:
> The outer wall can overlap with the inner walls due to the Outer Wall Inset setting, but you wouldn't want to compensate lines for that then.
Good point, I hadn't considered that at all.

smartavionics:
This is incorporated into #592, closing.

PR #580, opened 2017-08-07
For RepRap flavour gcode, output an initial T0 before any temperature setting occurs.

smartavionics:
Another request from the RepRap folks.

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -399,6 +399,11 @@ void FffGcodeWriter::processStartingCode(const SliceDataStorage& storage, const 
 
     gcode.writeComment("Generated with Cura_SteamEngine " VERSION);
 
+    if (gcode.getFlavor() == EGCodeFlavor::REPRAP)
+    {
+        gcode.writeCode("T0"); // required before any temperature setting commands
+    }
+
     if (gcode.getFlavor() != EGCodeFlavor::ULTIGCODE && gcode.getFlavor() != EGCodeFlavor::GRIFFIN)
     {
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Devs, see issue CURA-4220.

src/FffGcodeWriter.cpp
```diff
@@ -399,6 +399,11 @@ void FffGcodeWriter::processStartingCode(const SliceDataStorage& storage, const
 
     gcode.writeComment("Generated with Cura_SteamEngine " VERSION);
 
+    if (gcode.getFlavor() == EGCodeFlavor::REPRAP)
```
BagelOrb: Please also move the `T0` which is already being outputted for `GRIFFIN` here.

Line 444: `gcode.writeCode("T0"); // Toolhead already assumed to be at T0, but writing it just to be safe...`
should be moved here as well, by extending the if condition above.

BagelOrb:
Looks good to me. Are you still down with LGPL?

smartavionics:
I have signed the chit, LGPL is cool.

BagelOrb:
Oh wait I forgot. We split the Marling and RepRap gcode flavors, so this is not working for Marlin firmware.

I'm not sure why we don't just always put a T0 in there....

stuartpb:
Well, https://github.com/Ultimaker/Cura/issues/1819#issuecomment-301695644 suggested that some printers might have a problem with a superfluous T0 command, but it looks like #640 has all the appropriate carveouts - Cura 3.2 should be putting the appropriate T command at the top of all G-code for all the printers that can handle it.

PR #583, opened 2017-08-13
Added support for speed_print_layer_0_only_for_walls setting.

smartavionics:
If that setting is true, only the wall speeds are slowed down to speed_print_layer_0.

Engine component that matches https://github.com/Ultimaker/Cura/pull/2272.

### DIFF
--- a/src/settings/PathConfigStorage.cpp
+++ b/src/settings/PathConfigStorage.cpp
@@ -203,17 +203,20 @@ PathConfigStorage::PathConfigStorage(const SliceDataStorage& storage, int layer_
     }
 }
 
-void PathConfigStorage::MeshPathConfigs::smoothAllSpeeds(GCodePathConfig::SpeedDerivatives first_layer_config, int layer_nr, int max_speed_layer)
+void PathConfigStorage::MeshPathConfigs::smoothAllSpeeds(GCodePathConfig::SpeedDerivatives first_layer_config, int layer_nr, int max_speed_layer, bool only_smooth_wall_speeds)
 {
     inset0_config.smoothSpeed(              first_layer_config, layer_nr, max_speed_layer);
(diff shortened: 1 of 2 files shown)


BagelOrb:
Why would you want this?
Wouldn't this decrease adhesion?

smartavionics:
Hi Tim. What I have found is that the first layer walls sometimes need to be printed really slow otherwise they don't adhere well and the leading edge can curl up. Having printed the walls at, say, 20mm/S, I then found that I could print the skin at 50mm/S (the normal top/bottom speed for that profile) and the skin adheres perfectly well and the resulting print looks lovely.
I think that maybe the skin can be printed faster than the walls because (a) the skin lines start touching a wall and that "anchors" the end of the skin line so the extruder can move away quicker than if the line was only touching the build plate. (b) the skin lines (when using Lines or zig-zag styles) are straight, whereas walls lines tend to go round corners and that can cause the filament to become detached from the build plate if the speed is too high.
So, if there's a lot of skin, you can waste quite a lot of time when printing it at 20mm/S when you could be printing it at 50mm/S. That's why I would like to do this.

Ghostkeeper:
Decrease adhesion maybe but gain speed. If this decreases the adhesion only a bit but saves a lot of time...

Ghostkeeper:
Devs, see CURA-4247.

alekseisasin:
The reason of this solution to achieve higher printing speed by adjusting wall settings. If you will use a stick glue then it also can improve wall adhesion to the build plate.

BagelOrb:
@alekseisasin A glue stick *might* make the adhesion better, but that will make the adhesion better for all print features. The wall then still adheres worse to the build plate compared to the skin, so the original reasoning for wanting a different speed for the outer wall on the initial layer still stands.

smartavionics:
Closing, will replace with another PR.

Issue #577, opened 2017-08-02
Too much overlap compensation?


smartavionics:
Take a look at this example, the round feature consists of two outside walls. The outermost one was printed first and the innermost printed later and it has been overlap compensated down to almost nothing where it touches the outermost wall. In fact, it is drawn with such a thin line that you can hardly see it in this image.

![screenshot_2017-08-02_10-35-58](https://user-images.githubusercontent.com/585618/28867665-b6371bdc-776e-11e7-9cbb-00d1ec2006ce.png)

But the (inside) outline that it should have is actually the same as the outline on the layer below which is quite visible. I realise that the centre line of the very thin wall may not be lined up with the centre line of the corresponding wall below so it's hard to tell exactly but it looks to me that the thin wall is much too thin!

I don't think it's just a glitch in the layer view because this is what S3D shows the gcode to be:

![screenshot_2017-08-02_10-49-24](https://user-images.githubusercontent.com/585618/28868094-4aedaaec-7770-11e7-87c8-b61fafc4d22c.png)

This is just one example, I can see many other places where it looks like the overlap compensation is too aggressive.

smartavionics:
Well, of course, I couldn't just leave this problem unexplored so I have wasted a few hours trying to understand the marvels of the overlap compensation code. After making some changes in WallOverlapComputation::getApproxOverlapArea() it's now producing more sensible results:

![screenshot_2017-08-02_16-00-27](https://user-images.githubusercontent.com/585618/28880136-fff8b6f4-779b-11e7-9843-b40f3fe33c7f.png)

It's hard to see in the image but the wall in the centre is actually wider than it was before. I also have other examples which have improved.

Old code:
![screenshot_2017-08-02_16-07-40](https://user-images.githubusercontent.com/585618/28880437-f1f82b42-779c-11e7-8e37-5a55cbc80ae1.png)

New code:

![screenshot_2017-08-02_16-06-08](https://user-images.githubusercontent.com/585618/28880380-c8caa86c-779c-11e7-964f-50ab616b2634.png)

smartavionics:
I'm continuing to work on this but don't really understand how a lot of the code actually works.
Currently, I am getting best results if I disable the processing of zero-length segments (aka points) in handlePotentialOverlap(). Could someone who is familiar with this code please explain what the zero length segments represent and what effect they think those zero length segments should have in the overlap area calculation.

Ghostkeeper:
[...]

PR #586, opened 2017-08-17
[For discussion] Overlap area computation changes

smartavionics:
These are the changes I refer to in #577. I am currently using this code and it gives better results (when viewed in the layer view) than the original code. The original code is too aggressive when you have a cylinder with walls whose thickness is such that overlap compensation occurs. In that situation, one of the walls reduces to almost nothing when it shouldn't.
As I've mentioned before, I don't really understand the original code and the comments are not sufficient to really get a grip on what it is doing so it would be good if the original author(s) could chime in here.

### DIFF
--- a/src/wallOverlap.cpp
+++ b/src/wallOverlap.cpp
@@ -150,7 +150,54 @@ int64_t WallOverlapComputation::getApproxOverlapArea(const Point from, const Poi
         // O<------O   .            O------>O
         //         :   :                     \_
         //         '   O------->O             O------>O
-        return 0;
+
+        // It used to return 0 here but that doesn't work very well when you have a curved wall that overlaps
+        // with another wall that has a smaller radius (i.e. when you have concentric circular walls one inside the other.
+
+        // It was reducing the width of the segments between the "corners" but at each corner there was a spike because
+        // the width was not being reduced at all, so it looked like this: ____/\____/\____ (this shows the ___ segments in
+        // a straight line but, of course, in reality their orientation would be changing as they progress around the curved wall).
+
+        // So I have this "theory" that at each of the corners, there's a little diamond shaped area that wasn't getting flow reduced
+        // when 0 was being returned and, therefore, the code below tries to approximate an area for that little diamond shape.
+
+        // It's pure speculation but does appear to remove almost all of the spikes.
+
+        const int64_t len = vSize(to - from);
+        const int64_t other_len = vSize(other_to - other_from);
+        int64_t overlap_width_2 = line_width * 2;
+        if (len <= other_len)
+        {
+            // use length of from -> to
+            if (from_rel < 0)
+            {
+                // segment is closer to other_from
+                overlap_width_2 -= 2 * std::min(vSize(to - other_from), vSize(from - other_from));
+            }
+            else
+            {
+                // segment is closer to other_to
+                overlap_width_2 -= 2 * std::min(vSize(to - other_to), vSize(from - other_to));
+            }
+            const int64_t overlap_length_2 = len * 2;
+            return overlap_length_2 * overlap_width_2 / 4;
+        }
+        else
+        {
+            // use length of other_from -> other_to
+            if (other_from_rel < 0)
+            {
+                // segment is closer to from
+                overlap_width_2 -= 2 * std::min(vSize(other_to - from), vSize(other_from - from));
+            }
+            else
+            {
+                // segment is closer to to
+                overlap_width_2 -= 2 * std::min(vSize(other_to - to), vSize(other_from - to));
+            }
+            const int64_t overlap_length_2 = other_len * 2;
+            return overlap_length_2 * overlap_width_2 / 4;
+        }
     }
 
     if (from_rel != 0 && from_rel == other_from_rel && to_rel == 0 && other_to_rel == 0)
--- a/src/wallOverlap.cpp
+++ b/src/wallOverlap.cpp
@@ -133,21 +133,11 @@ int64_t WallOverlapComputation::getApproxOverlapArea(const Point from, const Poi
     // check whether the line segment overlaps with the point if one of the line segments is just a point
     if (from == to)
     {
-        if (LinearAlg2D::pointIsProjectedBeyondLine(from, other_from, other_to) != 0)
-        {
-            return 0;
-        }
-        const int64_t overlap_length_2 = vSize(other_to - other_from); //Twice the length of the overlap area, alongside the lines.
-        return overlap_length_2 * overlap_width_2 / 4; //Area = width * height.
+        return 0;
     }
     if (other_from == other_to)
     {
-        if (LinearAlg2D::pointIsProjectedBeyondLine(other_from, from, to) != 0)
-        {
-            return 0;
-        }
-        const int64_t overlap_length_2 = vSize(from - to); //Twice the length of the overlap area, alongside the lines.
-        return overlap_length_2 * overlap_width_2 / 4; //Area = width * height.
+        return 0;
     }
 
     short from_rel = LinearAlg2D::pointIsProjectedBeyondLine(from, other_from, other_to);
--- a/src/wallOverlap.cpp
+++ b/src/wallOverlap.cpp
@@ -170,7 +170,9 @@ int64_t WallOverlapComputation::getApproxOverlapArea(const Point from, const Poi
         const Point vec = from - to;
         const int64_t other_to_proj = dot(other_to - to, vec) / vSize(vec);
 
-        const int64_t overlap_length_2 = to_proj + other_to_proj; //Twice the length of the overlap area, alongside the lines.
+        // calculate the overlap area as the length of the region where the lines overlap times the width of that region which is
+        // the line width minus the average distance between the two lines
+        const int64_t overlap_length_2 = to_proj + other_to_proj;
         const int64_t overlap_width_2 = line_width * 2 - std::sqrt(LinearAlg2D::getDist2FromLineSegment(from, other_to, to)) - std::sqrt(LinearAlg2D::getDist2FromLineSegment(other_from, to, other_to));
         return overlap_length_2 * overlap_width_2 / 4; //Area = width * height.
     }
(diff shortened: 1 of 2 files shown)


Ghostkeeper:
Devs, see issue CURA-4191.

src/wallOverlap.cpp
```diff
@@ -62,6 +62,7 @@ float WallOverlapComputation::getFlow(const Point& from, const Point& to)
         //           to other
 
         bool are_in_same_general_direction = dot(from - to, to_other_it.p() - to_other_next_it.p()) > 0;
+#if 0
```
nallath: Please don't disable code this way. If it's no longer needed, remove it.

smartavionics: Hi, that's understood. But as the title says, this PR is up for discussion. I don't want to completely remove it yet because I don't feel 100% confident that I actually understand why it is there in the first place. I would very much like for people who understand how the original code worked to consider the changes I have made and suggest how this could move forward. I'm a bit stuck now.

Ghostkeeper: We've got our Git history for that if necessary :wink: Feel free to remove it.

That said, we will take a better look at it. Nallath's review was just something he noticed when passing by the notification in his e-mail, I think.

smartavionics:
I will be returning to this once the other pending PR's have been dealt with.

smartavionics:
Yes, I agree and, in fact, have already reverted that change. A better change was to increase the allowed gap between the ends of parallel segments (see 4872029). With that, I get much smoother results (less spiky).

src/wallOverlap.cpp
```diff
@@ -182,6 +171,7 @@ int64_t WallOverlapComputation::getApproxOverlapArea(const Point from, const Poi
         const int64_t other_to_proj = dot(other_to - to, vec) / vSize(vec);
 
         const int64_t overlap_length_2 = to_proj + other_to_proj; //Twice the length of the overlap area, alongside the lines.
+        const int64_t overlap_width_2 = line_width * 2 - std::sqrt(LinearAlg2D::getDist2FromLineSegment(from, other_to, to)) - std::sqrt(LinearAlg2D::getDist2FromLineSegment(other_from, to, other_to));
```
Ghostkeeper: This distance should already be stored in the link, which is read by `handlePotentialOverlap` and passed on via the `from_dist` and `to_dist` parameters here. If that distance is wrong, we should fix it when creating the link, not re-compute it here.

src/wallOverlap.cpp
```diff
@@ -202,15 +192,25 @@ int64_t WallOverlapComputation::getApproxOverlapArea(const Point from, const Poi
         const int64_t other_from_proj = dot(other_from - from, vec) / vSize(vec);
 
         const int64_t overlap_length_2 = from_proj + other_from_proj; //Twice the length of the overlap area, alongside the lines.
+        const int64_t overlap_width_2 = line_width * 2 - std::sqrt(LinearAlg2D::getDist2FromLineSegment(from, other_from, to)) - std::sqrt(LinearAlg2D::getDist2FromLineSegment(other_from, from, other_to));
```
Ghostkeeper: Same as above: The distance should already be stored in the link.

src/wallOverlap.cpp
```diff
@@ -202,15 +192,25 @@ int64_t WallOverlapComputation::getApproxOverlapArea(const Point from, const Poi
...
-    const Point to_middle = other_from + to; // don't divide by two just yet
-
-    const int64_t overlap_length_2 = vSize(from_middle - to_middle); //(An approximation of) twice the length of the overlap area, alongside the lines.
-    return overlap_length_2 * overlap_width_2 / 4; //Area = width * height.
+    // shortest segment is overlapped completely - use the length of whichever segment is shortest
+    const int64_t len = vSize(to - from);
+    const int64_t other_len = vSize(other_to - other_from);
+    if (len <= other_len)
+    {
+        const int64_t overlap_width_2 = line_width * 2 - std::sqrt(LinearAlg2D::getDist2FromLineSegment(other_from, from, other_to)) - std::sqrt(LinearAlg2D::getDist2FromLineSegment(other_from, to, other_to));
```
Ghostkeeper: Same as above.

src/wallOverlap.cpp
```diff
@@ -202,15 +192,25 @@ int64_t WallOverlapComputation::getApproxOverlapArea(const Point from, const Poi
...
+        const int64_t overlap_length_2 = len * 2;
+        return overlap_length_2 * overlap_width_2 / 4;
+    }
+    else
+    {
+        const int64_t overlap_width_2 = line_width * 2 - std::sqrt(LinearAlg2D::getDist2FromLineSegment(from, other_from, to)) - std::sqrt(LinearAlg2D::getDist2FromLineSegment(from, other_to, to));
```
Ghostkeeper: Same as above.

src/wallOverlap.cpp
smartavionics: But it's not the distances between to/other_to and from/other_from that I am computing here.

src/wallOverlap.cpp
smartavionics: Same as above reply.

src/wallOverlap.cpp
smartavionics: Same as above reply.

src/wallOverlap.cpp
smartavionics: Same as above reply.

src/wallOverlap.cpp
```diff
@@ -128,26 +129,14 @@ int64_t WallOverlapComputation::handlePotentialOverlap(const ListPolyIt from_it,
...
+        return 0;
     }
     if (other_from == other_to)
     {
-        if (LinearAlg2D::pointIsProjectedBeyondLine(other_from, from, to) != 0)
-        {
-            return 0;
-        }
-        const int64_t overlap_length_2 = vSize(from - to); //Twice the length of the overlap area, alongside the lines.
-        return overlap_length_2 * overlap_width_2 / 4; //Area = width * height.
```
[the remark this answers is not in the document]
smartavionics: Yeah, well, I didn't like the end results that the above code produced. It's a while now since I changed it but if my memory serves me, it was too aggressive. I will go back and check again the difference leaving that out makes.

smartavionics:
So here is an example. The first image shows the results of using the standard unchanged overlap code:

screenshot_2017-10-09_07-48-05

Notice how on the rounded edge the line segments get very much reduced in width in some places. Those places are not narrow regions so the width reduction is excessive.

Next is the result you get when the overlap computation code that handles the 0 length segment case is commented out:

screenshot_2017-10-09_07-48-48

Some of those excessively reduced thinner lines have gone away. It's not perfect but it's better.

The result I get with all of the mods in this PR applied is:

screenshot_2017-10-09_07-53-29

Still not perfect but it's as far as I have got given my meagre understanding of how this code functions.

Here's another part. With the original code, this one suffers from the "pineapple effect" which causes little spikes to occur on convex walls that are flow reduced. Here's a single layer and the spikes are clearly visible (notice how the inside, concave, flow-reduced wall does not have any artifacts):

screenshot_2017-10-09_07-51-06

With the PR code, it looks like this:

screenshot_2017-10-09_07-52-51

src/wallOverlap.cpp
Ghostkeeper: Indeed, you're computing the difference between the distances from `to` to the other line and from `other_to` to this line. That seems kind of random, now that you mention it. What's that for, and can you document it a bit better then? Why can't we store this one in the link?

smartavionics: Yes, I am computing the distances from each of the line ends to the other line because I want to determine the overlap area for exactly the region of overlap which I am calculating as the overlap length times the average distance between the two lines (in the overlap region).
That's a good one about the documentation, this whole subsystem (overlap compensation) is so badly documented nobody but the original author(s) have a chance of of understanding how any of it works (and even they probably can't remember what they did before).
As to why this distance can't be stored in the link, I honestly don't know because I don't understand how the link code works.

Ghostkeeper: I checked that math and it seems to be correct (I arrived at the same result calculating the area of two right triangles). It'd be good to document this area calculation still, even if the rest isn't documented well. And I don't know if we can accept it if it's not pre-calculated in the link.

smartavionics: OK, I will document this calculation. Well, this whole PR is really just for discussing some of the flaws in the overlap compensation. I'm using this PR and I believe it produces better results (at least visually on the layer view) than the original code. Whether you decide to incorporate it (or even just parts of it) really doesn't concern me. At least you are now aware of some of the problems. Every now and again when I am feeling up to the challenge I try to understand this part of Cura a little better but I still don't really "get it"!

jackha:
I'm having trouble to properly test this, here's a try.

Indeed in the provided cases the PR creates smoother results. However, I've bumped into a problem. Also, "I didn't like the end results" doesn't mean it will work better in all cases. I would at least expect some explanation why we don't need those parts.

I've created a half moon to test this PR to also add a sharp corner in the model. The result however, is that the moon didn't have any infill. Also (cause unknown) current master Cura didn't slice all the time and the newest master version would not display any layers while it did with the current master version of CuraEngine. So currently I'm not convinced that we should merge this PR.

Master:
wall-comp-orig

This PR version:
wall-overlap-mb

smartavionics:
Hi Jack, thanks for looking at this. Could you please make the model available so I can try it. Isn't that infill really top surface skin that's controlled by the Fill Gaps Between Walls setting? I wouldn't expect the changes in this PR to make any difference there. I am having problems with the current master of Cura causing the front end to crash and/or losing my existing profiles so I am currently a few days behind the current master branch.

jackha:
Hi Mark,
Thanks for your quick reply. Here's the model. And have fun with it :-) It would be nice if the wall compensation would improve in all cases.

[wall-comp.stl.zip](https://github.com/Ultimaker/CuraEngine/files/1490498/wall-comp.stl.zip)

smartavionics:
Hi Jack, 

Here's how the little model looks with overlap compensation enabled, not so bad:

screenshot_2017-11-21_08-21-40

As for the bigger model with the rounded edges, the sides are missing but they are missing even if I disable overlap compensation so there's something weird about that model!

smartavionics:
As I have mentioned above in other comments, the big problem I have with the overlap compensation code is that the original code (without this PR) has some obvious flaws which the PR tries to improve upon but I honestly do not know how a lot of the overlap compensation code actually works. The code is very poorly documented and it's very hard to figure out what is what. So I am not expecting this PR to be used as it is but rather, it exists to make the Ultimaker devs aware of the shortcomings of the existing implementation. The original authors should really be getting involved as they are probably the only people who will ever understand how this whole subsystem works.

jackha:
Thanks for all the work so far. For us this part of the code is also one of the more complex ones.
I forgot to mention, I also used the small moon at 50%.
I'll discuss what we should do with the issue and with this PR. I printed the rounded.stl and it actually looks quite ok with the current code, but improvements are always welcome :-)

smartavionics:
Surely, an issue gets closed if either the problem is fixed or it is proven not to be a real issue. Closing an issue that is valid and that hasn't been fixed just because the developers can't face fixing it is not good. By all means, leave it to be resolved in the future but leave the issue open to remind yourselves that this is still an area that needs attention.

jackha:
Sure I'll leave it open. Thanks again ^^

Ghostkeeper:
Seems the issue was accidentally closed anyway due to a miscommunication. I've made a new one tagged CURA-4631.

PR #589, opened 2017-08-25
Improve PathOrderOptimizer::getClosestPointInPolygon() handling of points on concave curves.

smartavionics:
The function used to adjust the computed distance of each point from prev_point by a
factor based on the angle between the line segments that pass through the point. It still
does this but with some differences. Now, it only adjusts the distance when the point
is on a concave curve and the amount of adjustment is now proportional to not only the
amount of curvature but also to the un-adjusted distance. The further away the point is
from prev_point, the more the computed distance is reduced when it lies on a concave curve.

### DIFF
--- a/src/pathOrderOptimizer.cpp
+++ b/src/pathOrderOptimizer.cpp
@@ -127,12 +127,17 @@ int PathOrderOptimizer::getClosestPointInPolygon(Point prev_point, int poly_idx)
         const Point& p1 = poly[point_idx];
         const Point& p2 = poly[(point_idx + 1) % poly.size()];
         int64_t dist = vSize2(p1 - prev_point);
-        float is_on_inside_corner_score = -LinearAlg2D::getAngleLeft(p0, p1, p2) / M_PI * 5000 * 5000; // prefer inside corners
-        // this score is in the order of 5 mm
-        if (dist + is_on_inside_corner_score < best_point_score)
+        const float corner_angle = LinearAlg2D::getAngleLeft(p0, p1, p2) / M_PI; // 0 -> 2
+        if (corner_angle > 1)
+        {
+            // p1 lies on a concave curve so reduce the distance to favour it
+            // the more concave the curve, the more we reduce the distance
+            dist -= (corner_angle - 1) * dist / 50;
+        }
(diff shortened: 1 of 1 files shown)


BagelOrb:
So simple, yet so brilliant!

src/pathOrderOptimizer.cpp
```diff
@@ -127,12 +127,17 @@ int PathOrderOptimizer::getClosestPointInPolygon(Point prev_point, int poly_idx)
...
         int64_t dist = vSize2(p1 - prev_point);
-        float is_on_inside_corner_score = -LinearAlg2D::getAngleLeft(p0, p1, p2) / M_PI * 5000 * 5000; // prefer inside corners
-        // this score is in the order of 5 mm
-        if (dist + is_on_inside_corner_score < best_point_score)
+        const float corner_angle = LinearAlg2D::getAngleLeft(p0, p1, p2) / M_PI; // 0 -> 2
+        if (corner_angle > 1)
+        {
+            // p1 lies on a concave curve so reduce the distance to favour it
+            // the more concave the curve, the more we reduce the distance
+            dist -= (corner_angle - 1) * dist / 50;
```
BagelOrb: Now the corner score only contributes 2% of the total distance. That seems to me to be a bit on the low side.

If we travel 1cm then this formula means we are willing to take a detour of 0.2mm.
However, when traveling so far I would like a more global optimum.

What about ` / 5` instead?

smartavionics: I tried various values for the divider and decided that I preferred the concave corners to have a weak influence because otherwise they tend to grab the seam rather too keenly. As I mention somewhere, you can adjust the strength of the pull by changing the position of the seam location hint point, as that gets further away the concave points are more likely to grab the seam.
I think you need to play with it and decide what the divisor should be. I started on 10 and preferred 50.

BagelOrb:
@Devs, see https://jira.ultimaker.com:8443/browse/CURA-4239

I would approve these changes, but extensive testing is required on the hardcoded divisor in the formula.

smartavionics:
#590 takes this further and is based on this PR. Closing...

PR #590, opened 2017-08-25
Add more z seam options

smartavionics:
This PR adds the ability to specify the z-seam location based on the curvature of the outlines. You can choose the seam to be located at the location that is most concave, most convex or most curved (concave or convex).

### DIFF
--- a/src/pathOrderOptimizer.cpp
+++ b/src/pathOrderOptimizer.cpp
@@ -13,17 +13,16 @@ namespace cura {
 */
 void PathOrderOptimizer::optimize()
 {
-    const EZSeamType type = (config)? config->type : EZSeamType::SHORTEST;
     bool picked[polygons.size()];
     memset(picked, false, sizeof(bool) * polygons.size());/// initialized as falses
     
     for (unsigned poly_idx = 0; poly_idx < polygons.size(); ++poly_idx) /// find closest point to initial starting point within each polygon +initialize picked
     {
         const ConstPolygonRef poly = *polygons[poly_idx];
-        switch (type)
+        switch (config.type)
         {
             case EZSeamType::USER_SPECIFIED:
-                polyStart.push_back(getClosestPointInPolygon(config->pos, poly_idx));
+                polyStart.push_back(getClosestPointInPolygon(config.pos, poly_idx));
                 break;
             case EZSeamType::RANDOM:
                 polyStart.push_back(getRandomPointInPolygon(poly_idx));
@@ -82,7 +81,6 @@ void PathOrderOptimizer::optimize()
 
 int PathOrderOptimizer::getClosestPointInPolygon(Point prev_point, int poly_idx)
 {
-    const EZSeamType type = (config)? config->type : EZSeamType::SHORTEST;
     ConstPolygonRef poly = *polygons[poly_idx];
 
     int best_point_idx = -1;
@@ -93,21 +91,20 @@ int PathOrderOptimizer::getClosestPointInPolygon(Point prev_point, int poly_idx)
         const Point& p1 = poly[point_idx];
         const Point& p2 = poly[(point_idx + 1) % poly.size()];
         // when type is SHARPEST_CORNER, actual distance is ignored, we use a fixed distance and decision is based on curvature only
-        int64_t dist = (type == EZSeamType::SHARPEST_CORNER)? 100 : vSize2(p1 - prev_point);
+        int64_t dist_score = (config.type == EZSeamType::SHARPEST_CORNER)? 100 : vSize2(p1 - prev_point);
         const float corner_angle = LinearAlg2D::getAngleLeft(p0, p1, p2) / M_PI; // 0 -> 2
-        const EZSeamCornerPrefType corner_pref = (config)? config->corner_pref : EZSeamCornerPrefType::Z_SEAM_CORNER_PREF_NONE;
-        if (type == EZSeamType::SHORTEST)
+        if (config.type == EZSeamType::SHORTEST)
         {
             // the more a corner satisfies our criteria the closer it appears to be
             const int64_t corner_shift = 10000 * 10000; // shift 10mm for a very acute corner
-            switch (corner_pref)
+            switch (config.corner_pref)
             {
                 case EZSeamCornerPrefType::Z_SEAM_CORNER_PREF_INNER:
                     if (corner_angle > 1)
                     {
                         // p1 lies on a concave curve so reduce the distance to favour it
                         // the more concave the curve, the more we reduce the distance
-                        dist -= (corner_angle - 1) * corner_shift;
+                        dist_score -= (corner_angle - 1) * corner_shift;
                     }
                     break;
                 case EZSeamCornerPrefType::Z_SEAM_CORNER_PREF_OUTER:
@@ -115,12 +112,12 @@ int PathOrderOptimizer::getClosestPointInPolygon(Point prev_point, int poly_idx)
                     {
                         // p1 lies on a convex curve so reduce the distance to favour it
                         // the more convex the curve, the more we reduce the distance
-                        dist -= (1 - corner_angle) * corner_shift;
+                        dist_score -= (1 - corner_angle) * corner_shift;
                     }
                     break;
                 case EZSeamCornerPrefType::Z_SEAM_CORNER_PREF_ANY:
                     // the more curved the region, the more we reduce the distance
-                    dist -= fabs(corner_angle - 1) * corner_shift;
+                    dist_score -= fabs(corner_angle - 1) * corner_shift;
                     break;
                 case EZSeamCornerPrefType::Z_SEAM_CORNER_PREF_NONE:
                 default:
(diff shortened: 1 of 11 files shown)


BagelOrb:
I like the user to be able to choose whether he wants to hide the z-seam, make it accessible, or do both.
However, the way you implemented it the user has to choose between those options and whether to take the shortest route.

I think that what you ultimately want is always a combination of nearby locations and seam hiding locations.
I was thinking you would change the distance score only.
So in `PathOrderOptimizer::getClosestPointInPolygon` you would do something like:
```
switch(corner_preference)
{
case OUTSIDE:
dist -= std::max(0.0, corner_angle - 1) * dist / 50;
break;
case INSIDE:
dist -= std::min(0.0, corner_angle - 1) * dist / 50;
break;
case BOTH:
dist -= std::abs(corner_angle - 1) * dist / 50;
break;
}
```

You would then have a separate setting in the frontend with the options you've now added to Z Seam Alignment...

What do you think about doing something like that instead?

smartavionics:
Yeah, I did think about combining the concave/convex preference with the user-specified location hint but didn't get as far as trying it out. Let me have a go at that.

BagelOrb:
I think you misunderstood me.

What I meant was to introduce a new setting - something like this
```
"z_seam_preference":
{
    "label": "Z Seam Preference",
    "options":
    {
        "most_curvature_convex":  "Hidden",
        "most_curvature_concave": "Exposed",
        "most_curvature":         "Hidden or Exposed"
```

smartavionics:
Yes, that's what I'm doing (well, similar, I think!)

BagelOrb:
I was talking to a colleague and we were pondering whether it would even be good to introduce even more settings:
```
"seam_corner_angle_divisor":
{
    "label": "Seam Detour Preference",
    "options":
    {
        "0":  "Shortest Path",
        "50":  "Nearby Hiding/Exposing Location",
        "10":  "Average Detour",
        "1":  "Best Hiding/Exposing Location"
```

src/pathOrderOptimizer.cpp
```diff
@@ -96,27 +97,16 @@ void PathOrderOptimizer::optimize()
...
         }
     }
 }
 
-int PathOrderOptimizer::getPolyStart(Point prev_point, int poly_idx)
-{
-    switch (type)
-    {
-        case EZSeamType::USER_SPECIFIED:    return getClosestPointInPolygon(z_seam_pos, poly_idx);
-        case EZSeamType::RANDOM:            return getRandomPointInPolygon(poly_idx);
```
BagelOrb: It seems like the call to `getRandomPointInPolygon(.)` got removed by this commit.

Is the random seam still working?

smartavionics: Yes.

BagelOrb:
@Devs see CURA-4244

src/pathOrderOptimizer.cpp
```diff
@@ -126,13 +92,78 @@ int PathOrderOptimizer::getClosestPointInPolygon(Point prev_point, int poly_idx)
...
         const Point& p2 = poly[(point_idx + 1) % poly.size()];
-        int64_t dist = vSize2(p1 - prev_point);
-        float is_on_inside_corner_score = -LinearAlg2D::getAngleLeft(p0, p1, p2) / M_PI * 5000 * 5000; // prefer inside corners
-        // this score is in the order of 5 mm
-        if (dist + is_on_inside_corner_score < best_point_score)
+        // when type is SHARPEST_CORNER, actual distance is ignored, we use a fixed distance and decision is based on curvature only
+        int64_t dist = (type == EZSeamType::SHARPEST_CORNER)? 100 : vSize2(p1 - prev_point);
+        const float corner_angle = LinearAlg2D::getAngleLeft(p0, p1, p2) / M_PI; // 0 -> 2
+        const EZSeamCornerPrefType corner_pref = (config)? config->corner_pref : EZSeamCornerPrefType::Z_SEAM_CORNER_PREF_NONE;
+        if (type == EZSeamType::SHORTEST)
```
BagelOrb: There's a lot of code duplication between this if branch and the else branch.
You can simplify it a lot by combining the two branches and merging the corner_shift with the dist/scaling_factor:
```
const int64_t corner_shift = (type == EZSeamType::SHORTEST)? 10000 * 10000 : dist / 50;
```

src/pathOrderOptimizer.cpp
```diff
@@ -126,13 +92,78 @@ int PathOrderOptimizer::getClosestPointInPolygon(Point prev_point, int poly_idx)
     {
         const Point& p1 = poly[point_idx];
         const Point& p2 = poly[(point_idx + 1) % poly.size()];
-        int64_t dist = vSize2(p1 - prev_point);
-        float is_on_inside_corner_score = -LinearAlg2D::getAngleLeft(p0, p1, p2) / M_PI * 5000 * 5000; // prefer inside corners
-        // this score is in the order of 5 mm
-        if (dist + is_on_inside_corner_score < best_point_score)
+        // when type is SHARPEST_CORNER, actual distance is ignored, we use a fixed distance and decision is based on curvature only
+        int64_t dist = (type == EZSeamType::SHARPEST_CORNER)? 100 : vSize2(p1 - prev_point);
```
BagelOrb: Please rename to `dist_score`. It's not only a distance measure after the corner scoring below.

src/pathOrderOptimizer.h
```diff
@@ -17,17 +37,15 @@ namespace cura {
 class PathOrderOptimizer
 {
 public:
-    EZSeamType type;
     Point startPoint; //!< A location near the prefered start location
-    Point z_seam_pos; //!< The position near where to create the z_seam (if \ref PathOrderOptimizer::type == 'back')
+    const ZSeamConfig* config;
```
BagelOrb: You made this member a pointer type so that it can be optional, while ZSeamConfig has a default constructor which looks as if it's meant exactly for that.

I think the whole charm of introducing the ZSeamConfig class is that you don't need a parameter with a default value which therefore has to be at the end of the function signature. I think that instead of working with `ZSeamConfig*` you should just work with `const ZSeamConfig&`.

src/pathOrderOptimizer.h
```diff
@@ -7,7 +7,27 @@
 #include "settings/settings.h"
 
 namespace cura {
- 
+
+class ZSeamConfig
```
BagelOrb: Please add documentation for this newly introduced class.

src/pathOrderOptimizer.h
```diff
@@ -17,17 +37,15 @@ namespace cura {
...
+    const ZSeamConfig* config;
     std::vector<ConstPolygonPointer> polygons; //!< the parts of the layer (in arbitrary order)
     std::vector<int> polyStart; //!< polygons[i][polyStart[i]] = point of polygon i which is to be the starting point in printing the polygon
     std::vector<int> polyOrder; //!< the optimized order as indices in #polygons
 
-    PathOrderOptimizer(Point startPoint, Point z_seam_pos = Point(0, 0), EZSeamType type = EZSeamType::SHORTEST)
-    : type(type)
-    , startPoint(startPoint)
-    , z_seam_pos(z_seam_pos)
+    PathOrderOptimizer(Point startPoint, const ZSeamConfig* config = nullptr)
```
BagelOrb: Following the comment above I'd like this function signature to be
```
PathOrderOptimizer(Point startPoint, const ZSeamConfig& config)
```
or
```
PathOrderOptimizer(Point startPoint, const ZSeamConfig& config = ZSeamConfig())
```
So you can just call it with
```
PathOrderOptimizer optimizer(startPoint, ZSeamConfig())
```
or
```
PathOrderOptimizer optimizer(startPoint)
```

src/pathOrderOptimizer.cpp
```diff
@@ -126,13 +90,77 @@ int PathOrderOptimizer::getClosestPointInPolygon(Point prev_point, int poly_idx)
...
         const Point& p1 = poly[point_idx];
         const Point& p2 = poly[(point_idx + 1) % poly.size()];
-        int64_t dist = vSize2(p1 - prev_point);
-        float is_on_inside_corner_score = -LinearAlg2D::getAngleLeft(p0, p1, p2) / M_PI * 5000 * 5000; // prefer inside corners
-        // this score is in the order of 5 mm
-        if (dist + is_on_inside_corner_score < best_point_score)
+        // when type is SHARPEST_CORNER, actual distance is ignored, we use a fixed distance and decision is based on curvature only
+        int64_t dist_score = (config.type == EZSeamType::SHARPEST_CORNER)? 100 : vSize2(p1 - prev_point);
+        const float corner_angle = LinearAlg2D::getAngleLeft(p0, p1, p2) / M_PI; // 0 -> 2
+        if (config.type == EZSeamType::SHORTEST)
```
BagelOrb: There's a lot of code duplication between this if branch and the else branch.
You can simplify it a lot by combining the two branches and merging the corner_shift with the dist/scaling_factor:
```
const int64_t corner_shift = (type == EZSeamType::SHORTEST)? 10000 * 10000 : dist / 50;
```

smartavionics: I think I would rather leave the duplicated code until we are happy that the two strategies are working OK.

BagelOrb: The way we work is that first we review the code of a PR and then we test whether it works alright. According to our workflow I cannot move on to testing until the code is okay.
You could also remove the code duplication like this:
```
float corrected_corner_angle;
(config.corner_pref)
{
  case EZSeamCornerPrefType::Z_SEAM_CORNER_PREF_INNER:
    corrected_corner_angle = std::max(0, corner_angle - 1);
...
}
if (config.type == EZSeamType::SHORTEST)
{
  dist -= corrected_corner_angle * corner_shift;
}
else
{  dist -= corrected_corner_angle * dist_score / scaling_divisor; }
```

PR #594, opened 2017-08-31
Change how the no-skin combing mode calculated the comb boundary.

smartavionics:
Before, when no-skin mode was active, the combing boundary was just the boundary of the
infill regions and it didn't include the inner walls that define the combing boundary when
the "all" mode is active.

Now, the combing boundary for no-skin mode is computed by taking the same boundary as used
for the "all" mode and subtracting the skin areas, that's what no-skin means!

### DIFF
--- a/src/LayerPlan.cpp
+++ b/src/LayerPlan.cpp
@@ -151,16 +151,20 @@ Polygons LayerPlan::computeCombBoundaryInside(CombingMode combing_mode)
             if (mesh.getSettingBoolean("infill_mesh")) {
                 continue;
             }
+            // comb to close to inside of the outer wall
+            layer.getSecondOrInnermostWalls(comb_boundary);
             if (mesh.getSettingAsCombingMode("retraction_combing") == CombingMode::NO_SKIN)
             {
+                // subtract "no go" skin areas
+                Polygons skin_part_outlines;
                 for (const SliceLayerPart& part : layer.parts)
                 {
-                    comb_boundary.add(part.infill_area);
+                    for (const SkinPart skin_part: part.skin_parts)
+                    {
+                        skin_part_outlines.add(skin_part.outline.outerPolygon());
+                    }
                 }
-            }
-            else
-            {
-                layer.getSecondOrInnermostWalls(comb_boundary);
+                comb_boundary = comb_boundary.difference(skin_part_outlines);
             }
         }
         return comb_boundary;


BagelOrb:
But then there's a small gap between the walls and the skin.
This means that combing can occur on top surfaces between the second wall and the skin.
The whole reason for no skin combing is that no combing is done on visible layers, right?

smartavionics:
OK, but then the setting has the wrong name, it should not be "no skin" it should be "no skin or walls". 

The reason this PR came about is that I have a part that has narrow regions (too narrow for infill) that require retractions to cross when the original no-skin mode is in use. Using "all" mode is not what I want because it then dribbles on the first layer. Perhaps, what would be better than this PR is to always use no-skin mode for the first layer when "all" mode is active?

BagelOrb:
Well, the walls are always avoided when using combing. That's the whole idea of combing ;)

Why is such oozing a problem for the first layer?
Why is that problem not present on other layers?

smartavionics:
> Well, the walls are always avoided when using combing. That's the whole idea of combing ;)

Really, this is using combing mode all, notice how some of the travel lines are on top of the inner walls!
screenshot_2017-08-31_10-29-35

> Why is such oozing a problem for the first layer?

Because when I look at the first layer I can see the ooze lines crossing between outlines rather than a nice regular skin fill.

> Why is that problem not present on other layers?

Obviously, it will ooze on other layers but as they will be invisible I don't care.

But, perhaps, also it would be good to not comb across, say, the top 2 skin layers as dribble on those will be visible also.

BagelOrb:
Okay, so only the outer wall is avoided - not all walls. If we could make no-skin combing avoid only the outer wall (+ half a wall) without waisting too much computation power I would be willing to merge a PR, but I have no idea how to do it efficiently.

Okay so it's just a cosmetic issue. 
Well, that's what no-skin combing is for exactly. I don't really see a reason to always apply no-skin combing on the first layer.

Let's focus on the real issue here: no-skin combing uses the infill areas which excludes all walls rather than only excluding the outer two walls only.

If we could efficiently compute the correct combing area from the second walls and the avoidance area we could fix the problem.

Then we could also improve the combing area when Top Surface Skin Layers are enabled. We should then only avoid the Top Surface Skin rather than all top/bottom skin.

Ghostkeeper:
Devs, see issue CURA-4307.

smartavionics:
Closing to reduce PR clutter!

PR #596, opened 2017-09-01
Now supports separate speed/accel/jerk for layer 0 walls.

smartavionics:
The idea here is that on layer 0, the walls can be printed slow to obtain good adhesion
and then the skin can go down using the normal layer 0 settings.


nallath:
CURA-4266

Ghostkeeper:
We've decided not to accept this change. See the discussion at https://github.com/Ultimaker/Cura/pull/2356 .

PR #597, opened 2017-09-01
Implemented separate flow setting for layer 0

smartavionics:
Now uses material_flow_layer_0 for everything on layer 0.

However, it doesn't alter the flow used for layers < 0, should it?

This PR is built on top of #596.

### DIFF
--- a/src/settings/PathConfigStorage.cpp
+++ b/src/settings/PathConfigStorage.cpp
@@ -27,7 +27,7 @@ std::vector<double> PathConfigStorage::getLineWidthFactorPerExtruder(const Slice
     return ret;
 }
 
-GCodePathConfig createPerimeterGapConfig(const SliceMeshStorage& mesh, int layer_thickness)
+GCodePathConfig createPerimeterGapConfig(const SliceMeshStorage& mesh, int layer_nr, int layer_thickness)
 {
     // The perimeter gap config follows the skin config, but has a different line width:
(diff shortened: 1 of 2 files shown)


nallath:
CURA-4268

nallath:
Layers below 0 are those reserved for the raft. I'm not a 100% sure what to do with those; One could argue that it should be seen as "layer 0".

smartavionics:
>Layers below 0 are those reserved for the raft. I'm not a 100% sure what to do with those; One could argue that it should be seen as "layer 0".

It only needs to be used for the very first layer of the raft, not all of it?

smartavionics:
I have changed the code so that the flow setting applies to either layer 0 (when no raft is used) or the first layer of the raft. It works but I wonder if it makes sense. After all, the first layer of the raft is printed using a different line width and spacing and doesn't (at least not with the default settings) have 100% fill. So before I commit the changed code, I'm looking to get some opinions from others. What do the Cura devs think?

Ghostkeeper:
I was going to say, it makes sense to apply this to the first printed layer, rather than always to the layer marked as "layer 0". It's the layer that touches the print bed that needs the improved adhesion, after all.

smartavionics:
OK, it now uses this setting for all items that touch the print bed (model, support, raft/skirt/brim).

smartavionics:
Closing as it's based on #596 which has been rejected.

PR #599, opened 2017-09-02
Update currentSpeed in writeZHopStart() and writeZHopEnd().

smartavionics:
It wasn't being updated and so the Z feedrate used was subsequently used for X/Y moves.

Should help with https://github.com/Ultimaker/Cura/issues/2371.

### DIFF
--- a/src/gcodeExport.cpp
+++ b/src/gcodeExport.cpp
@@ -816,7 +816,8 @@ void GCodeExport::writeZhopStart(int hop_height)
     if (hop_height > 0)
     {
         isZHopped = hop_height;
-        *output_stream << "G1 F" << PrecisionedDouble{1, current_max_z_feedrate * 60} << " Z" << MMtoStream{current_layer_z + isZHopped} << new_line;
+        currentSpeed = current_max_z_feedrate * 60;
+        *output_stream << "G1 F" << PrecisionedDouble{1, currentSpeed} << " Z" << MMtoStream{current_layer_z + isZHopped} << new_line;
         total_bounding_box.includeZ(current_layer_z + isZHopped);
         assert(current_max_z_feedrate > 0.0 && "Z feedrate should be positive");
     }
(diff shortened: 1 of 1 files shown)


smartavionics:
Hi, if you're going to release a 2.7.1, please include this.

nallath:
There won't be a 2.7.1 as far as I see. Unless this is a really major blocking issue (eg; Machines are going to break if we don't do this)

src/gcodeExport.cpp
```diff
@@ -816,7 +816,8 @@ void GCodeExport::writeZhopStart(int hop_height)
     if (hop_height > 0)
     {
         isZHopped = hop_height;
-        *output_stream << "G1 F" << PrecisionedDouble{1, current_max_z_feedrate * 60} << " Z" << MMtoStream{current_layer_z + isZHopped} << new_line;
+        currentSpeed = current_max_z_feedrate * 60;
```
nallath: Could you tell me how this is any different from what it already did? As far as I see this is just a different way of saying the exact same thing (eg; instead of doing current_max_z_feedrate * 60 inline, you first assign it to a variable)

smartavionics:
Yes, but that variable is tracking what the last feed rate was set to. It needs to be kept in sync with the feed rate last used in a G0 or G1 sent to the printer. Because writeZHopStart() used a low feed rate for the z-hop and it didn't update currentSpeed, the next X/Y movement that was output didn't contain a feed rate and so was much slower than it should have been.

smartavionics:
Two people have now reported this bug. It's pretty bloody annoying. Are you sure you don't want to release a 2.7.1 that includes this fix?

nallath:
Yes, we are going to release the next version (beta) soon-ish, so releasing another version doesn't make a whole lot of sense.

smartavionics:
OK, understood, thanks.

PR #600, opened 2017-09-03
Sync z-hop with layer changes

smartavionics:
Before, if there was z-hop at a layer change, any travel move in the the new layer before
the first extrusion would have been done with that much z-hop above the new layer z so it
would have involved changing z again.

See also #599 and https://github.com/Ultimaker/Cura/issues/2371.

### DIFF
--- a/src/gcodeExport.cpp
+++ b/src/gcodeExport.cpp
@@ -280,7 +280,23 @@ EGCodeFlavor GCodeExport::getFlavor() const
 
 void GCodeExport::setZ(int z)
 {
-    this->current_layer_z = z;
+    if (isZHopped)
+    {
+        // adjust the current z-hop height for the new layer z
+        isZHopped = current_layer_z + isZHopped - z;
+        if (isZHopped < 0)
+        {
+            isZHopped = 0;
+        }
+        if (isZHopped == 0)
+        {
+            // the new layer height is the same as the previous z-hop height
+            // so the z-hop height has become 0 and we need to update currentPosition.z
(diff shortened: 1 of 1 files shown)


smartavionics:
Devs, an unanswered question is why doesn't writeZHopStart() update currentPosition.z ? Should it?

nallath:
CURA-4280

Ghostkeeper:
> why doesn't writeZHopStart() update currentPosition.z ? Should it?

I'd expect that it would update the `currentPosition.z`.

Ghostkeeper:
Apart from that `currentPosition` thing which was always a problem, the fix itself looks good to me. I'll test it after I've done some Github ticket administration.

smartavionics:
OK, thanks, please consider testing #608 at the same time as that is fixing more of a problem in the same area.

diegopradogesto:
After some review in the behavior, your code fixes the issue when the `Z Hop Height` is bigger than the layer height. As you can see in the picture zhop_bigger_than_layer_height_ok in the right column is the new code and the line 7123 disappears as expected.

However, if you choose to do a Z Hop lower than the layer height, the gcode that updates the height in the new layer is deleted, and the printer will use the latest Z value that is the last Z hop. As you can see in the picture zhop_lower_than_layer_height_ko , the line 7123 disappears but also the line 7127, and this is not expected.

smartavionics:
Oh dear, you're absolutely right, sorry about that. Let's forget this PR and keep the old code.

PR #604, opened 2017-09-09
Use const ref setting strings rather than objects

smartavionics:
This PR should provide a little speedup by avoiding creating temporary objects to be returned by the getSettingString() method(s). Also, change quite a few of the callers of this method to use a const ref local variable rather than an object to avoid yet another construction/destruction.


Ghostkeeper:
Developers, see issue CURA-4330.

PR #606, opened 2017-09-11
Start polygon at closest point when SHARPEST_CORNER and Z_SEAM_CORNER_PREF_NONE both specified.

smartavionics:
The UI should not allow this combination as it doesn't make sense but it does. Before this
change, with this combination of settings, the first point in the polygon would be chosen
but I think it's better to use the closest point (to the current position) instead.


Ghostkeeper:
I agree with your initial assessment that it doesn't make sense to allow this in the UI. Let's just deactivate the combination there.

smartavionics:
> I agree with your initial assessment that it doesn't make sense to allow this in the UI. Let's just deactivate the combination there.
And how is that achieved?

Ghostkeeper:
> And how is that achieved?

Yeah it dawned on me too late :stuck_out_tongue: It's still close to dawn over here.

Ghostkeeper:
Devs, see issue CURA-4337.

PR #608, opened 2017-09-15
Fix the fix - assign correct value to currentSpeed when doing z-hop.

smartavionics:
Sorry, my previous fix to the z-hop speed bug was flawed, please merge this ASAP.


smartavionics:
This needs to be merged before 3.0!

Ghostkeeper:
We'll get it in, but this needs to be tested. Devs, see CURA-4350.

PR #609, opened 2017-09-18
Most line based infill patterns are now positioned relative to the centre of the model.

smartavionics:
This makes the infill immune to the position of the model on the build plate.

New settings infill_offset_x and infill_offset_y shift the infill pattern relative to the
model centre.

See https://github.com/Ultimaker/Cura/issues/2412.

### DIFF
--- a/src/infill.cpp
+++ b/src/infill.cpp
@@ -257,29 +257,30 @@ void Infill::addLineInfill(Polygons& result, const PointMatrix& rotation_matrix,
     }
 }
 
-void Infill::generateLineInfill(Polygons& result, int line_distance, const double& fill_angle, int64_t shift)
+int64_t Infill::getShiftOffsetFromInfillOriginAndRotation(const double& infill_rotation)
 {
     if (infill_origin.X != 0 || infill_origin.Y != 0)
     {
-        const double fill_angle_rads = fill_angle * M_PI / 180;
-        shift += infill_origin.X * std::cos(fill_angle_rads) - infill_origin.Y * std::sin(fill_angle_rads);
+        const double rotation_rads = infill_rotation * M_PI / 180;
+        return infill_origin.X * std::cos(rotation_rads) - infill_origin.Y * std::sin(rotation_rads);
     }
-    PointMatrix rotation_matrix(fill_angle);
+    return 0;
+}
+
+void Infill::generateLineInfill(Polygons& result, int line_distance, const double& infill_rotation, int64_t shift)
+{
+    shift += getShiftOffsetFromInfillOriginAndRotation(infill_rotation);
+    PointMatrix rotation_matrix(infill_rotation);
     NoZigZagConnectorProcessor lines_processor(rotation_matrix, result);
     bool connected_zigzags = false;
     generateLinearBasedInfill(outline_offset, result, line_distance, rotation_matrix, lines_processor, connected_zigzags, shift);
 }
 
 
-void Infill::generateZigZagInfill(Polygons& result, const int line_distance, const double& fill_angle)
+void Infill::generateZigZagInfill(Polygons& result, const int line_distance, const double& infill_rotation)
 {
-    int64_t shift = 0;
-    if (infill_origin.X != 0 || infill_origin.Y != 0)
-    {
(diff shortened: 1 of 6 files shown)


nallath:
CURA-4338

src/infill.cpp
```diff
@@ -259,6 +259,11 @@ void Infill::addLineInfill(Polygons& result, const PointMatrix& rotation_matrix,
 
 void Infill::generateLineInfill(Polygons& result, int line_distance, const double& fill_angle, int64_t shift)
 {
+    if (infill_origin.X != 0 || infill_origin.Y != 0)
+    {
+        const double fill_angle_rads = fill_angle * M_PI / 180;
+        shift += infill_origin.X * std::cos(fill_angle_rads) - infill_origin.Y * std::sin(fill_angle_rads);
+    }
```
Ghostkeeper: Could you make this in to a utility function in linearAlg2D instead? Just rotating a point could be useful elsewhere too.

We have a function called rotateAround() there but that might be too heavyweight for this sort of thing.

smartavionics: I don't mind packaging that code into a function but I'm struggling to think of what it should be called! It's not really rotating a point it's doing a calculation whereby the value of shift is adjusted based on the infill origin point and the infill angle.

smartavionics:
I made a new method to encapsulate that calculation and also renamed a couple of method parameters as they shadowed a class member which could be confusing.

In the future perhaps we should contemplate making the infill angles relative to the rotation of the mesh. That way, if you rotate a part on the build plate, the infill pattern would not change. Something to think about... (but not today)

smartavionics:
Ping! just checking if you're waiting on me

nickthetait:
Agree that infill angle/direction would be the next step improvement :wink: 

Just curious, what patterns cannot be adjusted by this PR? What sort of technical problems got in the way?

smartavionics:
Hi @nickthetait, in the UI, the ability to specify the offset for the infill is limited by this:
````
"enabled": "infill_pattern == 'grid' or infill_pattern == 'lines' or infill_pattern == 'triangles' or infill_pattern == 'cubic' or infill_pattern == 'tetrahedral' or infill_pattern == 'quarter_cubic' or infill_pattern == 'zigzag'",
````
That shows you the patterns that are adjusted by this PR, all others are not. Technically, those patterns that are created using a function called generateLinearBasedInfill() are supported and the patterns that are created using other functions are not. I guess, with more work, some or even all of the remaining patterns could also support this feature although as I never use those other patterns (and seriously question the need to use them at all), I am not too motivated to put time into that.

Ghostkeeper:
The other available patterns are Concentric, Concentric 3D, Cross, Cross 3D and Cubic Subdivision. Concentric and Concentric 3D already move along with the object by their nature. Cross and Cross 3D also already move along but more because of how they were implemented. Cubic Subdivision is then the only one that doesn't, but the infill density with that pattern is usually so high that it doesn't really matter I think.

smartavionics:
> Cubic Subdivision is then the only one that doesn't

That's right and I don't know how that could be altered to make it relative to the part's middle. If anyone really wants that capability, they can work out how to do it!

nickthetait:
Yeah, seems like those other patterns don't need this ability or it is irrelevant

Ghostkeeper:
Infill meshes get their own origin, but I guess that's expected behaviour.

smartavionics:
The infill patterns affected by this PR will now be positioned (in X and Y) relative to the middle of the part's BB with optional X/Y offsets supplied by the user.

Ghostkeeper:
Yeah, I got that, but for infill meshes it's a bit of a special case, since the infill is still within the normal mesh but modified by the infill mesh. The modification takes the centre of the infill mesh rather than the centre of the normal mesh, so the lines don't match up any more. The behaviour is largely fine though. In fact, perhaps it is the most logical behaviour, though the infill will be slightly weaker.

I just thought it might be worth mentioning. I think it's fine either way.

smartavionics:
OK, yes, good point. As the densities are most likely to be different between the infill mesh region and the part's region then the lines probably would not match up anyway. Using zigzag infill would probably give the best bond between the regions as shown in this example:

screenshot_2017-10-02_13-55-59

PR #611, opened 2017-09-23
Fixes SEGV when infill/support pattern is anything other than a cross fractal

smartavionics:
This PR fixes recently committed breakage whereby empty array objects were accessed causing a SEGV. You may not care for this solution but it's simple and effective.


smartavionics:
Damn, just realised that this isn't a complete solution, closing...!

PR #612, opened 2017-09-24
Only index into cross_fill_patterns[] when it contains elements!

smartavionics:
Fix recently introduced crash caused by accessing arrays that contain no elements.


smartavionics:
Usurped by a fix from the bug creator! Closing...

Ghostkeeper:
Oops, sorry. I didn't look at this fix. But thanks for the effort anyway :)

PR #613, opened 2017-09-27
Make the LineOrderOptimizer prefer to start at ends of a sequence of connected lines.

smartavionics:
Previously, when processing zigzag patterns of lines, it was possible for the lines to be
output as disconnected pairs of lines (a zig and a zag together). This commit adds code that
determines the vertices that are at the end of a sequence of lines and it selects the end
vertex that is closest to the current point as the next to be output.

This still doesn't guarantee that the lines are output optimally but it does at least appear
to stop the disconnected line pairs problem.

See https://ultimaker.com/en/community/51264-cura-27-toolpath?page=last#lastreply for a discussion
about the problem this PR targeted at.

**This one is going to require quite a lot of testing as it is used for infill, skin, (pretty much everything).**

### DIFF
--- a/src/pathOrderOptimizer.cpp
+++ b/src/pathOrderOptimizer.cpp
@@ -336,14 +336,14 @@ inline float LineOrderOptimizer::travelDistance(const Point& p0, const Point& p1
     CombPath comb_path;
     if (LinePolygonsCrossings::comb(*combing_boundary, *loc_to_line, p0, p1, comb_path, -40, 0, true))
     {
-        float dist2 = 0;
+        float dist = 0;
         Point last_point = p0;
         for (const Point& comb_point : comb_path)
         {
-            dist2 += vSize2f(comb_point - last_point);
+            dist += vSize(comb_point - last_point);
             last_point = comb_point;
         }
-        return dist2;
+        return dist * dist;
     }
     // fall back to direct distance
     return vSize2f(p0 - p1);
--- a/src/pathOrderOptimizer.cpp
+++ b/src/pathOrderOptimizer.cpp
@@ -222,7 +222,15 @@ void LineOrderOptimizer::optimize()
             }
         }
 
-        // if no line ends close to last_point have been found, see if we can find a point on a line that could be the start of a chain of lines
+        if (have_chains && best_line_idx != -1 && !pointsAreCoincident(prev_point, (*polygons[best_line_idx])[polyStart[best_line_idx]]))
+        {
+            // we found a point close to prev_point but it's not close enough for the points to be considered coincident so we would
+            // probably be better off by ditching this point and finding an end of a chain instead (let's hope it's not too far away!)
+            best_line_idx = -1;
+            best_score = std::numeric_limits<float>::infinity();
+        }
+
+        // if no line ends close to prev_point, see if we can find a point on a line that could be the start of a chain of lines
         if (best_line_idx == -1 && have_chains)
         {
             have_chains = false; // now assume that we don't have any chains and change back to true below if we find any joined line segments
--- a/src/pathOrderOptimizer.cpp
+++ b/src/pathOrderOptimizer.cpp
@@ -334,7 +334,7 @@ inline float LineOrderOptimizer::travelDistance(const Point& p0, const Point& p1
         return vSize2f(p0 - p1);
     }
     CombPath comb_path;
-    if (LinePolygonsCrossings::comb(*combing_boundary, *loc_to_line, p0, p1, comb_path, -40, 0, true))
+    if (LinePolygonsCrossings::comb(*combing_boundary, *loc_to_line, p0, p1, comb_path, -40, 0, false))
     {
         float dist = 0;
         Point last_point = p0;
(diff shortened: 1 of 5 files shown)


smartavionics:
Here's an example of how the original code can go wrong. The last move is a travel move in a NE direction from the lower end of the closest green extrusion line. The sequence of events was the N/S line that is the rightmost in that little group of lines was extruded from the bottom to the top, then the little extrusion from the top of that line that goes in the SW direction was output and then the last travel move in the NE direction. The next line in the sequence goes S, parallel to the others. It does a similar backwards/forwards shuffle at the other end of the long extrusion. And then this repeats for the following lines.

screenshot_2017-09-27_10-30-35

smartavionics:
This now makes a pretty good job of optimising the travel paths for zigzag infill. It's worth trying it out!

Ghostkeeper:
I know that this was one of the things that BagelOrb wanted to have done eventually but never got around to.

Developers, see CURA-4417.

src/pathOrderOptimizer.cpp
```diff
@@ -160,15 +162,23 @@ int PathOrderOptimizer::getRandomPointInPolygon(int poly_idx)
...
 */
-void LineOrderOptimizer::optimize()
+void LineOrderOptimizer::optimize(const Polygons* combing_boundary)
 {
     int gridSize = 5000; // the size of the cells in the hash grid. TODO
     SparsePointGridInclusive<unsigned int> line_bucket_grid(gridSize);
     bool picked[polygons.size()];
     memset(picked, false, sizeof(bool) * polygons.size());/// initialized as falses
+    loc_to_line = nullptr;
+    this->combing_boundary = combing_boundary;
```
LipuFei: I think it makes more sense to set `combinb_boundary` in the constructor.

src/pathOrderOptimizer.cpp
```diff
@@ -160,15 +162,23 @@ int PathOrderOptimizer::getRandomPointInPolygon(int poly_idx)
     return rand() % polygons[poly_idx]->size();
 }
 
+static inline bool pointsAreCoincident(const Point& a, const Point& b)
+{
+    return vSize2(a - b) < 25; // points are closer than 5uM, consider them coincident
```
LipuFei: I also found that this magic number in `infill/ZigzagConnectorProcessor.h` on line 204 `ZigzagConnectorProcessor::addLine()`. I guess this function can be moved to some common place like `utils/magic.h` so we only have one source of magic, although it is not really the work for this PR.

src/pathOrderOptimizer.cpp
smartavionics: OK, done that.

src/pathOrderOptimizer.cpp
smartavionics: I agree.

smartavionics:
Hi, thanks for merging this - I think it does a good job of removing unneeded travel moves when printing zigzag lines.

LipuFei:
Hi @smartavionics Unfortunately we just found some performance issue with this PR... There is a model which takes ages to slice with UM3 default settings.
image

We decide to revert this change. Sorry about that. Could you investigate this problem and make fix for it? Thanks.

smartavionics:
That's a real blow - please make the model available for me to investigate.

LipuFei:
@smartavionics 
[UM3_dee_dee_coaster.zip](https://github.com/Ultimaker/CuraEngine/files/1433904/UM3_dee_dee_coaster.zip) Here is the model.

LipuFei:
For more information: the slicing time also depends on where the model is located. In some places it doesn't take forever, but in some it does.

smartavionics:
I can't load that 3mf, cura dies with:

2017-11-01 11:11:49,433 - ERROR - 3MFReader.ThreeMFWorkspaceReader.preRead [176]: Expecting one machine definition container but got 0
2017-11-01 11:11:49,436 - ERROR - cura.CuraApplication.checkIsValidProjectFile [1436]: Could not check file file:///home/markb/Downloads/UM3_dee_dee_coaster.3mf: Expecting one machine definition container but got 0

smartavionics:
Please just provide the model (STL, OBJ)

LipuFei:
[UM3_dee_dee_coaster_stl.zip](https://github.com/Ultimaker/CuraEngine/files/1433929/UM3_dee_dee_coaster_stl.zip)

smartavionics:
Sorry, I just can't load that model without cura blowing up.

All I can really suggest is that we add a UI option to control this feature. It's far too valuable to throw away just because it's slow on very large models.

LipuFei:
@smartavionics Could you try this one then? This is the original cura project file. 
[UM3_dee_dee_coaster_project.zip](https://github.com/Ultimaker/CuraEngine/files/1433977/UM3_dee_dee_coaster_project.zip)

smartavionics:
Please put this back into the master branch as I have new code that improves the performance and although it does still take longer to slice, the potential benefit of very much reduced travel is worth the extra time taken to slice.

If you can't reinstate this, let me know and I will create a new PR from scratch.

ChrisTerBeke:
It's easier to create a new PR as more things have changed in master and we should be able to revert this entire change again.

LipuFei:
@ChrisTerBeke has a point. @smartavionics Could you make a PR with my revert reverted + your improvements? Thanks

smartavionics:
@LipuFei you really made life difficult for me by just reverting that code as I could no longer merge my development branch into master. Anyway, I have created a new branch for you to consider, it is #642.

PR #614, opened 2017-09-27
In addLine(), also test the vector length and reject zags < 5 uM long.

smartavionics:
Or are they zigs? Anyway, the point equality test is OK as far as it goes but lines of
lengths < 5uM were still being generated which would subsequently cause wacky travels to
far flung regions to print what are essentially, zero length extrusions. This change stops
those useless travels.

### DIFF
--- a/src/infill/ZigzagConnectorProcessor.h
+++ b/src/infill/ZigzagConnectorProcessor.h
@@ -201,8 +201,9 @@ inline void ZigzagConnectorProcessor::reset()
 
 inline void ZigzagConnectorProcessor::addLine(Point from, Point to)
 {
-    if (from == to)
+    if (from == to || vSize2(from - to) < 25)
     {
+        // don't add lines less than 5uM long
         return;
     }
     result.addLine(rotation_matrix.unapply(from), rotation_matrix.unapply(to));


Ghostkeeper:
I can't reproduce the bug that this is trying to solve. Do you know of a model and some settings with which this can happen?

smartavionics:
Hi. OK, when I slice this model with 90% zigzag infill with infill angles [45, 0, 135, 0] and infill linewidth 0.8, I get a whole bunch of short zags (mostly 1uM, some a little longer) trying to be added. Just put a debug print in ZigzagConnectorProcessor::addLine() and you should see them.

[bracket_bigheli.stl](https://www.dropbox.com/s/q926bot6nap33vn/bracket_bigheli.stl?raw=1)

Ghostkeeper:
Indeed, with that bracket it reproduces already by just setting it to zigzag, even without 90% infill, infill angles, etc. I kept the rest at default for Ultimaker 3 printers.

smartavionics:
Those are the infill settings I normally use for that part. So the short zags must be being created due to the geometry of the model rather than the infill settings as such. BTW - that part is the reason why I implemented the infill angles feature so that I could make the infill sequence be \ | / | \ | / ...

Ghostkeeper:
Developers, see issue CURA-4407.

diegopradogesto:
I'm reviewing this and I think it's a small change. However it's a patch that removes small lines, but in place it set tiny travels if I'm not wrong.
In my opinion it would be better to eliminate the second point (the end of the line) in case there are two points very close. I'm working on it.

smartavionics:
The travels are generated as the lines are being output by LayerPlan::addLinesByOptimizer() so what travels are created depends on the order the lines are output. Removing a small line segment does not guarantee that a small move will take its place.

smartavionics:
Adding the following removes a lot (but not all) of the short lines:
````
@@ -373,6 +373,10 @@ void Infill::generateLinearBasedInfill(const int outline_offset, Polygons& resul
         for(unsigned int point_idx = 0; point_idx < poly.size(); point_idx++)
         {
             Point p1 = poly[point_idx];
+            if (vSize2(p1 - p0) < 25) {
+                p0 = p1;
+                continue;
+            }
             if (p1.X == p0.X)
             {
                 zigzag_connector_processor.registerVertex(p1); 
````
That simply ignores points in the infill outline polygon that are too close to their predecessor. I should imagine that the other short lines are being created when a scan line hits the polygon very close to one of it's points?

diegopradogesto:
I will merge the branch today. I don't think it's the best solution, but it's OK to solve these small lines.

smartavionics:
Thanks. But you know, it's not such a terrible solution because it catches any short lines that have been created for whatever reason. In my experiments, I tried ignoring short segments in the infill outline polygon and that removed most but not all of the short lines that were created and I was left with a small number of short lines that I could not find what was generating them. So it seems to me that if you try to zap the short lines at the point that they are being created then there's always the possibility that some will get through due to a different reason for their creation. So my approach is let the short lines be created and then throw them away before actually turning them into plastic!

PR #616, opened 2017-09-30
Reduce distance nozzle is moved away from outer walls after each wall is completed

smartavionics:
This is a tweak to the wall optimisation code - not critical.

It was moving 2 x the width of the outer wall but 1.1 x is sufficient.

### DIFF
--- a/src/InsetOrderOptimizer.cpp
+++ b/src/InsetOrderOptimizer.cpp
@@ -25,8 +25,8 @@ void InsetOrderOptimizer::moveInside()
 {
     const coord_t outer_wall_line_width = mesh_config.inset0_config.getLineWidth();
     Point p = gcode_layer.getLastPlannedPositionOrStartingPosition();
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Should we instead use the `machine_nozzle_outer_diameter` setting to determine this? I imagine it's dependent on the shape of the nozzle.

smartavionics:
I'm thinking that there's a number of factors that decide on how big a blob will be left behind on retraction. Retraction speed/distance must play a part but also the line width and other intangible qualities like how much extra extrude you get at the end of a line segment due to lag in the extruder. So I think that almost any small value that gives nice looking results is going to be OK. Currently, I am happy using 1.1 x outer line width which is probably still quite generous.

Part of the move code checks that the move would not end up closer than an outer line width to the centre of the outer line. It does that so that if you have a narrow region and move inside from one edge you don't end up going over the other nearby edge.

PR #618, opened 2017-09-30
Some simple tweaks to improve the combing path

smartavionics:
Sometimes, the combing algorithm will produce a maximal length path that hugs the part
boundary when it could, in fact, be shorter and still be inside the part. For example,
Imagine a part shaped like the outline of a hand. The combing algorithm is quite happy to
create a path that goes up and down the fingers when travelling from one side of the palm
to the other when, perhaps, it could cut across the base of each finger.

A real optimizer would detect the excess travel and trim it to the minimum required by
"pulling tight" the path. This commit tries to achieve a similar result but with no claims
to being optimal. It's relatively cheap (computationally) and cheerful (can produce good
results).

What it does is look back at the sequence of points that have been inserted into the comb
path and see if they can be removed ("short-circuited") without causing the path to cross
the comb boundary. So paths that hug a concave boundary will be shortened. They can even be
shortened when they go around holes but that is not guaranteed to occur.

### DIFF
--- a/src/pathPlanning/LinePolygonsCrossings.cpp
+++ b/src/pathPlanning/LinePolygonsCrossings.cpp
@@ -215,6 +215,40 @@ bool LinePolygonsCrossings::optimizePath(CombPath& comb_path, CombPath& optimize
                 }
             }
             optimized_comb_path.push_back(comb_path[point_idx - 1]);
+
+            if (point_idx == 1)
+            {
+                // we have just inserted the first point in the path that touches the boundary
+                // and as the first line in the comb is pointing from the start point towards the
+                // end point, we can maybe save some travel by replacing this first point with another
+                // that also lies on the second path of the comb, like this:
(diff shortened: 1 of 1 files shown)


smartavionics:
Here's an example of what this PR can do:

Before:
screenshot_2017-10-01_10-23-45

After:
screenshot_2017-10-01_10-24-47

Ghostkeeper:
Developers, see issue CURA-4439.

alekseisasin:
Hi smartavionics, I overviewed your commit and it looks fine, as I understood you try to cut "corners" in combined path making path shorter. I did not try to print with "short-circuited" G-code, but I think it should be fine. 
Thanks.

smartavionics:
Hi, thanks for that. I am planning on taking a good look at the combing code when I have some time as it is far from optimal in terms of the routes produced at the moment.

PR #620, opened 2017-10-07
slicing mode

BagelOrb:
Options to determine how to slice a layer.

One of the reasons printed parts don't fit is because currently the middle of a layer is taken for the area of that layer, while that could mean half of the layer is sticking out of the model.

Exclusive:
[UM3N_cone_to_cylinder.3mf.txt](https://github.com/Ultimaker/CuraEngine/files/1364572/UM3N_cone_to_cylinder.3mf.txt)
image

Inclusive:
image

### DIFF
--- a/src/FffPolygonGenerator.cpp
+++ b/src/FffPolygonGenerator.cpp
@@ -89,8 +89,7 @@ bool FffPolygonGenerator::sliceModel(MeshGroup* meshgroup, TimeKeeper& timeKeepe
         logError("Layer height %i is disallowed.\n", layer_thickness);
         return false;
     }
-    int initial_slice_z = initial_layer_thickness - layer_thickness / 2;
-    int slice_layer_count = (storage.model_max.z - initial_slice_z) / layer_thickness + 1;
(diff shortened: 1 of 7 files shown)


ChrisTerBeke:
Very cool, I actually did a small experiment with this about a year ago to see if we could make screw holes fit better.

BagelOrb:
What was the result of your tests?

ChrisTerBeke:
If you want your screws etc to fit, use exclusive mode. If you then design without margin (like injection molded parts would have), it fits quite nicely (ofc not perfect, but a lot better than using the default 'middle' model and add the margins in your design).

Ghostkeeper:
Developers, see issue CURA-4459.

alekseisasin:
I think the name for the setting "SlicingMode" is very general. I propose:
-> SlicingModeCompensation
-> SlicingModeCorrection

ChrisTerBeke:
or "SlicingTolerance"?

BagelOrb:
I prefer Slicing Tolerance. It really shows the user what you would use this for.

ChrisTerBeke:
Not seeing any crashes anymore after latest changes.

PR #623, opened 2017-10-10
Bug fix - array index was checked against size of wrong array.

smartavionics:
This fixes a crash that can occur with certain models/profiles when wall order optimization is enabled - please merge ASAP!

For example, https://www.thingiverse.com/thing:1175684 cannot be sliced using the vanilla UM3 profile with optimization enabled without this fix.


Appesteijn:
Devs: CURA-4426

smartavionics:
Thanks for merging that. Don't forget that it needs to get into 3.0 as well!

LipuFei:
@smartavionics Also cherry-picked to 3.0.

smartavionics:
Thanks, I did the same in a new PR which I shall now delete as it is redundant.

PR #624, opened 2017-10-11
Inset order fixes for 3.0

smartavionics:
These are the same commits as those recently merged into the master branch so all you have to do is merge this into 3.0!


smartavionics:
No longer required as changes already have been merged.

PR #626, opened 2017-10-12
trihex infill

BagelOrb:
implementation to frontend PR: https://github.com/Ultimaker/Cura/pull/2591


BagelOrb:
Grid infill can shear easily, ie making "#" italic ;)

BagelOrb:
Comparing trihex against hexagon infill isn't fair. Trihex is a lot more
like triangle than like hexagon. We don't even have hexagonal infill.

Hexagonal infill is good if you want low stiffness, which means the object
can absorb a blow.

Trihex or triangle infill is very stiff, which means the objects don't
deform much under a given force.

Ghostkeeper:
Developers, see ticket CURA-4476.

PR #629, opened 2017-10-22
Join points whose distance is less than a threshold in ZigZag pattern

diegopradogesto:
Related with PR #614 from @smartavionics, the proposal here is to join the points whose distance is less than a threshold distance (for instance 5 microns), instead of deleting these tiny lines.
It can avoid creating those tiny lines or travels and create a continuous movement with no gaps in zigzag pattern.


Ghostkeeper:
Developers, see issue CURA-4512.

src/infill/ZigzagConnectorProcessor.cpp
```diff
@@ -134,21 +134,38 @@ void ZigzagConnectorProcessor::registerPolyFinished()
...
+        Point* to = &points[points.size() - 1];
+        if (!mergePointsByThreshold(from, to))
+            addLine(*from, *to);
     }
 }
+
+
+bool ZigzagConnectorProcessor::mergePointsByThreshold(Point* first_point, Point* second_point, int threshold)
+{
+    if (vSize2(*first_point - *second_point) < threshold*threshold)
```
alekseisasin: Some description, why do we check? threshold*threshold

smartavionics: Because vSize2() returns the square of length of the vector. We could use vSize() which returns the length of the vector but it's quicker to do a multiply than a square root.

smartavionics: I think it's fair to say that using squared distances in the engine code is well known technique and so I would not expect this to require a comment.

diegopradogesto: Yes, as Mark pointed, this is for avoid calculating the square root, and because it was the same strategy followed in PR #614.
I can add a comment anyway.

alekseisasin: Now it is clear, Personally, I try to follow the strategy to  comment "Why, I do check".

smartavionics:
It would be great if this could make it into 3.1, thanks.

smartavionics:
It would be great if this could make it into 3.1, thanks.

src/infill.h
```diff
@@ -87,6 +89,7 @@ class Infill
     , zag_skip_count(zag_skip_count)
     , apply_pockets_alternatingly(apply_pockets_alternatingly)
     , pocket_size(pocket_size)
+    , maximum_resolution(maximum_resolution)
```
alekseisasin: In infill the "maximum_resolution" is used only fro creating ZigzagConnectorProcessor, maybe it is better to define this variable in ZigzagConnectorProcessor class by default, for now it will help to prevent the number update from different places.

src/FffGcodeWriter.cpp
```diff
@@ -619,14 +644,28 @@ void FffGcodeWriter::processRaft(const SliceDataStorage& storage)
...
-        raft_outline_path.simplify(); //Remove those micron-movements.
+        raft_outline_path.simplify(maximum_resolution); //Remove those micron-movements.
         constexpr coord_t infill_outline_width = 0;
         Polygons raft_lines;
         int offset_from_poly_outline = 0;
         double fill_angle = 90 * raftSurfaceLayer;
         constexpr bool zig_zaggify_infill = true;
-        Infill infill_comp(EFillMethod::ZIG_ZAG, zig_zaggify_infill, raft_outline_path, offset_from_poly_outline, infill_outline_width, train->getSettingInMicrons("raft_surface_line_spacing"), fill_overlap, fill_angle, z, extra_infill_shift);
+
+        const Point& infill_origin = Point();
```
alekseisasin: Is necessary initialize an object? In other places in same class it is empty.

LipuFei: Although `infill_origin` is an optional parameter, it is not the last in the list. The code later needs to give other optional parameters that are after `infill_origin`, so it is still needed.

src/infill/ZigzagConnectorProcessor.cpp
```diff
@@ -134,21 +134,38 @@ void ZigzagConnectorProcessor::registerPolyFinished()
...
+                addLine(*from, *to);
         }
     }
     // only add the last line if:
     //  - it is not an end piece, or
     //  - it is an end piece and "connected end pieces" is enabled
     if ((!is_endpiece || (is_endpiece && this->connected_endpieces)) && points.size() >= 2)
     {
-        addLine(points[points.size() - 2], points[points.size() - 1]);
+        Point* from = &points[points.size() - 2];
```
LipuFei: Maybe this can be done in `addLine()` since this code and above are the same.

LipuFei:
I have added some more documentation and done a few refactoring.

PR #631, opened 2017-10-23
Bug fix - avoid unprintable roof skin areas

smartavionics:
This PR fixes https://github.com/Ultimaker/Cura/issues/2656.

### DIFF
--- a/src/skin.cpp
+++ b/src/skin.cpp
@@ -463,8 +463,15 @@ void SkinInfillAreaComputation::generateRoofing(SliceLayerPart& part)
                 // to be wholely or partly above air and it may not be printable so restrict
                 // the regions that have air above (the visibile regions) to not include any area that
                 // has air below (fixes https://github.com/Ultimaker/Cura/issues/2656)
-                Polygons no_air_below = getWalls(part, layer_nr - 1, wall_idx);
-                no_air_above = no_air_above.intersection(no_air_below);
+
+                // set no_air_below to the skin area for the current layer that does not have air below it
+                Polygons no_air_below = getWalls(part, layer_nr - 1, wall_idx).intersection(getWalls(part, layer_nr, wall_idx));
+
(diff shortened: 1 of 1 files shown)


smartavionics:
Oops! discovered a problem with this PR - will close for now and reopen when (if) I fix the problem...

smartavionics:
Back again!

So, this now works well inasmuch that it fulfils the original objective of fixing the problem where if top surface skin layers is non-zero and there is a hole covered by a single solid layer which then has another hole on top (like the example in the above issue where the OP is printing a countersunk hole) then the solid layer is not printed as one region but, instead, is printed as an inner region of "roof" surrounded by an outer region of normal skin. The problem is that the inner roof region cannot be printed on air (remember, it's above a hole). Here's an example of the problem, the inner region that is filled with horizontal lines is the roof region and it's surrounded by the non-roof skin. Unfortunately, it's all above a hole (which is of the same diameter as the visible hole to the left).

screenshot_2017-10-23_23-32-57

This next image shows what you get with this PR. The skin inside the two new walls is diagonal because it isn't roof. This is showing the layer above compared to the last image because I wanted to make the walls that are above the region of interest visible.

screenshot_2017-10-23_23-36-37

This PR fixes this problem by not allowing skin to be treated as roof when it is immediately above a hole.
The one known problem is that if the hole stops just 1 layer below the surface then the region immediately above the hole will not be roof and so is printed as a separate region. Look at this example, on the left side of the test object there is a hole that stops 1 layer down from the roof above and you can see the outline of the hole. If the hole stopped 2 layers down, you don't get the outline.

screenshot_2017-10-23_23-28-15

So is it better to keep the original behaviour which produces unprintable holes or is it better to have the behaviour produced by this PR which stops the unprintable holes being created but has the above described problem...?

Ghostkeeper:
Developers, see issue CURA-4532.

PR #635, opened 2017-10-28
In generateInsets(), if the last inset disappears try again using 99% of the line width.

smartavionics:
This makes it more likely that walls whose width is exactly the sum of the constituent wall
widths - e.g. a wall 1.5mm wide with line widths of 0.5mm will have all the walls
generated rather than the middle one getting missed out which often happened before.

### DIFF
--- a/src/WallsComputation.cpp
+++ b/src/WallsComputation.cpp
@@ -36,10 +36,23 @@ void WallsComputation::generateInsets(SliceLayerPart* part)
             part->insets[0] = part->outline.offset(-line_width_0 / 2 - wall_0_inset);
         } else if (i == 1)
         {
-            part->insets[1] = part->insets[0].offset(-line_width_0 / 2 + wall_0_inset - line_width_x / 2);
+            const int offset = -line_width_0 / 2 + wall_0_inset - line_width_x / 2;
+            part->insets[1] = part->insets[0].offset(offset);
+            if (part->insets[1].size() == 0)
+            {
+                // if a slightly thinner line would fit, allow it as the overlap is so small
+                // this helps when you have parts such as tubes whose walls are an exact number of line widths wide
+                // e.g. when the tube wall is 1.5mm wide and the line widths are 0.5mm
+                part->insets[1] = part->insets[0].offset(offset * 0.99f);
+            }
         } else
         {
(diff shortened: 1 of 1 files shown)


smartavionics:
Test tube!
````
$fn = 50;

difference() {
  cylinder(d = 20, h = 10);
  cylinder(d = 17, h = 10);
}
````

Before:
screenshot_2017-10-28_16-05-45

After:
screenshot_2017-10-28_16-06-25

smartavionics:
I realise that the number of vertices ($fn in the above example) influences what happens here. As the number of vertices reduces, the inner and outer lines get closer together and the middle line gets "squeezed out". So it could be argued that this fix isn't required and people should just get their model right (i.e. as the number of vertices is reduced they need to increase the gap between the inner and outer walls). However, by allowing the middle line to fit in a very slightly undersize gap it does ease the burden on the user. Without this mod, even if the number of vertices was very high the middle line would not fit in the gap but I think that people expect to be able to have structures such as the above example that will print correctly without them having to frig around modifying the model (or the line widths).

smartavionics:
As I mention in the comment in the last commit. You really need to have overlap compensation enabled when the width of the innermost region is less than 2 times the line width because even without this PR you will get over extrusion in that situation. As the region gets narrower, the walls get closer together and will overlap more and more.

Ghostkeeper:
Developers, see issue CURA-4535.

ianpaschal:
I've discussed this with @LipuFei and @Ghostkeeper.

We feel that this change should not be implemented because:

1. It relies on picking arbitrary numbers instead of heuristics. You value of 0.95 may work in this case, but why 5%? Why not 10%? Etc.
2. As an extension of this, it's only an effective solution in very specific cases, such as yours. If the gap is not uniform thickness a 5% overlap value may work near one corner, but not at the next corner. Another example is that depending on the mesh resolution you export at, the amount of overlap needs for a circle/tube with a given radius would have to be recomputed. For irregular shapes, it may not even be possible to recompute.

Better solutions to this would be either manually changing line widths to work, or adjusting the model slightly to adapt to the manufacturing process (i.e. make your tube walls 1.525 mm instead of 1.500 mm).

PR #636, opened 2017-10-28
When z-seam is located on sharpest corner align z-seam of single hole with outline z-seam.

smartavionics:
When printing parts that have a single hole that is close to the outline (e.g. a tube with
thin walls), it's nicer to have the z-seam of the hole aligned with the z-seam of the outer
wall. This happens for free when the z-seam type is nearest or user-specified but when
the z-seam type is sharpest corner, the z-seams for the hole and the outside wall don't
have to coincide. This mod handles that scenario and forces the z-seam of the hole to match
the z-seam of the outer wall.
Compare, without this mod:
screenshot_2017-10-28_15-41-28

and, with this mod:
screenshot_2017-10-28_15-41-09

### DIFF
--- a/src/InsetOrderOptimizer.cpp
+++ b/src/InsetOrderOptimizer.cpp
@@ -56,6 +56,7 @@ void InsetOrderOptimizer::processHoleInsets()
 {
     const coord_t wall_line_width_0 = mesh_config.inset0_config.getLineWidth();
     const coord_t wall_line_width_x = mesh_config.insetX_config.getLineWidth();
+    const coord_t max_gap = std::max(wall_line_width_0, wall_line_width_x) * 1.1f; // if polys are closer than this, they are considered adjacent
     const coord_t wall_0_wipe_dist = mesh.getSettingInMicrons("wall_0_wipe_dist");
     const bool retract_before_outer_wall = mesh.getSettingBoolean("travel_retract_before_outer_wall");
     const bool outer_inset_first = mesh.getSettingBoolean("outer_inset_first")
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
What happens when your example above has a dent somewhere in the outer wall and another dent somewhere in the inner wall but at a different position? Looking briefly at the code, it would move them together, defeating the setting of sharpest corner for one of the seams.

In my opinion this should only kick in if the corners all have the same angle, such that which corner is sharpest is ambiguous. And I'm not sure if that slim use case is worth the processing power.

Ghostkeeper:
Developers, see issue CURA-4534.

smartavionics:
> What happens when your example above has a dent somewhere in the outside wall and another dent somewhere in the inside wall but at a different position? Looking briefly at the code, it would move them together, defeating the setting of sharpest corner for one of the seams.

That's why it requires the walls to be pretty close together, if the walls are spaced far enough apart to allow appreciable dents, this tweak should not be triggered (that's the hope!)

alekseisasin:
@smartavionics , Hi, I tried to make a different models to get result which you showed on second image, I only see that infill and inner wall start apth close to each other, but the outer wall starts is not, Did try with different diameters.

smartavionics:
Hi, if the wall of the part are thick enough to have infill then the seams will not be aligned. This PR is really about stopping an unnecessary retract when printing tubes whose walls are thin.

LipuFei:
Hi @smartavionics , I tried to reproduce the result you have *without* your mod, but I can only get something like this:
image
There are many travel moves between the walls and there are only walls.

I merged with the latest master. Maybe some code in master broke something?

smartavionics:
You do need to have the wall order optimization enabled? I get the same as you do if I disable it.

PR #640, opened 2017-11-01
Fix start extruder heatup

fieldOfView:
This PR changes the temperature commands inserted before the start gcode if the start gcode does not contain heatup commands. For multiextrusion machines, it inserts a T command to activate the first used extruder. Only the first used extruder is heated up to printing temperatures, the other extruders used in the print are heated up to the standby temperature. Unused extruders are no longer heated up.

Griffin-based printers (such as the UM3) are not affected.

This PR fixes https://github.com/Ultimaker/Cura/issues/1819

I have tested the PR using a Custom FDM Printer with 4 extruders by slicing multiple objects printed with either of these extruders, with and without buildplate adhesion. All seems to behave correctly.


nallath:
CURA-4530

PR #641, opened 2017-11-01
Speedup the line order optimizer

smartavionics:
This PR requires that 82417aa45029871ac6aaf8b8c297d9aa161e7b3f is reverted so that the code that this PR modifies is back in place. Why did you just remove it without giving me a chance to come up with a solution? That was really very unfriendly.


Ghostkeeper:
Did you intend to close your pull request? You restored the branch but the PR didn't re-open.

We should've asked you for a fix. However ultimately it's our job to maintain the code base and we're not supposed to spend too much of Ultimaker's time in external contributions since they go outside of our planning. So we opted to revert the PR that was causing the issue instead of spending the time to fix it. It would've been good to notify you.

smartavionics:
I have already closed this PR (6 days ago it says above).

smartavionics:
This PR has been replaced with #642

PR #645, opened 2017-11-06
Now ensureInsideOrOutside() returns same value whether DEBUG is defined or not

smartavionics:
Before, when DEBUG was defined, after outputting some diagnostics it returned a default
constructed ClosestPolygonPoint() and if DEBUG was not defined it went on to do some other
processing.

By always returning ClosestPolygonPoint() irrespective of whether DEBUG is defined, the
function is more consistent.

This fixes the issue reported in https://github.com/Ultimaker/Cura/issues/2704.


Ghostkeeper:
This has got to be one of the more interesting bugs I've seen. Good find. That would've been hard to debug!

smartavionics:
We are lucky that someone managed to produce model/settings that triggered the problem and reported it and was willing to provide a project.

PR #649, opened 2017-11-09
Adaptive layer heights

ChrisTerBeke:
Adaptive layer heights is a new experimental feature that changes the layer height depending on mesh shape. If a layer section contains steeper angles on the outer wall, the layers will be thinner. For sections that are straight, the layers will be thicker. This is all done automatically.

To achieve this, the following changes were made:
* Added a boolean setting to enable adaptive layer heights.
* Added 3 parameters to influence the behaviour of adaptive layers while in experimental mode (materials & processing department can use these to build new profiles).
* The Gcode writer now uses the thickness and z position stored per layer instead of re-calculating again.
* The objects AdaptiveLayerHeights and AdapativeLayer were added to store the information and implement the algorithm.
* Some changes were made to FFFGcodeProcessor and Slicer to allow the implementation of the adaptive layer heights. These also clean up the code a bit and remove some duplication.

Known limitations:
* Only uses the 1st mesh now for calculating the layer heights.
* Print quality might vary on settings as the feedrate needs time to compensate when switching heights (hence the gradual change of heights per "variation step size"). Final values will be tweaked later by materials & processing.
* Not tested in combination with support and other special modes.

A test was done before and after these changes to make sure the outputted G-code does not change when slicing with these changes while adaptive layers is disabled (aka no regressions).

Bonus screenshot:

adaptive_layer_heights_example_1


nickthetait:
How was the no-regressions test case conducted?

ChrisTerBeke:
@nickthetait by slicing the same object with the same settings before and after the changes, then literally comparing the G-code texts in a diff checker.

nickthetait:
Nice. Seems kinda obvious now haha :smile:

PR #650, opened 2017-11-09
In addConnectingTravelMove() also force a retraction if retract_at_layer_change is true.

smartavionics:
See https://ultimaker.com/en/community/51226-rectract-before-outer-wall-not-working-zscar-issues


Ghostkeeper:
Looks good. Developers, see issue CURA-4569.

PR #652, opened 2017-11-14
Improve the order that lines are output to minimize travel (take 3)

smartavionics:
OK, this PR is back up and running again. I have been working to improve the efficiency so that it will produce good quality ordering without taking forever when the part outlines are complex. I believe the cost in CPU time is now acceptable given the potential savings in print times that good ordering can achieve.


Ghostkeeper:
Developers, see issue CURA-4418 (recycled from the previous PRs).

src/pathOrderOptimizer.cpp
```diff
@@ -249,24 +321,74 @@ void LineOrderOptimizer::optimize()
...
+{
+    // when looking for chains, just_point will be either 0 or 1 depending on which vertex we are currently interested in testing
+    // if just_point is -1, it means that we are not looking for chains and we will test both vertices to see if either is best
+
     const Point& p0 = (*polygons[poly_idx])[0];
     const Point& p1 = (*polygons[poly_idx])[1];
-    float dot_score = getAngleScore(incoming_perpundicular_normal, p0, p1);
+    float dot_score = (just_point >= 0) ? 0 : getAngleScore(incoming_perpundicular_normal, p0, p1);
+
+    if (just_point != 1)
```
LipuFei: This part can be slightly simplified (in a loop for example) so we don't need the duplicate the code, but I'm not sure how good its readability will be.

smartavionics: Hi, sorry, I am not exactly sure to which part of the code you are referring to there. If you want something changed, please rephrase your requirement, thanks.

jackha:
We found out that on some models this change actually takes a lot of time (i.e. from 8 seconds before to 80 seconds). See chainmail_coaster. @smartavionics if you have any ideas, you're welcome to share them :-) In the meanwhile, I'm looking at it.

[chainmail_coaster.stl.zip](https://github.com/Ultimaker/CuraEngine/files/1559515/chainmail_coaster.stl.zip)

smartavionics:
I can slice that with 0.1mm layers in less than 5 seconds. What machine/settings are you using?

smartavionics:
The setting that makes the difference is "fill gaps between walls" with that set to none, the speed is OK. I shall investigate what it is doing that takes so long...

jackha:
I'll investigate as well... "fill gaps between walls" indeed makes a huge difference.

On UM3 / AA 0.25 it's oddly enough a lot faster (but still slow, 24+ seconds) than UM3 / AA 0.4  (300+ seconds).

Turn off fill gaps between walls:
UM3 / AA 0.25: 20 seconds
UM3 / AA 0.4: 20 seconds

smartavionics:
I have to go out now but will work on this, this evening. I think it will be possible to improve the performance for this type of situation.

jackha:
Would be great! Thanks in advance, after today I will not be working on it till monday.

smartavionics:
Hi Jack, see #668.

PR #653, opened 2017-11-14
Bug fix - use floats to avoid loss of precision when determining sharpest corner.

smartavionics:
Fixes the issue discussed in https://ultimaker.com/en/community/51970-sharpest-corner-hide-seam-improvement-needed

### DIFF
--- a/src/pathOrderOptimizer.cpp
+++ b/src/pathOrderOptimizer.cpp
@@ -102,9 +102,9 @@ int PathOrderOptimizer::getClosestPointInPolygon(Point prev_point, int poly_idx)
         const Point& p1 = poly[point_idx];
         const Point& p2 = poly[(point_idx + 1) % poly.size()];
         // when type is SHARPEST_CORNER, actual distance is ignored, we use a fixed distance and decision is based on curvature only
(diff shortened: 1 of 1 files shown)


smartavionics:
It would be great if this could make it into 3.1, thanks.

smartavionics:
Hi @ChrisTerBeke, @Ghostkeeper this really needs to go in 3.1 - please process this ASAP!

ChrisTerBeke:
And must many other things... I don't think this will be added in the Beta, but maybe for the stable release depending on priorities, @Appesteijn ?

smartavionics:
I just don't understand how you guys decide priorities. You only ever seem to be concerned with adding fancy features rather than producing high quality gcode. I am beginning to understand why customers get pissed off with Ultimaker.

Appesteijn:
You don't understand or you don't agree? We are not only adding new features but also fixing a lot of bugs in every sprint. We are a little behind on the pull requests and working hard to catch up, but we do need to review and test every pull request.

Appesteijn:
Devs: CURA-4616

PR #655, opened 2017-11-15
Tree support

Ghostkeeper:
It's work in progress. See below.

screenshot from 2017-11-15 17-59-08

This is an algorithm to create a tree-like structure to support your models. The trees are specifically tuned to work well with FDM printing rather than similar sort of structures that are intended for SLA.

The basic idea of this algorithm is the following steps:

* Generate a load of contact points on the surface. These contact points are generated on a grid that is spaced such that the skin lines are all supported evenly when using the default settings. If the grid points have all missed the overhang area, then one additional point is added to support small pieces of overhang pointing downwards.
* On a layer, connect the points with a minimum spanning tree.
* Move the points to the next layer, but slightly move them. First they are slightly moved towards each other, so that the branches come together as you get to the lower layers. This relaxation sort of brings the points onto one line so they form a nice thick branch. Then they are also moved out of any collision area. The way that these collision areas are computed is quite complex, as they need to take into account the various radii that the tree's branches can take on. The computation of the collision areas is also the most significant part of the processing time.
* If branches join together, one of the nodes is removed. The node's metadata is resolved in some way. For instance, each node has a diameter that is dependent on the node's distance to its highest peak. This distance is stored with the node and upon merging, the highest distance needs to be chosen.
* A circle is drawn around every node on every layer with a certain width and shape. In the tips, this shape is an ellipsis that is oriented perpendicular to the skin lines that the branch tip supports. In the rest of the branch, it's just a circle with a gradually increasing radius (to make it stable).
* The circles are all joined together to form a layer of support. For support roofs, the circles are joined separately to form the roof layer. This layer is then filled with the normal support infill algorithms.

There's still some TODOs that need to be resolved before merging this:
* You get a crash when combining normal support with tree support.
* The support is not meant to touch the model, so it's basically always "Build plate only". For this reason also, support bottom is not implemented.
* If the support can't reach the build plate, it's left hanging in mid-air. This needs to be solved by linking each tree node to the tree nodes it was derived from in higher layers. Then deleting the tree node can delete the higher tree nodes as well.
* The main function in TreeSupport.cpp is a bit long and should be split up into constituents.


nickthetait:
Just curious, how many hours have you invested in this so far?

Ghostkeeper:
It was about two weeks work. We had a research sprint last two weeks where we could make a bit of a tech push.
According to my log, this took 67 hours and 55 minutes.

thopiekar:
Great! Told you, that you will manage, when I visited you at your desk! 😄

src/TreeSupport.cpp
```diff
@@ -0,0 +1,250 @@
...
+//CuraEngine is released under the terms of the AGPLv3 or higher.
+
+#include "utils/intpoint.h" //To normalize vectors.
+#include "utils/math.h" //For round_up_divide and PI.
+#include "utils/MinimumSpanningTree.h" //For connecting the correct nodes together to form an efficient tree.
+#include "utils/polygonUtils.h" //For moveInside.
+
+#include "TreeSupport.h"
+
+#define SQRT_2 1.4142135623730950488 //Square root of 2.
```
LipuFei: This can be a `const` and moved to some common header file.

src/utils/polygonUtils.cpp
```diff
@@ -329,6 +329,122 @@ unsigned int PolygonUtils::moveInside(const Polygons& polygons, Point& from, int
     return NO_INDEX;
 }
 
+//Version that works on single PolygonRef.
+unsigned int PolygonUtils::moveInside(const ConstPolygonRef polygon, Point& from, int distance, int64_t maxDist2)
```
LipuFei: Can this somehow be merged with the `moveInside` above?

src/TreeSupport.cpp
Ghostkeeper: It's already a preprocessor macro. It can't be made const.

I'm not really in favour of creating generic common header files because they tend to become a cesspool of stuff since there is no clearly defined scope for these header files. It's not like we need to adjust the value of SQRT_2 that often, so I'd rather keep it close to where it's needed.

src/utils/polygonUtils.cpp
Ghostkeeper: It's kind of difficult. The reason is documented down below.

src/TreeSupport.cpp
```diff
@@ -0,0 +1,468 @@
...
+    model_collision.resize((size_t)std::round((float)maximum_radius / radius_sample_resolution) + 1);
+
+    const coord_t xy_distance = storage.getSettingInMicrons("support_xy_distance");
+    constexpr bool include_helper_parts = false;
+    size_t completed = 0; //To track progress in a multi-threaded environment.
+#pragma omp parallel for shared(model_collision, storage) schedule(dynamic)
+    for (size_t radius_sample = 0; radius_sample < model_collision.size(); radius_sample++)
+    {
+        const coord_t diameter = radius_sample * radius_sample_resolution;
+        model_collision[radius_sample].push_back(storage.getLayerOutlines(0, include_helper_parts).offset(xy_distance + diameter, ClipperLib::JoinType::jtRound));
```
jackha: Isn't the same line (with 0 replaced with layer_nr) also in the loop below?

src/TreeSupport.cpp
```diff
@@ -0,0 +1,468 @@
...
+
+        for (size_t group_index = 0; group_index < nodes_per_part.size(); group_index++)
+        {
+            const MinimumSpanningTree& mst = spanning_trees[group_index];
+            //In the first pass, merge all leaf nodes.
+            for (std::pair<Point, Node> entry : nodes_per_part[group_index])
+            {
+                const Node node = entry.second;
+                std::vector<Point> neighbours = mst.adjacentNodes(node.position);
+                if (neighbours.size() == 1 && vSize2(neighbours[0] - node.position) < maximum_move_distance * maximum_move_distance) //This leaf is about to collapse. Merge it with the neighbour.
```
jackha: We have to find more ways to lose points. Maybe if a node has 2 neighbors that are really close?
If there are too many points (for instance with a large area or with a small Tree Support Branch Distance), then it currently doesn't really converge to a smaller base.

jackha: Ahhh the nodes were not removed from nodes_per_part, so they got inserted anyway in the loop below!

cura-4523-before2
cura-4523-after2

src/TreeSupport.cpp
```diff
@@ -265,58 +265,71 @@ void TreeSupport::dropNodes(const SliceDataStorage& storage, std::vector<std::un
...
+                            to_delete.insert(neighbour_node);
+                        }
                     }
                 }
             }
             //In the second pass, move all middle nodes.
             for (std::pair<Point, Node> entry : nodes_per_part[group_index])
             {
                 Node node = entry.second;
+                if (to_delete.find(node) != to_delete.end())
```
jackha: Why not delete everything from to_delete from nodes_per_part before entering this loop? Seems (a bit) cleaner to me.

src/TreeSupport.cpp
```diff
@@ -0,0 +1,535 @@
...
+                            to_delete.insert(neighbour_node);
+                        }
+                    }
+                }
+            }
+            //In the second pass, move all middle nodes.
+            for (std::pair<Point, Node> entry : nodes_per_part[group_index])
+            {
+                Node node = entry.second;
+                if (to_delete.find(node) != to_delete.end())
```
jackha: Why not delete everything from to_delete from nodes_per_part before entering this loop? Seems (a bit) cleaner to me.

src/TreeSupport.cpp
Ghostkeeper: Indeed, it looks like this creates one layer too many.

src/TreeSupport.cpp
```diff
@@ -0,0 +1,464 @@
...
+        {
+            Polygons previous_layer = model_avoidance[radius_sample][layer_nr - 1].offset(-maximum_move_distance); //Inset previous layer with maximum_move_distance to allow some movement.
+            previous_layer = previous_layer.unionPolygons(model_collision[radius_sample][layer_nr]);
+            model_avoidance[radius_sample].push_back(previous_layer);
+        }
+#pragma omp atomic
+        completed++;
+#pragma omp critical (progress)
+        {
+            Progress::messageProgress(Progress::Stage::SUPPORT, ((model_avoidance.size() >> 1) + (completed >> 1)) * PROGRESS_WEIGHT_COLLISION, model_avoidance.size() * PROGRESS_WEIGHT_COLLISION + storage.support.supportLayers.size() * PROGRESS_WEIGHT_DROPDOWN + storage.support.supportLayers.size() * PROGRESS_WEIGHT_AREAS);
```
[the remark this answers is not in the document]
Ghostkeeper: Done.

src/TreeSupport.cpp
```diff
@@ -405,7 +405,7 @@ void TreeSupport::generateContactPoints(const SliceMeshStorage& mesh, std::vecto
                 {
                     Point candidate = rotate(Point(x, y), rotate_angle) + bounding_box.min;
                     constexpr bool border_is_inside = true;
-                    if (overhang_part.inside(candidate, border_is_inside))
+                    if (overhang_part.inside(candidate, border_is_inside) and !collision_areas[layer_nr].inside(candidate, border_is_inside))
```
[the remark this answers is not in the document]
Ghostkeeper: I had already fixed the `and` in a later commit.

src/TreeSupport.cpp
Ghostkeeper: We can't modify `nodes_per_part` when we want to delete the nodes because we are at that point iterating over that set. It could be solved by making a third loop in between that goes over all nodes and copies them to a different set if they don't need to be deleted and then modifying the last loop to iterate over the new set, but then I think this is cleaner (and faster).

src/TreeSupport.cpp
Ghostkeeper: You had a fix but didn't push the commit when you went on vacation. In the meanwhile I also fixed it.

src/TreeSupport.cpp
```diff
@@ -0,0 +1,468 @@
...
+                    //Move towards the average position of all neighbours.
+                    Point sum_direction(0, 0);
+                    for (Point neighbour : neighbours)
+                    {
+                        sum_direction += neighbour - node.position;
+                    }
+                    if(vSize2(sum_direction) <= maximum_move_distance * maximum_move_distance) {
+                        next_layer_vertex += sum_direction;
+                    } else {
+                        next_layer_vertex += normal(sum_direction, maximum_move_distance);
```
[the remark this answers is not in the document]
Ghostkeeper: But you forgot to push the commit during your vacation. I've also fixed it now.

src/TreeSupport.cpp
```diff
@@ -0,0 +1,464 @@
...
+        rotated_bounding_box.include(rotate(bounding_box_size, -rotate_angle));
+        rotated_bounding_box.include(rotate(Point(0, bounding_box_size.Y), -rotate_angle));
+        rotated_bounding_box.include(rotate(Point(bounding_box_size.X, 0), -rotate_angle));
+        AABB unrotated_bounding_box;
+        unrotated_bounding_box.include(rotate(rotated_bounding_box.min, rotate_angle));
+        unrotated_bounding_box.include(rotate(rotated_bounding_box.max, rotate_angle));
+        unrotated_bounding_box.include(rotate(Point(rotated_bounding_box.min.X, rotated_bounding_box.max.Y), rotate_angle));
+        unrotated_bounding_box.include(rotate(Point(rotated_bounding_box.max.X, rotated_bounding_box.min.Y), rotate_angle));
+
+        for (const ConstPolygonRef overhang_part : outside_polygons)
```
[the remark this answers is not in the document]
Ghostkeeper: Fixed this with a different approach: To generate the original grid only once per mesh instead of per layer.

src/TreeSupport.cpp
```diff
@@ -0,0 +1,468 @@
...
+        {
+            bool added = false; //Did we add a point this way?
+            for (coord_t x = unrotated_bounding_box.min.X; x <= unrotated_bounding_box.max.X; x += point_spread)
+            {
+                for (coord_t y = unrotated_bounding_box.min.Y; y <= unrotated_bounding_box.max.Y; y += point_spread)
+                {
+                    Point candidate = rotate(Point(x, y), rotate_angle) + bounding_box.min;
+                    constexpr bool border_is_inside = true;
+                    if (overhang_part.inside(candidate, border_is_inside))
```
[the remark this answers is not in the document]
Ghostkeeper: I didn't find the cause of this. This was most likely an edge case, as we determined together. We won't fix this one.

src/TreeSupport.cpp
Ghostkeeper: We can't modify `nodes_per_part` when we want to delete the nodes because we are at that point iterating over that set. It could be solved by making a third loop in between that goes over all nodes and copies them to a different set if they don't need to be deleted and then modifying the last loop to iterate over the new set, but then I think this is cleaner (and faster).

PR #656, opened 2017-11-17
Add setting to determine whether or not to use firmware retractions

Ghostkeeper:
This way a printer definition can determine this themselves independently of the g-code flavour.

See also https://github.com/Ultimaker/Cura/pull/2797

Fixes https://github.com/Ultimaker/Cura/issues/1097

### DIFF
--- a/src/gcodeExport.cpp
+++ b/src/gcodeExport.cpp
@@ -37,6 +37,7 @@ GCodeExport::GCodeExport()
 
     isZHopped = 0;
     setFlavor(EGCodeFlavor::MARLIN);
+    firmware_retract = false;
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Developers, see issue CURA-4579.

dintid:
Does it mean we get it or not? Sorry for being a bit dense regarding how Github functions. :)

ChrisTerBeke:
It's now in master which mean it'll be in the next release branch (unless it's get cherry picked to 3.1). As we're basically in feature-freeze for 3.1 I expect this to go into 3.2 (beta) though.

Issue #661, opened 2017-11-28
Decorate Windows executable using "windres"


thopiekar:
Just a collection of pages I found around this topic.
* https://stackoverflow.com/questions/598633/gcc-on-windows-set-description-field-of-c-executable
* https://stackoverflow.com/questions/3526794/how-do-i-build-a-win32-app-with-a-resource-file-using-cmake-and-mingw
* http://fragglet.livejournal.com/4448.html

PR #663, opened 2017-12-01
Dress up the engine!

thopiekar:
Looks like that:

image

Is a mixture of the Cura icon + Terminal icon from the Breeze project.

Fixes: #661


krop:
You didn't add the license file for the breeze icon you use. Please also contact the copyright holder to check whether he accepts his work to be part of another icon.

thopiekar:
Good point. Only wanted to wait for feedback on this from the Ultimaker developers.

By the way, do you know how to add the licence correctly? Never put things together from other projects like this before.

krop:
The breeze icons tarball contain the license file (called COPYING-ICONS). Since you want to create a derivative work with one of these icons, you also need the author's agreement.

thopiekar:
Ok, thanks. So before this gets merged we have to care about getting either the agreement or designing our own icon.

Or could there already be consequences?

krop:
No, it's WIP. Just don't merge it until everything is sorted

Ghostkeeper:
Since Ultimaker will be publishing this executable and Ultimaker is licensed to use the icon, I don't foresee any licensing problems (as long as it is indeed the original Cura icon).

thopiekar:
He meant the console icon I embedded on top of the Cura icon.

Ghostkeeper:
Ah, I see. I thought that was an addition from your OS or something.

Maybe we can use the normal Cura icon for now and we'll see about getting a new icon for the engine.

thieraufc:
I'm no artist so feel free to shoot this down, but here's something I put together. I modified the existing cura logo found in [the cura repository](https://github.com/Ultimaker/Cura/blob/master/icons/cura-128.png). It's the cura logo with a terminal image in the same style:
cura-128
Because it's modified from an LGPL repository, there shouldn't be any licensing issues there or here.

Ghostkeeper:
I like the idea of that icon. The layout needs some work though. But for now I think we'll keep ourselves to the original PR.

ianpaschal:
Switched icon to use a gear in the C (see below). Discussed and voted upon internally.

icon_1024

PR #665, opened 2017-12-09
If retraction is enabled, do a retract between start gcode and first planned travel.

smartavionics:
Fix for https://github.com/Ultimaker/Cura/issues/2945.

BTW I added a call to resetExtrusionValue() because I noticed that when absolute extrusion is being used, I wasn't getting any G92 output at all which seems pretty dodgy as that would assume that the extruder was zeroed. Not a problem for me as I use relative extrusion.

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -501,11 +501,11 @@ void FffGcodeWriter::processStartingCode(const SliceDataStorage& storage, const 
     }
     if (gcode.getFlavor() != EGCodeFlavor::GRIFFIN)
     {
-        // ensure extruder is zeroed
-        gcode.resetExtrusionValue();
-
         if (getSettingBoolean("retraction_enable"))
         {
+            // ensure extruder is zeroed
+            gcode.resetExtrusionValue();
+
             // retract before first travel move
             gcode.writeRetraction(storage.retraction_config_per_extruder[start_extruder_nr]);
         }
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -499,6 +499,17 @@ void FffGcodeWriter::processStartingCode(const SliceDataStorage& storage, const 
     {
         gcode.writeExtrusionMode(true);
     }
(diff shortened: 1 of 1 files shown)


thopiekar:
Nice! 👍 
Hope someone will be able to take a look on it soon!

Ghostkeeper:
Developers, see issue CURA-4744.

Sorry that it took so long. I'm now 3 weeks behind again :(

diegopradogesto:
So far, it looks ok, but I have one question first: Is there a reason why you skip the printers with Griffin flavor (such as UM3)?

smartavionics:
Hi, well it's a while since this PR was created but I think the reason is that if you look at the lines above the inserted code you can see that there is a section which only gets executed when the flavour is griffin and that section ends with a retraction so for the griffin you don't need another. That make sense, doesn't it?

src/FffGcodeWriter.cpp
```diff
@@ -499,6 +499,17 @@ void FffGcodeWriter::processStartingCode(const SliceDataStorage& storage, const
     {
         gcode.writeExtrusionMode(true);
     }
+    if (gcode.getFlavor() != EGCodeFlavor::GRIFFIN)
+    {
+        // ensure extruder is zeroed
+        gcode.resetExtrusionValue();
```
diegopradogesto: I suggest to execute this line only if the retraction is going to be made. In this case, if the retraction is not enabled, the behavior is the same as before.

smartavionics: Yes, but as I mention above there doesn't appear to be any other G92 output to reset the extruder. If no G92 is output then it assumes the extruder is zeroed by some other means - is this safe? I agree that is how it is currently working so I don't have a big problem with your suggestion.

smartavionics: OK, I did as you asked.

michalka:
From the issue this was created for: "The other slicers draw the line, retract and travel move, then unretract and start the print cleanly."

This implementation never unretracts after the initial travel move of the print (layer 0), so now the print starts with an unprimed nozzle.

PR #668, opened 2017-12-14
Only pass combing boundary to LineOrderOptimizer when lines are infilll (not skin, support, etc.)

smartavionics:
The potential reduction in travel time achieved by passing the combing boundary for
non-infill lines is relatively small but the slicing time can increase greatly (e.g. when
processing the skin created when filling gaps between walls) so only pass the combing boundary
for infill lines where the travel time reduction can be well worth the extra slicing time.

Hi @jackha, this is a good fix because it now only does the expensive computation required to determine the correct travel distance for infill lines where it can make a big difference to the print time. For skin, there's little benefit in doing the extra computation and as we have seen with your example it can really increase the slicing time a lot.

### DIFF
--- a/src/LayerPlan.cpp
+++ b/src/LayerPlan.cpp
@@ -461,9 +461,9 @@ void LayerPlan::addPolygonsByOptimizer(const Polygons& polygons, const GCodePath
 void LayerPlan::addLinesByOptimizer(const Polygons& polygons, const GCodePathConfig& config, SpaceFillType space_fill_type, int wipe_dist, float flow_ratio, std::optional<Point> near_start_location)
 {
     Polygons boundary;
-    if (comb_boundary_inside.size() > 0)
+    if (config.type == PrintFeatureType::Infill && comb_boundary_inside.size() > 0)
     {
-        // use the combing boundary inflated so that all skin/infill lines are inside the boundary
+        // use the combing boundary inflated so that all infill lines are inside the boundary
         int dist = 0;
         if (layer_nr >= 0)
         {


jackha:
Ok this fixes the slicing speed for the test models, but I can still trigger the great computation time manually. 

I'll merge this change, but if we get a lot of computation time complaints we have to look at the optimization code again. Maybe make an option out of it or worst case we have to remove it again.

Thanks for looking into this! Cheers.

smartavionics:
> Ok this fixes the slicing speed for the test models, but I can still trigger the great computation time manually.

And how can you trigger the great computation time manually? Please let me have an example stl/settings and I will continue to investigate it.

jackha:
- Enable ironing

or: a model with a lot of islands and with infill.

- Select UM3 / PLA / AA0.25 / Fine (0.1mm) profile
- chainmail coaster at 200%
- Wall Line Count 1 (more space for infill)
- Top/Bottom Thickness 0mm (less top and bottom -> more infill)

On my machine the slicing takes about 600 seconds.

[chainmail_coaster.stl.zip](https://github.com/Ultimaker/CuraEngine/files/1567585/chainmail_coaster.stl.zip)

smartavionics:
Ironing should not slow down as it's printing skin not infill - do you have a sample model I can try that shows that?

As for the many small islands with infill, that is a problem. Without adding some extra params and complexity to addLinesByOptimizer() I am not sure I can avoid the slowdown. Perhaps we should add the setting to disable the use of combing in the infill travel optimization.

smartavionics:
I've added a setting to control the use of this optimization:

https://github.com/Ultimaker/CuraEngine/pull/671
https://github.com/Ultimaker/Cura/pull/3009

PR #670, opened 2017-12-16
Made gcode comment style consistent

thieraufc:
The gcode comments generated in gcodeExport.cpp were not all consistent: "; relative extrusion mode" and "; absolute extrusion mode" have spaces between the semicolon and text, while the rest do not. I've changed those two lines remove that space so that they match the rest of the comments.

I know this is kinda trivial, but I was reading through the gcode to learn and that was bugging me. It's also a chance for me to get more involved with contributing to these sorts of projects, which is good.

### DIFF
--- a/src/gcodeExport.cpp
+++ b/src/gcodeExport.cpp
@@ -486,11 +486,11 @@ void GCodeExport::writeExtrusionMode(bool set_relative_extrusion_mode)
 {
     if (set_relative_extrusion_mode)
     {
-        *output_stream << "M83 ; relative extrusion mode" << new_line;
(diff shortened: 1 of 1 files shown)


thieraufc:
I wanted to learn gcode, so I thought I would read the output of a slicer. I ended up noticing a typo that I was able to fix real quick. I'm just trying to help out here.

smartavionics:
Hi, please don't take me too seriously, I mean no harm. BTW, there's no typo there, that's how I wanted the comments to look, I like the space following the semi. It's all the other comments in the gcode that are wrong!

Ghostkeeper:
Consistency is nice, especially in the g-code output. Thanks.

PR #671, opened 2017-12-18
Add infill travel opt setting

smartavionics:
Companion to https://github.com/Ultimaker/Cura/pull/3009.

### DIFF
--- a/src/LayerPlan.cpp
+++ b/src/LayerPlan.cpp
@@ -461,7 +461,7 @@ void LayerPlan::addPolygonsByOptimizer(const Polygons& polygons, const GCodePath
 void LayerPlan::addLinesByOptimizer(const Polygons& polygons, const GCodePathConfig& config, SpaceFillType space_fill_type, bool enable_travel_optimization, int wipe_dist, float flow_ratio, std::optional<Point> near_start_location)
 {
     Polygons boundary;
-    if (enable_travel_optimization && config.type == PrintFeatureType::Infill && comb_boundary_inside.size() > 0)
+    if (enable_travel_optimization && comb_boundary_inside.size() > 0)
     {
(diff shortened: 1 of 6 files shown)


jackha:
Devs, see: CURA-4725

src/LayerPlan.cpp
```diff
@@ -458,10 +458,10 @@ void LayerPlan::addPolygonsByOptimizer(const Polygons& polygons, const GCodePath
         addPolygon(polygons[poly_idx], orderOptimizer.polyStart[poly_idx], config, wall_overlap_computation, wall_0_wipe_dist, spiralize, flow_ratio, always_retract);
     }
 }
-void LayerPlan::addLinesByOptimizer(const Polygons& polygons, const GCodePathConfig& config, SpaceFillType space_fill_type, int wipe_dist, float flow_ratio, std::optional<Point> near_start_location)
+void LayerPlan::addLinesByOptimizer(const Polygons& polygons, const GCodePathConfig& config, SpaceFillType space_fill_type, bool enable_travel_optimization, int wipe_dist, float flow_ratio, std::optional<Point> near_start_location)
 {
     Polygons boundary;
-    if (config.type == PrintFeatureType::Infill && comb_boundary_inside.size() > 0)
```
LipuFei: @smartavionics I suppose we should still check the line type because that flag is meant for infill?

smartavionics: That test is now redundant because whether to use the travel optimization is now controlled by the new boolean arg to addLinesByOptimizer() and that arg will only be true when it is called for infill. But in the future we may want to use the optimization for lines that aren't infill (support maybe?) so it's better not to require the line type to be infill - that's my view.

LipuFei: OK, makes sense.

PR #672, opened 2017-12-19
Repetier also expects the M203 param to be in mm/min rather than mm/sec.

smartavionics:
See https://github.com/Ultimaker/Cura/issues/2371.

### DIFF
--- a/src/gcodeExport.cpp
+++ b/src/gcodeExport.cpp
@@ -1127,7 +1127,7 @@ void GCodeExport::writeMaxZFeedrate(double max_z_feedrate)
 {
     if (current_max_z_feedrate != max_z_feedrate)
     {
-        if (getFlavor() == EGCodeFlavor::REPRAP)
+        if (getFlavor() == EGCodeFlavor::REPRAP || getFlavor() == EGCodeFlavor::REPETIER)
         {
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
The [documentation](http://reprap.org/wiki/G-code#M203:_Set_maximum_feedrate) says that this should result in some temperature monitor being set?

Oh well, it's better than what it is now. I believe you.

smartavionics:
Hi @Ghostkeeper , I just checked https://github.com/repetier/Repetier-Firmware/blob/master/src/ArduinoAVR/Repetier/Commands.cpp and the m203 command is indeed related to temperature monitoring. So I recommend that we revert this for now. Sorry about that, I should have checked myself earlier.

smartavionics:
Furthermore, perhaps it should really be:
````
        if (getFlavor() == EGCodeFlavor::REPRAP)
        {
            *output_stream << "M203 Z" << PrecisionedDouble{2, max_z_feedrate * 60} << new_line;
        }
        else if (getFlavor() != EGCodeFlavor::REPETIER)
        {
            *output_stream << "M203 Z" << PrecisionedDouble{2, max_z_feedrate} << new_line;
        }
````

i.e. don't output the m203 at all for repetier?

PR #680, opened 2018-01-04
Do not output M203 in Repetier flavour gcode.

smartavionics:
It appears that not all versions of Repetier firmware react the same to the M203 command
so it's best not to send it at all.

See http://reprap.org/wiki/G-code#M203:_Set_maximum_feedrate.

### DIFF
--- a/src/gcodeExport.cpp
+++ b/src/gcodeExport.cpp
@@ -1126,11 +1126,11 @@ void GCodeExport::writeMaxZFeedrate(double max_z_feedrate)
 {
     if (current_max_z_feedrate != max_z_feedrate)
     {
-        if (getFlavor() == EGCodeFlavor::REPRAP || getFlavor() == EGCodeFlavor::REPETIER)
+        if (getFlavor() == EGCodeFlavor::REPRAP)
         {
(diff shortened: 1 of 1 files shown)


jackha:
I'm not sure what to make from the docs, it looks like repetier:
- supports X, Y, Z, E for maximum feedrates
- and S to set the temperature monitor (whatever that is)

Ghostkeeper:
What documentation is that, Jackha? This one says that on Repetier the M203 command "sets the temperature monitor", which is not the intended behaviour: http://reprap.org/wiki/G-code#M203:_Set_maximum_feedrate

Ghostkeeper:
Ah, so you think it does both and Repetier's g-code flavour should behave normally as long as it doesn't output any S value?

smartavionics:
Hi, the problem is that not supplying an S value actually has an effect, here's the current code in their repo:

````
    case 203: // M203 Temperature monitor
        if(com->hasS())
            manageMonitor = com->S != 255;
        else
            manageMonitor = 0;
break;
````
Therefore, I don't think it is safe to send the M203 with feedrate values.

Ghostkeeper:
Good research.

I've added a little bit of documentation why the exception is made and pushed that to master.

PR #681, opened 2018-01-06
Fix generating (manual) support meshes when support enable is false

fieldOfView:
This PR restores the functionality of support meshes when support generation has been disabled. Currently setting support_enabled to false prevents any support mesh from creating layers. This kills the use case of only wanting to print custom supports but no regular "automatic" supports.


jackha:
Devs: CURA-4770

fieldOfView:
This seems to have already been fixed in 01b49ced858ea4193b19021becdf7e0bb0e627ea

jackha:
Thanks for helping cleaning up ^^

PR #688, opened 2018-01-24
Implement speed/flow/fan settings for bridge regions

smartavionics:
Not yet a merge candidate but it is usable if people wish to try it out.

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -1438,6 +1438,8 @@ bool FffGcodeWriter::processInsets(const SliceDataStorage& storage, LayerPlan& g
                 }
             }
             const int half_outer_wall_width = mesh_config.inset0_config.getLineWidth();
+            // remove those parts of the layer below that are narrower than a wall line width as they will not be printed
+            outlines_below = outlines_below.offset(-half_outer_wall_width).offset(half_outer_wall_width);
             // max_air_gap is the max allowed width of the unsupported region below the wall line
             // if the unsupported region is wider than max_air_gap, the wall line will be printed using bridge settings
             // it can be from 0 to 1/2 the outer wall line width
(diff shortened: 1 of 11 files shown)


smartavionics:
Associated Cura PR is https://github.com/Ultimaker/Cura/pull/3175.

Ghostkeeper:
Developers, see issue CURA-67.

ianpaschal:
Tested by myself and @jackha.

smartavionics:
Hi @ianpaschal & @jackha, thanks for merging this quickly. Now the fun starts! Much testing is going to be needed to come up with suitable settings for the various materials/qualities.
If you come across any big problems, please contact me and I will fix ASAP.

ianpaschal:
Yeah, to be honest my background is not even really in 3D printing; it's in product design and computer graphics, it was only a year ago that I started doing my thesis with a 3d printing company making CAD software and I've only been at Ultimaker for 2 months... soooo I still just use the stock profiles haha. But I twiddled your settings and did some prints and nothing exploded so... merged!

Ellecross:
Hey @smartavionics, I've enabled the bridge settings on a different model [flat_bridge.zip](https://github.com/Ultimaker/CuraEngine/files/1803019/flat_bridge.zip) and the print looked really stringy compared to the same print without it. 

**Here are sample images:** 
img_20180312_145521398

img_20180312_145537374

img_20180312_145603999

Is this because the model I used have the bases too far apart? I've used PLA AA0.4 with Fine quality settings.

smartavionics:
That bridge isn't so long, my guess is that the default values for the bridge settings are not right. You will need to try different values for the speeds and flows, etc. Fan amount is critical too. Also, the direction of the fan has an influence (not that we can change that).

The first layer of the bridge should have no (or very few) broken strands and they should be separate and not stuck together.

Issue #691, opened 2018-01-26
CuraEngine misreports version, always displays DEV


onitake:
Since 3.1, CuraEngine always identifies itself as `Cura_SteamEngine version DEV` instead of displaying its proper version.

This occurs because the copyright message is constructed in main.cpp, which is not compiled with the proper `VERSION` macro.

A fix is trivial: Add a `set_target_properties` statement to the `CuraEngine` target in CMakeLists.txt, similar to the one for `_CuraEngine`.

Ghostkeeper:
This reproduces and your PR fixed it. Thanks!

PR #692, opened 2018-01-26
Added missing version flag to CuraEngine build target

onitake:
Fixes: #691


Ghostkeeper:
Indeed, that was a bug and this fixes it. Thanks!

PR #693, opened 2018-01-26
Bug fix for problem in 3.2 beta (complex skin can cause a big slow down)

smartavionics:
Add find_chains arg to LineOrderOptimizer::optimize() to control whether chains are detected.

Because finding chains can be slow for some models (e.g. if the skin contains many holes),
find_chains will not be true for skin lines and only true for infill lines when the infill travel
optimization is enabled.

This provides a workaround for https://github.com/Ultimaker/Cura/issues/3188.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,7 +40,13 @@ if(CMAKE_BUILD_TYPE MATCHES DEBUG OR CMAKE_BUILD_TYPE MATCHES RelWithDebInfo)
 endif()
 
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall") # Add warnings
-set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -Ofast -funroll-loops")
+
+option (ENABLE_MORE_COMPILER_OPTIMIZATION_FLAGS
+    "Enable more optimization flags" ON)
+if (ENABLE_MORE_COMPILER_OPTIMIZATION_FLAGS)
+    message (STATUS "Compile with more optimization flags")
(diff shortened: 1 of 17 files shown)


ChrisRiddell:
I hope this gets merged for stable as I have hit this edge case a couple times because of the types of prints I'm doing atm.

smartavionics:
Hi, @jackha, @Ghostkeeper, etc. can we please evaluate this as a priority for 3.2, thanks.

jackha:
Devs, CURA-4900
If I understand correctly, the model  https://www.thingiverse.com/thing:2449890 currently also get stuck in slicing even when Optimize Wall Printing Order is off? (It seems so in my test). In that case I would say it's a good bugfix for 3.2

ChrisRiddell:
That seems correct @jackha

smartavionics:
Hi Jack, yes, the slowdown is in the processing of the skin not the walls. That model has over 6000 skin lines per layer in places so it is too much work to try and optimize the order of the skin lines. So this PR turns off the optimization if there are more than 1000 skin lines or if the infill travel optimization is not enabled.

jackha:
@smartavionics Can you set the merge branch to 3.2?

smartavionics:
> @smartavionics Can you set the merge branch to 3.2?
Sorry to be dumb but how do I do that? I can't see anything on this github page for changing the merge branch. Shall I make a new PR with the same changes based on 3.2?

jackha:
I didn't know I was able to change it, so it's changed :-)

jackha:
@smartavionics I see there are master items in this PR, like the machine_buildplate_type. Can you remove them for 3.2?

smartavionics:
Done, see https://github.com/Ultimaker/CuraEngine/pull/698

smartavionics:
This PR is now sort of redundant because of the existence of #699 but that PR assumes that this one has been applied first so I recommend that you apply this PR and  #699 together and just test the combination. Thanks.

PR #695, opened 2018-01-28
Bug fix for broken overlap compensation when optimizing walls and number of walls > 2

smartavionics:
Pass inner inset wall overlap computer a very slightly reduced line width.

Because the same overlap computer is used for all of the inner insets, we need to reduce
the line width to stop the computer from flow reducing adjacent insets that are not at
the same level but while still allowing overlapping insets that are at the same level
to be flow reduced.

Fixes https://github.com/Ultimaker/Cura/issues/3197.

### DIFF
--- a/src/InsetOrderOptimizer.cpp
+++ b/src/InsetOrderOptimizer.cpp
@@ -462,7 +462,8 @@ bool InsetOrderOptimizer::processInsetsWithOptimizedOrdering()
         {
             wall_x_polys.add(part.insets[inset_level]);
         }
-        wall_overlapper_x = new WallOverlapComputation(wall_x_polys, mesh_config.insetX_config.getLineWidth());
+        // use a slightly reduced line width so that compensation only occurs between insets at the same level (and not between insets in adjacent levels)
+        wall_overlapper_x = new WallOverlapComputation(wall_x_polys, mesh_config.insetX_config.getLineWidth() - 1);
     }
 
     // create a vector of vectors containing all the inset polys


smartavionics:
BTW - there is still a problem when the overlap compensation reduces a wall down to nothing because the extruder will be moved at print speed and will dribble for the duration of the line. It would be better to replace print moves with very low flows with a proper travel move. something like this:

````
        if (flow < 0.01)
        {
            addTravel(p);
        }
        else
        {
            getLatestPathWithConfig(config, space_fill_type, flow, spiralize, speed_factor)->points.push_back(p);
            last_planned_position = p;
        }
````
Where the flow lower limit probably needs to be a setting.

Ghostkeeper:
Developers, see issue CURA-4957.

smartavionics:
Further to my above comment about replacing extrusions that use a flow of 0 with moves, it's sometimes useful to keep that capability. The new bridging code uses this technique to reduce the nozzle pressure immediately before the bridge and, arguably, the existing wipe nozzle code is actually wrong in that it uses a travel move for the wipe but should, instead, be using a zero flow extrusion. So, for the situation where overlap compensation has reduced a wall to zero flow (or nearly zero), replacing the extrusion with a move is probably what is needed but for the two situations mentioned in this comment (pre-bridge pressure reduction and wiping the nozzle at the end of a wall), using the zero-flow extrusion would be best. Got that?

smartavionics:
Please consider this bug fix for inclusion in 3.3. Thanks.

jackha:
See https://github.com/Ultimaker/Cura/issues/3197
New PRs should solve these issues.

PR #699, opened 2018-02-08
Improve efficiency of detection of chained infill/skin lines

smartavionics:
It's now much better than before and I think it can be safely used on models of any complexity. Please test on large models that have very complex skin and/or infill.

### DIFF
--- a/src/LayerPlan.cpp
+++ b/src/LayerPlan.cpp
@@ -487,7 +487,7 @@ void LayerPlan::addLinesByOptimizer(const Polygons& polygons, const GCodePathCon
     {
         orderOptimizer.addPolygon(polygons[line_idx]);
     }
-    orderOptimizer.optimize(enable_travel_optimization || polygons.size() <= 1000);
+    orderOptimizer.optimize();
     for (int poly_idx : orderOptimizer.polyOrder)
     {
         ConstPolygonRef polygon = polygons[poly_idx];
--- a/src/pathOrderOptimizer.cpp
+++ b/src/pathOrderOptimizer.cpp
@@ -234,12 +234,56 @@ void LineOrderOptimizer::optimize(bool find_chains)
         line_bucket_grid.insert(poly[0], poly_idx);
         line_bucket_grid.insert(poly[1], poly_idx);
         picked[poly_idx] = false;
+    }
 
+    std::vector<int> chain_end_poly_idxs; // the indices of the lines (polys) that start/end a chain of lines
+    std::vector<int> chain_end_point_masks; // a mask value for each chain start/end line that indicates which of the line's points are at the end of the chain
+
+    if (find_chains)
+    {
+        // locate the chain ends by finding lines that join exactly one other line at one end and join either 0 or 2 or more lines at the other end
+
+        for (unsigned int poly_idx = 0; poly_idx < polygons.size(); poly_idx++)
+        {
+            int num_joined_lines[2];
+            for (unsigned point_idx = 0; point_idx < 2; ++point_idx)
+            {
+                std::set<int> joined_lines; // use a set because getNearbyVals() appears to return duplicates (?)
+                num_joined_lines[point_idx] = 0;
+                const Point& p = (*polygons[poly_idx])[point_idx];
+                // look at each of the lines that finish close to this line to see if either of its vertices are coincident this vertex
+                for (unsigned int close_line_idx : line_bucket_grid.getNearbyVals(p, gridSize))
+                {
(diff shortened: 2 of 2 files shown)


jackha:
Thanks!
Devs: CURA-4932

alekseisasin:
Hi @smartavionics , I tested you changes and they work for medium size model, so far  But it looks like have a problem for big models. 
Load attached model, set generate support with extruder 1. It will slice **three times** longer than original solution. Maybe because of "orderOptimizer.optimize(enable_travel_optimization || polygons.size() <= 1000);"
[nova_kruna_.STL.zip](https://github.com/Ultimaker/CuraEngine/files/1824865/nova_kruna_.STL.zip)

smartavionics:
> Maybe because of "orderOptimizer.optimize(enable_travel_optimization || polygons.size() <= 1000);"

Hold on, with this PR applied, that line is gone. Are you saying, it should have that line?

Also, please tell me what printer/profile you are using so I can recreate your setup.

Thanks.

alekseisasin:
I used Ultimaker 3, Fine profile.

smartavionics:
OK, with UM3, fine profile, support enabled, I get a slice time of 81 seconds with this PR in place and with the chain detection completely disabled (so it takes no time), I get a slice time of 76 seconds. The print time for both is around 36 hours because this optimization doesn't actually help this model.

So, it took 5 seconds longer to slice with the PR in place and the chain detection enabled but that really isn't a big deal given that it takes 75 seconds to slice anyway, is it?

alekseisasin:
Hmm... Actually I have different numbers, the slicing time takes around 240 sec. Without PR around 68 sec. To be sure I attach my project. I did not compare the print time for this model.
[test_nova_kruna_.zip](https://github.com/Ultimaker/CuraEngine/files/1825095/test_nova_kruna_.zip)

smartavionics:
The reason I got such a big time difference is because I rotated the model through 90 deg as I thought it would be better that way.

With the model orientated as per your project file, I do see a 3 times slowdown with the chain detection enabled (366 seconds compared to 121). However, the optimization does save 42 mins of print time. Sadly, that's not a big improvement because the total print time is now 85 hours! Nonetheless, the combined slice+print time with the chain detection enabled is still around 40 mins less that without.

So, yes, the chain finding is slow for this model when orientated as you have it but if the model is orientated more sensibly the difference in slicing time is negligible

smartavionics:
You have got me thinking, I'm currently working on improving the efficiency some more. WIll be back to you soon...

smartavionics:
Hi @alekseisasin , please continue your testing. I have reworked the chain detection code (again) and it is now much faster (135 seconds for the nova model).

smartavionics:
I changed the implementation to make it look nicer and it will be faster when there is a lot of chains. The actual algorithms used to detect the chains and their subsequent processing has not changed so the output
should be the same as before this PR.

smartavionics:
Hi @alekseisasin , I see you are continuing to test this, thanks. But I notice that as of 22 March, you are not testing the most recent version of this PR, please update and include the latest code. Thanks again.

alekseisasin:
@smartavionics , I keep track of changes in master branches. I made local branch from your pull request and merge master into it. Also build using latest Cura changes. During testing pull request I found downgrade in print time because of changes in printer profiles.
For now everything works fine, I see improvements in slicing time, the "test_nova_kruna" slices good.  I want to print some models to ensure that quality also did not change.

jackha:
Hi @smartavionics . Along with @alekseisasin I also did some performance testing. I'm using linux and apparently it's different. With his project it goes from around 100s to about 400s. Do you have any idea what can cause this?

smartavionics:
I am using linux too and the performance is good. Please let me have exactly the project file that @alekseisasin is using that has the performance slowdown. Thanks.

smartavionics:
OK, well I have that project and that's what I have been testing against. I don't believe the most recent version of the PR has a significant performance hit when slicing that.

smartavionics:
I just opened that project and sliced and it took 130 seconds and the resulting print time estimate is 3d 12h and 27m. I then disabled the chain detection and resliced and it took 123 seconds (not that much quicker) and the print time went up to 3d 12h 54m. So the chain detection saves 27m on the print time which isn't much but this print isn't that amenable to this optimization so I wouldn't expect to see that much benefit.

smartavionics:
Thanks for merging this. Please, if you come across any more performance problems with this, contact me before reverting. Thanks.

diegopradogesto:
As @jackha said, we tested with a build and we notice that there is no bad performance with that model. Now the ticket is in QA and the testing team will do further tests.

PR #700, opened 2018-02-08
Now pointsAreCoincident() allows points to be up to 10um apart.

smartavionics:
Zigzag infill is still being produced with the odd gap here and there between line end points.
This most often occurs when the boundary of the skin/infill is curved. By allowing the line
end points to be a little further apart, chains are continued which would otherwise be
terminated at the gap.

### DIFF
--- a/src/pathOrderOptimizer.cpp
+++ b/src/pathOrderOptimizer.cpp
@@ -199,7 +199,7 @@ int PathOrderOptimizer::getRandomPointInPolygon(int poly_idx)
 
 static inline bool pointsAreCoincident(const Point& a, const Point& b)
 {
-    return vSize2(a - b) < 25; // points are closer than 5uM, consider them coincident
+    return vSize2(a - b) < 100; // points are closer than 10uM, consider them coincident
 }
 
 /**


smartavionics:
Here's an example of what this PR fixes. The zigzag infill that starts above the oval hole should actually start on the right hand side of that grey region so that it butts up against the infill on the right but because the zigzag chain is broken due to a tiny (> 5um but < 10um) gap between the line end points, the chain is
shorter than it could be. 

screenshot_2018-02-08_14-29-29

So to complete that section of infill, another travel has to occur later to come back and fill in the mini chain that remains. The extra (long) travel is shown here:

screenshot_2018-02-08_14-33-35

With the PR in place, those two chains are merged into one and the long travel is avoided.

jackha:
Devs: CURA-4937

src/pathOrderOptimizer.cpp
```diff
@@ -199,7 +199,7 @@ int PathOrderOptimizer::getRandomPointInPolygon(int poly_idx)
 
 static inline bool pointsAreCoincident(const Point& a, const Point& b)
 {
-    return vSize2(a - b) < 25; // points are closer than 5uM, consider them coincident
+    return vSize2(a - b) < 100; // points are closer than 10uM, consider them coincident
```
Ghostkeeper: We think this is a good change, but could you instead use the setting `meshfix_maximum_resolution` for this? It acts effectively as the slicing resolution, so we think it would be fitting to have that instead of hard-coding this value.

smartavionics: That would be sensible. The only difficulty is that setting is not currently accessible within the LineOrderOptimizer class so it would require passing in via the constructor. Is that what you would like me to do?

smartavionics: Actually, it gets worse because where the LineOrderOptimizer is used in LayerPlan::addLinesByOptimzer(), the mesh isn't available either so it would have to be passed in to that method from wherever it is called from. All getting a bit messy. How should we proceed?

PR #707, opened 2018-02-23
fix: prime tower purge now starts away from middle

BagelOrb:
Starting in the middle causes a tower in the middle of the hollow prime tower,
while purging was meant to happen in the air without any back pressure of the previous layer.

Because sometimes the purge amount is very high, purging from the middle can lead to overextrusion,
which results in material sticking out above the current layer, which can cause collisions.


Appesteijn:
Devs: CURA-5099

PR #711, opened 2018-02-28
Feature: limit infill to where needed

BagelOrb:
This awesome feature is like a specialized support generation algorithm, but for the internal support structures ak infill.

With this enabled only infill will be generated where it's needed.

Doesn't work optimally with Infill meshes, but it works for 97% of infill mesh cases in combination with infill support.

image

### DIFF
--- a/src/skin.cpp
+++ b/src/skin.cpp
@@ -437,16 +437,7 @@ void SkinInfillAreaComputation::generateInfill(SliceLayerPart& part, const Polyg
     infill = infill.difference(skin);
     infill.removeSmallAreas(MIN_AREA_SIZE);
 
-    Polygons final_infill = infill.offset(infill_skin_overlap);
-
-    if (mesh.getSettingBoolean("infill_hollow"))
-    {
-        part.print_outline = part.print_outline.difference(final_infill);
-    }
-    else
-    {
-        part.infill_area = final_infill;
-    }
+    part.infill_area = infill.offset(infill_skin_overlap);
 }
 
 /*
--- a/src/skin.cpp
+++ b/src/skin.cpp
@@ -446,6 +446,7 @@ void SkinInfillAreaComputation::generateInfill(SliceLayerPart& part, const Polyg
     else
     {
         part.infill_area = final_infill;
+        part.infill_area = final_infill;
     }
 }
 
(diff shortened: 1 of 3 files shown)


BagelOrb:
Frontend PR: https://github.com/Ultimaker/Cura/pull/3413

Appesteijn:
Devs: CURA-5108

src/skin.cpp
```diff
@@ -446,6 +446,7 @@ void SkinInfillAreaComputation::generateInfill(SliceLayerPart& part, const Polyg
     else
     {
         part.infill_area = final_infill;
+        part.infill_area = final_infill;
```
ianpaschal: Looks like a mistake; why is this here twice?

BagelOrb: Yeah. fixed.

ianpaschal:
Looks good to me. Maybe @Ghostkeeper can offer the required 2nd opinion?

src/skin.cpp
```diff
@@ -496,6 +496,45 @@ void SkinInfillAreaComputation::generateRoofing(SliceLayerPart& part)
...
+        {
+            const Polygons& infill_area = part.infill_area;
+            if (infill_area.empty())
+            {
+                continue;
+            }
+
+            const Polygons unsupported = infill_area.offset(-max_dist_from_lower_layer);
+            const Polygons basic_overhang = unsupported.difference(inside_above);
+            const Polygons overhang_extented = basic_overhang.offset(max_dist_from_lower_layer + 50); // +100 for easier joining with support from layer above
```
Ghostkeeper: +100? :stuck_out_tongue:

Perhaps use `support_join_distance` instead? You'd be having a support setting that influences infill, but it probably needs to be equal in just about all cases anyway.

BagelOrb: I think it's counterintuitive to use the same option. It would then also have to be enabled even when support is not generated.

We could make it an extra option, but the whole idea of this PR was to have the functionality without introducing the millions of settings that support has.

If you find some cases which fail badly because of this magic value I could reconsider the implementation, but I'd rather just have it merged and tri s out by people.

Ghostkeeper: I'll just change the documentation then.

Ghostkeeper:
Looking at how the functionality works, what is the use case of this vs. the "Hollow Out Objects" feature?

It looks like the Hollow Out Objects feature is more powerful but harder to use. However its power is largely useless because for infill it doesn't matter that the "support" is easier to break off or leaves no traces or anything.

Ghostkeeper:
Tested and it seems to work fine for the rest.

BagelOrb:
Support is something different from infill; that's the easy answer. Hollow out objects was just a hack.

For a lot of prints I don't want to generate support because I know my model is better off without it, but that doesn't hold for infill of course.

Also, because of the support Z distance you would get more pillowing on top of the prints.

Ghostkeeper:
I discussed that with two colleagues and we agree that the Hollow Out Objects should be removed.

I'll remove it from the code and let another colleague review that. Removing a setting involves adding it to the version upgrade as well so we won't bother you with it.

BagelOrb:
Will you then automatically enable `infill_support_enabled` if any profile or project had Hollow Out Object enabled?

Ghostkeeper:
That's a good idea. Yeah, I think I will.

PR #715, opened 2018-03-09
Take support into consideration when evaluating a skin region to see if it is a bridge.

smartavionics:
Otherwise, an area of skin that is supported by a combination of model and support
could be considered a bridge region when it really shouldn't.

This is a bit of a grey area because some combinations of model+support could leave skin
regions that really should be printed using bridge settings. More work required sometime.

Please consider this for inclusion in 3.3 as it should be beta tested along with the main bridging PR that has already been merged. Thanks.


smartavionics:
Incorporated into another PR, closing...

PR #718, opened 2018-03-09
For RepRap firmware, the "Filament used" gcode comment now lists values for all extruders.

smartavionics:
The firmware expects the comment to look like this example:

; Filament used: 2.7m, 10.2m

i.e. the values are listed in extruder order separated by comma and space.

BTW - This PR does not affect Marlin gcode which uses the same string, should it?

### DIFF
--- a/src/gcodeExport.cpp
+++ b/src/gcodeExport.cpp
@@ -203,7 +203,24 @@ std::string GCodeExport::getFileHeader(const std::vector<bool>& extruder_is_used
         }
         else if (flavor == EGCodeFlavor::REPRAP || flavor == EGCodeFlavor::MARLIN)
         {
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
By the way that you implemented it, it looks like this does affect Marlin. It's in the same if-case. Right? I think that's fine. I guess that wherever we now output only one extruder we should output the information for all extruders.

Ghostkeeper:
Developers, see issue CURA-5194.

smartavionics:
It's the same if case but the loop limit is 1 for marlin. If you wish marlin to behave the same as reprap, it's trivial for me to change.

smartavionics:
Done that.

ianpaschal:
Smexy.

As always, thank you for your contributions!

PR #721, opened 2018-03-12
Wall line segments that have a flow < wall_min_flow are converted into travel moves.

smartavionics:
Companion to https://github.com/Ultimaker/Cura/pull/3479.

### DIFF
--- a/src/LayerPlan.cpp
+++ b/src/LayerPlan.cpp
@@ -655,6 +655,8 @@ void LayerPlan::addWall(ConstPolygonRef wall, int start_idx, const GCodePathConf
 
     const SettingsBaseVirtual* extr = getLastPlannedExtruderTrainSettings();
     const double min_bridge_line_len = extr->getSettingInMicrons("bridge_wall_min_length");
+    const double wall_min_flow = extr->getSettingInPercentage("wall_min_flow") / 100;
+    const bool wall_min_flow_retract = extr->getSettingBoolean("wall_min_flow_retract");
 
     // helper function to calculate the distance from the start of the current wall line to the first bridge segment
 
@@ -739,6 +741,8 @@ void LayerPlan::addWall(ConstPolygonRef wall, int start_idx, const GCodePathConf
         }
     };
 
+    bool travel_required = false; // true when a wall has been omitted due to its flow being less than the minimum required
+
     for (unsigned int point_idx = 1; point_idx < wall.size(); point_idx++)
     {
         const Point& p1 = wall[(start_idx + point_idx) % wall.size()];
@@ -749,9 +753,29 @@ void LayerPlan::addWall(ConstPolygonRef wall, int start_idx, const GCodePathConf
             computeDistanceToBridgeStart((start_idx + point_idx - 1) % wall.size());
         }
 
-        addWallLine(p0, p1, non_bridge_config, bridge_config, flow, non_bridge_line_volume, speed_factor, distance_to_bridge_start);
+        // HACK ALERT
+        // the overlap compensation is not perfect, it can produce short non-flow reduced line segments within a sequence of flow reduced
+        // line segments and so to try and avoid printing the spurious fat line segments we require that their lengths are above a threshold
+
+        const int64_t max_spurious_fat_segment_length = 50; // microns
+
+        if (flow >= wall_min_flow && (!travel_required || vSize(p0 - p1) > max_spurious_fat_segment_length))
+        {
+            if (travel_required)
+            {
+                addTravel(p0, wall_min_flow_retract);
+                travel_required = false;
+            }
+            addWallLine(p0, p1, non_bridge_config, bridge_config, flow, non_bridge_line_volume, speed_factor, distance_to_bridge_start);
+        }
+        else
+        {
+            travel_required = true;
+        }
+
         p0 = p1;
     }
+
     if (wall.size() > 2)
     {
         const Point& p1 = wall[start_idx];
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Developers, see issue CURA-5119.

src/LayerPlan.cpp
```diff
@@ -749,9 +754,30 @@ void LayerPlan::addWall(ConstPolygonRef wall, int start_idx, const GCodePathConf
             computeDistanceToBridgeStart((start_idx + point_idx - 1) % wall.size());
         }
 
-        addWallLine(p0, p1, non_bridge_config, bridge_config, flow, non_bridge_line_volume, speed_factor, distance_to_bridge_start);
+        // HACK ALERT
+        // the overlap compensation is not perfect, it can produce short non-flow reduced line segments within a sequence of flow reduced
+        // line segments and so to try and avoid printing the spurious fat line segments we require that their lengths are above a threshold
+
+        const int64_t max_spurious_fat_segment_length = 50; // microns
```
ianpaschal: Some notes about this: According to Ruben, this should be using coordinates, as well as the global resolution parameters.

We don't think this hack is too bad but are wondering what the consequences of not having it would be...

smartavionics: > Some notes about this: According to Ruben, this should be using coordinates, as well as the global resolution parameters.

Not sure I understand what you mean by using coordinates. As for the global resolution params, I don't think they are accessible in this method, are they?

> We don't think this hack is too bad but are wondering what the consequences of not having it would be...

If you don't have the hack then tiny line segments can be included which are not really wanted. Let me try and come up with a concrete example...

smartavionics: I can't recreate the situation that made me want to filter out the tiny fat line segments right now but I am sure that I added that because I was seeing a problem. I don't think I just put it in from the beginning in the hope that it would be useful.

ianpaschal: @Ghostkeeper

ianpaschal: > Not sure I understand what you mean by using coordinates.

I mean it should be a `coord_t`

Makes sense regarding the line segments. For me it's OK because it's logical, but was still worth asking.

smartavionics: Ah, understood. I've made that little change.

src/LayerPlan.cpp
```diff
@@ -646,15 +646,14 @@ void LayerPlan::addWall(ConstPolygonRef wall, int start_idx, const GCodePathConf
...
-    Point p0 = wall[start_idx];
-    addTravel(p0, always_retract);
-
     float non_bridge_line_volume = 0; // zero before first non-bridge line is output
     double speed_factor = 1.0; // start first line at normal speed
     double distance_to_bridge_start = 0; // will be updated before each line is processed
 
     const SettingsBaseVirtual* extr = getLastPlannedExtruderTrainSettings();
     const double min_bridge_line_len = extr->getSettingInMicrons("bridge_wall_min_length");
+    const double wall_min_flow = extr->getSettingInPercentage("wall_min_flow") / 100;
```
Ghostkeeper: We have a function `getSettingAsRatio` which does exactly what you need.

src/LayerPlan.cpp
```diff
@@ -749,9 +754,30 @@ void LayerPlan::addWall(ConstPolygonRef wall, int start_idx, const GCodePathConf
             computeDistanceToBridgeStart((start_idx + point_idx - 1) % wall.size());
         }
 
-        addWallLine(p0, p1, non_bridge_config, bridge_config, flow, non_bridge_line_volume, speed_factor, distance_to_bridge_start);
+        // HACK ALERT
+        // the overlap compensation is not perfect, it can produce short non-flow reduced line segments within a sequence of flow reduced
+        // line segments and so to try and avoid printing the spurious fat line segments we require that their lengths are above a threshold
+
+        const coord_t max_spurious_fat_segment_length = 50; // microns
```
Ghostkeeper: You should square this and compare with `vSize2(p0 - p1)` for performance.

src/LayerPlan.cpp
```diff
@@ -646,15 +646,14 @@ void LayerPlan::addWall(ConstPolygonRef wall, int start_idx, const GCodePathConf
         }
     }
 
-    Point p0 = wall[start_idx];
-    addTravel(p0, always_retract);
```
Ghostkeeper: I think that this was a much nicer/simpler solution than the business you have below with the `first_line` boolean variable. Why was it removed?

src/LayerPlan.cpp
```diff
@@ -749,9 +754,30 @@ void LayerPlan::addWall(ConstPolygonRef wall, int start_idx, const GCodePathConf
...
+            {
+                addTravel(p0, (first_line) ? always_retract : wall_min_flow_retract);
+                first_line = false;
+                travel_required = false;
+            }
+            addWallLine(p0, p1, non_bridge_config, bridge_config, flow, non_bridge_line_volume, speed_factor, distance_to_bridge_start);
+        }
+        else
+        {
+            travel_required = true;
```
Ghostkeeper: It seems illogical to set some `travel_required` variable here rather than just travelling immediately to `p1`.

src/LayerPlan.cpp
```diff
@@ -749,9 +754,30 @@ void LayerPlan::addWall(ConstPolygonRef wall, int start_idx, const GCodePathConf
...
         }
 
-        addWallLine(p0, p1, non_bridge_config, bridge_config, flow, non_bridge_line_volume, speed_factor, distance_to_bridge_start);
+        // HACK ALERT
+        // the overlap compensation is not perfect, it can produce short non-flow reduced line segments within a sequence of flow reduced
+        // line segments and so to try and avoid printing the spurious fat line segments we require that their lengths are above a threshold
+
+        const coord_t max_spurious_fat_segment_length = 50; // microns
+
+        if (flow >= wall_min_flow && (first_line || !travel_required || vSize(p0 - p1) > max_spurious_fat_segment_length))
```
Ghostkeeper: `addWallLine` is not called if the line is smaller than 50 microns now (unless it's the first line).

src/LayerPlan.cpp
smartavionics: Because, before, it would always do the first travel even when the whole wall is not printed because all the line segments have a flow that is too low. You only want to do the initial travel move if there are any segments that are to be printed.

src/LayerPlan.cpp
smartavionics: See reply above.

src/LayerPlan.cpp
smartavionics: That's true - I suppose you could get a first line printed that is shorter than 50 microns that you would otherwise not print but I don't think that is very likely and even if you did print it, so what?

ianpaschal:
This is ready for a re-review?

src/LayerPlan.cpp
ianpaschal: This one is still a bit of a problem because its possible to try and print a circle in which all the lines are less than 50 microns. I mean, unlikely because people often post files with stupidly low resolutions online, but it could happen. Then we'd end up skipping all the entire shape.

ianpaschal:
Just spoke with @Ghostkeeper and we're still concerned about one point.

src/LayerPlan.cpp
smartavionics: OK, I'll take the fat line avoidance stuff out for you (and keep it for me).

smartavionics:
Hi Ian, I have removed the hack so people can now happily print walls with lines < 50um long.

smartavionics:
You will, of course, now get spurious travels to random points on curved walls but, hey, that's a small price to pay.

ianpaschal:
That's better than having the curved wall not print at all, though, right? I mean, now I'm all conflicted trying to think what's more likely: long chains of 50 micron segments or random single 50 micron segments...

smartavionics:
Hi Ian, don't agonise about it. Go with how it is now and in the event of it being a real problem think of that as an incentive to fix the overlap compensation!

ianpaschal:
Yeah I guess neither is very common but we'd rather have small spurs that can be trimmed off with an x-acto knife than holes or a missing wall in the case of a chain of 50 micron .

> "Chain so light when a breeze roll by, man it flow-ow-ow, man it flow-ow-ow;"
>     - [The Lonely Island](https://youtu.be/dgDaWZAjefg?t=1m16s)

Before:
screen shot 2018-05-30 at 13 34 12


After:
screen shot 2018-05-30 at 13 34 23

jackha:
Hi @smartavionics , why are the settings Minimum Wall Flow and Prefer Retract under Compensate Wall Overlaps?

smartavionics:
> Hi @smartavionics , why are the settings Minimum Wall Flow and Prefer Retract under Compensate Wall Overlaps?

Because they are only relevant when the walls are being reduced in width by the overlap compensation. If there's no overlap compensation, all walls will be 100% of their normal width.

jackha:
We've discussed it and decided that the settings are not actual sub settings of wall overlap compensation (but related to), we're putting it just below. Visibility is the same as before.

smartavionics:
OK, that's fine. Will you still hide them when the overlap compensation is disabled?

jackha:
Yes, as you said these settings wouldn't make sense if overlap compensation is off.

PR #722, opened 2018-03-12
If the outer insets are being printed first, print the part outline before the holes.

smartavionics:
Makes for a nicer print when thin walls are being flow reduced (see #721).

### DIFF
--- a/src/InsetOrderOptimizer.cpp
+++ b/src/InsetOrderOptimizer.cpp
@@ -324,7 +324,7 @@ void InsetOrderOptimizer::processHoleInsets()
     }
 }
 
-void InsetOrderOptimizer::processOuterWallInsets()
+void InsetOrderOptimizer::processOuterWallInsets(const bool include_outer, const bool include_inners)
 {
     const coord_t wall_line_width_x = mesh_config.insetX_config.getLineWidth();
     const coord_t wall_0_wipe_dist = mesh.getSettingInMicrons("wall_0_wipe_dist");
@@ -335,11 +335,14 @@ void InsetOrderOptimizer::processOuterWallInsets()
     constexpr float flow = 1.0;
 
     // process the part's outer wall and the level 1 insets that it surrounds
+
+    ConstPolygonRef outer_wall = *inset_polys[0][0];
+    Polygons part_inner_walls;
+    int num_level_1_insets = 0;
+
+    if (include_inners)
     {
-        ConstPolygonRef outer_wall = *inset_polys[0][0];
         // find the level 1 insets that are inside the outer wall and consume them
-        Polygons part_inner_walls;
-        int num_level_1_insets = 0;
         for (unsigned int level_1_wall_idx = 0; inset_polys.size() > 1 && level_1_wall_idx < inset_polys[1].size(); ++level_1_wall_idx)
         {
             ConstPolygonRef inner = *inset_polys[1][level_1_wall_idx];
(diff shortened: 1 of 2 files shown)


Ghostkeeper:
Developers, see issue CURA-5119.

Ghostkeeper:
Developers, see issue CURA-5200.

ianpaschal:
Hey Mr. Avionics, we're trying to test this today. Can you explain what you mean by "insets", maybe with a before and after image?

smartavionics:
Insets are what walls are often referred to within the engine code (insets = walls). I guess they are called that because they are inset from the part's outline?

The reason that this change is beneficial is that when walls are closer together than the normal wall spacing one of the walls will be thinned down by the overlap compensation. The wall that is thinned down is the second wall to be printed (i.e. the first wall is printed full width and then where the second wall gets close to the first, it is narrowed). This narrowing can introduce a little visual artifact and so I thought that as the outside wall is more likely to be seen than an inside wall, it would be better to print the outside first so that it is the inside wall that gets narrowed. Does this make sense.

I need to go out now but I will try and get an image for you when I return.

ianpaschal:
Ah yes makes sense. So basically outer walls should print before inner walls.

If you have an ideal test model to create this situation, we'd be happy to see it.

smartavionics:
Here's a little project that shows the benefit of this PR 
[UM3_test7.curaproject.zip](https://github.com/Ultimaker/CuraEngine/files/2034844/UM3_test7.curaproject.zip)

Without this PR:

screenshot_2018-05-24_11-11-55

With this PR:

screenshot_2018-05-24_11-12-35

ianpaschal:
Thanks! Will test shortly and hopefully merge today!

ianpaschal:
I'm seeing the improvement in Cura but to be honest there's no difference in the actual print quality.

img_6196

smartavionics:
That's because the z-seam is hopping around in that area. Set the z-seam location to user-specified and it should come out prettier!

rickyzhang82:
Why do merge it even you don't see any physical differences?

smartavionics:
That particularly example maybe doesn't show the benefit as well as some others but this change can make the difference between seeing a visible rough patch on the outer wall of an object or not.

Here's another picture, again the difference isn't huge but the outer wall visual quality does improve. It's more noticeable when you rotate the objects

img_20180524_212020912

rickyzhang82:
I'm a conservative American on financial matters (also on programming) and a liberal on social matters. Please pardon me for my views.

If you accept a marginal (or even unnoticeable) improvement at the cost of introducing more ad-hoc code, chances are you may introduce more bugs and make future maintenance harder. The print might looks great in some STL model, but it might get bad result in most of STL model, which neither you nor I can prove or verify it for now.

In tool path generation, please don't and don't (I want to stress this) change it too often and too much like changing fashion clothes unless it is absolutely necessary. 

Premature optimization is the root of all evil in programming. That's what I experienced again and again in my daily jobs. The solution with fewest assumption is always the best. This has been proven again and again.

ianpaschal:
We take this pretty seriously.

We have about 150 employees at our office and a dozen of them are Cura team. A huge portion of the company is responsible for carrying out extensive testing on both hardware and software. This commit will now be use integration tests by Cura QA and then go through systems testing who like to show us all the prints with tiny defects that weren't there last week.

But if we try to go through that process before merging anything, nothing will ever get merged. This _is_ a careful approach. We work in small, minute, incremental improvements with lots of testing.

Cura is a widely used and mature piece of software so playing the "premature optimization" card doesn't really make sense here. In fact, it's kind of nonsensical whichever way you twist it. I am in favor of these incremental improvements in favor of trying to achieve a perfect result (or in other words, premature optimization) before merging and performing integration tests with the hardware.

rickyzhang82:
The size of team or company doesn't guarantee anything. In fact, more people you bring in, more bugs you introduce. It is engineering approach and philosophy that dictate quality.

You can deny this is not a premature optimization in any angle. But just answer me one simple question: After you merged multiple PR into upstream master branch, your QA bring up a new defective print compared to the regression test he did last time. How can you use git blame to figure out what PR cause the problem? There are so many and so many variable factors like settings here and there that make you trace the root cause impossible. 

If you are truly take the matter seriously, you would take the PR in an isolated regression test with all the bench print . But I doubt you can do it given the resource you had. In fact, I don't even see the integration test that run virtually on any benchmark model in Github. Maybe you don't release it on Github or maybe you haven't done it as well. I can imagine this is very difficult to benchmark regression test virtually as well.
 
I didn't intend to undermine Cura is not widely used nor immature. Please understand the fact that I want to get it better. 

I'd prefer to do this in a different way. For example, split the master branch into stable and experimental. For those features adding new hardware support, you can merge into stable branch ASAP. But **ONLY and ONLY** cherry pick the well tested tool path generation improvement features from experimental branch into stable branch after a long period of test. In such way, you can take whoever hand you whatever candy from the street. We still feel safe to use the latest release.

ianpaschal:
> The size of team or company doesn't guarantee anything. In fact, more people you bring in, more bugs you introduce. It is engineering approach and philosophy that dictate quality.

More eyes help find bugs, testers are not writing the code. 🙄

> I can imagine this is very difficult to benchmark regression test virtually as well.

This is a bit the problem. There's not a clear metric for quality and no universal benchmark print. The only reliable solution is try lots of stuff as much as possible, which is what we do. No one wants to be sloppy but we also need to keep the workflow moving along. This was unit tested, did what it was supposed to, and will now undergo extensive integration tests with hardware and humans.

That's kind of the end of the discussion. If you have further concerns you're welcome to reach out to the Cura team, or even open an issue we can label as "Discussion", but this PR is not the right place to have a huge discussion about development workflows.

Thank you for your feedback though.

rickyzhang82:
I disagree. My testers did make me add more bugs. Because she asked me to fix something that turns out to breaking another things. When solution to the problem becomes way too complex, bad things may happen.

OK. I will definitely do that in a discussion thread regarding to the branching. Thanks!

PR #727, opened 2018-03-19
Svg improvements

BagelOrb:
some improvements to output debugging

moved implementation to cpp file


BagelOrb:
Verified: still compiles.

Only affects the debug class, no calls had to be changed.

Appesteijn:
👍

BagelOrb:
:tada:

PR #728, opened 2018-03-19
More Svg improvements

BagelOrb:
I was having a lot of rounding errors, which produces visual problems which people have asked me about on several occasions.

In this PR the internal format the SVG class uses is FPoint3 (floating 3d) instead of Point (integer 2D)
(3d because there is no class in cura for a 2D float point and I didn't want to introduce one just for a debugging class).

### DIFF
--- a/src/utils/SVG.cpp
+++ b/src/utils/SVG.cpp
@@ -64,7 +64,12 @@ double SVG::getScale() const
 
 Point SVG::transform(const Point& p) 
 {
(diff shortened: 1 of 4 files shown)


BagelOrb:
Before:
image


After:
image

Ghostkeeper:
Wait, so we were casting the coordinates to integer in the SVG's coordinate space?

BagelOrb:
Yeah. I was lazy in the initial implementation. 

D:

PR #731, opened 2018-03-25
fix: Pathorderooptimizer used deleted object

BagelOrb:
The default object was destroyed right after the constructor call, but a reference to it remained.

### DIFF
--- a/src/pathOrderOptimizer.h
+++ b/src/pathOrderOptimizer.h
@@ -44,14 +44,14 @@ class PathOrderOptimizer
 {
 public:
     Point startPoint; //!< A location near the prefered start location
-    const ZSeamConfig& config;
+    const ZSeamConfig config;
     std::vector<ConstPolygonPointer> polygons; //!< the parts of the layer (in arbitrary order)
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Well, there is some value in having default argument values.

Passing this by value instead of by reference saves a few copies. In this case (looking at the usages of this constructor) it seems like this fix would add `3 * n_layers * n_models + 2 * n_layers` copies to the runtime where it's now passed by reference, which isn't all too much. But since the argument is `const`, GCC can optimize out the entire copy and makes the change to those calls effectively nothing when the argument is used.

When the argument is not used it now creates an extra `3 * n_layers * n_models` instances where it would otherwise have a reference to a destroyed object so that is the real performance impact. But I think that's small enough.

Ghostkeeper:
I couldn't find how to actually test this other than seeing whether it still compiles and slices. It's hard to see whether there is a reference to a deleted object since it seems that those 3 places where the function was called without the `zseamconfig` parameter were already working fine. Maybe the memory was deallocated but not yet overwritten by other data.

It still seems to compile and slice fine after the change.

Thanks.

PR #732, opened 2018-03-26
Bug fix when printing one at a time - move to next object location before waiting for bed temperature.

smartavionics:
It was waiting for the bed temperature to reach the first layer temp while the hotend
was still positioned on the top of the last part. Now it moves to the location of the next
part before waiting for the hotend to heat up.

Fixes issue reported at https://community.ultimaker.com/topic/22176-cura-321-is-melting-my-prints/

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -520,16 +520,16 @@ void FffGcodeWriter::processNextMeshGroupCode(const SliceDataStorage& storage)
     gcode.writeFanCommand(0);
     gcode.setZ(max_object_height + 5000);
 
+    CommandSocket::setSendCurrentPosition(gcode.getPositionXY());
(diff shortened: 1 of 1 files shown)


smartavionics:
Also.. there is still another related bug because if the first layer nozzle temperature is different from the temperature used for the other layers, it is only used for the first part. For all the remaining parts, the nozzle temperature stays the same for all layers. This could cause problems with first layer adhesion?

Ghostkeeper:
Developers, see issue CURA-5258.

BagelOrb:
@Ghostkeeper do you mean you want @smartavionics to do that before you will consider merging this PR?

smartavionics:
I know UM aren't keen on z movement but as this is only once per model I would have thought keeping the 5mm z-hop isn't going to be a big overhead and it keeps the nozzle out of the way?

PR #734, opened 2018-03-26
fix: escape newlines in settings

BagelOrb:
Before:
image

:(

After:
image

:)


Ghostkeeper:
Looks a bit neater, thanks.

PR #735, opened 2018-03-26
potential improvements to overlap compensation

BagelOrb:
- When a link cannot be introduced because the distance is zero, change the existing link instead, rather than doing nothing.

- When there are multiple links use the one with the most overlap. That means: only compensate for the overlap between a line segment and the most overlapping line segment. I'm not sure whether this is smart, because this means the amount of overlap being compensated depends on the processing order and thus on the Z seam.

I haven't checked if this fixes anything. The bugs I am encountering occur even though overlap compensation is off, so there is something weird going on here.

Anyway I thought I would post this here for other people to check whether this improves the horrid overlap compensation situation ;)

Maybe this is something @smartavionics would like to look at?

### DIFF
--- a/src/utils/PolygonProximityLinker.cpp
+++ b/src/utils/PolygonProximityLinker.cpp
@@ -369,6 +369,7 @@ void PolygonProximityLinker::addProximityEnding(const ProximityPointLink& link, 
     {
 //         addProximityLink(link.a, link.b, proximity_distance);
         // won't be inserted any way, because there already is such a link!
+        link.setDist(proximity_distance);
     }
 }
 
--- a/src/utils/ProximityPointLink.cpp
+++ b/src/utils/ProximityPointLink.cpp
@@ -17,4 +17,10 @@ bool ProximityPointLink::operator==(const ProximityPointLink& other) const
     return (a == other.a && b == other.b) || (a == other.b && b == other.a);
 }
 
+void ProximityPointLink::setDist(coord_t distance) const
+{
+    ProximityPointLink& thiss = *const_cast<ProximityPointLink*>(this);
+    thiss.dist = distance;
+}
+
 }//namespace cura 
--- a/src/utils/ProximityPointLink.h
+++ b/src/utils/ProximityPointLink.h
@@ -39,8 +39,9 @@ struct ProximityPointLink
 {
     const ListPolyIt a; //!< the one point (invalidated after list_polygons have been cleared!)
     const ListPolyIt b; //!< the other point (invalidated after list_polygons have been cleared!)
-    const int dist; //!< The distance between the two points
+    coord_t dist; //!< The distance between the two points
     const ProximityPointLinkType type; //!< The type of link; why/how it was created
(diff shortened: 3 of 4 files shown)


smartavionics:
Hi Tim, these days I'm thinking that very small overlaps, e.g. when a line goes into an acute corner and out again are simply not worth worrying about because the extruder lag is going to completely mask the flow reduction/increase. In the extreme case I could imagine that when the nozzle goes into such a corner, the dip in flow rate is actually going to cause a little narrowing of the line some time past the corner. So I am currently thinking that the only places that are really worth flow reducing are where you have line segments that are close together for some distance. i.e. it's the parallel lines that need to be flow reduced and not the corners. Or, if you are going to flow reduce the corners then it needs to take into account the extruder lag and start the flow reduction before the actual region where the lines are close is encountered.

BagelOrb:
The goal of overlap compensation is not so much to start printing thinner lines where there would be an overlap, but to take into account that we inevitable print thinner lines at some places. The goal is to not be extruding too much over a longer line. We need to prevent the overextrusion which would happen just after a couple of sharp corners.

If a corner is sharp then the line going into the corner will block material coming out for the line coming out of the corner.
This is especially so for very sharp corners of like 3*.
I could lower the angle limit....


----------------------

How about we make the overlap compensation a percentage? I think that would certainly help.

Ghostkeeper:
Developers, see issue CURA-5261.

src/utils/PolygonProximityLinker.cpp
```diff
@@ -369,6 +369,7 @@ void PolygonProximityLinker::addProximityEnding(const ProximityPointLink& link,
     {
 //         addProximityLink(link.a, link.b, proximity_distance);
         // won't be inserted any way, because there already is such a link!
```
jackha: outdated outcommented lines

BagelOrb:
This PR is not about creating a situation for which the code is better, but about creating code which deals better with difficult geometry. You should try a couple of models which have been reported to have weird overlap compensation behavior.

BagelOrb:
Just a random example. This seems to do some weird stuff due to overlap compensation in the round corners...
image
[UM3_01-03-00 Printhead with cartridges Copy.curaproject.3mf.renamed.zip](https://github.com/Ultimaker/CuraEngine/files/2048151/UM3_01-03-00.Printhead.with.cartridges.Copy.curaproject.3mf.renamed.zip)

I also recall stuff going wrong where the wall was modeled to be exactly 2x line width but it still did some weird compensation. Dunno which issue that is off hte top of my head.

BagelOrb:
Some more weirdness in the same print:
image

Also correct overlap compesnation:
image

PS:
It turns out this is a *really* good test model for overlap compensation. This model has it all! 
Ultimaker numba wan!

BagelOrb:
But don't put too much effort in this PR.
It was just an idea and I didn't really test it through.

jackha:
No accident here. In most cases I didn't see a difference, except for this weird Ultimaker numba wan case you provided. Here it's an improvement:

cura-5261-after
cura-5261-before

BagelOrb:
Oh sweet!!

I'm happy it was such an improvement! I didn't really expect that. Haha

PR #736, opened 2018-03-29
Modify combing to avoid supports

maximkulkin:
Combing feature has "avoid printed parts" option
which makes it go around other parts when travelling.
This patch also enables avoiding supports since they
are also kind of "printed".

The rationale here is that my printer (Creality Ender 2) has a problem
of hitting printed material when traveling. It might be related to little
nozzle block slack or not too tight belts which makes nozzle go slightly
down when traveling and thus hitting printed material. I know that it
should not happen but reality is that it does and I can't do much about
it (e.g. tightening belts results in much higher noise when moving which
is also undesired).
Combing "avoid printed parts" feature helps a lot. The only problem I have
is that it does not avoid supports which in some configurations (some models +
particular model orientation and layout) results in constantly hitting supports
which is noisy and in some cases can result in knocking supports off the bed.

This patch enables avoidance of supports as well as other printed parts which
solves that problem.

I understand that for some people this might not be a problem and increased
print time is unwanted. But at least it would be nice to have knobs to controls
that in cases when hardware is not perfect. I'm OK to rewrite it as a new
print setting.


Appesteijn:
CURA-5193

PR #740, opened 2018-04-06
Nix typo and untrue statement about Clipper

mesheldrake:
Clipper opperates on open paths now.
* operates
Key concept remains: slicing is prep for Clipper ops on polygons.

PR #741, opened 2018-04-06
Same fix as pull request #740

mesheldrake:


mesheldrake:
@BagelOrb
This is the same text fix but _in a different file_! Should have been clearer.
#740 updated /readme.md
This one, #741, updates the same text in /docs/overview.md

BagelOrb:
Ah ok

PR #748, opened 2018-04-22
Can now override the fan speed when printing the skin immediately above support

smartavionics:
Using a high fan speed to increase cooling appears to lessen the adhesion of the skin to the support and so that makes the support easier to remove. Useful when printing with materials like PETG.

Note that this PR is based on https://github.com/Ultimaker/CuraEngine/pull/738.


smartavionics:
BTW, with this PR I'm getting really good results with PETG over support even with 100% interface density. The support just peels away leaving a nice flat skin with no loose strands.

img_20180425_095618233

Ghostkeeper:
Developers, see issue CURA-5353.

smartavionics:
Hi Ian, I wouldn't bother looking at this PR until #738 has been accepted/rejected as this is based on that.

smartavionics:
OK, now for this PR... as I mention above, it's really good for printing PETG over support. Without 100% fan it can be very difficult to remove the support without damaging the part.

ianpaschal:
I'll be testing this now that the other PRs are taken care of.

Can you confirm though in the feedback above why the type was changed to float when it should be between 0-100 and not 0-1?

smartavionics:
Hi Ian, rather embarrassingly, I can't remember now why I changed it from double to float. It should be harmless though as the required range is 0-100 as you say (and -1 which gets used as a special value). Unless you have a big problem with the change, I would leave it as is for now.

ianpaschal:
It's fine. But it made me thing the change was not random and there was something to look out for there.

As long as the C++ is readable and the gcode prints well, I don't see it as an issue either.

ianpaschal:
Yeah high powered fan makes support a LOT easier to remove.

Spoke with @Ghostkeeper though and decided to use double instead of float since that's what we use almost everywhere else and because it's more precise than a float.

smartavionics:
LOL, you missed one, FffGcodeWriter::processSkinPrintFeature() still takes fan_speed as a float!

ianpaschal:
FML! I'll fix it tomorrow morning.

PR #749, opened 2018-04-22
Bridge wall mods

smartavionics:
This PR modifies the bridge wall code in two ways:

1 - it now assumes that at the start of a wall, the nozzle is fully pressurised. e.g. following a retract/prime sequence or a short move from the end of a preceding wall and so will use the maximum amount of coasting it can before the start of the first bridge segment. Before, it assumed that there was no pressure in the nozzle and so it did not coast before the start of the first bridge wall segment unless that was some distance from the start of the wall. Because it wasn't coasting when it should, the first bridge segment tended to droop more than it should have done.

2 - following a bridge segment it now immediately reverts the extrusion rate to the rate required by the line that follows the bridge but it still accelerates slowly so, in effect, it over-extrudes for a short period. This helps compensate for the underextrusion that occurs immediately following a bridge segment because the extruder can't increase the pressure in the nozzle quickly enough to keep up with the accelerating nozzle.


Ghostkeeper:
Developers, see issue CURA-5354.

alekseisasin:
Hi @smartavionics , could you please share the model/project, where the over-extrusion happens and I can print the model to see the difference.

smartavionics:
Hi @alekseisasin , these pictures show the effects of this PR. The first image is showing the old behaviour, the second image the behaviour with this PR. There are visible differences to the regions of the wall lines just before the bridge region starts and after the bridge region ends. The walls start in the top left corner and are printed anti-clockwise.

Without the PR there is no coasting immediately before the bridge region starts and that can cause the filament to dip low at the start of the bridge because the extruder pressure is still high. After the bridge region ends the line width is "normal" and this region will tend to be underextruded  because the extruder pressure lags.

screenshot_2018-07-22_07-50-45

With the PR, there's coasting before the bridge region starts and that will reduce the nozzle pressure which should make it less likely that the filament will dip low at the start of the bridge segment. The extrusion width is increased immediately following the end of the bridge section and this will reduce the underextrusion that normally happens here. It looks way too much in the image but when you look at the printed part it doesn't look overextruded. In fact, even with that extra flow at the start of the non-bridge line segment, there is generally underextrusion still visible. The temporary increase of the flow just reduces the underextrusion.

screenshot_2018-07-22_07-45-42

Here's the project file. I don't really recommend printing it as the bridge settings have not been tweaked for that filament so there's no guarantee that the bridge will print well even with the PR in place. 

[UMS5_multi-bridge.curaproject.3mf.zip](https://github.com/Ultimaker/CuraEngine/files/2216775/UMS5_multi-bridge.curaproject.3mf.zip)

alekseisasin:
Hi @smartavionics , Thank you for a good explanation. I have tested and it works. I will merge your pull request.

jackha:
In this bridging it doesn't produce the wanted output. In the "bend" it already does the extra extrusion... can you have a look @smartavionics ?

cura-5354

[bridging-rounding.curaproject.3mf.zip](https://github.com/Ultimaker/CuraEngine/files/2240987/bridging-rounding.curaproject.3mf.zip)

jackha:
Yeah I know that the banana isn't going to bridge well. By the extra extrusion I mean the increased extrusion width on the left in the middle. Is there any way that you can prevent that increased width in that part? If it's hard then we can leave it for now and see if people complain about it.

smartavionics:
Oh yeah, I see what you mean now. That shouldn't be there. I will look into it.

smartavionics:
OK, mystery solved. The fat wall lines are there because the individual line segments are shorter than the "Min Bridge Wall Length" setting (5mm in your profile). A quick workaround is to reduce that setting to 1 or 2 mm and then those line segments will be considered to be bridges. As those segments are in the middle of a bridge it would be more sensible if the min bridge length setting applied to the total length of the bridge rather than the lengths of the individual line segments. Future work...

jackha:
Thanks for the effort!

PR #750, opened 2018-04-23
Fix brim wall order bug

smartavionics:
Replacement for https://github.com/Ultimaker/CuraEngine/commit/2cdd76160f1c4574a20e29b560afd1f0b13d2c8d 

It's better to fix the bug in the optimizer than to avoid using the optimizer!

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -1531,7 +1531,6 @@ bool FffGcodeWriter::processInsets(const SliceDataStorage& storage, LayerPlan& g
         // Only spiralize the first part in the mesh, any other parts will be printed using the normal, non-spiralize codepath.
         // This sounds weird but actually does the right thing when you have a model that has multiple parts at the bottom that merge into
         // one part higher up. Once all the parts have merged, layers above that level will be spiralized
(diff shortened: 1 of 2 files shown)


diegopradogesto:
This is what I get working with your branch:
image

smartavionics:
Hmm, I haven't touched the brim code. That's weird...

diegopradogesto:
It happens in UM3 but not in Custom FDM printer. Maybe because other settings are affecting.

smartavionics:
OK, I will check using UM3 and get back...

smartavionics:
It still screws up even when the wall optimization is not enabled so it can't be the optimizer that's failing.

smartavionics:
246ed65c3733dc709d10966eeb07020beb02db72 is the culprit!

smartavionics:
Over to you @Ghostkeeper...

smartavionics:
I fixed the regression introduced in 246ed65 (see https://github.com/Ultimaker/CuraEngine/pull/752).

smartavionics:
Great, thanks!

diegopradogesto:
Thank you!

PR #751, opened 2018-04-24
Ignore cool_lift_head when a layer contains spiralized paths.

smartavionics:
Moving the nozzle away from the print at the end of each layer is incompatible
with spiralization, so don't do it.

Implements https://github.com/Ultimaker/Cura/issues/3700

### DIFF
--- a/src/LayerPlan.cpp
+++ b/src/LayerPlan.cpp
@@ -1195,6 +1195,8 @@ void LayerPlan::writeGCode(GCodeExport& gcode)
 
         bool update_extrusion_offset = true;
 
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
This is not the sort of solution we're looking for. If we were to merge this, the setting becomes broken. The user tries to enable it but nothing happens.

Normally Cura gives a warning colour when a setting is expected to give bad print results. We should implement that.

smartavionics:
> Normally Cura gives a warning colour when a setting is expected to give bad print results. We should implement that.

Over to you then...

PR #757, opened 2018-05-02
Improve outline accuracy  and reduce z-seam wobble by not discarding so many points

smartavionics:
A user on the forum was having problems with a wobbly z-seam and I have been looking into it. As far as I can tell, the problem is caused by the simplify() method discarding points that it really shouldn't do if the z-seam is to stand much chance of being straight. That method has two criteria to decide whether a point in a polygon outline can be removed while still preserving the shape of the polygon. (1) distance of the point from the points before and after and (2) distance of the point from the line that joins the before and after points (error distance).

One of the changes in this PR is to not use the error distance criteria when simplifying the outline polygons. This is enough to vastly improve the location of the z-seam, here's a couple of images that show the difference, first, the results for the original code:

screenshot_2018-05-02_14-22-05

and now the results for the PR code:

screenshot_2018-05-02_14-20-24

One downside of not throwing away points is that for models that have a lot of points, the slicing time is going to increase. If this PR is deemed unacceptably slow then one solution would be to introduce another setting that can go with Maximum Resolution (BTW, shouldn't this be called Minimum Resolution as it is specifying the min size of a line segment?) that can specify the max allowed error distance and that could be used to reduce the numbers of points.

Another change in this PR is that I modified the simplify() code so that it no longer accesses elements of a vector using invalid iterators. It may be that the original code does work OK but it can't be guaranteed to do so.

### DIFF
--- a/src/WallsComputation.cpp
+++ b/src/WallsComputation.cpp
@@ -74,8 +74,8 @@ void WallsComputation::generateInsets(SliceLayerPart* part)
             }
         }
 
-        //Finally optimize all the polygons. Every point removed saves time in the long run.
-        part->insets[i].simplify();
+        //Previously, simplify() was called here with the default params but as the polygons have already been simplified using meshfix_maximum_resolution
+        //it's better to not simplify again as it can introduce z-seam wobble
         part->insets[i].removeDegenerateVerts();
         if (i == 0)
         {
--- a/src/slicer.cpp
+++ b/src/slicer.cpp
@@ -779,7 +779,8 @@ void SlicerLayer::makePolygons(const Mesh* mesh, bool keep_none_closed, bool ext
 
     //Finally optimize all the polygons. Every point removed saves time in the long run.
     const coord_t line_segment_resolution = mesh->getSettingInMicrons("meshfix_maximum_resolution");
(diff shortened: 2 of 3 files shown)


smartavionics:
Yes, this PR does have a bad effect on slicing speed when the model has a lot of outlines but, then again, it does actually produce a more accurate result. Here's the same part sliced with the same resolution 0.04 with and without the error distance criteria.

When using the criteria (original code) the slice took around 35 seconds and the output looked like this:

screenshot_2018-05-02_15-56-36

When not using the error distance criteria the slice took around 180 seconds (much longer) but the output quality is noticeably improved:

screenshot_2018-05-02_15-55-08

smartavionics:
Damn, pressed the wrong button, reopened!

smartavionics:
However, playing more with the Maximum Resolution setting, you can trade off slicing time with output fidelity. With a resolution of 0.25mm, the slicing time is around 66 seconds and the output still looks much better than the result obtained with this PR so I think that perhaps the slowdown is acceptable given that the user can adjust the resolution.

screenshot_2018-05-02_16-13-59

smartavionics:
Hmm, not sure now about this one, going to think about it some more...

PR #758, opened 2018-05-03
Rewrite polygon simplify() to avoid illegal vector access

smartavionics:
I noticed that the simplify() method was being naughty because it was accessing the elements of a vector using iterators after elements had been removed from the vector using erase(). The C++ standard says that after an element is erased from a vector, any pointers or iterators that reference a later element in the vector are no longer valid.

This PR also has a little change in WallsComputation::generateInsets() in that it no longer calls simplify() because the slicer has already done the simplification using the resolution specified by the user.

Finally, why is the setting called "Maximum Resolution" when it's actually specifying the minimum line segment length? The blurb says "The minimum size..." so I find the setting name very confusing.

### DIFF
--- a/src/WallsComputation.cpp
+++ b/src/WallsComputation.cpp
@@ -74,8 +74,8 @@ void WallsComputation::generateInsets(SliceLayerPart* part)
             }
         }
 
-        //Finally optimize all the polygons. Every point removed saves time in the long run.
-        part->insets[i].simplify();
+        //Previously, simplify() was called here with the default params but as the polygons have already been simplified using meshfix_maximum_resolution
+        //it's better to not simplify again as it can introduce z-seam wobble
         part->insets[i].removeDegenerateVerts();
         if (i == 0)
         {
--- a/src/utils/polygon.cpp
+++ b/src/utils/polygon.cpp
@@ -411,7 +411,7 @@ void PolygonRef::simplify(int smallest_line_segment_squared, int allowed_error_d
                 if ( vSize2(here - prev) < min_length_2 && vSize2(next - here) < min_length_2 )
                 {
                     // don't add [here] to the result but add the next point
-                    if (++poly_idx < size())
+                    if (++poly_idx < result_list_poly.size())
                     {
                         prev = next;
                         new_result_list_poly.emplace_back(next);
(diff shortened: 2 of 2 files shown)


BagelOrb:
`erase` wasn't called on a vector!
The fact that we used `erase` was the whole reason for using a `ListPolygon` which is list-based, rather than a `Polygon` which is vector based.

BagelOrb:
According to my reasoning this PR should therefore be closed.
We can always reopen if my reasoning is wrong.

smartavionics:
I accept that this PR is based on a false assumption but, interestingly, it gives different results to the original. If I can be bothered I will go back and see where that difference is coming from and whether there is still some other issue to be fixed.

BagelOrb:
That difference causes me quite some concern. We tested the heck out of this function back when we introduced it. Also we had a lot of trouble with the old simplify. Simplification is quite difficult! :P

BagelOrb:
I'm not sure about the call to `simplify()` which was superfluous according to you.
That might be a separate PR...

smartavionics:
OK, tomorrow I will check my code again and see why it gives different results. It could easily be a flaw in my code, I'll let you know.

smartavionics:
I think that call to simplify() with default params is bad because at a higher level the polygons have already been simplified using the user's desired values so, potentially, they could have gone for a higher resolution which is then zapped by that other call to simplify().

smartavionics:
The differences in the output from the original simplify() and this new version are very small, most just a few microns so I don't really think it's worth spending any more time on this as the code doesn't need to be used so I am abandoning it.

BagelOrb:
The maximum resolution feature has been implemented a bit sloppy. `simplify` is called in all sorts of places and with different parameters now!

Somebody needs to fix this.
Do you want to 'bother'? I'm afraid that asking the Cura team won't really get a response any time soon... :S

I've had a look at their TODO list; I understand why they have become so unresponsive lately. There's a ton of pull requests, bug fixes and required functionality on their plate!

smartavionics:
Having visited them back in Feb, I understand how busy they all are so I am reluctant to increase the workload. I'll leave it for now.

BagelOrb:
Oh wait, no I was looking at an old state where the maximum resolution feature wasn't finished yet.

BagelOrb:
The extra `simplify` at the location you were talking about just simplifies new line segments smaller than 10 micron which were introduced due to the offset from the outline. They may stay in.

So yeah, let's not bother with another PR ;)

Ghostkeeper:
> The maximum resolution feature has been implemented a bit sloppy. simplify is called in all sorts of places and with different parameters now!

I have no qualms with `simplify` being called in various places. It removes vertices which can really improve performance.
I do agree though that the Maximum Resolution feature should probably have been implemented in the planning stage to remove lines after they have been generated, rather than at the beginning of the process...

PR #759, opened 2018-05-05
Important bug fix for regression in 3.3 that can vastly increase combed travel distances

smartavionics:
Commit 9c688e63d changed the result type of ConstPolygonRef::size() from unsigned int to size_t. Unfortunately, this badly broke the combing travel direction calculation. This PR provides a fix.

Personally, I think that changing size() to return a size_t is a mistake because we are never going to need more than 32 bits to represent the number of points in a polygon and it just makes the code bigger and slower.

I have also changed another couple of very similar expressions just in case they are similarly broken.

### DIFF
--- a/src/pathPlanning/LinePolygonsCrossings.cpp
+++ b/src/pathPlanning/LinePolygonsCrossings.cpp
@@ -146,7 +146,7 @@ void LinePolygonsCrossings::getBasicCombingPath(PolyCrossings& polyCrossings, Co
 {
     ConstPolygonRef poly = boundary[polyCrossings.poly_idx];
     combPath.push_back(transformation_matrix.unapply(Point(polyCrossings.min.x - std::abs(dist_to_move_boundary_point_outside), transformed_startPoint.Y)));
-    if ( ( polyCrossings.max.point_idx - polyCrossings.min.point_idx + poly.size() ) % poly.size() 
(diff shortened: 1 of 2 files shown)


smartavionics:
BTW, I think this is bad enough to warrant a 3.3.2 release.

smartavionics:
I've figured out why it was going wrong - polyCrossings.max.point_idx - polyCrossings.min.point_idx was sometimes producing a negative value but because both those are of 32 bit unsigned type, the result was also a 32 bit unsigned type and when that value was converted to 64 bits to be  added to the value of poly.size(), the conversion didn't sign extend through the top 32 bits (because both types are unsigned the top 32 bits are set to zero) and the resulting value came out close to the maximum value for a 32 bit unsigned number.

E.g if polyCrossings.max.point_idx was, say 5 and polyCrossings.min.point_idx was, say, 8, the result of the subtraction would be -3 if considered as signed but because those values are unsigned, it would be 0xfffffffd which would then be zero extended to 64 bits to give 0x00000000fffffffd which would then be added to poly.size().

jackha:
@smartavionics thanks for investigating and providing a fix in such a short notice! I hope we can keep up. Devs: CURA-5333

jackha:
Thank you!

PR #765, opened 2018-05-09
Fix long standing bug that would cause fan speed gcode to be duplicated for non-integer fan speeds.

smartavionics:
The current fan speed was being stored as an int but was being compared to the new value
which is a double so the comparison would often fail when it shouldn't have done and so
the fan speed gcode was output when it need not have been.


BagelOrb:
Fixed with extra robustness: https://github.com/Ultimaker/CuraEngine/commit/9472f6e5897df235bcef61eefd89c75d98faed08

BagelOrb:
Thanks!

PR #768, opened 2018-05-11
Ensure lib_CuraEngine is static.

krop:
Packagers usually use CMake macros in their build system to set default values
that work in most cases.
If BUILD_SHARED_LIBS is ON or BUILD_STATIC_LIBS is OFF, the CuraEngine
executable won't run.


Ghostkeeper:
Yeah, that library was never meant for dynamic linking. Thanks.

PR #767, opened 2018-05-11
Multiply infill

BagelOrb:
Very much wow.

Self-explanatory (hopefully).

image
image
image
image
image
concentric_3d_multiply_infill
cross_3d_multiply_infill

PR #769, opened 2018-05-12
fix: zigzaggify should put connections next to border, not on it

BagelOrb:
These fixes are cherry-picked from https://github.com/Ultimaker/CuraEngine/pull/767


BagelOrb:
Before:
image

After:
image

BagelOrb:
The problem seems mostly fixed by a commit by you: 5fa54dcfdb94b1eae0c443715cfb4b943f2b242c

BagelOrb:
Your fix is incorrect and introduces problems for cross infill.

The outline_offset is defined as the offset to obtain the outline of the area to fill with infill, so it's the Infill class who should be respondible for performing the correct offset, it should not be the caller who is responsible for fixing zigzag infill by providing different parameters.

BagelOrb:
Before:
image


After:
image

jackha:
Devs, see: CURA-5397

ianpaschal:
Code looks fine to me. Second opinion @Ghostkeeper?

PR #771, opened 2018-05-14
When build plate adhesion is brim, don't fuzzy the skin on the first layer.

smartavionics:
Having fuzzy skin on the first layer greatly reduces the bonding between the brim and the skin.

Is it really worth introducing another setting to enable this tweak? I don't think so but, then again, I don't use fuzzy skin and don't often use brim so, personally, I don't care.

Provides a fix for https://github.com/Ultimaker/Cura/issues/3807.


BagelOrb:
Oh oops. I didn't see you already had a PR for it.
+1

https://github.com/Ultimaker/CuraEngine/commit/657609e74387349c5a163adf2fd4a2eaa05a9811

smartavionics:
Why am I bothering?

BagelOrb:
Because it's low hanging fruit. It's hanging so low that it's touching the ground! Nobody can resist that!

BagelOrb:
I would have accepted your PR if I would have seen it earlier than the issue, you know.

PR #779, opened 2018-05-20
Small features slowdown

smartavionics:
Implements https://github.com/Ultimaker/Cura/pull/3837 so that small features (small holes and other outlines) will be printed slower.

Note that this PR is based on https://github.com/Ultimaker/CuraEngine/pull/721.


Ghostkeeper:
What is it about small features that they need to be printed slower? Why do small features break?

It seems to me that this issue could just be fixed by setting the jerk/acceleration a bit lower instead.

smartavionics:
> What is it about small features that they need to be printed slower? Why do small features break?

This is mostly about first layer adhesion but having implemented this I don't think it hurts for higher layers. I think what is happening is that when the radius is very small, the extruder pulls the still molten line it has just printed from the side (rather than in the direction just printed) and that tends to make it come unstuck. So by printing really slow, there's more time for the just printed line to harden a little.

> It seems to me that this issue could just be fixed by setting the jerk/acceleration a bit lower instead.

I think that changing accel/jerk is to be avoided as much as possible and you certainly wouldn't want to make that change for the whole print.

Here's an example object that I could not print on my mini Kossel without this PR working:

screenshot_2018-05-25_14-05-34

I normally print on blue tape and get excellent adhesion but still needed to print those small outlines even slower than normal.

smartavionics:
Slic3r has had a similar feature for years and I know that people have been asking for it in Cura for quite a while now. I didn't think about implementing it until the day I found I needed it!

smartavionics:
Here's the stl for the above model if you want to try it.

[magni-pitch-dial.zip](https://github.com/Ultimaker/CuraEngine/files/2042745/magni-pitch-dial.zip)

jackha:
Devs: CURA-5503

smartavionics:
Sorry, no, it didn't make it into 3.6.

damon-atkins:
@jackha  what is happen with this helpful PR request?

andreypopov:
also want to know

smartavionics:
This PR was available for 2 months and the Cura devs did nothing with it so I closed it. I currently have other PRs that have been hanging around for months and they too will get closed soon. The Cura team are only interested in sexy "fashion features" like the gyroid infill. Boring PRs that simply improve the print quality or fix a regression or bug are considered very low priority as they don't have the same marketing appeal. I am no longer submitting any new features to Cura, it's simply not worth the effort.

damon-atkins:
I would assume this comment means its been tracked `Devs: CURA-5503` on some internal system.  However it does come across as being difficult process to get a PR accepted.

smartavionics:
People still want this so reopening.

Ghostkeeper:
Developers, see the new ticket CURA-6464, created after re-opening.

rburema:
I can clearly see a (positive) difference in the models I tested with. Thanks again @smartavionics !

(You can see a sort of line where it switches over, but that's way less noticable than the blobs in the control model.)

PR #781, opened 2018-05-21
Fix no skin combing so that it doesn't cross skin areas!

smartavionics:
The problem was that previously it considered the part's infill area(s) to define the
boundary of the combing area but then travel moves that didn't actually cross the infill
areas were considered to be OK to do with no retraction and so those travels could cross
skin areas!

So now, it adds the part's walls into the combing area so that any travel moves that cross
the part's outline will be combed correctly.

Provides a solution for https://github.com/Ultimaker/Cura/issues/3840.

Without this PR:

screenshot_2018-05-21_14-32-25

With this PR:

screenshot_2018-05-21_18-57-47

This PR will require quite a bit of testing but a fossil fish printed on the S5 using this PR showed no artifacts that were obviously due to this change.

### DIFF
--- a/src/pathPlanning/LinePolygonsCrossings.cpp
+++ b/src/pathPlanning/LinePolygonsCrossings.cpp
@@ -153,7 +153,9 @@ void LinePolygonsCrossings::generateBasicCombingPath(PolyCrossings& polyCrossing
     std::vector<Point> fwd_points;
     Point prev = combPath.back();
     coord_t fwd_len = 0;
-    for (unsigned int point_idx = polyCrossings.min.point_idx; point_idx != polyCrossings.max.point_idx; point_idx = (point_idx < poly.size() - 1) ? (point_idx + 1) : (0))
+    for (unsigned int point_idx = polyCrossings.min.point_idx
+        ; point_idx != polyCrossings.max.point_idx
+        ; point_idx = (point_idx < poly.size() - 1) ? (point_idx + 1) : (0))
     {
         const Point p = PolygonUtils::getBoundaryPointWithOffset(poly, point_idx, dist_to_move_boundary_point_outside);
         fwd_points.push_back(p);
@@ -167,7 +169,9 @@ void LinePolygonsCrossings::generateBasicCombingPath(PolyCrossings& polyCrossing
     coord_t rev_len = 0;
     unsigned int min_idx = (polyCrossings.min.point_idx == 0)? poly.size() - 1 : polyCrossings.min.point_idx - 1;
     unsigned int max_idx = (polyCrossings.max.point_idx == 0)? poly.size() - 1 : polyCrossings.max.point_idx - 1;
-    for (unsigned int point_idx = min_idx; point_idx != max_idx; point_idx = (point_idx > 0) ? (point_idx - 1) : (poly.size() - 1))
+    for (unsigned int point_idx = min_idx
+        ; point_idx != max_idx
+        ; point_idx = (point_idx > 0) ? (point_idx - 1) : (poly.size() - 1))
     {
         const Point p = PolygonUtils::getBoundaryPointWithOffset(poly, point_idx, dist_to_move_boundary_point_outside);
         rev_points.push_back(p);
--- a/src/pathPlanning/LinePolygonsCrossings.cpp
+++ b/src/pathPlanning/LinePolygonsCrossings.cpp
@@ -107,7 +107,7 @@ bool LinePolygonsCrossings::lineSegmentCollidesWithBoundary()
 }
 
 
-bool LinePolygonsCrossings::getCombingPath(CombPath& combPath, int64_t max_comb_distance_ignored, bool fail_on_unavoidable_obstacles)
+bool LinePolygonsCrossings::generateCombingPath(CombPath& combPath, int64_t max_comb_distance_ignored, bool fail_on_unavoidable_obstacles)
 {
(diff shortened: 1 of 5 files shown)


ChrisTerBeke:
You're pumping our PR's at high speed man, kudos!

I hope we can find the time to start reviewing all of em.

smartavionics:
I have added another change to this PR - it now chooses the path to take around a polygon based on the shortest distance travelled rather than the smallest number of segments traversed. This fix isn't specific to the no skin case but will improve all combing. The effect is modest but for some models can make a difference. It didn't seem worth creating a new PR for this as it can be tested together with the no-skin fix.

qwerty8224:
**smartavionics** 
Thank you for your efforts and efforts and the desire to make this product which many enjoy better. )))

smartavionics:
Thanks @qwerty8224 , I'm glad you appreciate my contributions. Thanks should also go to UM for continuing to support open source software. It can't be easy for them balancing their commercial needs and priorities against encouraging community contributions.

qwerty8224:
yes the Ultimaker team is working very well they made a big jerk for which they too thank you. But sometimes they have their priorities over work. )

jackha:
Many thanks for contributing and having patience with us :-)
Devs: CURA-5505

smartavionics:
Updated so that it copes with thin walls better - see https://community.ultimaker.com/topic/23836-cura-341-is-combing-realy-working/

src/pathPlanning/LinePolygonsCrossings.cpp
```diff
@@ -144,29 +144,59 @@ void LinePolygonsCrossings::getBasicCombingPath(CombPath& combPath)
 
 void LinePolygonsCrossings::getBasicCombingPath(PolyCrossings& polyCrossings, CombPath& combPath)
```
alekseisasin: I know the method name "getBasicCombingPath" is there for a long time, usually getters return value. What about rename to "generateBasicCombingPath"

src/pathPlanning/LinePolygonsCrossings.cpp
```diff
@@ -144,29 +144,59 @@ void LinePolygonsCrossings::getBasicCombingPath(CombPath& combPath)
...
-            ; point_idx = (point_idx < poly.size() - 1) ? (point_idx + 1) : (0))
-        {
-            combPath.push_back(PolygonUtils::getBoundaryPointWithOffset(poly, point_idx, dist_to_move_boundary_point_outside));
-        }
+
+    // follow the path in the same direction as the winding order of the boundary polygon
+    std::vector<Point> fwd_points;
+    Point prev = combPath.back();
+    coord_t fwd_len = 0;
+    for (unsigned int point_idx = polyCrossings.min.point_idx; point_idx != polyCrossings.max.point_idx; point_idx = (point_idx < poly.size() - 1) ? (point_idx + 1) : (0))
```
alekseisasin: I know the for loop works, can you make multiple lines of instructions ...( as was before) or simplify
 the "point_idx = (point_idx < poly.size() - 1) ? (point_idx + 1) : (0)"

src/settings/settings.cpp
```diff
@@ -490,6 +490,10 @@ CombingMode SettingsBaseVirtual::getSettingAsCombingMode(std::string key) const
     {
         return CombingMode::NO_SKIN;
     }
+    if (value == "infill")
```
alekseisasin: we also need to add an extra option for the "retraction_combing" setting

src/pathPlanning/LinePolygonsCrossings.cpp
```diff
@@ -144,29 +144,59 @@ void LinePolygonsCrossings::getBasicCombingPath(CombPath& combPath)
...
-        unsigned int max_idx = (polyCrossings.max.point_idx == 0)? poly.size() - 1: polyCrossings.max.point_idx - 1;
 
-        for(unsigned int point_idx = min_idx; point_idx != max_idx; point_idx = (point_idx > 0) ? (point_idx - 1) : (poly.size() - 1))
+    // follow the path in the opposite direction of the winding order of the boundary polygon
+    std::vector<Point> rev_points;
+    prev = combPath.back();
+    coord_t rev_len = 0;
+    unsigned int min_idx = (polyCrossings.min.point_idx == 0)? poly.size() - 1 : polyCrossings.min.point_idx - 1;
+    unsigned int max_idx = (polyCrossings.max.point_idx == 0)? poly.size() - 1 : polyCrossings.max.point_idx - 1;
+    for (unsigned int point_idx = min_idx; point_idx != max_idx; point_idx = (point_idx > 0) ? (point_idx - 1) : (poly.size() - 1))
```
alekseisasin: Same, "point_idx = (point_idx > 0) ? (point_idx - 1) : (poly.size() - 1)"

src/pathPlanning/LinePolygonsCrossings.cpp
smartavionics: I don't mind. I can do that.

src/pathPlanning/LinePolygonsCrossings.cpp
smartavionics: Really? OK, if it will make you happy.

src/settings/settings.cpp
smartavionics: Don't know anything about "retraction_combing" setting.

src/pathPlanning/LinePolygonsCrossings.cpp
smartavionics: As above.

smartavionics:
OK, done.

PR #782, opened 2018-05-23
[WIP] Optimize path planning for parts

rickyzhang82:
# Why
Current path order optimization algorithm is sub par optimizer for parts (a.k.a island).

1. For parts order optimization, it didn't take into account of an entry point and an exit point of the parts. 
1. Used sub par nearest neighbor algorithm.

# How
Assuming we ignore the path planning within the parts, solving the path planning optimization for parts is an asymmetry traveling salesman problem (ATSP).

With commercial and the best academic mixed integer linear programming (MILP) solver, we can easily solve 100 cities complete graph of TSP within less than 0.5 seconds given the improvement of branch and cut algorithm and beefy machine.

To improve Cura engine path planning for parts, I proposed to replace the current algorithm with the following change:

1. Change the optimization problem from TSP into ATSP. 
1. Try on different ATSP algorithm either exact or heuristic.  

# What

I'm not an OR expert. I don't plan to reinvent the wheel unless it is a must. I already experimented commercial and academic MILP solver (like SAS, CbC and Concorde). They did easily solve the small size problem efficiently in exact TSP (except CbC).

The best exact TSP solver is [Concorde](http://www.math.uwaterloo.ca/tsp/concorde.html). However, the owner denied my request for use in Cura under GPL.

Currently, I looked into heuristic approach like 3-OPT and more sophisticated  [LKH](http://www.akira.ruc.dk/~keld/research/LKH/)

# Progress
I forked my work under legacy branch. This is due to the fact that I'm lazy. I prefer work on something constant and small. But after reading master branch code I found that there is little change in path planning for parts. It should be easily merged back from legacy branch to master branch.

I did several improvement so far:

- Added CMake
- Added a better logging G3LOG.
- Output parts as closed polygons in the form of text file defined by the format [here](https://gist.github.com/rickyzhang82/33831c6d5a3eaaa3de3ffb5122f15b69). This should open an opportunity for OR people to look into the problem.
- Created a [visualization tool](https://github.com/rickyzhang82/CuraEngine/tree/feature-optimize-path-planning-for-parts/utils) for viewing parts by layer.
 
What's next:
- Output starting point and exit point of parts per each layer.
- Implement ATSP solver either exact or heuristic.
- If everything looks OK, merge back to master branch.


rickyzhang82:
For those reading legacy code, here is class UML I generated by a Clang tool.
uml

rickyzhang82:
I decided to withdraw this PR and move on to work on it under my repository.

PR #783, opened 2018-05-24
Bug fix - use the correct line width when adding perimeter gaps between skin walls.

smartavionics:
It was using the inner wall line width when it should have been using the skin wall line width. To see the difference, use a narrower line width for inner walls than for skin and then specify 2 or more skin walls and you will see the gaps between those skin walls get gap filled (when they shouldn't).


smartavionics:
Looks like this is no longer required due to https://github.com/Ultimaker/CuraEngine/commit/57eca0e773f41d94fca3813505d5045886606511

PR #791, opened 2018-06-06
Now getInnermostWalls() will use 1st wall only for those regions where 2nd wall is missing.

smartavionics:
When a part narrows so that there is only enough room for the outer wall, the combing code
fell back to use the 1st wall as the combing region boundary when combing mode was "all"
rather than using the 2nd wall. This new code recognises when the 2nd wall has missing regions
and builds a polygon that is the combination of the regions of the 2nd wall that are present
along with regions from the outer wall as replacements for the missing 2nd wall regions.

This keeps the combing boundary away from the outer wall as much as possible.

Provides a fix for https://github.com/Ultimaker/Cura/issues/3890

Without this PR:

screenshot_2018-06-06_13-45-25

With this PR:

screenshot_2018-06-06_13-44-19

### DIFF
--- a/src/LayerPlan.cpp
+++ b/src/LayerPlan.cpp
@@ -162,7 +162,7 @@ Polygons LayerPlan::computeCombBoundaryInside(CombingMode combing_mode, int max_
             }
             else
             {
-                layer.getInnermostWalls(comb_boundary, max_inset);
+                layer.getInnermostWalls(comb_boundary, max_inset, mesh);
             }
         }
         return comb_boundary;
--- a/src/infill/SubDivCube.cpp
+++ b/src/infill/SubDivCube.cpp
@@ -227,7 +227,7 @@ int SubDivCube::distanceFromPointToMesh(SliceMeshStorage& mesh, int layer_nr, Po
         *distance2 = 0;
     }
     Polygons collide;
-    mesh.layers[layer_nr].getInnermostWalls(collide, 2);
+    mesh.layers[layer_nr].getInnermostWalls(collide, 2, mesh);
     Point centerpoint = location;
     bool inside = collide.inside(centerpoint);
     ClosestPolygonPoint border_point = PolygonUtils::moveInside2(collide, centerpoint);
--- a/src/sliceDataStorage.cpp
+++ b/src/sliceDataStorage.cpp
@@ -70,13 +70,13 @@ void SliceLayer::getOutlines(Polygons& result, bool external_polys_only) const
     }
 }
 
-void SliceLayer::getInnermostWalls(Polygons& layer_walls, int max_inset) const
+void SliceLayer::getInnermostWalls(Polygons& layer_walls, int max_inset, const SliceMeshStorage& mesh) const
 {
     for (const SliceLayerPart& part : parts)
     {
         switch (max_inset) {
             case 1:
-                // take the inner wall
+                // take the 1st wall
                 if (part.insets.size() >= 1) {
                     layer_walls.add(part.insets[0]);
                     continue;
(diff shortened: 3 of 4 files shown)


diegopradogesto:
For Devs.: CURA-5481

smartavionics:
Hi @LipuFei , thanks for looking at this PR. Just this evening I have modified it so that it can cope with the situation where the walls are so thin that the combing region would otherwise disappear completely. This solves the problem discussed in https://community.ultimaker.com/topic/23836-cura-341-is-combing-realy-working/

Sorry for the late change, please continue testing this.

smartavionics:
OK, let me look at it for a few minutes and if I can't see how it can be made quicker, you can revert it.

smartavionics:
One of the issues here is that someone else changed the combing boundary code to call computeCombBoundaryInside() twice for each layer and so that means it's doing twice as much work as it needs to.

smartavionics:
I think you should revert this PR and I'll just use it myself.

smartavionics:
The new implementation of the "no skin" combing mode is going to be similarly slow. I hope we can keep it  because if people find it too slow they can choose to use the old implementation (now known as "within infill").

rburema:
We'll have a look at it. We'd very much like to have this resolved before the 3.5 full release, preferably without the removal of any existing improvements. Devs: CURA-5765

smartavionics:
Another thing I noticed is that print that's really slow uses an infill that calls getInnermostWalls() which could almost certainly be made quicker by just using the part's outer boundary?

smartavionics:
What I have learnt so far.

If I make a change like shown below, the slowdown is not as bad but it's still slower. I get a slice time of around 37 seconds for Cura 3.4 and 55 seconds for 3.5 beta. Without this change, it's much slower still. I don't really know what the original author of distanceFromPointToMesh() was trying to achieve so I can't be sure that this change is actually right.

````
diff --git a/src/infill/SubDivCube.cpp b/src/infill/SubDivCube.cpp
index 2bf847f6..c85acd8f 100644
--- a/src/infill/SubDivCube.cpp
+++ b/src/infill/SubDivCube.cpp
@@ -227,7 +227,7 @@ int SubDivCube::distanceFromPointToMesh(SliceMeshStorage& mesh, int layer_nr, Po
         *distance2 = 0;
     }
     Polygons collide;
-    mesh.layers[layer_nr].getInnermostWalls(collide, 2, mesh);
+    collide = mesh.layers[layer_nr].getOutlines(false);
     Point centerpoint = location;
     bool inside = collide.inside(centerpoint);
     ClosestPolygonPoint border_point = PolygonUtils::moveInside2(collide, centerpoint);
````

smartavionics:
Hello again.

Further experimentation shows that getInnermostWalls() is not a major slow down. With the above patch in place to avoid calling getInnermostWalls() from the infill generation, I see only 2 or 3 seconds of slow down between the current getInnermostWalls() and a "simplified" version that simply returns part.outline. i.e. don't don't any of the polygon manipulation. So the slice time goes 56 to 53 when doing that. The 3.4 slicer can process the same job in 37 seconds so there is some time going somewhere but it's not getInnermostWalls() that is the major culprit here.

rburema:
Hi @smartavionics, Thanks for your continued assistance. It seems right now that this issue is blocked until we've solved another issue related to slicing time, which was introduced more recently. After that is resolved (hopefully _very_ soon), this'll have my (or someone else's) full attention. cheers, Remco.

smartavionics:
Hi, I'm at the TCT show today so can't really respond much. Please try out
the PR I created yesterday as it makes that destroyer print much quicker.

rburema:
Hello again @smartavionics,

I had a look at the PR you mentioned in the previous comment, and specifically at the line you also changed in the comment in this thread that starts with _'What I have learnt so far'_. I asked around (being pretty new here) and it turns out that you'd really _do_ need the inner walls there.

My tests with your PR, that line and subsequent discussion did point me towards a (/the?) solution however, which is to cache the results of `getInnerMostWalls`. PR for that is #859. If that goes through, it seems we can keep this in as is, but without the slowdown experienced in 3.5-BETA by users that needed cube subdivision.

cheers, Remco.

PR #795, opened 2018-06-13
Fix intpoint

BagelOrb:
The type of int used in Point3 was 32bit instead of 64 bit!!

Also I cleaned some related int types problems.

I also refactored coord_t into its own class.


Ghostkeeper:
I saw this when I moved the class to a separate file recently. I thought it was probably no problem if we didn't do multiplication of these coordinates or distance calculation. But honestly, we probably do. Or at least interpolation which could also overflow.

Developers, see issue CURA-5475.

PR #797, opened 2018-06-18
Now passes machine_cooling_fan_number as parameter to M106/M107.

smartavionics:
Until now, M106/M107 (cooling fan speed control) didn't get a P parameter and so defaulted
to fan 0. Now it can be specified through the machine_cooling_fan_number setting.

### DIFF
--- a/src/gcodeExport.cpp
+++ b/src/gcodeExport.cpp
@@ -41,6 +41,8 @@ GCodeExport::GCodeExport()
 
     extruder_count = 0;
 
+    fan_number = 0;
+
     total_bounding_box = AABB3D();
 }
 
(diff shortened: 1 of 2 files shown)


jackha:
Devs: CURA-5502

ianpaschal:
I spoke with @alekseisasin about this PR pair. We're OK to merge it but lately the whole team has been mentioning the complexity of Cura Engine's code and its danger of eventually becoming an uninterpretable myth. For this reason, we'd like to see a bit more documentation/commenting in the code explaining what this setting is (easy to confuse with an extruder number) and where it's coming from (never set in the UI unless a certain plugin is installed).

smartavionics:
No problem, I only produced these PRs because someone on the forum needed a solution but then they came up with another fix in their printer firmware setup so this feature is now redundant. So let's just close these PRs and forget about them!

ianpaschal:
Hmm OK. Agreed in that case. Cura team has been cursed with an evil spell that makes an incision in our skin every time another setting is added to Cura and it's getting painful.

PR #800, opened 2018-06-27
Avoid crashing in optimizingInsetsIsWorthwhile() when part has no walls.

smartavionics:
When a part has no walls, it just returns false.

This fixes the crash reported in https://github.com/Ultimaker/Cura/issues/3983.


LipuFei:
@smartavionics Thanks for the fix!

PR #803, opened 2018-07-04
fix slowdown regression introduced in #738

smartavionics:
Use explicit const reference types for loop variables rather than auto to avoid costly construction of temp objects.

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -1467,7 +1467,7 @@ bool FffGcodeWriter::processInsets(const SliceDataStorage& storage, LayerPlan& g
             {
                 if (m.isPrinted())
                 {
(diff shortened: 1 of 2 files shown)


jackha:
The slowness is fixed!

ianpaschal:
Smort.

Also a nice lesson for me to keep an eye on slicing time when testing engine stuff. I'm carefully inspecting the resulting gcode but I think pretty much ignoring slicing time as long as it doesn't hang for 5+ minutes.

smartavionics:
Yes, it's real easy to miss slowdowns when testing with smallish models which is what I tend to do because I want a quick turnaround when slicing many times while testing.

Appesteijn:
Tested on MacOS, looks good

PR #809, opened 2018-07-23
CURA-5455 Add addPolylines() for line-based infills to fix tiny movements in infill.

LipuFei:
Line-based infill patterns generate polylines which contain connected line segments. `addLinesByOptimizer()` reorders all lines and treats them as individual lines with extra wiping extrusions, and this can cause tiny movements. `addPolylines()` adds each polyline continuously and only add an extra wiping extrusion (if needed) in the end of each polyline.


LipuFei:
@smartavionics Make sense. I was a bit afraid to break other things by modifying `addLinesByOptimizer()`. I will try some simpler solutions.

smartavionics:
I just realised that my example above is wrong because I have initialised last_end to a pair of huge values so it will try and do an extremely long wipe before the first point! So don't try the example above without modification.

I think it's worth continuing to use the LineOrderOptimizer because it works hard to join up the individual lines into chains and that really cuts down on the travels required. So if you just modify the wipe code to something like what I have above (without the long wipe bug) it may be sufficient. Worth trying I think. Cheers.

LipuFei:
Closing this in favour of  #811

PR #814, opened 2018-07-26
Improve processing of overhanging walls

smartavionics:
The rational behind this PR is to separate the handling of overhung walls from the bridge handling. This allows overhung walls to be detected (and potentially printed with different speed) without having the bridge code enabled.

Detecting overhung walls is beneficial because it allows the wall start vertex to be moved if it is in an overhang region. Without doing that, poor print quality may result.

See https://github.com/Ultimaker/Cura/issues/3875 for a (bad tempered) discussion of this problem!

This PR makes the following changes:

1 - avoids starting outer wall lines over an overhung region. Instead, the wall line will start on the first vertex following the overhung region.

2 - the portions of an outer wall line that are considered to be overhung may be printed at a lower speed than normal.

3 - removes the wall overhang detection in the bridge code as it is no longer required.


diegopradogesto:
Hey @smartavionics thanks again for your contribution. We will take a look at it soon.
Devs. CURA-5633

src/FffGcodeWriter.cpp
```diff
@@ -1534,27 +1535,53 @@ bool FffGcodeWriter::processInsets(const SliceDataStorage& storage, LayerPlan& g
...
+                gcode_layer.setOverhangMask(Polygons());
+            }
+            else
+            {
+                // the overhang mask is set to the area of the current part's outline minus the region that is considered to be supported
+                // the supported region is made up of those areas that really are supported by either model or support on the layer below
+                // expanded to take into account the overhang angle, the greater the overhang angle, the larger the supported area is
+                // considered to be
+                double overhang_width = layer_height * std::tan(overhang_angle / (180 / M_PI));
+                Polygons overhang_region = part.outline.offset(-half_outer_wall_width).difference(outlines_below.offset(10+overhang_width-half_outer_wall_width)).offset(10);
```
alekseisasin: Can you please explain this calculation, the number 10?

src/LayerPlan.cpp
```diff
@@ -878,7 +871,40 @@ void LayerPlan::addWalls(const Polygons& walls, const GCodePathConfig& non_bridg
...
+    if (bridge_wall_mask.empty() && overhang_mask.empty())
+    {
+        return start_idx;
+    }
+
+    Polygons air_below(bridge_wall_mask.unionPolygons(overhang_mask));
+
+    unsigned curr_idx = start_idx;
+
+    for(;;)
```
alekseisasin: I know it is not wrong, but can make use "while" loop

alekseisasin:
It seems work, but I do not have a model which where the fix is obvious. Do you have any model?

smartavionics:
Hi, I'm away for a week, please send my PRs to the back of the queue.
Cheers, Mark

LipuFei:
Hi @smartavionics , I tested this with the vertical calibration mode in Ultimaker/Cura#3875. It still seems to start wall lines in an overhang area on the left branch. See the screenshot below. I tested with an Ultimaker 3 + the default 0.1 quality.
image

rburema:
Hello @smartavionics,

I tried to review the fix in (the) Windows(-build), but unfortunately, slicing hangs for me. As in, the progress bar empties basically right after start & then nothing happens. (I tried this multiple times, cleaned out the build, restarted, etc., and at one point left it running for about 10 ~15 mins.).

Neither the commit this was based on, nor the current master have that particular problem (for me at least).

cheers, Remco.

PR #816, opened 2018-07-31
Get heating temperatures

alekseisasin:
Original issue is here: https://github.com/Ultimaker/Cura/issues/3741


Ghostkeeper:
To be clear, this doesn't fix https://github.com/Ultimaker/Cura/issues/3741.

PR #817, opened 2018-08-01
CURA-5560 Return back the top skin

alekseisasin:
The issue comes from: https://github.com/Ultimaker/Cura/issues/4053


ChrisTerBeke:
Approved but not merged? That's how features get lost...

ianpaschal:
I run into this a lot and blame it entirely on the fact that our JIRA workflow puts code review before testing. This results in a scenario where code can be reviewed but still needs extensive testing before it can be merged.

PR #818, opened 2018-08-02
Add user editable snap_distance

cubiq:
Requires this PR https://github.com/Ultimaker/Cura/pull/4177 for Cura

### DIFF
--- a/src/slicer.cpp
+++ b/src/slicer.cpp
@@ -773,7 +773,7 @@ void SlicerLayer::makePolygons(const Mesh* mesh, bool keep_none_closed, bool ext
     }
 
     //Remove all the tiny polygons, or polygons that are not closed. As they do not contribute to the actual print.
-    int snapDistance = MM2INT(1.0); // TODO: hardcoded value
+    int snapDistance = mesh->getSettingInMicrons("snap_distance");
     auto it = std::remove_if(polygons.begin(), polygons.end(), [snapDistance](PolygonRef poly) { return poly.shorterThan(snapDistance); });
     polygons.erase(it, polygons.end());
 


cubiq:
Sure! Thanks for asking @ChrisTerBeke I was actually going to add some real-life examples. From my tests on a laser SLA printer a good starting point for snap distance is 0.1 (instead of the 1.0 default).

Let's start from some screenshots. On the left the 3D model, in the center sliced at snap distance 1.0 on the right snap distance 0.1. In this particular case the print would just fail with 1.0 because some details are simply cut away by the slicer.

snap-distance-diff

Apart from that the overall surface resolution (again at least on a SLA printer) is so much better pumping the snap distance a little. I'll post some prints too!

cubiq:
@ChrisTerBeke I printed the following model in Snap Distance 1.0 (the default) and 0.1. Layer height is 0.04mm. The overall height of the model is about 25mm

Snap Distance 0.1
sd01

Snap Distance 1.0
sd10

As you can see you lose a lot of vertical details with the default SD and for SLA printers a lower value is basically mandatory.

ianpaschal:
My comment from the Cura PR:

> My opinion is that we're already deep in the deep end when it comes to settings users probably don't need to mess with; after all, that's why we have recommended mode.
> Considering what a small and clean tweak this is, and removes a magic hard-coded value from the engine (a TODO anyway, apparently) I'm in favor of a merge, @Ghostkeeper.

ianpaschal:
@cubiq I spoke with @Ghostkeeper more about this. We really appreciate the removal of magic numbers from the engine, but don't feel adding another setting is the way to go. This should instead be a function of the line width, removing the magic number, but still working automatically without requiring a hard coded value in the printer definitions/quality profiles specifying its value.

For this reason, I'm going to close the Cura PR and we can continue the discussion and changes here.

src/slicer.cpp
```diff
@@ -773,7 +773,7 @@ void SlicerLayer::makePolygons(const Mesh* mesh, bool keep_none_closed, bool ext
     }
 
     //Remove all the tiny polygons, or polygons that are not closed. As they do not contribute to the actual print.
-    int snapDistance = MM2INT(1.0); // TODO: hardcoded value
+    int snapDistance = mesh->getSettingInMicrons("snap_distance");
```
ianpaschal: Let's make this parametric and driven by line width instead of using a setting.

cubiq:
thanks for reviewing this PR @ianpaschal 

As you noted 1.0 looks like an arbitrary value. I haven't tested this option on an FDM printer but it is very important to lower snapdistance when the "nozzle" (or laser) width is in the 0.07 range (as you can see from the pictures above).

From my tests I've been able to get very reliable gcode with a 0.1 snap distance on a 0.067 line width, so if I had to guess I'd say snapDistance should be approx two times line width. That would make snapDistance 0.8 for a standard 0.4 nozzle size and 1.2 for a 0.6. Both values are very close to the current 1.0 hard-coded default and so it shouldn't impact the average user. On a 0.067 line width we would have 0.134 that looks totally reasonable.

ianpaschal:
Doesn't that mean it should be closer to 1.5x? 0.067 x 1.5 = 1.005.

To use your examples:

| snapDistance | nozzle size |
|---------------|------------|
| 0.1005             | 0.067         |
| 0.6                   | 0.4              |
| 0.9                   | 0.6              |

So that would really be putting on a diet. Would 2.0x instead of 1.5x work in your cases?

cubiq:
1.5 would totally work for me (the smaller the better) but .1005 or .134 would be more or less the same honestly. I was trying to go closer to the current 1.0 default that worked for you so far.

I'm not 100% sure I understand how snap distance work but wouldn't be easier to just set it equal to the line width (or nozzle size) and let the toolpath generator filter out whatever is too small?

cubiq:
I did some more extensive tests. Generally speaking line_width × 2 seems to be working but there is no golden value here. It's really on a case by case basis.

I tried to slice some "benchmark" I've created, have a look at what happens with snap distance 0.3

snapdistance-tst

That is totally unpredictable and could happen with any value depending on the model and angle of the feature. Unfortunately I don't see any other way if not making it user editable with a default value that could be line_width times 2, but if that's not possible any value seems arbitrary and let's proceed with line_width × 2 or 1.5.

ianpaschal:
Hmm our default line width for printing with PLA is 0.35. A value of 3 would actually get us very close to 1.0. We will need to discuss this more internally and also think how we're going to test it, since this change would affect all models.

If it that conversation yields "Ughhh" we may need to elect to either not change anything or reopen the pull request with the setting.

ianpaschal:
> This hardcoded value should have very little impact on the print quality;
if it does then maybe the function should be replaced by a new one which is
more robust.

That's sort of the point. We don't want to add more settings that 1 in 1000 people use, or change something globally, without knowing its impact on print quality. And for that we need to run tests. That kind of testing is actually a pretty massive part of what goes on at our office.

@cubiq has already done some tests on this and found that smaller values are improving high detail prints so...

> ...if it does then maybe the function should be replaced.

cubiq:
I don't think I'm able to refactor the function from scratch. The best I can do at the moment is to make snapDistance = line_width × [something] as it is now (1.0) it really doesn't work for high resolution prints. If that has some chances to be merged I'll proceed with that

ianpaschal:
Hmm in this case maybe it's better to close it. I know that's unfortunate for you as you would benefit from a smaller value but if we write an improvement that benefits a broader segment of users its maybe not worth doing (since any change will impact *all* users).

cubiq:
@ianpaschal it's not that a lower snap distance is going to impact negatively the quality of lower resolution printers. I totally understand that it's not ideal but lowering snap distance is something that doesn't change anything for current users and benefits who wants to print at higher resolution. It's a patch, I give you that :) but if it works...

jackha:
Discussed this in the team. We renamed it to Minimum polygon circumference and merged it :-)

PR #821, opened 2018-08-09
CURA-5535 Refactor line merging

jackha:
Algorithm to try and merge potentially many tiny lines into a more straight line and also lose the travels between these lines. Apply only on skin and infill lines. This may reduce the g-code by a lot, in a test case with just the right wall thickness, the complete g-code was reduced to less than 1/3.


src/MergeInfillLines.cpp
```diff
@@ -1,237 +1,316 @@
...
-    }
+        if (first_is_already_merged)
+        {
+            // take second point of path, first_path.points[0]
+            average_first_path = first_path.points[0];
+        }
+        else
+        {
+            average_first_path += first_path_start;
+            for (const Point point : first_path.points)
```
LipuFei: Use `const Point& point`, faster.

src/MergeInfillLines.cpp
```diff
@@ -1,237 +1,316 @@
...
-    return isConvertible(a, b, c, d, line_width, first_middle, second_middle, resulting_line_width, use_second_middle_as_first);
-}
-
-bool MergeInfillLines::isConvertible(const Point& a, const Point& b, const Point& c, const Point& d, int64_t line_width, Point& first_middle, Point& second_middle, int64_t& resulting_line_width, bool use_second_middle_as_first)
-{
-    use_second_middle_as_first = false;
-    const coord_t max_line_width = nozzle_size * 3 / 2;
+        coord_t second_path_length = calcPathLength(second_path_start, second_path);
+        Point average_second_path = second_path_start;
+        for (const Point point : second_path.points)
```
LipuFei: Same here

PR #822, opened 2018-08-12
Spiralization tweaks

smartavionics:
This PR contains some tweaks to spiralization mode. The changes are:

1 - Simplified findSpiralizedLayerSeamVertexIndex() - it was using findClosest() but should have been using findNearestVert() which makes the code simpler.

2 - It now starts the single wall outline that is printed immediately before the first spiral layer at the same vertex as the spiral. Previously it didn't and so it could do a big travel between the two walls.

3 - Now, when the spiral is not being smoothed, at the start of each spiral layer it immediately does an extrude move to get from the end of the last layer onto the closest point on the first line segment of the current layer. It then outputs the remainder of the first line segment and carries on as before. By doing this, the bulk of the first line segment will be orientated correctly instead of being at some slight angle to the correct direction. This stops an ugly ridge appearing. The downside is that the short extrude can produce a little step but as this tends to be hidden if the start/end vertex is positioned at a noticeable bend in the mode, the overall quality seems to improve. YMMV.

The inspiration for these mods came from trying to make a better job of models such as can be found in https://github.com/Ultimaker/Cura/issues/4215. To be honest, these are very difficult to spiralize without some artifacts occurring. I think that maybe a solution could be to add extra vertices in places where the inter-layer kink would not be visible but it's not possible at the moment because the insets are read-only. Could be a lot of work for little gain....


smartavionics:
Added a new feature that reduces the step you get when the last spiralized layer finishes.

As discussed in https://github.com/Ultimaker/Cura/issues/4221.

smartavionics:
Now also tapers the bottom spiral layer to avoid the big squidge of plastic that normally occurs.

Ghostkeeper:
Uh, I think this has been overlooked. Sorry! Looks like something we should incorporate.

Developers, see issue CURA-5816.

alekseisasin:
@Hi @smartavionics ,  we discussed your PR request and I would like to ask you, can we split it into two at least? ( In the first PR - the first three, second PR -  the top and bottom layer for Spririalize). It will be easier to test and see the quality changes. Also not sure about the (3 - Now, when the spiral is not being smoothed..), it might not work for multiple models or if closest point not so "close".

smartavionics:
OK, I'll rework this stuff and submit new PRs.

alekseisasin:
Thanks,

PR #824, opened 2018-08-14
Cura 5614 support minimum x y

alekseisasin:
This fix should fix a case when a round arc shape with support enabled, support overhang angle set to 0, the support minimum distance set to 5mm, the support disappears.

[SupportXYDistance.curaproject.3mf.zip](https://github.com/Ultimaker/CuraEngine/files/2288418/SupportXYDistance.curaproject.3mf.zip)

### DIFF
--- a/src/utils/SVG.cpp
+++ b/src/utils/SVG.cpp
@@ -1,6 +1,7 @@
 /** Copyright (C) 2017 Tim Kuipers - Released under terms of the AGPLv3 License */
 #include "SVG.h"
-
+#include <exception>
+#include <stdexcept>
 
 namespace cura {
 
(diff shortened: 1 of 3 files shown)


ChrisTerBeke:
The CI fails on this PR (but not on other PRs). I really think that should be fixed.

src/utils/SVG.cpp
```diff
@@ -45,6 +46,12 @@ SVG::SVG(const char* filename, AABB aabb, Point canvas_size)
         fprintf(out, "<?xml version=\"1.0\" encoding=\"UTF-8\" standalone=\"no\"?>\n");
     }
     fprintf(out, "<svg xmlns=\"http://www.w3.org/2000/svg\" version=\"1.1\" style=\"width:%llipx;height:%llipx\">\n", canvas_size.X, canvas_size.Y);
+    
+    if(background != Color::NONE)
+    {
+        fprintf(out, "<rect width=\"100%\" height=\"100%\" fill=\"%s\"/>\n", toString(background).c_str());
+    }
```
jackha: Adding fill-opacity is better way to solve this, so you have color and opacity instead.

PR #826, opened 2018-08-16
Fix recently introduced mistakes in SVG.cpp

smartavionics:
If I had submitted a PR with these mistakes it would have been thrown back at me. But the Cura devs can merge any old crap they like. There doesn't appear to be much quality control being applied to non-contributed code, does there?


alekseisasin:
@smartavionics , Please do not take personally. No one does not want to judge you or anybody else. The code from contributed and non-contributed is reviewed same. Only this can seem different when you speak directly or through GitHub.

PR #828, opened 2018-08-20
Initial commit for including mesh names as comments in gcode

paukstelis:
Modification to use mesh names from slicing message in comments.

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -1177,9 +1177,9 @@ void FffGcodeWriter::addMeshLayerToGCode(const SliceDataStorage& storage, const 
     }
 
     const ExtruderTrain* train = storage.meshgroup->getExtruderTrain(extruder_nr);
-
-    gcode_layer.setMesh(reinterpret_cast<size_t>(&mesh));
-
+    //PJP
+    gcode_layer.setMesh(mesh.mesh_name);
+    
     ZSeamConfig z_seam_config(mesh.getSettingAsZSeamType("z_seam_type"), mesh.getZSeamHint(), mesh.getSettingAsZSeamCornerPrefType("z_seam_corner"));
     Point layer_start_position = Point(train->getSettingInMicrons("layer_start_x"), train->getSettingInMicrons("layer_start_y"));
     PathOrderOptimizer part_order_optimizer(layer_start_position, z_seam_config);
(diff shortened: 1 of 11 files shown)


src/FffGcodeWriter.cpp
```diff
@@ -1201,8 +1201,8 @@ void FffGcodeWriter::addMeshLayerToGCode(const SliceDataStorage& storage, const
     {
         addMeshOpenPolyLinesToGCode(mesh, mesh_config, gcode_layer);
     }
-
-    gcode_layer.setMesh(0);
+    //remove this as wipe movements would get put into this category
+    //gcode_layer.setMesh("NONMESH");
```
Ghostkeeper: Removing this causes the next stuff to be marked as belonging to this mesh. The next stuff could be a prime tower or brim or basically anything.

src/FffGcodeWriter.cpp
```diff
@@ -1177,9 +1177,9 @@ void FffGcodeWriter::addMeshLayerToGCode(const SliceDataStorage& storage, const
     }
 
     const ExtruderTrain* train = storage.meshgroup->getExtruderTrain(extruder_nr);
-
-    gcode_layer.setMesh(reinterpret_cast<size_t>(&mesh));
-
+    //PJP
```
Ghostkeeper: What does PJP stand for? Why is this here?

src/FffGcodeWriter.cpp
paukstelis: I believe these were the extra movements following the skin to minimize oozing. Its not a huge deal, but it does require more travel moves if an object is cancelled. I'll fix this, but perhaps there is a way for these moves to be included in the mesh block.

src/FffGcodeWriter.cpp
paukstelis: This was just a place holder for me to do something when I made some changes and was not in an IDE. I will remove.

src/FffGcodeWriter.cpp
Ghostkeeper: Ah, probably coasting then.

I don't think this is a big deal, but you're right in that it should belong to the mesh.

paukstelis:
One of the issues that is a bit sticky with the "NONMESH" identifier is that supports for an object become NONMESH. In nearly every case it is going to be desirable to have the supports for a mesh be modified (cancelled) along with the mesh itself. Haven't looked yet, but how feasible would it be to separate/tag supports from other types of non-mesh features?

Ghostkeeper:
Infeasible. The support area can be shared with multiple multiple models. If you place multiple models close together or even on top of each other (think of dual extrusion) then a single block of support can support multiple meshes.

PR #829, opened 2018-08-20
Reorder constructor expressions to avoid compilation warnings.

smartavionics:
Fix another recent change that introduced compilation warnings.


Ghostkeeper:
I also fixed this here: https://github.com/Ultimaker/CuraEngine/pull/830

PR #831, opened 2018-08-21
Use coord_t instead of double

alekseisasin:
This Pull Request is for reviewing the changes related not appropriate variable type. The question is, do we want to continue using double, int64_t instead of coord_t ? Like for the function vSize()

In CuraEngine we already have many places where we use different (not right) types for distance. Instead of coord_t we use double or int64_t, float.


smartavionics:
Irrespective of the merits of this change, I would like to point out that it conflicts with #814 which could possibly be merged.

alekseisasin:
I will check the PR #814 at the end of this week.

Ghostkeeper:
Developers, see issue CURA-5706.

Some conflicts have to be resolved first. I don't know if anything will be left once they are resolved?

alekseisasin:
# Resolved.

PR #838, opened 2018-09-03
CURA-5672 Use insets as infill for prime tower on the first layer

LipuFei:


diegopradogesto:
Done in 317e3a9fc6065225b1acb9ff15f6b4262d67f900

PR #841, opened 2018-09-06
CURA-5666 Spread start locations on prime tower to avoid Z-seam blob

LipuFei:
Z-seam blob can build up if with the current prime tower because it always starts at the same location. This commit spreads the start location evenly around the prime tower to avoid the blob being built up.

### DIFF
--- a/src/PrimeTower.cpp
+++ b/src/PrimeTower.cpp
@@ -85,7 +85,7 @@ void PrimeTower::generatePaths(const SliceDataStorage& storage)
     if (enabled)
     {
         generatePaths_denseInfill(storage);
-        generateStartLocations(storage);
+        generateStartLocations();
     }
 }
 
@@ -141,12 +141,12 @@ void PrimeTower::generatePaths_denseInfill(const SliceDataStorage& storage)
 }
 
 
-void PrimeTower::generateStartLocations(const SliceDataStorage& storage)
+void PrimeTower::generateStartLocations()
 {
-    PolygonsPointIndex segment_start; // from where to start the sequence of wipe points
-    PolygonsPointIndex segment_end; // where to end the sequence of wipe points
-
-    segment_start = segment_end = PolygonUtils::findNearestVert(middle, outer_poly);
+    // Evenly spread out a number of dots along the prime tower's outline. This is done for the complete outline,
+    // so use the same start and end segments for this.
+    PolygonsPointIndex segment_start = PolygonsPointIndex(&outer_poly, 0, 0);
+    PolygonsPointIndex segment_end = segment_start;
 
     PolygonUtils::spreadDots(segment_start, segment_end, number_of_prime_tower_start_locations, prime_tower_start_locations);
 }
--- a/src/PrimeTower.cpp
+++ b/src/PrimeTower.cpp
@@ -16,6 +16,7 @@
 
 #define CIRCLE_RESOLUTION 32 //The number of vertices in each circle.
 
+
 namespace cura 
 {
 
@@ -84,6 +85,7 @@ void PrimeTower::generatePaths(const SliceDataStorage& storage)
     if (enabled)
     {
         generatePaths_denseInfill(storage);
+        generateStartLocations(storage);
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/PrimeTower.cpp
```diff
@@ -139,6 +141,17 @@ void PrimeTower::generatePaths_denseInfill(const SliceDataStorage& storage)
 }
 
 
+void PrimeTower::generateStartLocations(const SliceDataStorage& storage)
+{
+    PolygonsPointIndex segment_start; // from where to start the sequence of wipe points
+    PolygonsPointIndex segment_end; // where to end the sequence of wipe points
+
+    segment_start = segment_end = PolygonUtils::findNearestVert(middle, outer_poly);
```
Ghostkeeper: Two things about this:
* If we're spreading out the points anyway, it doesn't matter that it starts with any particular vertex.
* Isn't the middle equally far away from all vertices? I think the result of this will be random based on rounding errors or the bias in the `findNearestVert` algorithm.

It's easier to start at vertex 0.

src/PrimeTower.cpp
```diff
@@ -139,6 +141,17 @@ void PrimeTower::generatePaths_denseInfill(const SliceDataStorage& storage)
 }
 
 
+void PrimeTower::generateStartLocations(const SliceDataStorage& storage)
```
Ghostkeeper: The `storage` parameter is unused.

src/PrimeTower.cpp
```diff
@@ -224,4 +242,20 @@ void PrimeTower::subtractFromSupport(SliceDataStorage& storage)
 }
 
 
+void PrimeTower::gotoStartLocation(const SliceDataStorage& storage, LayerPlan& gcode_layer, const int extruder_nr) const
+{
+    int current_start_location_idx = ((extruder_nr + 1) * gcode_layer.getLayerNr()) % number_of_prime_tower_start_locations;
```
Ghostkeeper: Since `gcode_layer.getLayerNr()` can result in a negative layer index, this integer can become negative. That then results in a segfault in the line below when looking it up in the `prime_tower_start_locations` vector.

To fix, I suggest to do a positive-only modulus wraparound: `((x % n) + n) % n`

src/PrimeTower.cpp
```diff
@@ -164,6 +177,11 @@ void PrimeTower::addToGcode(const SliceDataStorage& storage, LayerPlan& gcode_la
         post_wipe = false;
     }
 
+    // Go to the start location if it's not the first layer
+    if (layer_nr != 0) {
```
Ghostkeeper: Code style: Brackets on new line.

src/PrimeTower.cpp
LipuFei: Done in 13c94392. Also updated the function's documentation.

src/PrimeTower.cpp
LipuFei: Removed in 13c94392

src/PrimeTower.cpp
LipuFei: Fixed in 1e5bf956

src/PrimeTower.cpp
LipuFei: Improved this for negative layers in d4957ca1

src/PrimeTower.h
```diff
@@ -40,6 +40,10 @@ class PrimeTower
 
     Point post_wipe_point; //!< Location to post-wipe the unused nozzle off on
 
+    std::vector<ClosestPolygonPoint> prime_tower_start_locations; //!< The differernt locations where to pre-wipe the active nozzle
+    const unsigned int prime_tower_start_location_skip = 13; //!< How big the steps are when stepping through \ref PrimeTower::wipe_locations
```
diegopradogesto: This in never used.

src/PrimeTower.h
```diff
@@ -40,6 +40,10 @@ class PrimeTower
...
+    const unsigned int number_of_prime_tower_start_locations = 21; //!< The required size of \ref PrimeTower::wipe_locations
```
diegopradogesto: Where this value is coming from?

LipuFei: The value from before.

PR #850, opened 2018-09-13
Implement gyroid infill pattern

smartavionics:
Add implementation of gyroid infill.

Basic implementation works. High densities (> 50%) can be slow but this is due to other parts of Cura choking on the large number of infill line segments rather than the generation of those line segments.
    
Doesn't support connecting the lines where they reach the edge of the infill area.


smartavionics:
Here's an example of incorrect connecting lines. In the triangular part near the centre of the build plate, on the right hand side of the part, there's two groups of lines, each group is connected at both ends to make a continuous loop. But the 4 gyroid lines should be connected end to end to make a single line rather than as two loops.

screenshot_2018-09-15_07-29-13

smartavionics:
But, then again, the lines infill pattern also gets it wrong!

screenshot_2018-09-15_08-32-05

smartavionics:
I reworked the connecting lines and now I think it only creates loops in one specific situation where a gyroid line gets cut into a sequence of "bumps" when it hits the boundary and those bumps are connected in series and then the ends of the connected sequence of bumps are joined to form a loop. I know how that could be fixed but it would add complexity and at the moment I don't think it's worth doing as the vast majority of lines do not form loops. YMMV

src/pathOrderOptimizer.cpp
```diff
@@ -207,8 +207,8 @@ static inline bool pointsAreCoincident(const Point& a, const Point& b)
 */
 void LineOrderOptimizer::optimize(bool find_chains)
 {
-    int gridSize = 5000; // the size of the cells in the hash grid. TODO
-    SparsePointGridInclusive<unsigned int> line_bucket_grid(gridSize);
+    const int grid_size = 2000; // the size of the cells in the hash grid. TODO
+    SparsePointGridInclusive<unsigned int> line_bucket_grid(grid_size);
```
alekseisasin: What was the reason to change 5000 to 200 grid_size?

src/pathOrderOptimizer.cpp
```diff
@@ -260,7 +260,7 @@ void LineOrderOptimizer::optimize(bool find_chains)
                 num_joined_lines[point_idx] = 0;
                 const Point& p = (*polygons[poly_idx])[point_idx];
                 // look at each of the lines that finish close to this line to see if either of its vertices are coincident this vertex
-                for (unsigned int close_line_idx : line_bucket_grid.getNearbyVals(p, gridSize))
+                for (unsigned int close_line_idx : line_bucket_grid.getNearbyVals(p, 10))
```
alekseisasin: The number 10, what was the reason to change it?

src/pathOrderOptimizer.cpp
smartavionics: It was changed to 2000 not 200.

I wanted to reduce the number of hits returned by getNearbyVals() as it was returning far more than is needed.

src/pathOrderOptimizer.cpp
smartavionics: There, we're only interested in finding points that are coincident and so there is no  benefit in asking for points that are further away than 10 um. I believe that getNearbyVals() returns points that can be as far away as gridSize+queryRadius so it's probably still returning points that are up to 2mm away when we only want points that are a few um away.

alekseisasin:
So far the Gyroid infill worked for all tested models.
I would like to ask you. What about making a new class GyroidInfill.cpp and move all logic to there. And split generateGyroidInfill() method into two. One for "horizontal" lines and another for "vertical" lines.
I see the tests are failing, it looks we have an issue on the server.

src/infill.cpp
```diff
@@ -234,6 +238,390 @@ void Infill::multiplyInfill(Polygons& result_polygons, Polygons& result_lines)
...
+    if (zig_zaggify)
+    {
+        outline_offset -= infill_line_width / 2; // the infill line zig zag connections must lie next to the border, not on it
+    }
+
+    const Polygons outline = in_outline.offset(outline_offset);
+    const AABB aabb(outline);
+
+    int pitch = line_distance * 2.41; // this produces similar density to the "line" infill pattern
+    int num_steps = 4;
```
alekseisasin: The number 4. Why 4 and not 2, I am curious

smartavionics:
> I would like to ask you. What about making a new class GyroidInfill.cpp and move all logic to there. And split generateGyroidInfill() method into two. One for "horizontal" lines and another for "vertical" lines.

I cannot see any benefit in doing that.

src/infill.cpp
smartavionics: The line is based on a sine wave and so I figured that at the lowest resolution you still need 4 points to model a sine wave (0,+peak, 0, -peak).

smartavionics:
> I see the tests are failing, it looks we have an issue on the server.

Yeah, it's got some problem with the new arcus comms.

alekseisasin:
The benefit, the  infill.cpp is more like a manager, it uses the rest infill types. The new infill already has 383 code lines, at the end the infill.cpp we get so big. I do not know why we did not do it before.
We already have SpaghettiInfill, maybe it not so good example because it is used in another class.

smartavionics:
> The benefit, the infill.cpp is more like a manager, it uses the rest infill types. 

You know that's not really true. A couple of infill types have their own classes but a lot do not.

I simply do not see how taking a single function and making a class out of it (two more files) that has a number of functions in it makes the code any easier to understand or maintain.

So, I'm not going to do that. You can either use the PR with it's current structure or we can close the PR without merging it. If you merge it and subsequently want to move the gyroid implementation to its own class then I can't stop you but it won't be with my blessing.

Appesteijn:
@smartavionics We added the classes ourself on this branch: https://github.com/Ultimaker/CuraEngine/tree/CURA-5761_gyroid_infill could you have a look?

Of course you will be contributed for this work, thank you very much!

smartavionics:
It's pretty much how I expected it would look. If it makes you happy, use it.

Appesteijn:
I'll close this pull request as the other branch is merged. We expect this to be in Cura 3.6. Thank you for your contribution!!

PR #853, opened 2018-09-18
Cura 5709 disabled 2nd extruder primeblob prints in air

rburema:


src/gcodeExport.cpp
```diff
@@ -997,12 +997,26 @@ void GCodeExport::writePrimeTrain(double travel_speed)
...
         {
             command += " S1";  // use S1 to disable prime blob
+            should_correct_z = true;
         }
         *output_stream << command << new_line;
+
+        // There was an issue with the S1 strategy parameter, where it would only change the material-position,
+        //   as opposed to 'be a prime-blob maneuvre without actually printing the prime blob', as we assumed here.
+        // After a chat, the firmware-team decided to change the S1 strategy behaviour,
+        //   but since people don't update their firmware at each opportunity, it was decied to fix it here as well.
```
ChrisTerBeke: typo: decied -> decided

LipuFei:
I think this should be in 3.5

src/gcodeExport.cpp
LipuFei: Fixed

PR #856, opened 2018-09-20
fix: omit threading funcitonality if not compiling with open MP

BagelOrb:


BagelOrb:
This should make (cross)compiling easier if you only want CuraEngine without any advanced features.

PR #859, opened 2018-09-26
Cura 5765 fix slicing time increase

rburema:


smartavionics:
Hi @rburema, just looked at your caching code and it looks to me that it doesn't take into account that the getInnermostWalls() method can be called with max_inset either 1 or 2 and so it seems to just cache the result of the first call and always return that on subsequent calls even if max_inset is different from the first time the method is called.

Also, I seriously doubt that this caching is needed at all. In your comment in the other thread you said:
> I asked around (being pretty new here) and it turns out that you'd really do need the inner walls there.

Well, that's all screwy because in the example print we have been looking at, the profile is actually calling for 3 walls and the getInnermostWalls() method only provides the "right" result for either 1 or 2 walls so in that example, it's not using a polygon that matches either the model outline or the outline of the infill area. So I simply don't believe that the polygons returned by that method are that critical. I still think that using the part outline would be fine. Please ask your colleagues for the justification for doing the extra calculation.

smartavionics:
Also, as not everyone uses that infill type, always caching in getInnermostWalls() is wasting a little time and possibly a fair amount of memory depending on the number of layers, outline complexity, etc. Maybe we don't care about memory these days but think about a huge print?

smartavionics:
OK,  so that's fixed the problem with not taking max_inset into account but I have to say that I still think this is completely the wrong solution to this problem and certainly doesn't meet with my approval. If this gets merged to master I shall be reverting it for my own installation.

rburema:
Hi @smartavionics,
Thanks for your review.
1. Yikes! I don't know how I missed accounting for the arguments. I've fixed that now in two commits.
2. I've asked for more clarification from @Ghostkeeper. The concern is that the results are different with the replacement. Apparently it also ignores the 'cubic subdivision shell' (in the code as 'sub_div_rad_add') setting that way.
3. (Also discussed:) The memory-use does increase this way, but it should not matter much as compared what we do in some other places in the code. Tree-support came up as an example.
4. _(Just saw your most recent comment.)_ It's true that caching is not the most elegant solution, but changing this the other way would mean a deep dive into the SubDivCube code, and a lot of testing.

smartavionics:
Hi. I agree that using my simple fix does give different results in terms of the infill generated but I bet that nobody was concerned with the difference that would also have occurred when the combing boundary was changed from the outer wall to the 2nd wall by a Cura dev a few months before I came up with the current implementation. Also, as I pointed out above, the current use of the 2nd wall as the boundary doesn't even match the real boundary of the infill anyway when the profile is using more than 2 walls.

Ghostkeeper:
I believe that the intent is to check whether we need to subdivide a cube based on whether that cube exits the infill area. Any solution that approaches this intent closely is fine with me. 

The current solution to take the 2nd wall (or the outer wall if there is no 2nd wall) is indeed not ideal but better than taking the outer shell. There could be a better solution that is still efficient by looking at the border of the infill area itself, but it also does something tricky with the skin so there was probably a reason for this.

> Apparently it also ignores the 'cubic subdivision shell' (in the code as 'sub_div_rad_add') setting that way.

I meant that this Cubic Subdivision Shell setting is wrong if this check for subdividing cubes is not on the inner wall. If you say you want a 3mm shell of fully subdivided cubes, then it should not leave cubes unsubdivided that are closer than 3mm from the wall.

src/infill/SubDivCube.cpp
```diff
@@ -226,8 +226,7 @@ int SubDivCube::distanceFromPointToMesh(SliceMeshStorage& mesh, int layer_nr, Po
         return 2;
         *distance2 = 0;
     }
-    Polygons collide;
-    mesh.layers[layer_nr].getInnermostWalls(collide, 2, mesh);
+    Polygons& collide = mesh.layers[layer_nr].getInnermostWalls(2, mesh);
```
nallath: I know it was already there, but the "2" is a bit of a magic number. Could you add some documentation as to why the 2 is used?

nallath:
Documentation in the header needs updating for getInnermostWalls

smartavionics:
This is so bad, it fills me with dismay. it may be a lovely bit of code but it's completely the wrong solution.

PR #861, opened 2018-09-27
Feat woodfill

BagelOrb:
Looks nice.

Works very well if you set the Top Surface Layers to 1, set the Top Surface Pattern to Concentric and set the Speckle Density to something like 5 speckles per square centimeter.

image


nallath:
And the point of this is?

BagelOrb:
> And the point of this is?

Aesthetics.

BagelOrb:
"aesthetics"

BagelOrb:
Maybe my bed wasnt really level.

here's another print:


With connect top/bottom polygons enabled (on the branch where the feature is fixed).

BagelOrb:
Maybe this feature is just not cool enough, that's okay.

But don't act like aesthetics is not a valid reason for wanting some type of top/bottom pattern.

BagelOrb:
I don't think this is pretty any more.

PR #864, opened 2018-10-05
Feature: "Minimum Support Area" and "Minimum Support Interface Area"

vgribinchuk:
Feature: Minimum Support Area / Minimum Interface Area.
Backend implementation of the feature: https://github.com/Ultimaker/Cura/pull/4486

### DIFF
--- a/src/support.cpp
+++ b/src/support.cpp
@@ -1393,7 +1393,8 @@ void AreaSupport::generateSupportBottom(SliceDataStorage& storage, const SliceMe
         {
             mesh_outlines.add(mesh.layers[std::round(layer_idx_below)].getOutlines());
         }
-        Polygons bottoms = generateSupportInterfaceLayer(global_support_areas_per_layer[layer_idx], mesh_outlines, bottom_line_width, minimum_bottom_area);
+        Polygons bottoms;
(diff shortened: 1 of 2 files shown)


src/support.cpp
```diff
@@ -1418,19 +1424,22 @@ void AreaSupport::generateSupportRoof(SliceDataStorage& storage, const SliceMesh
...
-    interface_polygons = support_areas.intersection(model);
+    Polygons interface_polygons = support_areas.intersection(model);
     interface_polygons = interface_polygons.offset(safety_offset).intersection(support_areas); //Make sure we don't generate any models that are not printable.
-    interface_polygons.removeSmallAreas(1.0);
+    if (minimum_interface_area > 0.0)
+    {
+        interface_polygons.removeSmallAreas(minimum_interface_area);
+    }
     support_areas = support_areas.difference(interface_polygons);
+    return interface_polygons;
```
nallath: This does increase the number of copies when compared to the previous situation. Returning the polygons (without reference) means that an additional copy is done.

vgribinchuk: Here is the place where C++ compiler is able to apply NRVO eliminating copying of the object, thus no additional overhead.

nallath:
for our reference; CURA-5903

rburema:
Hi @vgribinchuk, Same as for the frontend, some conflicts got introduced. Could you fix those? cheers, Remco.

PR #865, opened 2018-10-09
[bug fix] Correct optimisation in combing path generation that can cut corners.

smartavionics:
It would be good if this could be merged ASAP as it can cause combing travels to cut corners.

### DIFF
--- a/src/pathPlanning/LinePolygonsCrossings.cpp
+++ b/src/pathPlanning/LinePolygonsCrossings.cpp
@@ -190,11 +190,6 @@ void LinePolygonsCrossings::generateBasicCombingPath(PolyCrossings& polyCrossing
         rev_points.push_back(p);
         rev_len += vSize(p - prev);
         prev = p;
-        if (rev_len > fwd_len)
-        {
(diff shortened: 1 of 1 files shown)


smartavionics:
I have reinstated the optimisation and fixed the problem by ensuring that the length of the fwd path is correct before comparing the reverse path length to it. This is better than just dropping the optimisation completely.

smartavionics:
Without this PR, you can get unretracted travels that cut across holes, like this:

screenshot_2018-10-10_07-43-20

With the PR, only travels that use retraction will cross a hole:

screenshot_2018-10-10_07-44-36

smartavionics:
You really want to be including this in 3.5.1.

alekseisasin:
Hi @smartavionics , can you please share your model?

smartavionics:
[UMS5_Pocket Tool Box V2.zip](https://github.com/Ultimaker/CuraEngine/files/2479190/UMS5_Pocket.Tool.Box.V2.zip)

alekseisasin:
I will be added to 3.5.1. Thank you.

PR #866, opened 2018-10-09
CURA-5777 Fix point operators

LipuFei:
Because Point is a clipperlib point, which has a constructor Point(int x = 0, int y = 0), which can implicitly convert a int/unsigned int "c" to a Point(c, y = 0), and this causes undesired behaviour. So, we remove
the /= and *= operators in Point to avoid this implicit conversion to happen.

### DIFF
--- a/src/LayerPlan.cpp
+++ b/src/LayerPlan.cpp
@@ -231,7 +231,7 @@ Polygons LayerPlan::computeCombBoundaryInside(const size_t max_inset)
             }
             else
             {
-                layer.getInnermostWalls(comb_boundary, max_inset, mesh);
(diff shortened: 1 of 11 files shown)


rburema:
@LipuFei Did you add the extra documentation that was added to the 3.5.1-version of this PR here as well? I can't find it right now.

LipuFei:
@rburema We can merge 3.5 back to master then the doc there will be in as well. Or you can also cherry-pick that doc commit.

rburema:
@LipuFei Good point. Let's wait for the 3.5 merge then, by then the build-server will be able to do the checks for master again as well.

LipuFei:
Well... There's no need to merge this I guess. Those changes are in `master` already.

PR #868, opened 2018-10-10
CURA-5798 Fix a segfault in tree support

LipuFei:
Fixes https://github.com/Ultimaker/Cura/issues/4491

 - Delete all allocated nodes at the end to avoid invalid address access
   in the loop.
 - Add destructor to struct Node so it resets its pointers and cleans
   its data properly.

### DIFF
--- a/src/TreeSupport.h
+++ b/src/TreeSupport.h
@@ -109,7 +109,7 @@ public:
          * can't be on the model and the path to the buildplate isn't clear),
          * the entire branch needs to be known.
          */
-        Node * parent;
+        Node *parent;
 
         /*!
         * \brief All neighbours (on the same layer) that where merged into this node.
--- a/src/TreeSupport.h
+++ b/src/TreeSupport.h
@@ -59,11 +59,13 @@ public:
          , parent(parent)
         {}
 
+#ifdef DEBUG // Clear the delete node's data so if there's invalid access after, we may get a clue by inspecting that node.
         ~Node()
         {
             parent = nullptr;
             merged_neighbours.clear();
         }
+#endif // DEBUG
 
         /*!
          * \brief The number of layers to go to the top of this branch.
--- a/src/TreeSupport.h
+++ b/src/TreeSupport.h
@@ -59,6 +59,12 @@ public:
          , parent(parent)
         {}
 
+        ~Node()
+        {
+            parent = nullptr;
+            merged_neighbours.clear();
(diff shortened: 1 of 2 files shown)


src/TreeSupport.h
```diff
@@ -59,6 +59,12 @@ class TreeSupport
          , parent(parent)
         {}
 
+        ~Node()
+        {
+            parent = nullptr;
+            merged_neighbours.clear();
```
Ghostkeeper: This has no effect. The instance is about to be deleted and this memory is about to be unallocated, so whether the unallocated memory says 0 or not makes no difference.

src/TreeSupport.h
```diff
@@ -103,7 +109,7 @@ class TreeSupport
          * can't be on the model and the path to the buildplate isn't clear),
          * the entire branch needs to be known.
          */
-        Node *const parent;
+        Node * parent;
```
Ghostkeeper: Code style: No space between class name and `*` pointer operator.

src/TreeSupport.h
LipuFei: If there's no potential segfault, this has no effect, but this leave a trace of clue if a segfault happens. If a deleted object is picked up later (which shouldn't happen), it may not lead to a direct segfault immediately, but when you try to access something inside, it can. Like `node->parent` works (by chances), but `node->parent->parent` doesn't (segfault by chance.)

I added this reset code here is to "cleanup" the deleted object which shouldn't be accessed later. If in case something happens and we are debugging at this location, this "empty" node object can give us a good clue of what might have happened. Otherwise, if we pick up a random Node with old data in it while we shouldn't, it's impossible to tell at the first glance because that Node contains a valid old data.

src/TreeSupport.h
LipuFei: Fixed in a3d7485f

src/TreeSupport.h
Ghostkeeper: I think we should work from the assumption that no segfault occurs though.

Especially since this is live code that is run during tree support. Zeroing all that memory is just a waste of processing power for our users. If you really need this we should at least put this between `#ifdef DEBUG` statements and document why we do this.

LipuFei: As discussed, I put that into a DEBUG block so it won't affect release builds. Commit: ac1fcea

Issue #825, opened 2018-08-15
Inefficient grid generation in TreeSupport::generateContactPoints


bjude:
The grid generation in `TreeSupport::generateContactPoints()` claims to be creating a grid pattern at an angle of 22 degrees but what the code actually does is different. 

The code as it stands:
- take the model bounding box and rotate it 22 degrees
- take the AABB of that rotated box and rotate it back 22 degrees
- create a grid of point across the AABB of the re-rotated box
- rotate those points 22 degrees and translate them to the model coordiantes

![image](https://user-images.githubusercontent.com/4530480/44130347-a391aade-a07f-11e8-835b-275c92c4a84c.png)

This is wildly inefficient and means that the box that the points are generated over is 1.7 times bigger in each dimension (for a square initial AABB).
This is easy to fix and i can generate a PR if desired.

However, i have another idea for point sampling that will be more efficient still. The comments state that the 22 degree rotated grid is employed so better support diagonal lines, which presumably are likely to fall between the lines of an aligned grid. But this just shifts the problem to poorly supporting lines that are 22 degrees rotated (which may well be uncommon but we're not really addressing the real problem this way).

The sampling problem could be addressed by using a low-discrepancy sequence like the [Halton Sequence](https://en.wikipedia.org/wiki/Halton_sequence) to generate 2d points This could either be done by simply generating a sequence of Halton points across the entire AABB until the density is sufficient (eg mean distance between points is `support_tree_branch_distance`) or by splitting the AABB into a grid and generating a smaller sequence of Halton points in each grid cell. Both approaches should fix the issue of having "channels" of empty space through the grid.

rburema:
Hi @bjude,
Thank you for your report.

As you've noticed, TreeSupport is a relatively new area of the code, with many possible future optimisations. The option as a whole is still in the 'experimental' part of the settings, and there might be relatively large changes to that part of the code base still to come.

Moreover, as you've already mentioned as well, the grid generation will only run for 'generateContactPoints', which only runs once for each branch of the tree. Right now, if we where about to fix inefficiencies in TreeSupport we'd look into the parts of the code which run for each layer.

That said: If you, or anyone else in the community wants to create a PR, please go ahead. However, our attention is required elsewhere at the moment.

cheers, Remco.

bjude:
Thanks for the reply Remco.

While it does only run once per branch, it causes 190% more point-polygon intersection checks so by fixing this we could go for much higher precision at the same computational effort.

[...]

PR #869, opened 2018-10-11
TreeSupport: Simplify grid point generation

bjude:
Addresses issue #825
Previous implementation was causing the rectangle of generated points to be 1.7 times too large in each dimension (for a square inital AABB).
This implementation works as follows:
* Rotate the AABB of the mesh by 22 degrees
* Take AABB of this rotated box and generate a grid of points
* Rotate the points so they are angled (compared to the original AABB)
* Check if the points are inside the original AABB
* If so, add to the grid_points vector

### DIFF
--- a/src/TreeSupport.cpp
+++ b/src/TreeSupport.cpp
@@ -562,26 +562,36 @@ void TreeSupport::generateContactPoints(const SliceMeshStorage& mesh, std::vecto
 
     //First generate grid points to cover the entire area of the print.
     AABB bounding_box = mesh.bounding_box.flatten();
-    //We want to create the grid pattern at an angle, so compute the bounding box required to cover that angle.
-    constexpr double rotate_angle = 22.0 / 180.0 * M_PI; //Rotation of 22 degrees provides better support of diagonal lines.
+    // We want to create the grid pattern at an angle, so compute the bounding
+    // box required to cover that angle.
+    // Rotation of 22 degrees provides better support of diagonal lines.
+    constexpr double rotate_angle = 22.0 / 180.0 * M_PI;
     const Point bounding_box_size = bounding_box.max - bounding_box.min;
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Developers, see issue CURA-5897.

Looks good at first glance. I hope it makes it slightly more efficient (but the real work is in the collision area generation so it won't matter visibly on the progress bar).

src/TreeSupport.cpp
```diff
@@ -562,26 +562,36 @@ void TreeSupport::generateContactPoints(const SliceMeshStorage& mesh, std::vecto
...
+    // Store centre of AABB so we can relocate the generated points
+    const auto centre = bounding_box.getMiddle();
+    const auto sin_angle = std::sin(rotate_angle);
+    const auto cos_angle = std::cos(rotate_angle);
+    // Calculate the dimensions of the AABB of the mesh AABB after being rotated
+    // by `rotate_angle`. Halve the dimensions since we'll be using it as a +-
+    // offset from the centre of `bounding_box`.
+    const auto rotated_dims = Point(
+        bounding_box_size.X * cos_angle + bounding_box_size.Y * sin_angle,
+        bounding_box_size.X * sin_angle + bounding_box_size.Y * cos_angle) / 2;
```
Ghostkeeper: This can't be correct. You're only rotating one point, so somewhere it's going to map to 0. For instance, if I put an angle of 110 in there and a bounding box of 100x300, it'll output a rotated bounding box of 247x-9.

bjude: This will only work for angles <90 as the either the sine or cosine will be <0. This isnt a problem for this code since the angle is hardcoded as 22 degrees but will obviously be an issue if that angle becomes a user set angle. I'll add a comment to make it more clear.

What i'm trying to achieve is better shown below. Left is the original bounding box, middle is rotated, right is the AABB of the rotated bounding box. The `rotated_dims` calculation directly gives us the dimensions of the right hand box (halved since i'm using it as an offset from the centre of the AABB)

image

PR #872, opened 2018-10-12
feat: ability to omit bottom skin for inital layers

BagelOrb:
I quite often find myself setting the bottom layers to zero because of the first layer only, so I made an option for only the very bottom.

Disabling the bottom skin is nice when:
- you want to save material / print time
- you want to remove the infill afterwards (e.d. when using PVA infill)
- you want to fill the print afterward with some material to make it a solid part
- you want to show what the infill looks like
- you cut a big model into parts and the bottom of this part is not the bottom of the eventual assembled product 

The build plate is counted as a solid model, so infill is generated for the bottom.

Pictures taken from below:
image

image

Angled cube from top:
image


diegopradogesto:
As discussed personally with the whole team, it will be better if the setting allow the user to choose the thickness of the very bottom layer instead of just remove it completely. If you can do that, then we will consider you contribution 😄

BagelOrb:
Done

diegopradogesto:
Devs. CURA-5918

BagelOrb:
:tada:

PR #874, opened 2018-10-12
Feat: flow per print-feature

BagelOrb:
Have separate flow settings per print feature.

This way we can control the line spacing separately from the flow.
This has been a long standing deficit. 
Because Cura models the cross-section of a printed line as a square thing, while it is in fact rounded, lines are wider than computed.
This is a problem in densely filled regions (skin, walls, prime tower, brim, etc etc) because it leads to overextrusion.

Because the materials team reminded me of this deficit I decided to fix what should have been fixed a long time ago!

The only way to combat this problem before this PR is to set the global flow to something lower, or to set the steps_per_mm in the firmware to something which is in fact incorrect.

Now we have flow settings per print feature.

I was doubting whether I should make settings for the line spacing instead, but that proved to be quite difficult as there are spacings between several print features and we wouldn;t want to introduce setting for the spacing of ewach possible combination of two print features (e.g. skin-walls, skin-infill, etc)

Because the flow settings are now per print-feature it doesn't make sense to have a global flow for the first layer only. Therefore, the initial layer flow is a multiplier on top of the material_flow, instead of replacing it for the first layer. Luckiuly no single material profile set the material_flow_layer_0, so it's not a problem.
Now the initial layer flow doesn';t inherit from the material_flow anymore, because they are multiplied together.

There are no flow settings for the raft, but I don;t know whether it's really useful to introduce those.
The initial layer flow does work on the first raft layer in the same way as it used to, though.

Also the perimeter gaps now get their multiplier directly from the wall_x_extruder, so there's no more need to use a weird work-around which was in the code previously.

I hope you see this is something which Cura needs. This is not something I want and made in my free time, but something which was requested by the material team which I decided to make a headstart in for the Cura team.
Hope this is a nice kick-start!

### DIFF
--- a/src/TopSurface.cpp
+++ b/src/TopSurface.cpp
@@ -71,17 +71,16 @@ bool TopSurface::ironing(const SliceMeshStorage& mesh, const GCodePathConfig& li
 
     //Add the lines as travel moves to the layer plan.
     bool added = false;
-    const Ratio ironing_flow = mesh.settings.get<Ratio>("ironing_flow");
     if (!ironing_polygons.empty())
     {
         constexpr bool force_comb_retract = false;
         layer.addTravel(ironing_polygons[0][0], force_comb_retract);
-        layer.addPolygonsByOptimizer(ironing_polygons, line_config, nullptr, ZSeamConfig(), 0, false, ironing_flow);
+        layer.addPolygonsByOptimizer(ironing_polygons, line_config, nullptr, ZSeamConfig());
         added = true;
     }
     if (!ironing_lines.empty())
     {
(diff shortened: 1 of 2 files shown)


BagelOrb:
I talked to the materials team and they weren't sure whether these settings were the right conceptualisation of the functionality.

Let's put this PR on hold or use it for the materials team to do some preliminary testing on.

BagelOrb:
Is there any commit msg rule which I violate which bothers you specifically, or are you just updating me that there is now an official guide?

nallath:
Pre-fixing commit messages with the type of fix is not needed and you should use better interpunction. Those are the most important ones at least.

Note that we've had these rules for committing for quite some time, but no-one was really checking for them.

PR #876, opened 2018-10-15
Remove all max(0U, x)

bjude:
Taking the max of 0 and an unsigned number is a no-op and is misleading


Ghostkeeper:
Thanks. I agree that it's only misleading and tested it. There are no differences in behaviour.

PR #877, opened 2018-10-16
Fix Mingw build failure

bjude:
The refactoring in commit 8089744 broke the build on windows under mingw due to a mismatch in the dllexport attributes. This commit adds the macro used in libArcus to ensure that theyre the same.

### DIFF
--- a/tests/arcus/MockSocket.h
+++ b/tests/arcus/MockSocket.h
@@ -14,7 +14,7 @@ namespace cura
  * \brief Mocks a socket connection from libArcus such that we can test with it
  * without creating an actual connection.
  */
-class MockSocket : public Arcus::Socket
+class ARCUS_EXPORT MockSocket : public Arcus::Socket
 {
 public:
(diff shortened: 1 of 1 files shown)


bjude:
apparently the gcc version on the CI server doesnt like this as a fix :/

bjude:
So this is caused by the CI server using a version of libArcus prior to this commit: https://github.com/Ultimaker/libArcus/commit/74e19d54cfeddee76d73a765387c8ef986740b06

When the functions in Arcus::Socket are not marked virtual we get a compile error for CuraEngine

Ghostkeeper:
We've now fixed the socket virtual method issue (finally; I had uncommitted changes for so long to make my local tests compile).

I'm going to reopen this and still give it a try. I hope you don't mind.

Ghostkeeper:
It still compiles and runs fine locally for me. If the build server thinks it's okay, I'm going to merge it in.

PR #879, opened 2018-10-17
Feature: "Support Interface Offset"

vgribinchuk:
Feature: "Support Interface Offset"
Backend implementation of the feature: https://github.com/Ultimaker/Cura/pull/4596


nallath:
For our reference; CURA-5880

PR #882, opened 2018-10-17
Feature: "Nozzle Switch Extra Prime Distance"

vgribinchuk:
Feature: "Nozzle Switch Extra Prime Distance"
Backend implementation of the feature: https://github.com/Ultimaker/Cura/pull/4599


nallath:
CURA-5883

PR #883, opened 2018-10-18
Bug fix: reinstate test for multiple bridge layers enabled which got removed

smartavionics:
Must have got lost when the code was refactored.


Ghostkeeper:
Developers, see issue CURA-5894.

Also note that we'll probably want to fix this in 3.6 so we'd have to cherry-pick this commit in.

nallath:
I've merged them into 3.6 first and then into master. Thanks for fixing it :)

PR #884, opened 2018-10-18
Bug fix: correct setting type

smartavionics:
This was broken when the new setting scheme was introduced.


smartavionics:
Hey @Ghostkeeper , this PR is fixing your regression. How can you possibly release a 3.6 beta without this and the other bug fix PRs being merged first? Are you all completely mad?

Ghostkeeper:
Yes, we are all completely mad. We also have 181 other open bugs, so we're going to have to release with some bugs still open. Yes, even the stable version. Even deciding on, reviewing, testing and integration testing the pull requests takes time.

The fix itself is a no-brainer. Thanks for finding it! We've done 2 passes of code review over that refactor and 3 layers of tests but it's very hard to test all settings thoroughly.

Ghostkeeper:
I can't reproduce the bug though. I seems to work fine on 3.6 even though it interprets it as an integer.

Since the branch is based off master, we'll merge it into master as a style fix.

smartavionics:
It breaks for me because the strings "True" or "False" when passed to atof() both produce zero so it always appears false. This really should be in 3.6.

PR #888, opened 2018-10-24
Bug Fix: regression in skin/infill line ordering that came with the gyroid infill

smartavionics:
This PR fixes a recently introduced regression that can occasionally cause the line ordering to choose a line that is further away from a better candidate. Here's an example that shows that when printing some skin, the bug caused it to skip over a couple of lines (which do get filled in later).

Without this PR:

screenshot_2018-10-24_07-21-22

With this PR:

screenshot_2018-10-24_07-22-23


rburema:
Devs: As part of CURA-5761.

smartavionics:
Thanks.

PR #889, opened 2018-10-24
[3.6] [CURA-5831] Use takeNextMessage as a blocking call to fix busy loop

alekseisasin:
…a new message is received. It will

reduce performance issue, it usess less CPU then constantly requesting for a new message.


rburema:
For anyone just tuning in:
See the more involved reviews/discussion in https://github.com/Ultimaker/libArcus/pull/77 .
(And be aware that you need to update libArcus (yes, yet again!) for this to work!)

PR #890, opened 2018-10-25
[3.6] Use support infill extruder instead of adhesion extruder or the brim_replaces_support setting

diegopradogesto:
Use the support infill extruder to get the value of the brim_replaces_support setting, since after the commit https://github.com/Ultimaker/Cura/commit/56fdab276e383992866a5e93a338473a2f4d7b49 it is now limited to the support infill extruder instead of the adhesion extruder.


BagelOrb:
Oh yes I forgot to change that!

Code looks good.

PR #900, opened 2018-11-04
TreeSupport: Lazily calculate the collision, avoidance and internal model guide volumes

bjude:
The ModelVolumes object lazily calculates the volumes as required, singificantly reducing the amount of work done since most radius samples were never read in the old code.
The object is not thread-safe and we lose the parallelism of generating the volumes with OpenMP but this is still faster because only the required areas are genereated.

There is a slight change in behaviour due inconsistent behaviour in the old code regarding radius sample index calculation. The new code consistently uses ceiling when calculating the sample radius whereas the old code used ceiling for generating the volumes and rounding when accessing later on.

### DIFF
--- a/src/TreeSupport.cpp
+++ b/src/TreeSupport.cpp
@@ -16,8 +16,7 @@
 
 //The various stages of the process can be weighted differently in the progress bar.
 //These weights are obtained experimentally.
-#define PROGRESS_WEIGHT_COLLISION 50 //Generating collision areas.
-#define PROGRESS_WEIGHT_DROPDOWN 1 //Dropping down support.
+#define PROGRESS_WEIGHT_DROPDOWN 50 //Dropping down support.
 #define PROGRESS_WEIGHT_AREAS 1 //Creating support areas.
 
 namespace cura
@@ -82,9 +81,19 @@ TreeSupport::TreeSupport(const SliceDataStorage& storage)
     }
     actual_border = actual_border.offset(-adhesion_size);
 
+    Polygons machine_volume_border;
     machine_volume_border.add(actual_border.offset(1000000)); //Put a border of 1m around the print volume so that we don't collide.
     actual_border[0].reverse(); //Makes the polygon negative so that we subtract the actual volume from the collision area.
     machine_volume_border.add(actual_border);
+
+    const coord_t xy_distance = mesh_group_settings.get<coord_t>("support_xy_distance");
+    const coord_t layer_height = mesh_group_settings.get<coord_t>("layer_height");
+    const AngleRadians angle = mesh_group_settings.get<AngleRadians>("support_tree_angle");
+    const coord_t maximum_move_distance
+        = (angle < TAU / 4) ? (coord_t)(tan(angle) * layer_height) : std::numeric_limits<coord_t>::max();
+    const coord_t radius_sample_resolution = mesh_group_settings.get<coord_t>("support_tree_collision_resolution");
+
+    volumes_ = ModelVolumes(storage, machine_volume_border, xy_distance, maximum_move_distance, radius_sample_resolution);
 }
 
 void TreeSupport::generateSupportAreas(SliceDataStorage& storage)
@@ -107,22 +116,6 @@ void TreeSupport::generateSupportAreas(SliceDataStorage& storage)
         return;
     }
 
-    //Generate areas that have to be avoided.
-    std::vector<std::vector<Polygons>> model_collision; //For every sample of branch radius, the areas that have to be avoided by branches of that radius.
-    collisionAreas(storage, model_collision);
-    std::vector<std::vector<Polygons>> model_avoidance; //For every sample of branch radius, the areas that have to be avoided in order to be able to go towards the build plate.
-    propagateCollisionAreas(storage, model_collision, model_avoidance);
-    std::vector<std::vector<Polygons>> model_internal_guide; //A model to guide branches that are stuck inside towards the centre of the model while avoiding the model itself.
-    for (size_t radius_sample = 0; radius_sample < model_avoidance.size(); radius_sample++)
-    {
-        model_internal_guide.emplace_back();
-        for (size_t layer_nr = 0; layer_nr < model_avoidance[radius_sample].size(); layer_nr++)
-        {
-            Polygons layer_internal_guide = model_avoidance[radius_sample][layer_nr].difference(model_collision[radius_sample][layer_nr]);
-            model_internal_guide[radius_sample].push_back(layer_internal_guide);
-        }
-    }
-
     std::vector<std::unordered_set<Node*>> contact_nodes;
     contact_nodes.reserve(storage.support.supportLayers.size());
     for (size_t layer_nr = 0; layer_nr < storage.support.supportLayers.size(); layer_nr++) //Generate empty layers to store the points in.
@@ -135,14 +128,14 @@ void TreeSupport::generateSupportAreas(SliceDataStorage& storage)
         {
             continue;
         }
-        generateContactPoints(mesh, contact_nodes, model_collision[0]);
+        generateContactPoints(mesh, contact_nodes);
     }
 
     //Drop nodes to lower layers.
-    dropNodes(contact_nodes, model_collision, model_avoidance, model_internal_guide);
+    dropNodes(contact_nodes);
 
     //Generate support areas.
-    drawCircles(storage, contact_nodes, model_collision);
+    drawCircles(storage, contact_nodes);
 
     for (auto& layer : contact_nodes)
     {
@@ -157,40 +150,7 @@ void TreeSupport::generateSupportAreas(SliceDataStorage& storage)
     storage.support.generated = true;
 }
 
-void TreeSupport::collisionAreas(const SliceDataStorage& storage, std::vector<std::vector<Polygons>>& model_collision)
-{
-    const Settings& mesh_group_settings = Application::getInstance().current_slice->scene.current_mesh_group->settings;
-    const coord_t branch_radius = mesh_group_settings.get<coord_t>("support_tree_branch_diameter") / 2;
-    const coord_t layer_height = mesh_group_settings.get<coord_t>("layer_height");
-    const double diameter_angle_scale_factor = sin(mesh_group_settings.get<AngleRadians>("support_tree_branch_diameter_angle")) * layer_height / branch_radius; //Scale factor per layer to produce the desired angle.
-    const coord_t maximum_radius = branch_radius + storage.support.supportLayers.size() * branch_radius * diameter_angle_scale_factor;
-    const coord_t radius_sample_resolution = mesh_group_settings.get<coord_t>("support_tree_collision_resolution");
-    model_collision.resize((size_t)std::round((float)maximum_radius / radius_sample_resolution) + 1);
-
-    const coord_t xy_distance = mesh_group_settings.get<coord_t>("support_xy_distance");
-    constexpr bool include_helper_parts = false;
-    size_t completed = 0; //To track progress in a multi-threaded environment.
-#pragma omp parallel for shared(model_collision, storage) schedule(dynamic)
-    for (size_t radius_sample = 0; radius_sample < model_collision.size(); radius_sample++)
-    {
-        const coord_t radius = radius_sample * radius_sample_resolution;
-        for (size_t layer_nr = 0; layer_nr < storage.support.supportLayers.size(); layer_nr++)
-        {
-            Polygons collision = storage.getLayerOutlines(layer_nr, include_helper_parts);
-            collision = collision.unionPolygons(machine_volume_border);
-            collision = collision.offset(xy_distance + radius, ClipperLib::JoinType::jtRound); //Enough space to avoid the (sampled) width of the branch.
-            model_collision[radius_sample].push_back(collision);
-        }
-#pragma omp atomic
-        completed++;
-#pragma omp critical (progress)
-        {
-            Progress::messageProgress(Progress::Stage::SUPPORT, (completed / 2) * PROGRESS_WEIGHT_COLLISION, model_collision.size() * PROGRESS_WEIGHT_COLLISION + storage.support.supportLayers.size() * PROGRESS_WEIGHT_DROPDOWN + storage.support.supportLayers.size() * PROGRESS_WEIGHT_AREAS);
-        }
-    }
-}
-
-void TreeSupport::drawCircles(SliceDataStorage& storage, const std::vector<std::unordered_set<Node*>>& contact_nodes, const std::vector<std::vector<Polygons>>& model_collision)
+void TreeSupport::drawCircles(SliceDataStorage& storage, const std::vector<std::unordered_set<Node*>>& contact_nodes)
 {
     const Settings& mesh_group_settings = Application::getInstance().current_slice->scene.current_mesh_group->settings;
     const coord_t branch_radius = mesh_group_settings.get<coord_t>("support_tree_branch_diameter") / 2;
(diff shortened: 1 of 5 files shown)


Ghostkeeper:
I had tried this before and couldn't get it to work correctly. If you did, good job! Maybe we can figure something out with locking to fix the multithreading issue? I guess a review will tell.

Developers, see issue CURA-5912.

src/TreeSupport.cpp
```diff
@@ -650,32 +610,119 @@ void TreeSupport::insertDroppedNode(std::unordered_set<Node*>& nodes_layer, Node
...
+
+    if (layer == 0)
+    {
+        avoidance_cache_[key] = collision(radius, 0);
+        return avoidance_cache_[key];
+    }
+
+    // Avoidance for a given layer depends on all layers beneath it so could have very deep recursion depths if
+    // called at high layer heights
+    constexpr auto max_recursion_depth = 100;
```
BagelOrb: I find it hard to understand this piece of code.
Please add more comments in the code to explain what you are doing here.

It took me 15 minutes and a rubber duck to understand! (Am I dumb?)

I think it's not too important to prevent high recursion depths. You're only putting two ints and a pointer on the stack for every call.

Perhaps it's best to rewrite this code so that it just uses a for-loop instead of recursion.

src/TreeSupport.cpp
```diff
@@ -650,32 +610,119 @@ void TreeSupport::insertDroppedNode(std::unordered_set<Node*>& nodes_layer, Node
...
+    collision_areas = collision_areas.offset(xy_distance_ + radius, ClipperLib::JoinType::jtRound);
+    const auto ret = collision_cache_.insert({key, std::move(collision_areas)});
+    assert(ret.second);
+    return ret.first->second;
+}
+
+const Polygons& ModelVolumes::calculateAvoidance(const RadiusLayerPair& key) const
+{
+    const auto& radius = key.first;
+    const auto& layer = key.second;
```
BagelOrb: Please be more explicit in your variable names.
```suggestion
    const LayerIndex& layer_idx = key.second;
```

src/TreeSupport.cpp
```diff
@@ -650,32 +610,119 @@ void TreeSupport::insertDroppedNode(std::unordered_set<Node*>& nodes_layer, Node
...
+    avoidance_areas = avoidance_areas.unionPolygons(collision(radius, layer));
+    const auto ret = avoidance_cache_.insert({key, std::move(avoidance_areas)});
+    assert(ret.second);
+    return ret.first->second;
+}
+
+const Polygons& ModelVolumes::calculateInternalModel(const RadiusLayerPair& key) const
+{
+    const auto& radius = key.first;
+    const auto& layer = key.second;
```
BagelOrb: Me and my colleague discourage the over-use of `auto`.
On this specific line it's quite unclear what `layer` is - especially because it's not a layer, but a layer *number*.
```suggestion
    const LayerIndex& layer_index = key.second;
```

src/TreeSupport.h
```diff
@@ -7,10 +7,92 @@
...
 
+
 namespace cura
 {
+/*!
+ * \brief Lazily generates tree guidance volumes.
+ *
+ * NOTE: This class is not currently thread-safe and should not be accessed in OpenMP blocks
+ */
+class ModelVolumes
```
BagelOrb: why is this functionality in a separate class?

src/TreeSupport.h
```diff
@@ -130,28 +212,12 @@ class TreeSupport
...
-     * There will be a volume for each sample of branch radius. The radii of the
-     * branches are unknown at this point (there will be several radii at any
-     * given layer too), so a collision area is generated for every possible
-     * radius.
-     *
-     * \param storage The settings storage to get settings from.
-     * \param model_collision[out] A vector to fill with the output collision
-     * areas.
+     * Lazily computes volumes as needed.
+     *  Note: This class is NOT currently thread-safe and should not be accessed in OpenMP blocks
```
BagelOrb:
```suggestion
     *  \warning This class is NOT currently thread-safe and should not be accessed in OpenMP blocks
```

src/TreeSupport.h
```diff
@@ -7,10 +7,92 @@
...
+    const Polygons& calculateCollision(const RadiusLayerPair& key) const;
+    const Polygons& calculateAvoidance(const RadiusLayerPair& key) const;
+    const Polygons& calculateInternalModel(const RadiusLayerPair& key) const;
+
+    Polygons machine_border_;
+    coord_t xy_distance_;
+    coord_t max_move_;
+    coord_t radius_sample_resolution_;
+    std::vector<Polygons> layer_outlines_;
+    mutable std::unordered_map<RadiusLayerPair, Polygons> collision_cache_;
```
BagelOrb: I think it's a shame you didn't implement these as `std::vector<std::vector<LazyInitialization<Polygons>>>`, but I find it a shame solely because I made that class and it has the word 'Lazy' in it ;)

src/TreeSupport.cpp
```diff
@@ -650,32 +610,119 @@ void TreeSupport::insertDroppedNode(std::unordered_set<Node*>& nodes_layer, Node
...
+    if (it != internal_model_cache_.end())
+    {
+        return it->second;
+    }
+    else
+    {
+        return calculateInternalModel(key);
+    }
+}
+coord_t ModelVolumes::roundRadius(coord_t radius) const
```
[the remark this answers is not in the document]
bjude: Agreed

src/TreeSupport.cpp
```diff
@@ -650,32 +610,119 @@ void TreeSupport::insertDroppedNode(std::unordered_set<Node*>& nodes_layer, Node
     conflicting_node->support_roof_layers_below = std::max(conflicting_node->support_roof_layers_below, p_node->support_roof_layers_below);
 }
 
-void TreeSupport::propagateCollisionAreas(const SliceDataStorage& storage, const std::vector<std::vector<Polygons>>& model_collision, std::vector<std::vector<Polygons>>& model_avoidance)
+ModelVolumes::ModelVolumes(const SliceDataStorage& storage, Polygons machine_border, coord_t xy_distance,
```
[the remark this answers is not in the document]
bjude: `machine_border` is constructed in the `TreeSupport` constructor and doesnt live beyond that so using a reference or pointer wont work.

Passing by value and moving allows some compiler optimisation since it knows it can modify the passed in value. What i forgot to do was actually call std::move, i'll add it in the appropriate places. Theres a decent chance the compiler will do a move anyway but better safe than sorry

src/TreeSupport.h
```diff
@@ -7,10 +7,92 @@
...
+     * \brief Creates the areas that have to be avoided by the tree's branches.
+     *
+     * The result is a 2D area that would cause nodes of radius \p radius to
+     * collide with the model.
+     *
+     * \param radius The radius of the node of interest
+     * \param layer The layer of interest
+     * \return Polygons object
+     */
+    const Polygons& collision(coord_t radius, int layer) const;
```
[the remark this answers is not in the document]
bjude: Agreed, will change.

src/TreeSupport.cpp
bjude: The stack frames are small but i've been bitten by this before so wanted to be a bit defensive.
I'll add some comments to explain whats going on.

src/TreeSupport.cpp
bjude: Understood, i'll be the first to admit I can be a bit over enthusiastic with auto.
However, the advantage here is that it makes it trivial to swap out the types (from int to LayerIndex as you recommend above) without having to modify the code because the type interfaces are the same.

I think the issue here is the poor variable naming on my part. I'll clean up the naming, let me know if that provides enough clarity

src/TreeSupport.h
bjude: It allows the details of generating the various model volumes to be abstracted out of the tree support generation code. This both helps the TreeSupport code become cleaner (as it doesnt have the collision volume generation mixed in) and makes the coding for the lazy generation much simpler.

Theres not reason it _couldn't_ be implemented inside the TreeSupport object, but i'd argue theres no good reason it _should_.

src/TreeSupport.h
bjude: Using a vector of vectors would mean that a `LazyInitialization<Polygons>` would be allocated for every combination of radius and layer index. Even though the `Polygons` object is lazily allocated, there are still an unnecessary number of vectors allocated.
The unordered_map expresses the intent better to my mind.

src/TreeSupport.h
```diff
@@ -7,10 +7,92 @@
...
+     * are unable to reach the build platform
+     *
+     * \param radius The radius of the node of interest
+     * \param layer The layer of interest
+     * \return Polygons object
+     */
+    const Polygons& internal_model(coord_t radius, int layer) const;
+
+private:
+    using RadiusLayerPair = std::pair<coord_t, int>;
```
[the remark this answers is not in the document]
bjude: I wasnt aware of the LayerIndex type, i'll convert ModelVolumes to use it

src/TreeSupport.h
BagelOrb: Yeah I know. I was just joking.

src/TreeSupport.cpp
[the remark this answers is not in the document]
BagelOrb: Actually I think you should move the whole object into your new class.
It's only used for the stuff which that class is responsible for.

src/TreeSupport.cpp
BagelOrb: Why not use a for-loop isntead of recursion?

BagelOrb:
There's still some threads open. If all discussions are handled we can consider merging.

src/TreeSupport.h
bjude: Long day, humour detector faulty :)

src/TreeSupport.h
```diff
@@ -7,10 +7,92 @@
...
+
 #include "sliceDataStorage.h"
 
+
 namespace cura
 {
+/*!
+ * \brief Lazily generates tree guidance volumes.
+ *
+ * NOTE: This class is not currently thread-safe and should not be accessed in OpenMP blocks
```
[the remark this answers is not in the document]
bjude: Done

src/TreeSupport.cpp
```diff
@@ -650,32 +610,119 @@ void TreeSupport::insertDroppedNode(std::unordered_set<Node*>& nodes_layer, Node
...
+                           coord_t max_move, coord_t radius_sample_resolution) :
+    machine_border_{machine_border},
+    xy_distance_{xy_distance},
+    max_move_{max_move},
+    radius_sample_resolution_{radius_sample_resolution}
 {
-    model_avoidance.resize(model_collision.size());
+    for (auto i = 0; i < static_cast<int>(storage.support.supportLayers.size()); ++i)
```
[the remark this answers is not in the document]
bjude: Done

src/TreeSupport.cpp
```diff
@@ -650,32 +610,119 @@ void TreeSupport::insertDroppedNode(std::unordered_set<Node*>& nodes_layer, Node
...
+    {
+        layer_outlines_.push_back(storage.getLayerOutlines(i, false));
```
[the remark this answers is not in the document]
bjude: Done

src/TreeSupport.cpp
```diff
@@ -650,32 +610,119 @@ void TreeSupport::insertDroppedNode(std::unordered_set<Node*>& nodes_layer, Node
...
+    return radius + delta;
+}
+
+const Polygons& ModelVolumes::calculateCollision(const RadiusLayerPair& key) const
+{
+    const auto& radius = key.first;
+    const auto& layer = key.second;
+
+    auto collision_areas = machine_border_;
+    if (layer < static_cast<int>(layer_outlines_.size())) {
```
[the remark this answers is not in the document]
bjude: Done

src/TreeSupport.h
```diff
@@ -7,10 +7,92 @@
...
+
+private:
+    using RadiusLayerPair = std::pair<coord_t, int>;
+
+    coord_t roundRadius(coord_t radius) const;
+    const Polygons& calculateCollision(const RadiusLayerPair& key) const;
+    const Polygons& calculateAvoidance(const RadiusLayerPair& key) const;
+    const Polygons& calculateInternalModel(const RadiusLayerPair& key) const;
+
+    Polygons machine_border_;
```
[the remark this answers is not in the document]
bjude: Done

src/TreeSupport.cpp
bjude: Variable name clarified. I've left auto for now since the variable name says enough IMO

src/TreeSupport.cpp
bjude: It complicates the logic a bit because we need to either loop backwards from `layer_idx` until we find the highest layer in the cache or loop upwards from zero and then run the loop forwards. Something like this (theres probably an off-by-one error in here):

```suggestion
    auto prev_layer = layer_idx - 1;
    for (; prev_layer > 0; --prev_layer)
    {
        if (avoidance_cache_.find({radius, prev_layer}) != avoidance_cache_.end())
        {
            break;
        }
    }
    for (; prev_layer < layer_idx + 1; ++prev_layer)
    {
        auto avoidance_areas = avoidance_cache_[{radius, prev_layer}].offset(-max_move_).smooth(5);
        avoidance_areas = avoidance_areas.unionPolygons(getCollision(radius, prev_layer + 1));
        const auto ret = avoidance_cache_.insert({key, std::move(avoidance_areas)});
    }

    return avoidance_cache_[key];
```

This isn't as clean to my mind but i guess the recursive solution isn't particularly clean when the defensive part is included. I'll leave it up to you, if the for loop is the preference i can do that instead.

src/TreeSupport.cpp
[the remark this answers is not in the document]
bjude: Thats a good point. I initially tried to modify the TreeSupport class as little as possible but it doesnt make sense for it to be outside ModelVolumes

I'll move it in.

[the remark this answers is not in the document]
BagelOrb: I think that in this instance the machine border can be thought of as a special model which is always there.

I'm not sure where you are putting it now, but it would be a bit too much to make a separate file for it - if that's what you were planning.

I would be okay with the code being in SliceDataStorage somewhere.

src/TreeSupport.h
BagelOrb: Okay

src/TreeSupport.cpp
```diff
@@ -22,10 +22,11 @@
 namespace cura
 {
 
-TreeSupport::TreeSupport(const SliceDataStorage& storage)
+namespace {
```
BagelOrb: The reasons you mention for making it anonymous namespace sound to me like perfect reasons to make this function a private class function.

By making it a static function in anonymous namespace you are bypassing the documentation rules.

Please make it into a private (static) function and document it in the header file.

bjude: Done

src/TreeSupport.cpp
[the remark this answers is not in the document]
bjude: Ok, i've moved the machine border code to SliceDataStorage and the machine border collision area to a private static function in ModelVolumes

BagelOrb:
Code looks good!

Moving this issue to [TEST].

rburema:
Hi

@BagelOrb, Do you also intend to do the Component Test (before it's merged and pushed to Q&A), or should someone else (me for example) pick that up?

at everyone: I see some conflicts have also been introduced, those should be fixed before pushing to master as well.

cheers, Remco.

Ghostkeeper:
Discovered no new bugs in the behaviour now. It's fun to see how the progress bar is initially still slow as it enters the support generation stage, but then gradually speeds up as the lazy evaluation kicks in.

On an example case that handles most ordinary cases, slicing times (N=1) are:
* Before this PR: 76.3s
* After this PR: 64.4s

Not as much as I expected, but still a good improvement.

PR #901, opened 2018-11-05
Remove unused imports

Ghostkeeper:
I've ran [cppclean](https://pypi.org/project/cppclean/) on CuraEngine to see what imports could be removed to improve compilation time. This is what it came up with.

It's also coming up with several places where we have static data (which could prevent us from slicing multiple times in the same run) and some unused variables. That is not fixed in this PR.

Compilation times recorded so far:
* My Ubuntu laptop: Reduced compile time from 81s to 68s (includes building tests).
* Msuurmond's MacOS: Increased compile time from 67s to 72s (noise?)
* Diegopradogesto's Windows: Reduced compile time from 121s to 99s.

It also reduces the amount of files that need to get re-compiled if you change one of the headers.

It's worth noting that our code style guidelines for C++ also specify that we should use forward declarations wherever possible. This implements that over the entire code base.


Ghostkeeper:
Developers, see issue CURA-5909.

PR #903, opened 2018-11-08
Feature: "Z Hop After Extruder Switch Height"

vgribinchuk:
Feature: "Z Hop After Extruder Switch Height"
Backend implementation of the feature: https://github.com/Ultimaker/Cura/pull/4753


diegopradogesto:
Devs. CURA-5920

LipuFei:
@vgribinchuk There are merge conflicts unfortunately.

vgribinchuk:
@LipuFei Merge conflicts have been fixed

src/LayerPlan.cpp
```diff
@@ -1463,7 +1473,8 @@ void LayerPlan::writeGCode(GCodeExport& gcode)
                 gcode.writeRetraction(retraction_config);
                 if (path.perform_z_hop)
                 {
-                    gcode.writeZhopStart(retraction_config.zHop);
+                    gcode.writeZhopStart(z_hop_height);
+                    z_hop_height = retraction_config.zHop; // back to normal z hop
```
rburema: It could be that I'm missing something, but I don't get this line: `z_hop_height` is not a reference, is not used below this line and is defined _within_ the scope of the enclosing for loop, so what effect is this supposed to have?

vgribinchuk: Yes, this is redundant. Thanks! Fixed.

src/LayerPlan.cpp
```diff
@@ -1463,7 +1473,7 @@ void LayerPlan::writeGCode(GCodeExport& gcode)
                 gcode.writeRetraction(retraction_config);
                 if (path.perform_z_hop)
                 {
-                    gcode.writeZhopStart(retraction_config.zHop);
+                    gcode.writeZhopStart(z_hop_height);
```
rburema: Hi @vgribinchuk , I did some testing, and it seems that (the change from original to) this line causes the 'height when switched' to be applied to every retraction z-hop, instead of just those for the nozzle switch. I don't think that's what you meant to do. cheers, Remco.

vgribinchuk: Hi, @rburema ! Thank you for noticing that. I will have a look and check as soon as I can.

vgribinchuk: `z_hop_height = retraction_config.zHop;`
got back this line, since it is inside the `for` loop, and `writeZhopStart` is called multiple times. To not apply the 'height when switched' to every retraction, `writeZhopStart` is called now with correct z-hop value.

rburema: My original point was that the line _after_ this one doesn't do anything, since `z_hop_height` isn't referred to after that point, and will just reset to the value given by the declaration of `z_hop_height` in the next iteration of the loop. Which had me worried you might've forgotten to commit something. That didn't/doesn't seem to be the case, so I think we can move ahead.

rburema: Ah I see, there was an inner for-loop after all. Not sure how I could have missed that. Sorry!

PR #904, opened 2018-11-12
Allow points to slightly overlap lines to avoid generating vast numbers of small line segments.

smartavionics:
Avoid generating a new point for each of the small overlaps that can occur when curved
walls lie next to each other and the vertices of the wall with the smaller radius
protrude into the line segments of the wall with the larger radius

This fix only helps when curved walls are adjacent and not actually overlapping.

Here's a project that benefits from this PR. Without the PR, the gcode contains very many small line segments which have a negative effect on print quality.

[CFDMP_80.zip](https://github.com/Ultimaker/CuraEngine/files/2571413/CFDMP_80.zip)

### DIFF
--- a/src/utils/PolygonProximityLinker.cpp
+++ b/src/utils/PolygonProximityLinker.cpp
@@ -189,6 +189,22 @@ void PolygonProximityLinker::findProximatePoints(const ListPolyIt a_point_it, Li
         return;
     }
 
+    if (dist2 > proximity_distance_2 * 0.9)
+    {
+        // avoid generating a new point for each of the small overlaps that can occur when curved walls lie next to each other
+        // and the vertices of the wall with the smaller radius protrude into the line segments of the wall with the larger radius
+
+        // doing this provides a double win in that (a) less points need to be processed from here on and, more importantly, (b) it reduces the
+        // likelyhood of very short line segments being created which can produce poor print quality
+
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Good idea. CURA-6062.

smartavionics:
The reason this was investigated was because a user was getting poor print quality. This change improved the print quality for them.

src/utils/PolygonProximityLinker.cpp
```diff
@@ -189,6 +189,22 @@ void PolygonProximityLinker::findProximatePoints(const ListPolyIt a_point_it, Li
         return;
     }
 
+    if (dist2 > proximity_distance_2 * 0.9)
```
Ghostkeeper: This doesn't filter out any short line segments. It filters out line segments that are sort-of-close but not very close. I think this is then not a good change to make because it negatively affects accuracy. If they are sort-of-close but not very close (<10% overlap) then the flow rate won't change much anyway, so the visual quality will not be affected much.

You might've seen a user here that judged print quality based on how layer view looked, or based on a single experiment with one model.

The original idea is good to me, but this implementation is not implementing that idea.

smartavionics:
I don't care anymore if you use this PR or not. I'll keep using it.

PR #906, opened 2018-11-13
Move optional to namespace cura

bjude:
Adding names to namespace std is undefined behaviour. This probably wont bite anyone until the project adopts a new c++ version, but is UB nonetheless.

This PR also cleans up the interface of optional to behave a bit more like the std::optional from c++17 which should make the upgrade a bit easier later on. Perhaps not as easy as `namespace cura { using optional = std::optional }`, but still easier.

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -1687,11 +1687,11 @@ bool FffGcodeWriter::processInsets(const SliceDataStorage& storage, LayerPlan& g
     return added_something;
 }
 
-std::optional<Point> FffGcodeWriter::getSeamAvoidingLocation(const Polygons& filling_part, int filling_angle, Point last_position) const
+cura::optional<Point> FffGcodeWriter::getSeamAvoidingLocation(const Polygons& filling_part, int filling_angle, Point last_position) const
 {
     if (filling_part.empty())
     {
-        return std::optional<Point>();
+        return cura::optional<Point>();
     }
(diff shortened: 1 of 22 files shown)


alekseisasin:
@rburema , Can you take a look at it?

rburema:
Hi @bjude,
Thanks for the time you took to try to help us out. I certainly understand that it seems ugly to have someone muck about with the standard namespace in this way. There's a decent explanation for this however!

First, keep in mind that we're currently at C++11 and plan to go to C++14 before we will think about C++17 (which might be a while!). Then, you might see that the current seemingly 'hacked in' `optional` will give _exactly_ the same behaviour as the C++14 one, as it's copied from the GCC C++14 one. You've probably realized where this is going: once we do the upgrade to C++14, we only have to remove that one file.

With all of that said, we'll have to reject this PR. I hope you'll understand.
cheers, Remco.

bjude:
Understood, practicality often trumps correctness :)

This version of std::optional is significantly different to the current one in the standard and does not propagate const correctly (amongst other things). Do you know which version of libstdc++ was the current implementation taken from? Relying on an experimental version from the libstdc++ standard library will preclude compiling with msvc or clang (if using libc++)

rburema:
We'll be discussing moving to C++14 soonish, so hopefully we won't need all this practicality much longer.

As to where it's from _exactly_, I'm not sure. It was added before my time. Maybe @Ghostkeeper can help with that?

PR #908, opened 2018-11-14
After merging ironing lines with infill lines the ironing skips some …

alekseisasin:
…parts

CURA-5855_ironing_skips_some_parts

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -851,6 +851,8 @@ LayerPlan& FffGcodeWriter::processLayer(const SliceDataStorage& storage, LayerIn
                 
                 // path optimization is currently broken when using gyroid infill
                 disable_path_optimisation = disable_path_optimisation || mesh.settings.get<EFillMethod>("infill_pattern") == EFillMethod::GYROID;
+                // path optimizer merges ironing lines with infill lines. Result of this the ironing skips some parts.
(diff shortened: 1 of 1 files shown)


src/FffGcodeWriter.cpp
```diff
@@ -851,6 +851,8 @@ LayerPlan& FffGcodeWriter::processLayer(const SliceDataStorage& storage, LayerIn
                 
                 // path optimization is currently broken when using gyroid infill
                 disable_path_optimisation = disable_path_optimisation || mesh.settings.get<EFillMethod>("infill_pattern") == EFillMethod::GYROID;
+                // path optimizer merges ironing lines with infill lines. Result of this the ironing skips some parts.
+                disable_path_optimisation = disable_path_optimisation || (mesh.settings.get<bool>("ironing_enabled") && !mesh.layers[gcode_layer.getLayerNr()].top_surface.areas.empty());
```
Ghostkeeper: This is not a good solution, since it disables path optimisation for the entire print if ironing is enabled in any mesh. This causes a lot of prints to become ugly, increase printing time by an order of magnitude in some cases, etc. That is actually also a problem with that Gyroid infill solution above.

rburema:
General note: The PR for the infill line fix (#911) also solves this issue.

rburema:
I'll close this pull request, as this will be fixed with that other PR (eventually) and we don't want people accidentally merging this one instead.

PR #911, opened 2018-11-15
CURA-5776 Fix line merging

LipuFei:


src/MergeInfillLines.cpp
```diff
@@ -248,9 +249,19 @@ MergeInfillLines::MergeInfillLines(ExtruderPlan& plan)
 
             if (second_path.config->isTravelPath())
             {
+                has_first_path = false;
                 continue; //Skip travel paths, we're looking for the first non-travel path.
             }
 
+            // Use the first non-travel path as the first path that can be used for merging. After we encounter
+            // a travel path, we need to find another first non-travel path for merging.
+            if (!has_first_path) {
```
Ghostkeeper: Code style: Brackets on new line.
```suggestion
            if (!has_first_path)
            {
```

Ghostkeeper:
Quite often it's not merging small gap lines even if they are completely perpendicular. This is Benchy scaled to 50%, sliced with the fast profile:
screenshot from 2018-12-03 15-15-03

Ghostkeeper:
It also doesn't seem to merge skin lines at all any more. That is kind of important, since the skin tends to be used more for models that have many thin walls, since they are often the kinds of models that people want to print at 100% infill.

rburema:
@Ghostkeeper, have you compared that Benchy-slice to the current master? I don't get that much different results. Even when I remove the check that was reintroduced in order to fix the 'initial' version of this bug, those lines remain unmerged.

PR #922, opened 2018-11-28
Improve the no-skin combing boundary

smartavionics:
This fixes a problem when no-skin combing on a part that has thin walls and no normal insets.

Without this PR, you can get travels that cross air when they could be over walls. Like this...

screenshot_2018-11-28_17-26-34

With this PR you get this...

screenshot_2018-11-28_17-22-29

### DIFF
--- a/src/LayerPlan.cpp
+++ b/src/LayerPlan.cpp
@@ -220,8 +220,14 @@ Polygons LayerPlan::computeCombBoundaryInside(const size_t max_inset)
                             inner = part.insets[num_insets - 1].offset(-10 - mesh.settings.get<coord_t>("wall_line_width_x") / 2);
                         }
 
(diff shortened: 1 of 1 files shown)


alekseisasin:
Ref: CURA-5989

smartavionics:
I have added a small change - it now merges the part's infill and perimeter gaps into the no-skin combing region. This can reduce the complexity of the combing region and give a little more area for combing. In the images below, the thin yellow lines are the combing boundary. Note the difference in print times for this model with this change included.

Perimeter gaps included in no-skin combing boundary...

screenshot_2019-01-10_10-53-40

Perimeter gaps not included in no-skin combing boundary...

screenshot_2019-01-10_10-56-26

PR #923, opened 2018-11-29
Feature interface generation

vgribinchuk:
Feature: "Build Interface If Support Is Present"
Backend implementation of the feature: https://github.com/Ultimaker/Cura/pull/4905


diegopradogesto:
Devs. CURA-6106

diegopradogesto:
This is actually a bug and it should be fixes without a new setting. Could you take a look if you can change this PR?

diegopradogesto:
Thanks, we will take a look!

src/support.cpp
```diff
@@ -19,6 +19,25 @@
...
+{
+    const cura::Settings& mesh_group_settings = cura::Application::getInstance().current_slice->scene.current_mesh_group->settings;
+    const cura::ExtruderTrain& infill_extruder = mesh_group_settings.get<cura::ExtruderTrain&>("support_infill_extruder_nr");
+    const cura::coord_t support_line_width = infill_extruder.settings.get<cura::coord_t>("support_line_width");
+    const size_t wall_line_count = infill_extruder.settings.get<size_t>("support_wall_count");
+    const cura::coord_t support_offset = wall_line_count > 0 ? -support_line_width / 2 : 0;
+    return support_offset;
+}
+
+}
```
Ghostkeeper: I don't like global functions in the global namespace. I'd move this one in the AreaSupport class...

src/support.cpp
```diff
@@ -1399,7 +1418,30 @@ void AreaSupport::generateSupportBottom(SliceDataStorage& storage, const SliceMe
...
+                {
+                    support_bottoms.add(bottom_part);
+                }
+            }
+            support_layers[layer_idx].support_bottom.add(support_bottoms);
+        }
+        else
+        {
+            support_layers[layer_idx].support_bottom.add(bottoms);
+        }
```
Ghostkeeper: This part should be moved into the `generateSupportInterfaceLayer` function or a separate one beneath it. That would also prevent the duplication here. The only changes really are the `+ 1` vs. `- 1`.

src/support.cpp
vgribinchuk: Thank you for review, I will have a look and make changes

vgribinchuk: Was moved to the AreaSupport scope

src/support.cpp
vgribinchuk: Refactored, thank you

vgribinchuk:
Hi, @Ghostkeeper ! I've submitted fixes, could you please review it once again? Thank you!

LipuFei:
Hi @vgribinchuk , I tested your PR and the bug still seems to be there. This the project I used for testing: 
[UMS5_testcube_05mm.zip](https://github.com/Ultimaker/CuraEngine/files/3481766/UMS5_testcube_05mm.zip). Turn off "auto drop onto build plate".

Here's a screenshot:
image

Ellecross:
Hey @vgribinchuk will you be able to take a look at @LipuFei 's comments ? Thanks

vgribinchuk:
Hi @Ellecross @LipuFei ! Yes, sure. I'm working on solution and will provide it soon.

vgribinchuk:
Could you please check changes? Thank you

AbeFM:
Instead of posting everything over there, since it seems to be part of this discussion I'll include it here.

ASIDE:
PLEASE let me know if I'm breaking etiquette. We're all on the same team, please assume I didn't waste my saturday to make you look bad. If something bothers you, please try to stick to the facts and not hold any offense against the information provided. Let me know if I'm out of line but don't clutter up the thread. It's the last I'll bring that up.

I have looked at 4.2.1 vs the current stuff and found this issue is only present in the later version.
I did some varied slicing, and found a lot of stuff appearing and disappearing at times that didn't make sense to me:

> Model Slopes of 15 and 30 degrees (75 and 60 from vertical)
Minimum support areas are both 0, x/y is 0.3 and min x/y is 0.
> 
> Interfaces Ex1, Infill Ex0, 65*
> image
> 
> Interfaces Ex1, Infill Ex1, 65*
> image
> 

Changing the Support Infill extruder still leaves small gaps at the middle roof of the model, I think they are min x/y distance issues, but it's too far from the wall as I only have the distance set to 0.3, 0.0 minimum (the images are from SA's build)

Otherwise is looks like the supports filled in with the change in support extruder selection.

> Interfaces Ex1, Infill Ex0, 55*
> image
This is pretty bizarre. Both of the lower slopes are > 55*, so perhaps that's doing it both in front and behind the bottom hill.

> 
> Interfaces Ex1, Infill Ex1, 55*
> image
Again, the issue resolves if you use a single extruder for support


> Interfaces Ex1, Infill Ex1, 30*
> image
> 
30 degrees. matched extruders behaves as expected, but....
> 
> Interfaces Ex1, Infill Ex0, 30*
> image
> 
Mismatched extruders leaves a gap just on the right side which is not present in the homo-extruder set up.
Ah, the gap comes from supporting or not supporting the bridge from the edge of the angled bit to the extreme of the overhang. Why that change with extruder selection I don't get.

Two final slices:
Interfaces Ex1, Infill Ex1, 78*
image
Odd, there is no support for the horizontal bits at all. Not sure why.

Interfaces Ex1, Infill Ex0, 78*
image
Same strange unsupported roofs on the left, weird support floor garbage in between the supports in the middle... Two more views.
image
image


> Ok, how about this for a theory: The roof or floor is being interpreted as the mirror of the angle (90-angle) instead of the angle, so they are doing different things?

I suspect this is WAY long enough, but I hope it helps someone see what's going on!

Ghostkeeper:
> PLEASE let me know if I'm breaking etiquette.

I think you're doing fine! No worries and thank you for the analysis as always. One thing - I don't know if you'd call it etiquette - but for us it's easier to have each bug reported separately. It makes it easier to oversee what has been handled and what has not.

We are currently still working on the interface generation since our QA team discovered issues with it as well after this PR.

Is the overhang angle different in the two extruders? Or the join distance maybe? If the two extruders have the same settings, I would expect that there can be no difference when switching the support infill extruder. Ergo, there must be some difference in the settings. Finding out what the difference is can shed a light on why it's reacting this way. Note that only one of the extruders' values is shown at a time: The one that should be used by slicing. The value that's being used switches to the value of the other extruder when you switch the support infill extruder.

> Odd, there is no support for the horizontal bits at all. Not sure why.

Maybe the support is too thin to print with the selected line width?

> Ok, how about this for a theory: The roof or floor is being interpreted as the mirror of the angle (90-angle) instead of the angle, so they are doing different things?

The angle of the part that the support rests on doesn't influence support generation above it. Only within the X/Y and Z distance and the Stair Step Height and Stair Step Width would I expect any changes. So higher up in the support it shouldn't remove support just because there is something sloped beneath it.

smartavionics:
I don't have anything to say regarding this issue other than to state that the changes I have made to the support code in my builds simply adjust the x/y distance of support from the model when the support is resting on a sloped surface. It makes the support and/floor come closer to the model. The PR for that is https://github.com/Ultimaker/CuraEngine/pull/1115 if you want to check it out. Apart from that, I think the support code in my builds are bug-for-bug compatible with the upstream code.

Ghostkeeper:
> Are you saying that if the values weren't linked there would be some value there specific to the extruder? If you set EX0 to "5" and EX1 to "8", then "8" would be displayed because it is the most recent, but "5" would still be stored somewhere w/r/t EX0?

Not exactly. "8" would be displayed because the Support Extruder setting is set to print with EX1. But indeed it's storing a different value for each extruder because the profiles can be different. E.g. if you have one extruder with PLA and the other with PVA, you'd want to get different settings if you change the support extruder to the one with PVA.

> Oh - but Support Pattern and Support Density ARE changing with the selected "Support Infill Extruder" - even though the values are linked, the USED value is changing and is not made obvious to the user that these are happening. Similarly, I believe Gradual Infill Steps is an issue here.... AND Support X/Y Distance... There's QUITE a lot changing when I change 'one' variable!

Precisely, because you change the extruder a lot of things have to change.

> Does it make sense to have a popup that details every setting being changed other than the one you clicked on?

Many settings change some other setting somewhere. For instance, if you change the infill density that's going to affect the infill line distance setting. It'd get very annoying to have pop-ups for everything. We've been trying to get rid of pop-ups as much as possible anyway. And if you change the infill density and it pops up with "Wait! This'll change the infill line distance!", what is the user supposed to do then? Saying "No I don't want that." will make the infill density change have no effect at all.

The screenshots with large swaths of support missing seems like a serious bug to me, but I haven't been able to reproduce it yet.

The screenshots with the red arrow pointing at a certain border doesn't seem like a problem to me, but it's hard to see which parts are interface and which parts are infill. Maybe it's a problem with the support interface resolution?

smartavionics:
Disabling this PR stops the support from disappearing when the interfaces are deep.

This is what you get when this PR is disabled:

Screenshot_2019-08-26_21-43-46

Screenshot_2019-08-26_21-43-15

Screenshot_2019-08-26_21-42-38

This is what you get when this PR is enabled:

Screenshot_2019-09-02_11-17-02

Screenshot_2019-09-02_11-20-00

Screenshot_2019-09-02_11-20-38

Ghostkeeper:
So today I get to investigate this topic again. I've discussed it with @nallath and @diegopradogesto as well and we've decided to revert this pull request. Since this has had a lot of investigation from smart people and it's a complex problem, I'm writing down very thoroughly what our reasons for it are!

Summary
----
We think that this change is not a complete solution to this problem. While a partial solution is normally better than no solution, this solution also causes many other problems. We think the additional problems outweigh the small case that is solved by this solution.

Partial solution
----
This solution is not a complete solution because the problem is not particular to interface. The original problem also occurs if interface is disabled. Here is a piece of overhang with X/Y distance set very high, Z distance and minimum X/Y distance set very low:
Screenshot from 2019-09-26 16-30-22
Since Z overrides X/Y, the Z distance is held where the overhang needs supporting, and the X/Y distance is only applied in the vertical parts where support is supporting other support. As such the support that supports the model is still there but the support that supports other support keeps my massive 20mm X/Y distance and then disappears completely. Since the interface is generated by separating it from the normal support areas, they have the same problem. But the problem is not particular to the interface.

Similarly you can also see that it's a partial solution by combining support and support interface:
Screenshot from 2019-09-26 16-35-03

The interface on the lowest layer gets trimmed because it's not being supported by anything. But the normal support body doesn't, even though they have the same problem.

Regressions
----
Apart from being a partial solution, it causes a number of regressions in real-world cases. The best example was found by @Ellecross:
image-2019-09-03-14-42-39-036
The interface on the right hand side was not supported by any support, and then it gets removed. I believe that the interface on the left hand side didn't get removed because there is a piece of support next to it that is invisible due to being smaller than a line width.

This is similar to the models from AbeFM, best seen in this screenshot:
64108048-bc900900-cd73-11e9-90ad-c785d72977f4
It's happening here on the front/left part, where the interface is the only part left because the roof is so close. After the first "bump" it occurs less because the second roof is high enough that after a few layers there is actual support infill which doesn't get removed, and all interface layers in subsequent layers don't get trimmed any more either. This case is less of a real-world example though.

New partial solution
----
I'm reverting the original solution and proposing a new one in this PR: (link coming right up).

The new pull request only gives a warning when the difference between the X/Y distance and the minimum X/Y distance becomes big. The difference between these two settings is the maximum overhang you're experiencing. This is an illustrative example where I set the X/Y distance to 8 and minimum X/Y distance to 7. This results in a 1mm overhang:
Screenshot from 2019-09-26 16-55-58
The new solution is to simply give a warning when this overhang is too big. It's not a good solution, but it doesn't break anything either.

PR #924, opened 2018-12-01
Mb spiralize fixes 1

smartavionics:
Spiralize bug fixes as described in https://github.com/Ultimaker/CuraEngine/pull/822.

Fixes https://github.com/Ultimaker/Cura/issues/4916.

PR #925, opened 2018-12-01
Mb spiralize fixes 2

smartavionics:
Tapers top and bottom spiral layers as described in https://github.com/Ultimaker/CuraEngine/pull/822.

This PR assumes that https://github.com/Ultimaker/CuraEngine/pull/924 has already been merged.

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -165,31 +165,32 @@ unsigned int FffGcodeWriter::findSpiralizedLayerSeamVertexIndex(const SliceDataS
         ConstPolygonRef last_wall = (*storage.spiralize_wall_outlines[last_layer_nr])[0];
         ConstPolygonRef wall = layer.parts[0].insets[0][0];
         const int n_points = wall.size();
-        const int last_layer_seam_vertex_idx = storage.spiralize_seam_vertex_indices[last_layer_nr];
-        const Point& last_layer_seam_vertex = last_wall[last_layer_seam_vertex_idx];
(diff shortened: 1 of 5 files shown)


diegopradogesto:
Same internal ticket for both PRs.

Devs. CURA-5816

rburema:
Hi @smartavionics,
I took some more time because I wanted to figure out a way to do what you did with the top-layer without your self-reported hack, but ultimately I couldn't come up with anything that would give the same gcode, that also wouldn't require bigger rewrites outside of the scope of a pull request. I decided to let it be, since it's documented and quite isolated.

I _did_ fix an issue where a vector didn't report the right size, and the `capacity` was queried instead of the `size`. Both the incorrect size and the reason `capacity` worked where because of the way the vector was initialized. (Which is the reason this PR is merged indirectly, with the additional https://github.com/Ultimaker/CuraEngine/commit/b5b17df3e0fd992429ea84a198a848f02efaa63a commit.)
cheers, Remco.

smartavionics:
Hello Remco,
> I took some more time because I wanted to figure out a way to do what you did with the top-layer without your self-reported hack, but ultimately I couldn't come up with anything that would give the same gcode, that also wouldn't require bigger rewrites outside of the scope of a pull request. I decided to let it be, since it's documented and quite isolated.

Yes, it's not ideal but there was no other easy way to avoid the jump in extrusion rate that would occur and the whole point of this is to improve the visual look at the top of the spiral.

> I did fix an issue where a vector didn't report the right size, and the capacity was queried instead of the size. Both the incorrect size and the reason capacity worked where because of the way the vector was initialized. (Which is the reason this PR is merged indirectly, with the additional b5b17df commit.)

That's a good catch. Thanks for that.

PR #933, opened 2018-12-21
Cura 5864 prime tower brim enable

rburema:


src/FffPolygonGenerator.cpp
```diff
@@ -991,18 +991,29 @@ void FffPolygonGenerator::processPlatformAdhesion(SliceDataStorage& storage)
...
     ExtruderTrain& train = mesh_group_settings.get<ExtruderTrain&>("adhesion_extruder_nr");
+
+    Polygons first_layer_outline;
+    coord_t primary_line_count;
     switch(mesh_group_settings.get<EPlatformAdhesion>("adhesion_type"))
     {
     case EPlatformAdhesion::SKIRT:
-        SkirtBrim::generate(storage, train.settings.get<coord_t>("skirt_gap"), train.settings.get<size_t>("skirt_line_count"));
+        primary_line_count = train.settings.get<size_t>("skirt_line_count");
+        SkirtBrim::getFirstLayerOutline(storage, primary_line_count, true, first_layer_outline);
```
Ghostkeeper: Why did you move this call here instead of from `generate()`? You could also get the line count from the `generate` function, right?

rburema: We need the line count for the `getFirstLayerOutline` call as well. (Which was split off from `generate` because we can then differentiate between the case where the first layer outline is the entire plate versus just the tower.)

src/PrimeTower.h
```diff
@@ -48,7 +48,9 @@ class PrimeTower
 
 public:
     bool enabled; //!< Whether the prime tower is enabled.
+    bool specific_brim; //<! Whether the 'brim for prime tower' option is enabled (as opposed to brim in general).
```
LipuFei: If so, I would use `brim_for_prime_tower_enabled` as the name because it's more self-explanatory.

rburema: Turns out that variable wasn't even needed any more. I'll make a new commit.

Issue #930, opened 2018-12-12
[3.6.0] Slicing for second extruder does not work like before.


ipepe:
**Application Version**
3.6.0

**Platform**
Mac OS X - Darwin

**Qt**
Official Ultimaker website

**PyQt**
Official Ultimaker website

**Display Driver**
Intel Iris Pro 1536 MB graphics

**Steps to Reproduce**
`CuraEngine slice -j dual_extruder_printer.def.json -s adhesion_extruder_nr=1 -e1 -l 'file.stl' -o 'output.gcode'`

**Actual Results**
output.gcode is printing on `T0` extruder

**Expected results**
output.gcode is printing on `T1` extruder

**Additional Information**
Slicing on second extruder was working fine on version 3.5.1. I can create repository with example `dual_extruder_printer.def.json` if needed. I suppose there were changes regarding `extruders_enabled_count` and probably some more values that I should consider in my def.json? The worst part is that it's not failing, but it proceeds to generate gcode with print on T0 silently.

ipepe:
Well that sounds great... But that also means that I should compile Windows, Linux and Mac version of library. And that does not sound so great...

If You are aware of this issue, then why it's not merged?

rburema:
_Possible_ duplicate of (frontend) [issue](https://github.com/Ultimaker/Cura/issues/5103), in which case it was fixed already (see #974, will be in the _second_ 4.0-BETA, which isn't out yet).

@ipepe, Please note that @WinstonMao isn't part of the team and can't merge anything into our repo here.

ipepe:
Thanks for information.

ipepe:
Today I tried slicing on second extruder with supports and adhesion on T1 and I did not succeed.

rburema:
Hi @ipepe, Could you send `dual_extruder_printer.def.json` and the input .stl? You can post them here if you rename the files to have a .zip extension, so you don't have to make an entire repository just for a bug-report.

ipepe:
Hi @rburema. 
I created https://github.com/ipepe/CuraEngine-second-extruder-slicing for You to look into.

rburema:
Hi @ipepe Thanks for the effort. I can see that in the `machine_start_gcode` value hardcodes values for `T`, the tool-switch command. We give a `T` command _before_ the start-gcode, so that the start-gcode can run on the selected extruder.

[...]

PR #935, opened 2018-12-25
fix error of slicing dual model

WinstonMao:
It should be the last setting of  extruder,not the extruder[0]

### DIFF
--- a/src/communication/CommandLine.cpp
+++ b/src/communication/CommandLine.cpp
@@ -100,7 +100,6 @@ void CommandLine::sliceNext()
 
     slice.scene.extruders.reserve(arguments.size() >> 1); //Allocate enough memory to prevent moves.
     slice.scene.extruders.emplace_back(0, &slice.scene.settings); //Always have one extruder.
-    ExtruderTrain& last_extruder = slice.scene.extruders[0];
 
     for (size_t argument_index = 2; argument_index < arguments.size(); argument_index++)
     {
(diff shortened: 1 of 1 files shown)


WinstonMao:
There is an issue that describes the details of this error:
https://github.com/Ultimaker/CuraEngine/issues/930

> Can you provide a project file where we can easily see the bug?
> 
> Devs. CURA-6109

src/communication/CommandLine.cpp
```diff
@@ -205,7 +204,7 @@ void CommandLine::sliceNext()
 
                         const FMatrix3x3 transformation = last_settings->get<FMatrix3x3>("mesh_rotation_matrix"); //The transformation applied to the model when loaded.
 
-                        if (!loadMeshIntoMeshGroup(&slice.scene.mesh_groups[mesh_group_index], argument.c_str(), transformation, last_extruder.settings))
+                        if (!loadMeshIntoMeshGroup(&slice.scene.mesh_groups[mesh_group_index], argument.c_str(), transformation, *last_settings))
```
Ghostkeeper: I think, if you're slicing two models with per-object settings, this would result in the settings from the first object bleeding over into the second object. Like this:

    CuraEngine -l first.stl -s infill_line_distance=1.2 -l second.stl

`second.stl` would also get sliced with 1.2mm infill line distance, right?

PR #937, opened 2018-12-28
fix error

schneider007:
If the coordinate is negative, it will be converted to an unsigned positive number before divided by an unsigned number, which will get a wrong result.

### DIFF
--- a/src/infill.cpp
+++ b/src/infill.cpp
@@ -388,7 +388,8 @@ void Infill::generateCrossInfill(const SierpinskiFillProvider& cross_fill_provid
 
 void Infill::addLineSegmentsInfill(Polygons& result, Polygons& input)
 {
-    ClipperLib::PolyTree interior_segments_tree = in_outline.lineSegmentIntersection(input);
+	ClipperLib::PolyTree interior_segments_tree;
+	in_outline.lineSegmentIntersection(input, interior_segments_tree);
     ClipperLib::Paths interior_segments;
     ClipperLib::OpenPathsFromPolyTree(interior_segments_tree, interior_segments);
     for (size_t idx = 0; idx < interior_segments.size(); idx++)
--- a/src/infill/SubDivCube.cpp
+++ b/src/infill/SubDivCube.cpp
@@ -37,7 +37,7 @@ SubDivCube::~SubDivCube()
 void SubDivCube::precomputeOctree(SliceMeshStorage& mesh)
 {
     radius_addition = mesh.settings.get<coord_t>("sub_div_rad_add");
-    AngleRadians infill_angle = M_PI / 4.0;
+    AngleRadians infill_angle = 45;
 
     const coord_t furthest_dist_from_origin = std::sqrt(square(mesh.settings.get<coord_t>("machine_height")) + square(mesh.settings.get<coord_t>("machine_depth") / 2) + square(mesh.settings.get<coord_t>("machine_width") / 2));
     const coord_t max_side_length = furthest_dist_from_origin * 2;
--- a/src/utils/polygon.h
+++ b/src/utils/polygon.h
@@ -759,16 +759,14 @@ public:
     /*!
      * Clips input line segments by this Polygons.
      * \param other Input line segments to be cropped
-     * \return the resulting interior line segments
+     * \param segment_tree the resulting interior line segments
      */
-    ClipperLib::PolyTree lineSegmentIntersection(const Polygons& other) const
+    ClipperLib::PolyTree lineSegmentIntersection(const Polygons& other, ClipperLib::PolyTree& segment_tree) const
     {
-        ClipperLib::PolyTree ret;
         ClipperLib::Clipper clipper(clipper_init);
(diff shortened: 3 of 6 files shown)


src/infill/SubDivCube.cpp
```diff
@@ -37,7 +37,7 @@ SubDivCube::~SubDivCube()
 void SubDivCube::precomputeOctree(SliceMeshStorage& mesh)
 {
     radius_addition = mesh.settings.get<coord_t>("sub_div_rad_add");
-    AngleRadians infill_angle = M_PI / 4.0;
+    AngleRadians infill_angle = 45;
```
Ghostkeeper: This still says `AngleRadians`, so the 45 gets wrapped around 2*pi using clock arithmetic. It then ends up at 1.01, which is about 58 degrees. Not what you wanted, I think. But you're right that `PointMatrix` accepts the angle in degrees. You just need to make the type an `AngleDegrees`.

schneider007: Yes, indeed!

schneider007:
Hello,I have made a pull request right now,but it says "All checks have failed",what's the reason?

The following is the modification:

remove support nodes in holes

At present, overhang areas in holes will always generate support, but sometimes such holes may be very large, whis is a waste of material.
I have deleted such nodes in my case, but I'm not sure is the code correct.
Happy Chinese new year!


-----原始邮件-----
发件人:Ghostkeeper <notifications@github.com>
发送时间:2019-01-23 16:47:19 (星期三)
收件人: Ultimaker/CuraEngine <CuraEngine@noreply.github.com>
抄送: schneider007 <hexuejun@zju.edu.cn>, Author <author@noreply.github.com>
主题: Re: [Ultimaker/CuraEngine] fix error (#937)



@Ghostkeeper commented on this pull request.

In src/infill/SubDivCube.cpp:

> @@ -37,7 +37,7 @@ SubDivCube::~SubDivCube()
 void SubDivCube::precomputeOctree(SliceMeshStorage& mesh)
 {
     radius_addition = mesh.settings.get<coord_t>("sub_div_rad_add");
-    AngleRadians infill_angle = M_PI / 4.0;
+    AngleRadians infill_angle = 45;


This still says AngleRadians, so the 45 gets wrapped around 2*pi using clock arithmetic. It then ends up at 1.01, which is about 58 degrees. Not what you wanted, I think. But you're right that PointMatrix accepts the angle in degrees. You just need to make the type an AngleDegrees.

schneider007:
fix_error

diegopradogesto:
Hi @schneider007, we are currently reviewing your PR, but there are some conflicts that need to be resolved. Could you take a look?

Thanks!

schneider007:
Hi, I clicked the button "Resolve conflicts", but only saw a cat image in waiting state.

Ghostkeeper:
Yeah, I'm seeing the same. Github's web editor seems to be a bit buggy. It's the wrong tool for the job anyway though, since it won't try to compile and test your code before committing the merge.

src/infill.cpp
```diff
@@ -388,7 +388,8 @@ void Infill::generateCrossInfill(const SierpinskiFillProvider& cross_fill_provid
 
 void Infill::addLineSegmentsInfill(Polygons& result, Polygons& input)
 {
-    ClipperLib::PolyTree interior_segments_tree = in_outline.lineSegmentIntersection(input);
+	ClipperLib::PolyTree interior_segments_tree;
+	in_outline.lineSegmentIntersection(input, interior_segments_tree);
```
Ghostkeeper: Code style: Use spaces instead of tabs.

This is also where the merge conflict lies. You can fix the merge conflict and the tab indenting by replacing these lines with:

```
    ClipperLib::PolyTree interior_segments_tree;
    in_outline.offset(infill_overlap).lineSegmentIntersection(input, interior_segments_tree);
```

src/utils/polygon.h
```diff
@@ -759,16 +759,14 @@ class Polygons
     /*!
      * Clips input line segments by this Polygons.
      * \param other Input line segments to be cropped
-     * \return the resulting interior line segments
+     * \param segment_tree the resulting interior line segments
      */
-    ClipperLib::PolyTree lineSegmentIntersection(const Polygons& other) const
+    ClipperLib::PolyTree lineSegmentIntersection(const Polygons& other, ClipperLib::PolyTree& segment_tree) const
```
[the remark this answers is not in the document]
Ghostkeeper: Fixed here: https://github.com/Ultimaker/CuraEngine/commit/6bbd81b46ecddc675114399cdfc911978674edc4

src/infill.cpp
Ghostkeeper: Fixed in the merge commit here: https://github.com/Ultimaker/CuraEngine/commit/05e93dabce9e863b8742fd69ed87717e1594e7a9

Ghostkeeper:
This fixed the bug on the models I made to hang my cooking spices: https://github.com/Ghostkeeper/SlicerTestModels/blob/master/my/spice_bracket.scad

PR #939, opened 2018-12-28
Fix maximum resolution being too aggressive

Ghostkeeper:
The maximum resolution setting was being too aggressive in removing vertices.

I've rewritten the algorithm to a form that's somewhat simpler. Just one stage.

Fixes CURA-5924.

### DIFF
--- a/src/utils/polygon.cpp
+++ b/src/utils/polygon.cpp
@@ -340,7 +340,7 @@ void PolygonRef::simplify(int smallest_line_segment_squared, int allowed_error_d
     }
 
     //For the last/first vertex, we didn't check the connection that closes the polygon yet. Add the first vertex back if this connection is too long.
-    if(vSize2(new_path.back() - new_path.at(0)) > smallest_line_segment_squared)
+    if(!new_path.empty() && vSize2(new_path.back() - new_path[0]) > smallest_line_segment_squared)
     {
         new_path.push_back(path->at(0));
     }
--- a/src/utils/polygon.cpp
+++ b/src/utils/polygon.cpp
@@ -337,12 +337,14 @@ void PolygonRef::simplify(int smallest_line_segment_squared, int allowed_error_d
             break; //New polygon also doesn't have any vertices yet, meaning we've completed the loop without adding any vertices. The entire polygon is too small to be significant.
         }
         accumulated_area_removed += current.X * next.Y - current.Y * next.X; //Shoelace formula for area of polygon per line segment.
-        const coord_t area_removed_so_far = std::abs(accumulated_area_removed + next.X * previous.Y - next.Y * previous.X); //Close the polygon.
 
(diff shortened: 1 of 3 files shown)


src/utils/polygon.cpp
```diff
@@ -304,158 +304,72 @@ void PolygonRef::simplify(int smallest_line_segment_squared, int allowed_error_d
...
-            std::vector<ListPolyIt> new_skipped_verts;
+        }
+        //Don't remove the vertex.
 
-            for (unsigned int skipped_vert_idx = 0; skipped_vert_idx < skipped_verts.size(); skipped_vert_idx++)
+        //But we may need to remove the previous vertex if it happens to be almost exactly in line with the new vertex.
+        if (new_path.size() >= 2)
+        {
+            const coord_t distance = LinearAlg2D::getDist2FromLineSegment(new_path[new_path.size() - 2], new_path[new_path.size() - 1], current);
+            if (distance < 25) //Less than 5 microns off, so this falls within our rounding error range.
```
rburema: While I don't think it's serious enough to send it back for, I don't really like the hardcoded value here. Maybe it could be a constexpr?

tests/utils/PolygonTest.cpp
```diff
@@ -183,5 +183,57 @@ void PolygonTest::getEmptyHolesTest()
...
+        std::stringstream ss_short;
+        ss_short << "Segment " << (point_index - 1) << " - " << point_index << " is too short! " << segment_length << " < " << minimum_segment_length;
+        CPPUNIT_ASSERT_MESSAGE(ss_short.str(), segment_length >= minimum_segment_length);
+        std::stringstream ss_long;
+        ss_long << "Segment " << (point_index - 1) << " - " << point_index << " is too long! " << segment_length << " > " << maximum_segment_length;
+        CPPUNIT_ASSERT_MESSAGE(ss_long.str(), segment_length <= maximum_segment_length);
+    }
+}
+
+void PolygonTest::simplifyZigzag()
```
rburema: Nice! More UnitTests! :-) Maybe also add a test where it should _not_ simplify something further?

src/utils/polygon.cpp
Ghostkeeper: It turned out that this was dysfunctional when dealing with polygons that started off with 5 micron parts. I've reworked this part.

tests/utils/PolygonTest.cpp
Ghostkeeper: Yes, added two of those! One for each limiting parameter.

Timocop:
This seems to cause problems. Its way too loose now. My printer cant print walls more than 15mm/s without stuttering. Using @smartavionics version of Cura he already merged this and released it. @smartavionics stated he didnt change any behaviour regrading to this issue.
More here:
https://github.com/smartavionics/CuraEngine/issues/5

rburema:
@Timocop We've noticed some issues like I think you may have as well, and have started a slight re-work on this too: See #1037

PR #972, opened 2019-01-08
Bug Fix - Only add S1 param to G10 when generating repetier flavour gcode.

smartavionics:
The other firmwares can interpret the S1 differently so don't let them see it.


diegopradogesto:
Devs. CURA-6111

PR #973, opened 2019-01-09
Bug Fix: reworked line polygon crossings so that combing copes with complex boundaries

smartavionics:
The original line polygon crossings implementation did not cope with complex combing regions well because it assumed that each polygon in the combing boundary could only have a single crossing pair. This is not true for complex combing regions that have multiple holes.

Before this PR...

screenshot_2019-01-08_08-50-18

With this PR...

screenshot_2019-01-08_09-07-35

This PR fixes https://github.com/Ultimaker/Cura/issues/5116.

Note that the problem is not only with no-skin combing, it can occur with any combing where the combing region is sufficiently complex.


diegopradogesto:
Devs. CURA-6112

theWaldschrat:
I'll like to second that PR, because I ran into the same issue on a 3Dbenchy (!?!).

rburema:
> Looks like a good approach to the problem. Reduces the administration with that minimum and maximum, too. I'll give it a good test...

Good that I saw you post this, I had this one pegged as the next one to-do, but couldn't get to it yesterday. Almost also started a review :-)

Ghostkeeper:
Well it's got a Jira ticket so Jira would've resolved the conflict anyway :)

Ghostkeeper:
Took a while, but indeed I found some cases where it improved combing. It doesn't seem to have any noticeable performance impact.

PR #974, opened 2019-01-09
CURA-6093 unneeded retract with 2nd extruder

rburema:
Fixes (frontend!) github-issue #5103


src/FffGcodeWriter.cpp
```diff
@@ -33,13 +33,14 @@ FffGcodeWriter::FffGcodeWriter()
 
 void FffGcodeWriter::writeGCode(SliceDataStorage& storage, TimeKeeper& time_keeper)
 {
-    gcode.preSetup(getStartExtruder(storage));
+    size_t start_extruder_nr = getStartExtruder(storage);
```
smartavionics: Make this const to avoid it getting changed later by mistake?

rburema: Good point.

PR #982, opened 2019-01-27
CMake: Depend on Protobuf

thopiekar:
When building the engine with Arcus, we are in the following lines generating C code using protoc.
However, if your installation is not complete and you've got all libs and headers installed, but no protoc installed, then the build fails.
At least that is the case on Windows, when protoc can't be found in %PATH%.

find_package should ensure here, that all libs, headers and command line tools are installed and found. If not this should throw an error and prevent the situation I ran into.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,7 +11,11 @@ option (ENABLE_ARCUS
 
 if (ENABLE_ARCUS)
     message(STATUS "Building with Arcus")
-    find_package(Protobuf REQUIRED)
+    # We want to have access to protobuf_generate_cpp and other FindProtobuf features.
+    # However, if ProtobufConfig is used instead, there is a CMake option that controls
+    # this, which defaults to OFF. We need to force this option to ON instead.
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
It's true that we depend on Protobuf directly as well. The `protobuf_generate_cpp` function is used right in the same CMake file. Good find!

In libArcus though we require Protobuf in a slightly more extensive way that I think is better:

```
# We want to have access to protobuf_generate_cpp and other FindProtobuf features.
# However, if ProtobufConfig is used instead, there is a CMake option that controls
# this, which defaults to OFF. We need to force this option to ON instead.
set(protobuf_MODULE_COMPATIBLE ON CACHE "" INTERNAL FORCE)
find_package(Protobuf 3.0.0 REQUIRED)
```

Could you change it to require 3.0.0 and set the `protobuf_MODULE_COMPATIBLE` variable?

thopiekar:
I'm pretty sure I made another change to this, which should do kind of the same as you suggest.
Give me some time..

Ghostkeeper:
Works! Let's hope that this helps some folks to find build errors earlier.

PR #986, opened 2019-01-30
Bug fix: round z to nearest micron when spiralizing

smartavionics:
As discussed in https://github.com/Ultimaker/Cura/issues/5243.


smartavionics:
Ooops, this looks wrong, try again!

PR #987, opened 2019-01-30
When spiralizing, round the interpolated z value to nearest micron.

smartavionics:
As discussed in https://github.com/Ultimaker/Cura/issues/5243.


src/LayerPlan.cpp
```diff
@@ -1533,7 +1533,7 @@ void LayerPlan::writeGCode(GCodeExport& gcode)
                         Point p1 = path.points[point_idx];
                         length += vSizeMM(p0 - p1);
                         p0 = p1;
-                        gcode.setZ(z + layer_thickness * length / totalLength);
+                        gcode.setZ(std::round(z + layer_thickness * length / totalLength));
```
Ghostkeeper: I don't see how this would help, because `z` and `layer_thickness` are already integers, the result should be an integer as well, right? It'll still round it down.

smartavionics: But length and totalLength are doubles so layer_thickness * length / totalLength doesn't necessarily equal layer_thickness  when length == totalLength.

PR #1000, opened 2019-02-25
Feature: "Wipe Script"

vgribinchuk:
Feature: "Wipe Script"
Backend implementation of the feature: https://github.com/Ultimaker/Cura/pull/5356


Ghostkeeper:
Developers, see issue CURA-6249.

src/LayerPlan.cpp
```diff
@@ -1387,12 +1391,21 @@ void LayerPlan::writeGCode(GCodeExport& gcode)
...
-            // only do the retract if the paths are not spiralized
-            if (!mesh_group_settings.get<bool>("magic_spiralize"))
+            const WipeScriptConfig& wipe_config = storage.wipe_config_per_extruder[extruder_plan.extruder_nr];
+            if (wipe_config.clean_between_layers && gcode.getExtrudedVolumeAfterLastWipe(extruder_nr) > wipe_config.max_extrusion_mm3)
             {
-                gcode.writeRetraction(retraction_config);
+                gcode.insertWipeScript(wipe_config);
+                gcode.ResetLastEValueAfterWipe(extruder_nr);
+            }
+            else if (layer_nr != 0 && Application::getInstance().current_slice->scene.extruders[extruder_nr].settings.get<bool>("retract_at_layer_change"))
```
rburema: This of course makes `retract_at_layer_change` incompatible with `clean_between_layers`, which is fine, but I don't see any handling of this in the frontend (see my comment there).

src/FffGcodeWriter.cpp
```diff
@@ -309,6 +311,38 @@ void FffGcodeWriter::setConfigRetraction(SliceDataStorage& storage)
...
+        WipeScriptConfig& wipe_config = storage.wipe_config_per_extruder[extruder_index];
+
+        wipe_config.retraction_enable = train.settings.get<bool>("wipe_retraction_enable");
+        wipe_config.retraction_config.distance = train.settings.get<double>("wipe_retraction_amount");
+        wipe_config.retraction_config.speed = train.settings.get<Velocity>("wipe_retraction_retract_speed");
+        wipe_config.retraction_config.primeSpeed = train.settings.get<Velocity>("wipe_retraction_prime_speed");
+        wipe_config.retraction_config.prime_volume = train.settings.get<double>("wipe_retraction_extra_prime_amount");
+        wipe_config.retraction_config.retraction_min_travel_distance = 0;
+        wipe_config.retraction_config.retraction_extrusion_window = 99999.9;
+        wipe_config.retraction_config.retraction_count_max = 9999999;
```
rburema: You can use `<numeric_limits>` for these kind of 'largest value possible' numbers.

src/gcodeExport.cpp
```diff
@@ -404,6 +405,17 @@ double GCodeExport::mmToE(double mm)
     }
 }
 
+double GCodeExport::eToMm3(double e, size_t extruder)
+{
+    if (is_volumatric)
```
rburema: (Spelling: volumetric.)

rburema:
Hi @vgribinchuk 
Sorry it took a while, despite the priority we gave this ticket above your other PR's it still took a while to get here. 

My comments are small enough that I ordinarily would fix them myself, but our QA people have requested that you write some unit-tests as well. (This is just a friendly request, not a requirement for merge.) Do you have some time to do this, or should I fix the remaining comments & go ahead with the component-test (which could deliver additional comments before the merge)?
cheers, Remco.

vgribinchuk:
> Hi @vgribinchuk
> Sorry it took a while, despite the priority we gave this ticket above your other PR's it still took a while to get here.
> 
> My comments are small enough that I ordinarily would fix them myself, but our QA people have requested that you write some unit-tests as well. (This is just a friendly request, not a requirement for merge.) Do you have some time to do this, or should I fix the remaining comments & go ahead with the component-test (which could deliver additional comments before the merge)?
> cheers, Remco.

Hi, @rburema !
Surely, I will fix points mentioned in comments. But for implementing some unit-tests I need more time to first understand current structure/architecture of already implemented unit-tests and provide additional unit tests for wipe_script. 
Regards, Vlad

rburema:
@vgribinchuk  Thanks for your quick reply! It's great that/if you want to make some more time for those tests. For C++ (I wouldn't really bother with tests for the frontend, since you only had to change the config-files there) we've recently switched to the GoogleTest framework. I've recently added some very simple tests in one commit (https://github.com/Ultimaker/CuraEngine/commit/e456ce2241316e1bac9b00e71198a07cf3687211) that you can use as inspiration.

src/FffGcodeWriter.cpp
vgribinchuk: Good point, thanks! This was fixed.

src/gcodeExport.cpp
vgribinchuk: Fixed

src/LayerPlan.cpp
vgribinchuk: The following condition will not always true even if `clean_between_layers` was turned on:
`if (wipe_config.clean_between_layers && gcode.getExtrudedVolumeAfterLastWipe(extruder_nr) > wipe_config.max_extrusion_mm3)`

I mean, if clean_between_layers is ON, it will not necessarily be working at every layer changing. For those layers when clean between layers was not activated, because last extruded volume value has not exceeded the max_extrusion_mm3 yet, `retract_at_layer_change` may work (it it was also enabled).

Does this logic make sense to you?

vgribinchuk:
Thank you! I will take a look.

Vandrasc:
Hi @vgribinchuk , could you please create a separate PR for those tests (when ready)? It will help us more in our administration, thank you!

rburema:
... it'll also help to get this PR merged quicker if we devide it :-)

src/LayerPlan.cpp
LipuFei: @vgribinchuk I think what @rburema means is the `else if` part:
```
if (clean_between_layers and volume > max_mm3)
{
    ...
}
<<else>> if (retract_at_layer_change)
{
    ...
}
```
The `else if` here makes `clean_between_layers` and `retract_at_layer_change` exclusive to each other. If this is not intentional, we should remove the `else`.

rburema: Hi @Lipu,
Yes that is what I meant, but I also understand what @vgribinchuk is saying here:
Namely that it is mut.exclusive _only_ per layer, and that a model _as a whole_ still may be effected by both settings, which means disabling one if the other is on in the front-end will limit possibilities.
I can see that point, but think it should be documented somewhere (something like 'this will influence retract at layer change') where the user can see it.
cheers, Remco.

vgribinchuk: Hi @rburema, @LipuFei ! 
Thank you, gentlemen, for clarifications!
So, I see two options here I can implement.
1. remove `<<else>>` making `clean_between_layers` and `retract_at_layer_change` not mutually exclusive

But this might be a bit confusing, at least from my point of view. Because wipe script provides its own settings for retraction, which could lead to two consecutive retractions at layer change.

So, since wipe script has own settings for retraction, those settings could be applied for retraction at layers where script would be working. On other layers, where script would not be working, `retract_at_layer_change` could be applied. This is the the second option:
2. leave logic as it is, but change description for `clean_between_layers` in front end, like this:
```
"clean_between_layers": {
"description": "Whether to include nozzle wipe G-Code between layers. Enabling this setting could influence behavior of retract at layer change. Please use Wipe Retraction settings to control retraction at layers where the wipe script will be working."
...
```

Let me please know which option works best for you. Or if you see other more appropriate solution.

rburema: Hi @vgribinchuk,
I think we should leave the logic as it is and change the description, which seems like the option least likely to casue a mess later on.
cheers, Remco.

vgribinchuk: OK, got it. Thank you, @rburema !

vgribinchuk: Description was changed. Please see commit:
https://github.com/Ultimaker/Cura/pull/5356/commits/6e955f780b115a86197b1c8b2e2d892efbfe8cf6

rburema:
Now Q.A. just needs to do their pass, then it's in to stay :-)

vgribinchuk:
Hi, @rburema, @Vandrasc  !
Separate PR for unit tests has been opened. Please see https://github.com/Ultimaker/CuraEngine/pull/1026.

PR #1003, opened 2019-02-27
When infill is not printed every layer, reduce the amount the outline is shrunk.

smartavionics:
Normally, the outline is shrunk by 1/2 the infill line width but when infill isn't printed
every layer, the line width is multiplied accordingly and so the outline would be shrunk
too much.

Fixes https://github.com/Ultimaker/Cura/issues/5345.


Ghostkeeper:
Developers, see issue CURA-6258.

radekkpl:
Can I ask, when this will be implemented? I am printing a lot with 0,05mm height and saving time with printing infill only 1/3 layer is enormous. I know that in preview window it will look ugly (infill will be visible outside sometimes due to increased visual width, but i can live with it :) )

Ghostkeeper:
We don't give time estimates on this sort of thing. I'm sorry, but the Cura developers are torn between a lot of stakeholders. We keep these things alive for a long time these days. It's still on our backlog and I'd like to work through the backlog of pull requests soon. We need to give more priority on that to prevent code rot.

rburema:
Again, sorry for the delay!

PR #1005, opened 2019-03-04
Feature slice data import export

vgribinchuk:
Feature: Slice Data Import/Export
This feature makes CuraEngine running in a mode when it will perform only slicing and generating supports for the input part. As a result, CuraEngine will output slice data which can be imported into it later (avoiding slicing and generating supports steps since slices will be already pre-computed and stored in the input slice data). With this feature CuraEngine can output slices of the part and its supports. This is helpful for using slice data in some 3rd party applications or for performing some processing on slice data before loading it back to CuraEngine for generating gcode.

This feature introduces new command line parameters used for exporting/importing slice data. There is ability to import supports and parts slices separately. Following command line parameters are used for exporting slice data:
- ‘--export’ – key for enabling exporting slice data mode;
- ‘-o <output_supports_slices_file>’ – path to the file for exporting slices of the supports;
- ‘-o1 <output_part_slices_file>’ – path to the file for exporting slices of the part.

Following command line parameters are used for importing slice data:
- ‘-x <input_part_slices_file>’;
- ‘-u <input_supports_slices_file>’.

Slice data is serialized in proto format described in the SliceData.proto file. Paths of the part itself have type=PART. Paths in supports slices will have other types: SUPPORT, SUPPORT_ROOF, SUPPORT_BOTTOM. If raft is turned on, it will also be outputted to the supports slices. No matter how many layers of the raft are defined by settings, there will be only one layer with polygons with type = RAFT. Z coordinate of this raft layer will be equal to the total height of the raft.


rburema:
We left this too long, sorry.

Unfortunately, as the situation stands now, we expect significant changes to CuraEngine architecture to occur overwriting (and eventually, hopefully even superseding) these changes. Therefore, at the moment, we don't see the use of pulling it in (any more).

PR #1015, opened 2019-03-18
Bug fix - fetch retract_at_layer_change value from extruder in addConnectingTravelMove().

smartavionics:
It was incorrectly fetching a value from the mesh which always true.

Fixes https://github.com/Ultimaker/Cura/issues/5444


Ghostkeeper:
Yep, that seems to fix it.

PR #1016, opened 2019-03-19
Take current position into account when planning infill mesh part print order.

smartavionics:
Before, the ordering was anchored at layer start x/y rather than the current position and
didn't take into account the end location of the previous part printed.

This was particularly noticeable when printing infill mesh after the main mesh had been printed.

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -1177,17 +1177,23 @@ void FffGcodeWriter::addMeshLayerToGCode(const SliceDataStorage& storage, const 
 
     gcode_layer.setMesh(mesh.mesh_name);
 
-    std::vector<SliceLayerPart> parts(layer.parts);
+    std::vector<const SliceLayerPart*> parts; // use pointers to avoid recreating the SliceLayerPart objects
+
+    for(const SliceLayerPart& part_ref : layer.parts)
+    {
+        parts.emplace_back(&part_ref);
+    }
+
     while (parts.size())
     {
         PathOrderOptimizer part_order_optimizer(gcode_layer.getLastPlannedPositionOrStartingPosition());
-        for (const SliceLayerPart& part : parts)
+        for (auto part : parts)
         {
-            part_order_optimizer.addPolygon((part.insets.size() > 0) ? part.insets[0][0] : part.outline[0]);
+            part_order_optimizer.addPolygon((part->insets.size() > 0) ? part->insets[0][0] : part->outline[0]);
         }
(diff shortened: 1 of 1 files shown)


smartavionics:
I realised that for regular meshes with parts that have walls, the new scheme could go wrong so I have reinstated the original code for "printable" meshes and only use the new code for the other mesh types, i.e. infill meshes.

smartavionics:
> So (without looking at the code) how does this determine the first part of an extruder plan?

This change doesn't alter the order the meshes are printed, it can change the order the meshes' parts are printed when the mesh is "non-printable", i.e an infill mesh.

Ghostkeeper:
I'm wondering why the scheme didn't work when you have walls. :confused:

smartavionics:
When you have walls and the z-seam config forces the z-seam to a particular location it will do the ordering based on the positions of the z-seams of each of the parts. When you have an infill mesh you really just want to order based on the distance from the current location to the nearest vertex in the infill outlines, i.e. go to the next lump of infill whose boundary is closest to the current location.

BTW i believe there is still a bug in pathOrderOptimizer.cpp. The initialization of prev_point (lines 46-57) is wrong. It should set it to startPoint for all values of config.type. That would make the first polygon chosen be the one whose z-seam vertex is closest to the current location rather than the polygon whose z-seam vertex is closest to the z-seam hint location.

Ghostkeeper:
This one slipped through the cracks, but it was found again. Developers, see CURA-6676.

ninovanhooff:
We finally moved this into the sprint, sorry for the delay. 
Just looking at the code, I'd say it could be beneficial.
I was able to create a scenario were a single travel move was saved, which doesn't seem valuable enough for the (granted, minor) increase in code complexity.

If you could provide us with an example project we'd welcome this contribution. 

print_order_after
print_order_before

[UM3_tables_print_order_test.zip](https://github.com/Ultimaker/CuraEngine/files/3818872/UM3_tables_print_order_test.zip)

smartavionics:
It's true that this PR doesn't make that much difference to the overall print time but it does help make Cura look a little more intelligent in its travel planning. It's easy to think that small gains are not worthwhile but people do notice when the gcode is dumb. Anyway, the project that this PR was designed to help is attached.

smartavionics:
Oops, here's that file...

[KXL_combo-mount-ela.zip](https://github.com/Ultimaker/CuraEngine/files/3818971/KXL_combo-mount-ela.zip)

rburema:
Thanks! It's clear what's happening now.

PR #1017, opened 2019-03-21
Account for the fact that narrow skin regions aren't printed when calculating skin gaps.

smartavionics:
The gap region is calculated by subtracting the skin/infill areas from the inside
edge of the inner wall but if the skin regions are too narrow to be printed (less than
a skin line width wide) then we want to include the narrow region in the gap.

Look at the front window ledge on this benchy, first without this PR and then with the PR...

Screenshot_2019-03-21_15-09-47

Screenshot_2019-03-21_15-10-36


Ghostkeeper:
Looks useful.
Developers, see issue CURA-6409.

Timocop:
Doesnt seem there is a max. limit of how much much gap filling it does. Its litterly over extruding with this fix. Over 5x bigger than Wall thickness. And if the hole is too small Cura doesnt even bother filling it (Cura without this fix filled those holes).

smartavionics:
Hello @Timocop , if you are commenting on the behaviour of my master branch build please post the issue on my github repo and not Ultimaker's, it's not their problem to fix.

But I don't think what you are describing is a real problem at all. The reason that small hole isn't filled is because my release contains settings Minimum Gap Area and Inner Layer Minimum Gap Area which set a lower limit on the areas of the filled gaps.

As for the overextrusion, so what?

smartavionics:
Actually, thinking about it. Those are not skin gaps anyway in your images above. They are inter-wall gap fill so this PR that modifies the behaviour of skin gaps doesn't make any difference to the behaviour you are seeing. It's the fact that you are using my Cura which has a different implementation of the wall gap filling that is the difference.

Timocop:
Oohh, sorry. I thought this was related somehow. Because when i put 'Fill gaps between walls' to 'Nowhere' the blobs dissapear. As for overextrusion, i wasnt sure at which speed its printing those giant blobs. I dont want the same behavior as spagetti infill ;) So i guess its intended? Otherwise ill create an issue on your fork

smartavionics:
No problem. The "giant blobs" are sized based on the area to be covered so the're not really overextrusions. Still, if you print that part and there's a problem with those areas, please open an issue on my Cura repo. Thanks.

smartavionics:
I'm closing this because I now think the real problem is in the way the skin regions are computed so I will look into that and maybe submit another PR some day if I come up with a fix.

smartavionics:
I'm reopening this in a slightly modified form.

rburema:
Hi @smartavionics,
I can't seem to reproduce the original problem in current master (without the fix). I think something else was changed/fixed. Would you mind attaching a 3mf?
cheers, Remco.

smartavionics:
I'll try and find an example that this PR fixes.

smartavionics:
OK, here you go. It's a benchy, I don't know if it is the same file as I used for the image at the top but anyway. Without this PR, this is what you get when sliced with skin pattern lines...

Screenshot_2019-11-12_17-21-19

Note how the window sill in the cabin has skin between the walls.

If you then slice with skin pattern zig-zag, that skin gets omitted (presumably this is because the zig-zags connectors reduce the skin area so much that it goes away completely).

Screenshot_2019-11-12_17-20-56

With this PR, it will fill that gap (as long as fill gaps is enabled)...

Screenshot_2019-11-12_17-22-13

Here's the project file zipped...

[KMMB_3DBenchy.zip](https://github.com/Ultimaker/CuraEngine/files/3837403/KMMB_3DBenchy.zip)

rburema:
Thanks. I did try with the Benchy (before as well), but (still) no luck. It seems I'll need the profile as well in order to properly load the file. (I lost the old one you send me when my last laptop crashed.)

smartavionics:
Don't you get the profile in the .3mf file?

rburema:
(Some delay: It seems the hectic days around the release are done for now.)

I remembered you send me a mail last year, so I could dig up the definitions and so open the project. I can now reproduce the problem you saw.

On further testing, unfortunately, I found a problem: The print time can increase around 2 percent max. In and of itself this might not have been a problem, but then I found out what it was spending time on, and I think you might want to see this:

On master (gif):
normal_skin_on_master

With the branch (gif):
extra_skin_on_branch

Project file is attached (zipped, not renamed):
[UM3_boardgame_test1.zip](https://github.com/Ultimaker/CuraEngine/files/3880328/UM3_boardgame_test1.zip)

smartavionics:
Oops! Good catch.
I think it's fixed now, please test again. Thanks.

rburema:
That did it. Thanks once again for your contributions!

PR #1022, opened 2019-04-01
When coasting was enabled, the path.speed_factor was being ignored.

smartavionics:
This PR makes the path.speed_factor effective when the path is being coasted.

### DIFF
--- a/src/LayerPlan.cpp
+++ b/src/LayerPlan.cpp
@@ -1647,7 +1647,7 @@ bool LayerPlan::writePathWithCoasting(GCodeExport& gcode, const size_t extruder_
     coord_t coasting_min_dist_considered = 100; // hardcoded setting for when to not perform coasting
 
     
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
We may want to remove this thing entirely. As far as I could find, the only place where this `speed_factor` was set was here:

https://github.com/Ultimaker/CuraEngine/blob/c1c8bcb550805aae8fa091a2581ddef1e55e1244/src/LayerPlan.cpp#L785

When I change that value to be `constexpr`, it compiles fine...

smartavionics:
Ah, I actually use it in my code that reduces the wall print speed when the wall length is short (e.g. small circles). I would be grateful if you don't remove it. And, who knows, maybe one day when PRs are being used again, that code (or something based on that) could be in Ultimaker Cura.

smartavionics:
It also gets used in the PRs that are still waiting for your attention that improve the spiralization by tapering the top (and bottom) layer heights (and fixes other problems too).

PR #1024, opened 2019-04-03
CURA-6002 fix support brim

rburema:
Print support-brim later than other skirt/brim, so if skirt/brim is also used to prime, then this option works without impacting support.

### DIFF
--- a/src/FffGcodeWriter.h
+++ b/src/FffGcodeWriter.h
@@ -270,7 +270,7 @@ private:
      * \param extruder_nr The extruder train for which to process the skirt or
      * brim.
      */
-    void processSkirtBrim(const bool for_support, const Polygons& skirt_brim, LayerPlan& gcodeLayer, unsigned int extruder_nr) const;
+    void processSkirtBrim(const Polygons& skirt_brim, LayerPlan& gcodeLayer, unsigned int extruder_nr, bool for_support) const;
 
     /*!
      * Adds the ooze shield to the layer plan \p gcodeLayer.
--- a/src/FffGcodeWriter.h
+++ b/src/FffGcodeWriter.h
@@ -264,12 +264,13 @@ private:
      * This function should be called for only one layer;
      * calling it for multiple layers results in the skirt/brim being printed on multiple layers.
      * 
-     * \param storage where the slice data is stored.
+     * \param for_support Whether or not this is brim specially made underneath support-infill.
+     * \param skirt_brim The outline of the polygons that'll be skirt/brim.
      * \param gcodeLayer The initial planning of the g-code of the layer.
      * \param extruder_nr The extruder train for which to process the skirt or
      * brim.
      */
-    void processSkirtBrim(const SliceDataStorage& storage, LayerPlan& gcodeLayer, unsigned int extruder_nr) const;
+    void processSkirtBrim(const bool for_support, const Polygons& skirt_brim, LayerPlan& gcodeLayer, unsigned int extruder_nr) const;
 
     /*!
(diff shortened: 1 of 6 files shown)


src/LayerPlan.h
```diff
@@ -380,14 +381,14 @@ class LayerPlan : public NoCopy
      */
     void setPrimeTowerIsPlanned(unsigned int extruder_nr);
 
-    bool getSkirtBrimIsPlanned(unsigned int extruder_nr) const
+    bool getSkirtBrimIsPlanned(const bool for_support, const unsigned int extruder_nr) const
```
LipuFei: I would give `for_support` a default value `false`. Another thing is adding that extra `const` doesn't do much because those 2 variables are local copies anyway. So, I would suggest a function signature as below:
```
bool getSkirtBrimIsPlanned(unsigned int extruder_nr, bool for_support = false) const
```

src/LayerPlan.h
```diff
@@ -380,14 +381,14 @@ class LayerPlan : public NoCopy
...
     {
-        return skirt_brim_is_processed[extruder_nr];
+        return (for_support ? support_brim_is_processed : skirt_brim_is_processed)[extruder_nr];
     }
 
-    void setSkirtBrimIsPlanned(unsigned int extruder_nr)
+    void setSkirtBrimIsPlanned(const bool for_support, const unsigned int extruder_nr)
```
LipuFei: Same as above, I'd use
```
bool setSkirtBrimIsPlanned(unsigned int extruder_nr, bool for_support = false)
```

src/FffGcodeWriter.h
```diff
@@ -264,12 +264,13 @@ class FffGcodeWriter : public NoCopy
...
      * 
-     * \param storage where the slice data is stored.
+     * \param for_support Whether or not this is brim specially made underneath support-infill.
+     * \param skirt_brim The outline of the polygons that'll be skirt/brim.
      * \param gcodeLayer The initial planning of the g-code of the layer.
      * \param extruder_nr The extruder train for which to process the skirt or
      * brim.
      */
-    void processSkirtBrim(const SliceDataStorage& storage, LayerPlan& gcodeLayer, unsigned int extruder_nr) const;
+    void processSkirtBrim(const bool for_support, const Polygons& skirt_brim, LayerPlan& gcodeLayer, unsigned int extruder_nr) const;
```
LipuFei: Same as below, I'd make `bool for_support = false` and put it at the end of the parameter list.

rburema:
I don't have a strong preference for the order of the variables (or the default to false), so I'll fix that. I guess I can loose the const modifier as well in this case (though it's not just about efficiency and also about readability).

I'll try to fix it in the next 5 ~10 minutes, then I'll have to leave :-)

src/FffGcodeWriter.h
rburema: Done in https://github.com/Ultimaker/CuraEngine/pull/1024/commits/5748ac125f8e025874de29e41c4fab96a18e89ff

src/LayerPlan.h
rburema: Done in https://github.com/Ultimaker/CuraEngine/pull/1024/commits/5748ac125f8e025874de29e41c4fab96a18e89ff

src/LayerPlan.h
rburema: Done in https://github.com/Ultimaker/CuraEngine/pull/1024/commits/5748ac125f8e025874de29e41c4fab96a18e89ff

rburema:
Closed because #1029 fixes this in a more elegant way.

PR #1026, opened 2019-04-05
Feature Wipe Script Unit Tests

vgribinchuk:
Related pull request: https://github.com/Ultimaker/CuraEngine/pull/1000
Added several test cases for the submitted functionality related to Wipe Script.


rburema:
Hi @vgribinchuk, bit of a late reply, as I was at the Hannover Messe. I've seen this and will ask if this can be merged without a ticket (so we don't get into the situation that this keeps lagging while the other one is already merged). cheers, Remco.

PR #1029, opened 2019-04-07
CURA-6002 Don't place support brim polygons on the back of skirt_brim as that upsets the ordering.

smartavionics:
When the skirt/brim lines are printed, it is assumed that the polygon at the back of
skirt_brim is furthest away from the outline of the model. By adding support brim lines
to the back of skirt_brim, if any of those lines were inside of the model (i.e. support
inside a hole in the model) it would print those inner support brim lines first and then
go on to print the brim around the outside of the model from inner to outer (wrong!)

This change ensures that the support brim lines are before the original lines in skirt_brim
thus ensuring that the back polygon doesn't change.

Fixes https://github.com/Ultimaker/Cura/issues/5503

### DIFF
--- a/src/SkirtBrim.cpp
+++ b/src/SkirtBrim.cpp
@@ -247,6 +247,8 @@ void SkirtBrim::generateSupportBrim(SliceDataStorage& storage)
     const Polygons brim_area = support_outline.difference(support_outline.offset(-brim_width));
     support_layer.excludeAreasFromSupportInfillAreas(brim_area, AABB(brim_area));
 
+    Polygons support_brim;
+
     coord_t offset_distance = brim_line_width / 2;
     for (size_t skirt_brim_number = 0; skirt_brim_number < line_count; skirt_brim_number++)
     {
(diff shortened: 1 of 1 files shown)


smartavionics:
Just noticed that you already have a PR for fixing this problem (https://github.com/Ultimaker/CuraEngine/pull/1024) - looks rather complicated compared to mine. Anyway, would you like me to close this PR?

rburema:
Simpeler (if it works) is better, so please don't close this yet. I'll have a look if this works for the issue(s) we're trying to solve with the other PR. If so, we'll merge this one and close the other one instead.

smartavionics:
OK, I'll leave it to you to decide.

As far as I can see, there's two issues that need to be solved and this PR does both.

1 - outer brim needs to be printed before support brim so that support brim doesn't miss any lines

2 - outer brim always needs to be printed outer line first

rburema:
Devs, please review/test/merge this branch instead of #1024 for CURA-6002.

nallath:
As your solution is far more elegant (eg; simple), we're using that one.

PR #1035, opened 2019-04-12
RepRap firmware ignores jerk during travel moves so there's no benefit in changing it.

smartavionics:
This PR omits jerk changes for travel moves when generating RepRap flavour gcode as that firmware
ignores jerk for non-printing moves. This can save outputting many M566 lines when printing skin and infill.

### DIFF
--- a/src/LayerPlan.cpp
+++ b/src/LayerPlan.cpp
@@ -1455,7 +1455,11 @@ void LayerPlan::writeGCode(GCodeExport& gcode)
             }
             if (jerk_enabled)
             {
-                gcode.writeJerk(path.config->getJerk());
+                // RepRap firmware ignores jerk for travel moves so don't bother changing it when travelling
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Maybe those printer definitions need to set the travel jerk to be equal to the printing jerk. RepRap is a rather large family of printers to conclude that they don't support jerk during travels.

smartavionics:
I discussed this with the RepRap firmware developer and he says that jerk is ignored during travel moves.

smartavionics:
People use different values of jerk for printing walls, skin, infill, etc. and so it's not always possible to set the travel jerk equal to the printing jerk.

Ghostkeeper:
We merged it but also made sure to hide the settings from the front-end in this case, because they had no effect any more then: https://github.com/Ultimaker/Cura/commit/d5fa0b395e5185968c70147cf24cb421369fa2f0

smartavionics:
Understood, thanks for doing that.

Issue #1032, opened 2019-04-11
[Cura 4.0] Some Infill Pattern ignore 'Infill Overlap Percentage'


Timocop:
Infill Pattern that ignore 'Infill Overlap Percentage':
 - Cubic Subdivision
 - Cross
 - Cross 3D
 - Gyroid (Fix already merged #899)

If this is intended behavior, why not hide 'Infill Overlap Percentage' like when 'Concentric' infill pattern is selected?

EDIT: I just noticed that all of those infill pattern listed above dont have any overlap at all. You can see it clearly when having 'Connect Infill Lines' enabled. Seems like its always 0%.

**Application Version**
Cura 4.0

**Platform**
Windows 10 x64

**Steps to Reproduce**
Select one of aboves infill. Set it more than 0% infill overlap. Notin diff. Wat givs?

Cubic works as intended:
![](https://i.imgur.com/gcQs7qM.png)
Gyroid doesnt:
![](https://i.imgur.com/pTEgsHE.png)

Gyroid (with 'Connect infill lines') has no overlap at all:
![](https://i.imgur.com/5DdKcUC.png)
Lines (with 'Connect infill lines') has overlapping:
![](https://i.imgur.com/bbltrpJ.png)

Timocop:
Seems like @smartavionics Cura version fixed this issue. I guess its time to switch.
His comment and release link can be found here: https://github.com/Ultimaker/CuraEngine/issues/1031#issuecomment-481013028

smartavionics:
The fix for gyroid should be in the next Ultimaker release.

rburema:
... and another two lines changed for the other 3 cases. Made a PR, #1036 my colleages may (or may not) pick it up on monday :-)

PR #1036, opened 2019-04-12
Fix 'infill overlap percentage' was ignored for some infills.

rburema:
Should fix the other 3 cases of issue #1032


LipuFei:
CURA-6459

PR #1038, opened 2019-04-17
Update extruder_prime_layer_nr for all extruders used in a layer, not just the last used.

smartavionics:
As discussed in https://github.com/Ultimaker/Cura/issues/5545.


Ghostkeeper:
I still don't know what problem this issue is fixing. The `extruder_order_per_layer` doesn't contain the starting extruder so that one would still be missing. And as far as I know this `extruder_prime_layer_nr` can only contain 0, the start of the air gap or the start of the raft. On layers below 0 it will only have one extruder (either raft or support) so it wouldn't make any difference.

In Ultimaker/Cura#5545 the issue that was being fixed was an erroneous travel move, but that should be fixed by https://github.com/Ultimaker/CuraEngine/pull/1039.

smartavionics:
I'm simply making the contents of extruder_prime_layer_nr reflect reality. If you don't care that the value is wrong for some extruders then ignore this PR.

Ghostkeeper:
I don't think that it is reality. At least for 2 or fewer extruders it will not be reality due to the starting extruder not being present.

PR #1039, opened 2019-04-17
Only call planPrime() if prime_blob_enable is true.

smartavionics:
As discussed in https://github.com/Ultimaker/Cura/issues/5545.


Ghostkeeper:
Developers, see issue CURA-6485.

PR #1041, opened 2019-04-18
Treat skin over sparse infill like a first bridge layer skin

smartavionics:
Engine PR that goes with https://github.com/Ultimaker/Cura/pull/5635.


Ghostkeeper:
Seems like a good addition to us. Developers, see issue CURA-6486.

PR #1045, opened 2019-04-25
Remove non-standard double suffix

bjude:
The d suffix is not standard c++ and breaks the build for MSVC. All the places it's used are already doubles without the suffix


Ghostkeeper:
Ah, I didn't know this! This was probably my doing. I'll try to keep it in mind.

PR #1048, opened 2019-05-06
Implement wall_overhang_fan_speed to override fan speed while printing overhung walls.

smartavionics:
Allow overhung wall lines to have their own cooling fan speed.


Ghostkeeper:
What use case would you have in mind for this feature?

smartavionics:
Being able to alter the fan speed while printing overhung walls was a feature request when the wall overhang detection was introduced.

akshaal:
@smartavionics why did you close your own PR? This sounds like quite a useful feature for printing with filaments that can't be cooled all the time (PETG, ABS) but still need cooling for printing overhungs. I know that you have the feature in your fork, just curious why not here..

smartavionics:
Hi @akshaal , I can't remember now, that was so long ago. Anyway, I am no longer contributing new features to Cura, only bug fixes.

PR #1049, opened 2019-05-06
Implement cool_fan_lag setting.

smartavionics:
This PR makes print cooling fan speed overrides happen earlier so that the requested speed change has been achieved by the time the nozzle reaches the override zone.

### DIFF
--- a/src/LayerPlan.cpp
+++ b/src/LayerPlan.cpp
@@ -1623,12 +1623,8 @@ void LayerPlan::writeGCode(GCodeExport& gcode)
                             if (path_time >= fan_speed_override_at)
                             {
                                 // time for the fan override to kick in
-                                if (default_fan_speed != fan_override_events[0].second)
-                                {
-                                    gcode.writeComment("FAN OVERRIDE");
-                                    default_fan_speed = fan_override_events[0].second;
-                                    gcode.writeFanCommand(default_fan_speed);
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
To be consistent with the structure of CuraEngine, this sort of thing would normally be implemented in the LayerPlanBuffer.

smartavionics:
OK, let's forget it then,  I can't be bothered to haggle with you guys any more.

Ghostkeeper:
Yeah, I'm sorry. It's just that we need to keep the code quality high, especially if there is a high ratio of contributions from outside vs. from the maintainers themselves, or we'll lose control over this code.

The LayerPlanBuffer was introduced in order to insert pre-heat commands in advance in order to make sure that an extruder has the correct temperature when necessary. That is the same use case as inserting fan speed commands in advance in order to make sure that a fan has the correct speed when necessary. It wouldn't make sense to implement that logic in two different places.

smartavionics:
> That is the same use case as inserting fan speed commands in advance in order to make sure that a fan has the correct speed when necessary.

Actually, not true. The fan can change speed very much quicker than the nozzle can change temperature and so the granularity of the existing layer plan inserts mechanism is far too coarse. For optimal fan speed changes, you need to insert the fan commands within paths not on path boundaries.

PR #1056, opened 2019-05-20
Bridge bug fixes

smartavionics:
This PR contains a fix for the over large extrusions that can follow bridges when the bridge wall speed is very much lower than the normal wall speed (see https://community.ultimaker.com/topic/28053-cura-36-overextrusion-after-bridge-lines-solved-in-40/ and https://community.ultimaker.com/topic/28012-experimental-bridging-in-40-causes-big-blobs-in-slicing/).


Ghostkeeper:
Seems like a serious problem. Developers, see CURA-6541.

PR #1063, opened 2019-06-03
Add support for printers that have multiple extruders that share a single heater

smartavionics:
See https://github.com/Ultimaker/Cura/issues/5770


Ghostkeeper:
Developers, see issue CURA-6579.

PR #1069, opened 2019-06-06
extend processing order to carve multi volumes as well

BagelOrb:
I was having trouble in getting my mouse ears custom brim not carving away my model.

Now the option for infill meshes is extended such that it also works for normal meshes.


BagelOrb:
My current use case of mouse ears:
My model has a lot of complex geometry and it would be annoying to remove the brim afterwards, but the model doesn't have a lot of surface area, so I do need extra adhesion.

Before:
image


After: (setting the Mesh Processing Rank to 1 for the cylinder)
image

BagelOrb:
Colleagues of mine at TU Delft have also mentioned they experienced the annoyance.

BagelOrb:
New behavior is like this.
image

Setting 1:
top two cubes have a rank of 0 (default behavior)
bottom cube has rank 1

Top two cubes have alternate carve order doing its magic.
Bottom cube is disregarded.
image

Setting 2:
bottom cube has rank -1
image
bottom cube wins

Setting 3:
setting 2 plus en extra rank -1 cube:
image

Ghostkeeper:
Developers, see issue CURA-6580.

BagelOrb:
Example of person needing this:
https://www.reddit.com/r/3Dprinting/comments/do87x5/cura_430_printing_helper_disks_separately_the/

src/FffPolygonGenerator.cpp
```diff
@@ -309,7 +309,7 @@ void FffPolygonGenerator::slices2polygons(SliceDataStorage& storage, TimeKeeper&
         std::multimap<int, size_t> order_to_mesh_indices;
         for (size_t mesh_idx = 0; mesh_idx < storage.meshes.size(); mesh_idx++)
         {
-            order_to_mesh_indices.emplace(storage.meshes[mesh_idx].settings.get<size_t>("infill_mesh_order"), mesh_idx);
+            order_to_mesh_indices.emplace(storage.meshes[mesh_idx].settings.get<int>("infill_mesh_order"), mesh_idx);
```
rburema: It's not a big enough issue to stop me from merging this if I'm through with the component test, but I'm just curious to why you use `int` here. Or more specifically, why you'd replace `size_t`, and necessitating a new get method in the settings.

BagelOrb: A user might want to put a negative value. Would be combersome if you would have to increase all settings values of all other meshes if theres only one mesh you would like to have processed earlier than all the other.

rburema: Ok, makes sense!

PR #1076, opened 2019-06-14
Always move Z-axis separately [CURA-6582]

rburema:


smartavionics:
As I mentioned, this doesn't fix the issue it was intended to fix but if you add something like the code below it looks better (not well tested, yet)...
```
diff --git a/src/LayerPlan.cpp b/src/LayerPlan.cpp
index 0652d529..06347ad5 100644
--- a/src/LayerPlan.cpp
+++ b/src/LayerPlan.cpp
@@ -82,6 +82,7 @@ LayerPlan::LayerPlan(const SliceDataStorage& storage, LayerIndex layer_nr, coord
 : storage(storage)
 , configs_storage(storage, layer_nr, layer_thickness)
 , z(z)
+, final_travel_z(z)
 , mode_skip_agressive_merge(false)
 , layer_nr(layer_nr)
 , is_initial_layer(layer_nr == 0 - static_cast<LayerIndex>(Raft::getTotalExtraLayers()))
@@ -1752,6 +1753,10 @@ void LayerPlan::writeGCode(GCodeExport& gcode)
             }
             if (path.config->isTravelPath())
             { // early comp for travel paths, which are handled more simply
+                if (!path.perform_z_hop && final_travel_z != z && extruder_plan_idx == (extruder_plans.size() - 1) && path_idx == (paths.size() - 1))
+                {
+                    gcode.setZ(final_travel_z);
+                }
                 for(unsigned int point_idx = 0; point_idx < path.points.size(); point_idx++)
                 {
                     gcode.writeTravel(path.points[point_idx], speed);
diff --git a/src/LayerPlan.h b/src/LayerPlan.h
index 4606298c..38d16a83 100644
--- a/src/LayerPlan.h
+++ b/src/LayerPlan.h
@@ -236,7 +236,8 @@ private:
 
 public:
     const PathConfigStorage configs_storage; //!< The line configs for this layer for each feature type
-    int z;
+    coord_t z;
+    coord_t final_travel_z;
     bool mode_skip_agressive_merge; //!< Wheter to give every new path the 'skip_agressive_merge_hint' property (see GCodePath); default is false.
 
 private:
diff --git a/src/LayerPlanBuffer.cpp b/src/LayerPlanBuffer.cpp
index faaee55f..a35acbcd 100644
--- a/src/LayerPlanBuffer.cpp
+++ b/src/LayerPlanBuffer.cpp
@@ -100,6 +100,7 @@ void LayerPlanBuffer::addConnectingTravelMove(LayerPlan* prev_layer, const Layer
         const bool force_retract = extruder_settings.get<bool>("retract_at_layer_change") ||
           (mesh_group_settings.get<bool>("travel_retract_before_outer_wall") && (mesh_group_settings.get<bool>("outer_inset_first") || mesh_group_settings.get<size_t>("wall_line_count") == 1)); //Moving towards an outer wall.
+        prev_layer->final_travel_z = newest_layer->z;
         prev_layer->addTravel(first_location_new_layer, force_retract);
     }
 }
```

smartavionics:
I think there is a further problem with your commits in that when the z only move is done first, it will use the F value that was actually intended for the x/y part of the move and so it can end up much too quick (or slow  but more likely too quick). i.e if the travel speed is, say, 300 mm/S then that's the speed it will use the for z movement. My printer wouldn't mind but printers with moving platforms probably won't be happy.

smartavionics:
All in all, I think it would be better to revert these commits and try again.

rburema:
I'm inclined to agree, it seems I misinterpreted the original issue. Thank you for catching this!

smartavionics:
Yeah, sorry to be negative but I think the solution needs more thought.

rburema:
I appreciate the sentiment, but this is on me, not you. I sould have read the original ticket more carefully at the very least.

PR #1077, opened 2019-06-15
Fix helper-struct brim overlap with model when brim for prime-tower is on.

rburema:
Aims to fix https://github.com/Ultimaker/Cura/issues/5886


src/FffPolygonGenerator.cpp
```diff
@@ -1046,7 +1046,7 @@ void FffPolygonGenerator::processPlatformAdhesion(SliceDataStorage& storage)
     // If brim for prime tower is used, add the brim for prime tower separately.
     if (should_brim_prime_tower)
     {
-        SkirtBrim::generate(storage, storage.primeTower.outer_poly, 0, train.settings.get<size_t>("brim_line_count"));
+        SkirtBrim::generate(storage, storage.primeTower.outer_poly, 0, train.settings.get<size_t>("brim_line_count"), false);
```
Ghostkeeper: This is a boolean trap. It should be replaced with something like:

    constexpr allow_helpers = false;
    SkirtBrim::generate(..., allow_helpers);

I'll fix that up for you.

rburema: Thanks!

PR #1082, opened 2019-06-20
Syntax issues on paragraphs 81 & 137

kant:


Ghostkeeper:
Thanks! Very attentive.

PR #1087, opened 2019-07-01
Use correct setting when computing infill flow.

smartavionics:
Fixes bug in recent upstream merge.


Ghostkeeper:
Developers, see issue CURA-5892.

nallath:
That was my bad, should have seen that. Thanks for fixing it so fast!

PR #1093, opened 2019-07-13
infill: Remove stb basedir from include

thopiekar:
For some reason, someone decided to include stb_image.h from stb. But CMake already does the job well and will look for this directory during build configuration.
During build, it appears in the compilation command and therefore telling to look into "stb" will even fail, if there is no stb directory in any detected include directory.
Therefore, let's include stb_image.h directly and use all stb headers from one place as found by CMake.

### DIFF
--- a/src/infill/ImageBasedDensityProvider.cpp
+++ b/src/infill/ImageBasedDensityProvider.cpp
@@ -4,7 +4,7 @@
 
 #define STBI_FAILURE_USERMSG // enable user friendly bug messages for STB lib
 #define STB_IMAGE_IMPLEMENTATION // needed in order to enable the implementation of libs/std_image.h
-#include <stb/stb_image.h>
+#include <stb_image.h>
 
 #include "ImageBasedDensityProvider.h"
 #include "SierpinskiFill.h"


Ghostkeeper:
Hmm, well it worked for me previously and in this branch it doesn't work any more:

```
ruben@UM-MBA-187:~/Projects/CuraEngine/build$ cmake ..
-- Building with Arcus
-- Building Stb from source.
-- Created Stb installation at: /home/ruben/Projects/CuraEngine/build/stb-prefix/src
-- Building debug release of CuraEngine.
-- Building tests...
-- Configuring done
-- Generating done
-- Build files have been written to: /home/ruben/Projects/CuraEngine/build
ruben@UM-MBA-187:~/Projects/CuraEngine/build$ make
[  1%] Built target clipper
[  2%] Skipping update step for 'stb'
[  2%] No configure step for 'stb'
[  3%] No build step for 'stb'
[  4%] No install step for 'stb'
[  5%] Completed 'stb'
[  7%] Built target stb
Scanning dependencies of target _CuraEngine
[  8%] Building CXX object CMakeFiles/_CuraEngine.dir/src/infill/ImageBasedDensityProvider.cpp.o
/home/ruben/Projects/CuraEngine/src/infill/ImageBasedDensityProvider.cpp:7:10: fatal error: stb_image.h: No such file or directory
 #include <stb_image.h>
          ^~~~~~~~~~~~~
compilation terminated.
make[2]: *** [CMakeFiles/_CuraEngine.dir/build.make:617: CMakeFiles/_CuraEngine.dir/src/infill/ImageBasedDensityProvider.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:1023: CMakeFiles/_CuraEngine.dir/all] Error 2
make: *** [Makefile:163: all] Error 2
```

The `stb_image.h` file is located in `~/Projects/CuraEngine/build/stb-prefix/src/stb/stb_image.h`. CMake has added `~/Projects/CuraEngine/build/stb-prefix/src` to the include directories, as the `cmake` output says. So it seems that this was working fine.

Maybe there is a difference when building from source vs. finding stb on a system library or something?

thopiekar:
Guess it is not "maybe" but "for sure". If I'm not wrong CMake is looking in the default include directories for stb.h.
(By the way, I dislike the fact that you are downloading and installing software here in the engine. We have cura-build-environment for that in my point of view...)

I think it depends on what you want to see in the code. If you like it better to use `include <stb/stbimage.h>`, then we should make CMake look exactly for this relative path, too.
So the only required change would be to adjust the lookup for stb (which is my case) in CMake to fit to your manual download.

What do you prefer?

Ghostkeeper:
I'd prefer `<stb/stbimage.h>` as it's easier on the namespace.

thopiekar:
Ok. So I would appreciate if you as Ultimaker fix that issue with the
system lookup.
Could you share any priority on this? Any expected time for this to be
fixed?
If that takes too long and I have time to hack around I'll try to fix it on
my own.

Cheers!

Ghostkeeper:
Does this work for you? https://github.com/Ultimaker/CuraEngine/commit/cd3fa7419b027426630cdc03ca97ee80a529be09

thopiekar:
Works like a charm! Thanks!

PR #1094, opened 2019-07-19
Ensure skins have suitable thickness when adaptive layers are used

smartavionics:
As discussed in https://community.ultimaker.com/topic/22998-top-thickness-not-respected-when-using-adaptive-layers/.


rburema:
Hi @smartavionics,

Sorry for the delay (again). We've noticed we're falling behind on PR's again, so we're making a renewed push to pick a few of them up again.

We've discussed this PR internally, and we have a few remarks on it.
 - This seems to be a partial solution to a more general problem, since the same sort of behaviour is seen in the bottom layers, or interface, etc.
 - The question of whether the value of thickness in mm or the 'raw' amount of layers should be 'leading' will also become more complicated: In the front-end, the number of top-layers is calculated _from_ the total thickness of the top. In this PR, the total thickness is calculated from the number of layers. What will (or more importantly _should_) happen where the user (in the front-end) to input the number of top layers directly (overwriting the calculated value there) instead of the thickness in mm?

cheers,
Remco.

Dim3nsioneer:
I understand that related issues should ideally be solved with a common solution. However, the adaptive layer height feature is something that sees more and more interest and other slicers are currently introducing similar features.
This feature here is flagged as 'experimental' for a pretty long time already. As I understand (please correct me if I'm wrong) this PR would allow to make it a non-experimental feature. Therefore I would like to ask to consider this partial solution and to finally finish that feature started a long time ago.

PR #1097, opened 2019-07-27
Fix polygon.simplify

BagelOrb:
The goal of simplify was to remove verts as long as it doesn't introduce too much error.

Currently collinear line segments will not be simplified. It explicitly states in the code that collinear line segments should only be simplified if they are not too long.

Some algorithms may depend on that there are no collinear line segments in the input, so it is quite important that this is fixed. It might even happen that code crashes because of this.

Please compare this fix against the intent of the coders who wrote this new version of the `simplify()` function.

### DIFF
--- a/src/utils/polygon.cpp
+++ b/src/utils/polygon.cpp
@@ -355,7 +355,7 @@ void PolygonRef::simplify(const coord_t smallest_line_segment_squared, const coo
         //h^2 = (2A)^2 / b^2  [factor the divisor]
         //h^2 = 4A^2 / b^2    [remove brackets of (2A)^2]
         const coord_t height_2 = (4 * area_removed_so_far * area_removed_so_far) / base_length_2;
-        if (length2 < smallest_line_segment_squared && height_2 <= allowed_error_distance_squared) //Line is small and removing it doesn't introduce too much error.
+        if (height_2 <= allowed_error_distance_squared) // removing the vert doesn't introduce too much error. Both connected line segments are (nearly) collinear
         {
             continue; //Remove the vertex.
         }


BagelOrb:
This is the polygon which caused a crash for me:
`(57202,12388), (56402,12394), (56393,12394), (56382,13625), (57632,13633), (57621,12385), (57602,12385)`

`(57202,12388)` should have been removed.

BagelOrb:
There's some wrong commits in here.

BagelOrb:
(Had to remove some accidental commits from another branch.)

BagelOrb:
On another note:
According to the explanation of the function it should remove the following polygon fully:
`(15200,15647), (15201,15647), (15201,15650)`
But it doesn't.

Ghostkeeper:
This pull request breaks the Maximum Resolution setting. It has no effect any more except on the closing line segment which had an exception case in code below.

It also causes the unit tests to fail:

```
[ RUN      ] PolygonTest.simplifyLimitedLength
/home/ruben/Projects/CuraEngine/tests/utils/PolygonTest.cpp:286: Failure
Expected equality of these values:
  spiral.size()
    Which is: 0
  11 - 3
    Which is: 8
Should merge segments of length 1100 with 1200, 1300 with 1400 and first with last.
[  FAILED  ] PolygonTest.simplifyLimitedLength (1 ms)
```

The expected behaviour right now is that vertices with less than 5 microns deviation are removed anyway regardless of the Maximum Resolution setting. This is implemented here:

https://github.com/Ultimaker/CuraEngine/blob/ade1b43fda54292b8239cb60d8e837dea94532e3/src/utils/polygon.cpp#L363

I'd expect that exactly colinear lines are combined by that. Perhaps you'd like to add a test case for it?

BagelOrb:
Ah I guess the `&&` should have been an `||` then, instead of throwing away the first argument.

I'm currently a bit busy to implement a test case. I've mentioned two test cases above, though. You could make those into a test case.

BagelOrb:
This fix is simply faulty.

I've made a bug report instead:
https://github.com/Ultimaker/Cura/issues/6079

PR #1098, opened 2019-07-27
When adding lines, ignore lines less than 5um long.

smartavionics:
Those lines occasionally get created when skins and infill are generated and cause
unwanted travels to occur to their start locations.

As discussed in https://community.ultimaker.com/topic/28980-unnecessary-travel-moves-at-the-end-of-the-layer/


rburema:
devteam: CURA-6895

ninovanhooff:
Analysis by @rburema :

> Very small. Magic value, would prefer a constexpr + comment, but we can fix that ourselves.

I agree with this. So I'll just go ahead and merge this, and make the suggested change as a follow-up.

PR #1099, opened 2019-07-30
Avoid unwanted travel move when ironing is enabled but no lines/polygons are generated.

smartavionics:
When ironing is enabled it could generate a travel move even when no ironing lines/polygons were
going to be output. Now it doesn't.

As discussed in https://community.ultimaker.com/topic/28980-unnecessary-travel-moves-at-the-end-of-the-layer/

### DIFF
--- a/src/TopSurface.cpp
+++ b/src/TopSurface.cpp
@@ -50,6 +50,11 @@ bool TopSurface::ironing(const SliceMeshStorage& mesh, const GCodePathConfig& li
     Polygons ironing_lines;
     infill_generator.generate(ironing_polygons, ironing_lines);
 
+    if (ironing_polygons.empty() && ironing_lines.empty())
+    {
+        return false; //Nothing to do.
+    }
+
     layer.mode_skip_agressive_merge = true;
 
     if (pattern == EFillMethod::LINES || pattern == EFillMethod::ZIG_ZAG)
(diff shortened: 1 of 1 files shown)


rburema:
devteam: CURA-6896

ninovanhooff:
Sorry for the delay @smartavionics. 
Your PR is concise and clean and well documented in commit messages. I also read the forum discussion. A
ll seems well. We'd like to merge it. 

When patching the current master with your change though, I don't see any change in the gcode, using the supplied `logo2.3mf` by @kulfuerst as an example, both CE Master and the patched master produce a binary equal gcode file.

Could you please verify the attached results? the gcode file being both the output of master and the patched version (considering they are equal I did not package both)

Regardless, there might be some merit in merging the change, for models with other characteristics. I will also try the cage model mentioned in the forum.

[logo_branch.zip](https://github.com/Ultimaker/CuraEngine/files/3744107/logo_branch.zip)

smartavionics:
Hi @ninovanhooff , thanks for looking at this PR. I think you're using the wrong project to test this PR, you need to use Cage.3mf which the OP provided on the forum thread. The logo project shows a different issue. Anyway, here's layer 2 of the Cage model without and with this PR in place, spot the difference!

Screenshot_2019-10-18_15-58-24

Without the PR, the travel move from the bottom right area back up to the top of the left hand hole goes via the right hand corner of the bottom circular cutout. With the PR, the travel move is direct from lower right to top of left hand hole.

Screenshot_2019-10-18_15-57-55

ninovanhooff:
Thanks for the clarification. Merge approved.

PR #1107, opened 2019-08-14
Only calculate the retraction region once rather than for every outer wall inset.

smartavionics:
This was wasting a lot of time when the part had a lot of holes or the outer wall was
fragmented.

Part 1 of the fixes required for https://github.com/Ultimaker/Cura/issues/6164.

Part 2 to follow and is not related to this PR.

### DIFF
--- a/src/InsetOrderOptimizer.h
+++ b/src/InsetOrderOptimizer.h
@@ -40,7 +40,8 @@ public:
     z_seam_config(mesh.settings.get<EZSeamType>("z_seam_type"), mesh.getZSeamHint(), mesh.settings.get<EZSeamCornerPrefType>("z_seam_corner")),
     added_something(false),
     wall_overlapper_0(nullptr),
-    wall_overlapper_x(nullptr)
+    wall_overlapper_x(nullptr),
+    retraction_region_calculated(false)
     {
     }
 private:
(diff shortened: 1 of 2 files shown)


src/InsetOrderOptimizer.h
```diff
@@ -40,7 +40,8 @@ class InsetOrderOptimizer
     z_seam_config(mesh.settings.get<EZSeamType>("z_seam_type"), mesh.getZSeamHint(), mesh.settings.get<EZSeamCornerPrefType>("z_seam_corner")),
     added_something(false),
     wall_overlapper_0(nullptr),
-    wall_overlapper_x(nullptr)
+    wall_overlapper_x(nullptr),
+    retraction_region_calculated(false)
```
Ghostkeeper: In general it's best to group boolean fields together so that they also get stored together. This helps with packing the class so that it'll use less memory and constructs faster.

Not really important in this case, but useful to know!

Ghostkeeper:
Developers, see issue CURA-6944.

Ghostkeeper:
Quick performance test. With 3 normal-type models I saw no difference. On a model engineered towards the case that this performance fix should be good at, slicing time went down by about 11%.

Ghostkeeper:
Thanks for the tweak! Seems like all is working well.

PR #1108, opened 2019-08-16
CURA-6675 Retract Minimum Travel Distance should always work

LipuFei:


nallath:
The proposed fix also doesn't work, since it doesn't take the min travel distance into account at all. 

[retraction_minimum_travel_bug.3mf.zip](https://github.com/Ultimaker/CuraEngine/files/3544843/retraction_minimum_travel_bug.3mf.zip)

smartavionics:
> The proposed fix also doesn't work, since it doesn't take the min travel distance into account at all.

Are you referring to my suggested fix? If so, it does take the min travel distance into account.

path->retract = retract || (retract_threshold > 0 && distance > retract_threshold && distance >= retraction_config.retraction_min_travel_distance);

src/LayerPlan.cpp
```diff
@@ -459,7 +459,15 @@ GCodePath& LayerPlan::addTravel(Point p, bool force_comb_retract)
             }
         }
     }
-    
+
+    // CURA-6675:
+    // Retraction Minimal Travel Distance should work for all travel moves. If the travel move is larger than the
```
smartavionics: comment is wrong: move is shorter not larger

src/LayerPlan.cpp
```diff
@@ -459,7 +459,15 @@ GCodePath& LayerPlan::addTravel(Point p, bool force_comb_retract)
...
+    // Retraction Minimal Travel Distance, retraction should be disabled.
+    if (!is_first_travel_of_layer && last_planned_position && shorterThen(*last_planned_position - p, retraction_config.retraction_min_travel_distance))
```
smartavionics: This is duplicating the test that follows with a minor change. If you are going to do this you should refactor the test below and avoid the duplication.

src/LayerPlan.cpp
```diff
@@ -459,7 +459,15 @@ GCodePath& LayerPlan::addTravel(Point p, bool force_comb_retract)
...
+    {
+        path->retract = false;
+    }
+
```
smartavionics: Sorry again. I think this is OK.

src/LayerPlan.cpp
LipuFei: Fixed in de1f4506

kazooless:
Is there any information as to what this bug actually is? I think the disabling of this in the fdm printer is the culprit of why we can't get the creality printers to print as well as v4.0 with our old profiles and the legacy definitions I created for 4.3. I was hoping this would be fixed by the 4.4 release. 

Is it disabled for all printers? Or just the generic? Is there anything I can do to help with this?

Ghostkeeper:
The original bug report came from internal (actually from me as I was trying to create screenshots of the Retraction Minimum Travel Distance setting for my Settings Guide), so it's not here on Github. The bug report states that the entire setting has no effect any more on the print. This started happening in version 4.0. In 3.6 it's still working properly. The original bug report was found using Custom FFF Printer and also reproduced with the Ultimaker 3, so it's likely an issue for all printers.

smartavionics:
Just did a quick test and when combing is enabled, the min travel distance doesn't appear to have any effect. When combing is disabled, it does have an effect but I haven't verified that the behaviour is completely sensible.

smartavionics:
Sorry, I'm looking at my current source which is based on the current master.

smartavionics:
Looking at the combing code I see something odd. In LayerPlan::addTravel(), it updates last_planned_position and then later, uses that variable again assuming it has the original value...
````
                }
                for (Point& comb_point : combPath)
                {
                    if (path->points.empty() || vSize2(path->points.back() - comb_point) > maximum_travel_resolution * maximum_travel_resolution)
                    {
                        path->points.push_back(comb_point);
                        distance += vSize(last_point - comb_point);
                        last_point = comb_point;
                    }
                }
                last_planned_position = combPath.back();       <<<<<<<<<<<<<<<<<<<<<<<<
                distance += vSize(last_point - p);
````

I'm going to comment out that assignment and test what happens.

nallath:
Closing this in favor of the pull request of @smartavionics

PR #1112, opened 2019-08-20
First travel move on first layer should be z-hopped if applicable.

rburema:
Users might have start gCode that depends on z-hopping, even if there is only one extruder.
part of CURA-6718


src/LayerPlan.cpp
```diff
@@ -382,6 +382,11 @@ GCodePath& LayerPlan::addTravel(Point p, bool force_comb_retract)
         bypass_combing = true; // first travel move is bogus; it is added after this and the previous layer have been planned in LayerPlanBuffer::addConnectingTravelMove
         first_travel_destination = p;
         first_travel_destination_is_inside = is_inside;
+        if (layer_nr == 0 && extruder->settings.get<bool>("retraction_hop_enabled"))
```
LipuFei: I think this needs to check both `retraction_enable` and `retraction_hop_enabled`.

rburema: Adressed in bd4d9ab

PR #1113, opened 2019-08-20
Bugfix for getAngleLeft

BagelOrb:
I copied code from Cura to somewhere and we found a bug in that code, which we tracked back to this origin.

Based on some simple reasoning it should be easy to understand that if the determinant is zero than the lines are collinear:
```
a.x * c.y - a.y * c.X = 0
so  a.x * c.y = a.y * c.X
so  a.X / a.Y = c.X / c.Y
```

We can easily disambiguate between whether the two line segments are oriented in the same direction or in the opposite direction using the logic presented in this commit.


src/utils/LinearAlg2D.cpp
```diff
@@ -18,6 +18,20 @@ float LinearAlg2D::getAngleLeft(const Point& a, const Point& b, const Point& c)
     const Point bc = c - b;
     const coord_t dott = dot(ba, bc); // dot product
     const coord_t det = ba.X * bc.Y - ba.Y * bc.X; // determinant
+    if (det == 0)
+    {
+        if (
+            (ba.X != 0 && (ba.X > 0) == (bc.X > 0))
+            || (ba.X == 0 && (ba.Y > 0) == (bc.Y > 0))
+            }
```
smartavionics: This bracket should be }, not ).

BagelOrb: lol. I used the Github interface for this fix.

BagelOrb: I've found out how to change code in a PR from within Github. I'm somewhat of a master-hacker, you know. :P

Ghostkeeper:
Developers, see issue CURA-6900.

Ghostkeeper:
It looks like those cases were explicitly avoided in our set of tests for this function. Maybe we already knew that they were failing...

PR #1114, opened 2019-08-20
Fix simplify

smartavionics:
A few changes to simplify() that fix the issue reported in https://github.com/Ultimaker/Cura/issues/6215.

However, it fails the simplifyLimitedError test. Sorry, I don't understand what the test is doing so I don't know if it's the test or the new simplify() that is in error.


smartavionics:
Now includes fix for https://github.com/Ultimaker/CuraEngine/issues/1151.

smartavionics:
I know that you really want to merge these fixes but are put off by the fact that it fails PolygonTest.simplifyLimitedError. So I have been looking at why that test fails and have come to the following conclusions:

1 - the test code incorrectly calculates the error height allowed as 4 * area / diagonal_length but, according to the formula in the comments above and also the code in simplify() that should be 2 not 4.

2 - a bigger problem is that the height being calculated is based simply on the 1400 and 1500 lines and doesn't take into account the error accumulated by removing the previous lines. simplify() works with accumulated error but this test code doesn't.

I don't believe that this is a valid test at the moment.

Ghostkeeper:
Instead of removing dysfunct tests we should fix them.

We've been having some trouble with that test since it's hard to verify the ground truth. You can't really work out the ground truth automatically without repeating the code under test, which would make the test contain the same bugs as the code itself. So in the test the ground truth is worked out manually and the result is hard-coded (in the test code this is the `11 - 5` and `11 - 4` you're seeing). But this is then hard to verify if the result of `simplify()` changes such that the test fails: Why should the result contain 6 or 7 segments?
I think the best outcome is to still hard-code it but document the steps that led to that answer. The manual work has been lost now.

smartavionics:
Obviously, the person who writes the test should not be the same person who coded the facility being tested.

Dim3nsioneer:
It seems to me the process here is a bit stuck. If this is the case, can we somehow get it running again? @Ghostkeeper what should be the next logical step from your point of view?

Ghostkeeper:
Yeah, true. In my point of view the next steps are to write a new test to verify the algorithm being limited by the error margin, as well as for Ultimaker to review and test the code, which we haven't done yet either.

BagelOrb:
This PR fixes JIRA ticket CURA-7243 and CURA-6955, which was for https://github.com/Ultimaker/Cura/issues/6604.

Please merge this soon, guys!

BagelOrb:
@smartavionics Why is this PR titled "**Maybe** fix simplify"? Are you insecure that this PR fixes it? Or is it meant to make the title look like a suggestion?

smartavionics:
This PR evolved. It started as a fix for a particular issue and I wasn't 100% certain at the time that what I had done was good. Subsequently, I added a fix for another issue. Having been using this code for many months with no problems I would now be happy to say that the PR does indeed fix the known problems with simplify(). I think the insecurity really arose from the fact that the above mentioned test fails with this code but given that the test is bogus, I no longer am concerned about that.

BagelOrb:
I'm looking into writing the test.

The process of dealing with PRs like this seems fickle to me. I am taking matters into my own hands and dealing with this issue the way it should be dealt with.

smartavionics:
I think it is good you are writing the test. The test needs to be written by someone who wasn't directly involved in writing the thing being tested.

BagelOrb:
Either that or you write the test before writing the code..

BagelOrb:
Still, writing the test yourself is better than not having a test. Writing the test yourself just bares the risk of having overfitting test criteria, but it doesn't neccesarily mean that the tests you write are always bad.

smartavionics:
> Either that or you write the test before writing the code..

I don't think so because if the same person(s) writes the test and the code, then if they have a misunderstanding as to the function of the code then that will be reflected in both the test and the code. The code can pass the test but the code may not actually do what it should.

smartavionics:
With small teams, it's inevitable that tests will get written by the same people who did the coding. At the very least, someone else should review the code and tests with a very critical eye.

BagelOrb:
Sure. We have a component test phase and a QA test phase. We test quite thoroughly. However, I'm not sure as to the procedure for writing the automated tests. I think they are written by the coder himself ATM.

BagelOrb:
I've modified the test to something that makes sense, but the test only passes if I call simplify() on it 6 times. You could simply make simplify() call itself until it doesn't remove any point any more.

smartavionics:
Can you post the new test code, please?

BagelOrb:
It's still uncommited changes, with all kinds of debug code in it atm.

[test_debug_changed.diff.txt](https://github.com/Ultimaker/CuraEngine/files/4301667/test_debug_changed.diff.txt)

The most important thing is to add two unsimplifyable nodes:
```
     PolygonRef spiral = spiral_polygons.newPoly();
-    spiral.add(Point());
+    spiral.emplace_back(-11000, 11000);
+    spiral.emplace_back(-11000, 1000);
+    spiral.emplace_back(0, 0);
```

And change the acceptance criteria accordingly.

BagelOrb:
However, the very first point at 0,0 from which we want to force the algorithm to start the simplification is by definition simplifiable, because the length of the first spiral segment is less than the allowed height. The test is therefore inherently flawed.

Moreover, a spiral is a very contrived example, because it is a non-euclidean solid; it's not watertight because of the self-intersections of the last segment which is connected to the first point.

I'm writing a new test altogether.

smartavionics:
> I'm writing a new test altogether.

Excellent!

BagelOrb:
I am rather worried by the fact that my temporary test doesn't pass with a single call to `simplify()` though. Could you please fix the function?

smartavionics:
Did you see this:

> 1 - the test code incorrectly calculates the error height allowed as 4 * area / diagonal_length but, according to the formula in the comments above and also the code in simplify() that should be 2 not 4.

You haven't changed the 4 to 2 so you don't agree with my analysis?

BagelOrb:
Uhm. Idk about the calculation, but the value of the variable height is correct: approx 2051, i.e. pythagoras on 1400,1500

BagelOrb:
Either way the function `simplify()` should never perform any change to the polygons **_the second time it is called in a row_**.

smartavionics:
> Either way the function simplify() should never perform any change to the polygons the second time it is called in a row.

I wonder, because simplify() uses accumulated error, doesn't that mean that if you call it multiple times you can get different results? That's because a point that doesn't get removed the first time because the accumulated error would be too high, gets removed on the next call because the accumulated error would be zero at that point? And so on...

BagelOrb:
Uhm yes, you are right. Although the area calculations are an implementation detail.

I noticed that when performing some tests I might end up with collinear segments in the final polygon. I think that could be problematic for some algorithms.

BagelOrb:
> However, the very first point at 0,0 from which we want to force the algorithm to start the simplification is by definition simplifiable, because the length of the first spiral segment is less than the allowed height. The test is therefore inherently flawed.

This wasn't true. A point may never be removed if it is connected to a segment longer than `smallest_line_segment`. The point 0,0 should therefore always remain in the output.

BagelOrb:
> Uhm. Idk about the calculation, but the value of the variable height is correct: approx 2051, i.e. pythagoras on 1400,1500

I hadn't eaten yet :S. This comment was also wrong. You were right. It should be approx half the diagonal.

BagelOrb:
Please see https://github.com/Ultimaker/CuraEngine/pull/1214

BagelOrb:
Ok it took me a while before I really understood the requirements of the function signature.

1. Simplify simplifies the geometry, but doesn't guarantee the amount of simplification. (The implementation uses an area based heuristic to estimate the deviation distance, and it doesn't always remove a vertex even though it could)

2. Simplify may never remove a vertex if either connected segment is longer than `smallest_line_segment`

3. Simplify may only remove a vertex if the distance from the vertex removed to the final polygon is less than `allowed_error_distance`

smartavionics:
Thanks for #1214. I still don't think that's right because the height value is not based on the cumulative area. So here's a version based on your branch that  does use the cumulative area...

```
TEST_F(PolygonTest, simplifyLimitedError)
{
    //Generate a square spiral with increasingly large corners until the area exceeds the limit.
    Polygons spiral_polygons;
    PolygonRef spiral = spiral_polygons.newPoly();
    spiral.emplace_back(-15000, 11000); // introduce points that may never be simplified so that we know the simplification should start at 0,0
    spiral.emplace_back(-15000, 1000);
    spiral.emplace_back(0, 0);

    //Generate a square spiral, 90 degree corners to make it easy to compute the area loss while retaining a positive area per corner.
    coord_t segment_length = 1000;
    Point last_position(0, 0);
    double angle = 0;
    coord_t accumulated_area = 0;
    for (size_t i = 0; i < 10; i++)
    {
        const coord_t dx = std::cos(angle) * segment_length;
        const coord_t dy = std::sin(angle) * segment_length;
        accumulated_area += last_position.X * (last_position.Y + dy) + last_position.Y * (last_position.X + dx);
        last_position += Point(dx, dy);
        spiral.add(last_position);
        segment_length += 100;
        angle += M_PI / 2;
    }

    Polygon spiral_before = spiral;
    Polygons spirals_before = spiral_polygons;

    //Take the accumulated_area and plug it into the formula for the height to get at the baseline height, which is our allowed error.
    const coord_t diagonal_length = vSize(last_position);
    //A = 0.5 * b * h. diagonal_length is the base line in this case.
    //2A = b * h
    //2A / b = h
    const coord_t height = 2 * accumulated_area / diagonal_length; //Error of the first vertex we want to keep, so we must set the limit to something slightly lower than this.
    spiral_polygons.simplify(10000, 20 * height - 10);

    EXPECT_THAT(spiral.size(), testing::AllOf(testing::Ge(3), testing::Le(spiral_before.size() - 9))) << "Should simplify all spiral points except those connected to far away geometry.";
}
```

Also,  I would expect the result to have exactly 3 points.

BagelOrb:
Just force pushed a new version of the last commit. Please git reset --hard on my branch.

BagelOrb:
The function signature doesn't mention anything about area, so the test should not mention anything about area. That is an implementation detail.

Please refer to my 3 criteria I mentioned above.

BagelOrb:
> Also, I would expect the result to have exactly 3 points.

That spiral polygon needs to have 4 segments, because 3 of those have a length larger than `smallest_line_segment`.

smartavionics:
> That spiral polygon needs to have 4 segments, because 3 of those have a length larger than smallest_line_segment.

If 3 segments have a length > smallest_line_segment, that should leave 3 points in the polygon.

At the end, it should be these points that remain:

```
    spiral.emplace_back(-15000, 11000); // introduce points that may never be simplified so that we know the simplification should start at 0,0
    spiral.emplace_back(-15000, 1000);
    spiral.emplace_back(0, 0);
```

BagelOrb:
Those 3 segments have 4 points, so those 4 points should remain in.
`Point (1400,1500)` should also remain in.

BagelOrb:
When checking whether a vertex can be removed **both** of its segments should be checked whether they are short enough to allow for the removal.

smartavionics:
> When checking whether a vertex can be removed both of its segments should be checked whether they are short enough to allow for the removal.

That sounds reasonable, let me look at simplify()...

BagelOrb:
As it currently stands, there is no automated test which this PR fixes, while we have manually verified that this PR fixes at least 4 known bugs. We should check where those bugs came from and write automated tests for those.

smartavionics:
So simplify could be altered with the following diff to take into account the length of the line segment following the current point of interest...
```
@@ -332,6 +332,7 @@ void PolygonRef::simplify(const coord_t smallest_line_segment_squared, const coo
         Point next = path->at((point_idx + 1) % size());
 
         const coord_t length2 = vSize2(current - previous);
+        const coord_t next_length2 = vSize2(current - next);
 
         //Check if the accumulated area doesn't exceed the maximum.
         accumulated_area_removed += current.X * next.Y - current.Y * next.X; //Shoelace formula for area of polygon per line segment.
@@ -350,7 +351,7 @@ void PolygonRef::simplify(const coord_t smallest_line_segment_squared, const coo
         //h^2 = (2A)^2 / b^2  [factor the divisor]
         //h^2 = 4A^2 / b^2    [remove brackets of (2A)^2]
         const coord_t height_2 = (4 * area_removed_so_far * area_removed_so_far) / base_length_2;
-        if (length2 < smallest_line_segment_squared && height_2 <= allowed_error_distance_squared) //Line is small and removing it doesn't introduce too much error.
+        if (length2 < smallest_line_segment_squared && next_length2 < smallest_line_segment_squared && height_2 <= allowed_error_distance_squared) //Line is small and removing it doesn't introduce too much error.
         {
             continue; //Remove the vertex.
         }
```

BagelOrb:
Yeah. It also needs to be added to the last two checks for completing the polygon, though.

smartavionics:
Perhaps, this...
```
@@ -332,6 +332,7 @@ void PolygonRef::simplify(const coord_t smallest_line_segment_squared, const coo
         Point next = path->at((point_idx + 1) % size());
 
         const coord_t length2 = vSize2(current - previous);
+        const coord_t next_length2 = vSize2(current - next);
 
         //Check if the accumulated area doesn't exceed the maximum.
         accumulated_area_removed += current.X * next.Y - current.Y * next.X; //Shoelace formula for area of polygon per line segment.
@@ -350,7 +351,7 @@ void PolygonRef::simplify(const coord_t smallest_line_segment_squared, const coo
         //h^2 = (2A)^2 / b^2  [factor the divisor]
         //h^2 = 4A^2 / b^2    [remove brackets of (2A)^2]
         const coord_t height_2 = (4 * area_removed_so_far * area_removed_so_far) / base_length_2;
-        if (length2 < smallest_line_segment_squared && height_2 <= allowed_error_distance_squared) //Line is small and removing it doesn't introduce too much error.
+        if (length2 < smallest_line_segment_squared && next_length2 < smallest_line_segment_squared && height_2 <= allowed_error_distance_squared) //Line is small and removing it doesn't introduce too much error.
         {
             continue; //Remove the vertex.
         }
@@ -373,7 +374,7 @@ void PolygonRef::simplify(const coord_t smallest_line_segment_squared, const coo
     }
 
     //For the last/first vertex, we didn't check the connection that closes the polygon yet. Remove it if it's too short.
-    if(new_path.size() > 2 && (vSize2(new_path.back() - new_path[0]) < smallest_line_segment_squared || vSize2(new_path.back() - new_path[new_path.size() - 2]) < smallest_line_segment_squared))
+    if(new_path.size() > 2 && vSize2(new_path.back() - new_path[0]) < smallest_line_segment_squared && vSize2(new_path.back() - new_path[new_path.size() - 2]) < smallest_line_segment_squared)
     {
         if (LinearAlg2D::getDist2FromLine(new_path.back(), new_path[new_path.size() - 2], new_path[0]) < allowed_error_distance_squared)
         {
```

BagelOrb:
That code change already fixes https://github.com/Ultimaker/Cura/issues/6604.

I can't reproduce https://github.com/Ultimaker/Cura/issues/6215 or https://github.com/Ultimaker/CuraEngine/issues/1151.

So currently I am inclined to just incorporate that fix without merging the rest of this PR.

Do you have any other testing models which produce those bugs?

BagelOrb:
[simplify_fix.diff.txt](https://github.com/Ultimaker/CuraEngine/files/4302023/simplify_fix.diff.txt)

This is the way I fixed `simplify()`.

smartavionics:
I still believe the other changes in this PR are beneficial, especially the change that stops too many points being removed when removing co-linear points. So I will be continuing to use this myself and will very likely incorporate the extra test for the length of the following segment as just discussed.

BagelOrb:
I was just looking at the commit on co-linearity, but I don't really understand your reasoning in the commit msg.

Could you please explain more?

BagelOrb:
You say "we need to test against the original line direction", but the goal of this code is to prevent co-linearity in the **_output_**, not to remove co-linearity in the input.

smartavionics:
The problem there was that when the model had very many tiny line segments, the co-linearity removal was too eager because what it was doing was looking to see if a point was nearly in line with the previous segment, if it was, the point was dropped. But on the next iteration, when it does the same check it would be using a different line to test co-linearity against. In effect, as points were dropped, the test line would move around the polygon and so points would be considered to be co-linear and dropped when they should not have been.

BagelOrb:
Aha. So the co-linearity collapsing can escalate over the whole polygon. However, your fix might still leave one co-linearity in the output, because you are checking against the input instead of the output.

Let me think...

BagelOrb:
Some of the escalation is already prevented by `else if (length2 >= smallest_line_segment_squared` 

However, if you set `smallest_line_segment` really high you would still expect the escalation to happen. However, I am unable to write an automated test which triggers the escalation. I've tried quite a lot of settings of my circle test and my sine test.

BagelOrb:
I also noticed a mistake in the area calculation. It should be `1 * ` instead of `4 *`.
I've fixed that in my simplify fix PR and made a frontend PR to update that setting here: https://github.com/Ultimaker/Cura/pull/7247

smartavionics:
I saw that you have been busy.

Have you also incorporated the fix to stop the co-linear point removal doing too much as we discussed on Saturday? I think that is still needed.

BagelOrb:
Based on the code I also still think it is needed, but I first want to find an automated test to prove that it is a problem. I have been trying loads of situations, but I cannot for the life of me find a problem case.

It seems to me that with lower settings of max resolution does result in coarser meshes sometimes, but the deviation is always less than `allowed_error_distance + 5`, so the function is still up to spec.

The user of https://github.com/Ultimaker/CuraEngine/issues/1151 used very low setting values which are below 10 micron. If we simply require that the Max resolution is larger than 10 micron we prevent that issue altogether.

smartavionics:
With the original code, I think the problem would always occur when there is a sequence of segments that were longer than max resolution but whose (individual) removal did not cause a deviation greater than the allowed value.

BagelOrb:
Hm ok, but that problem doesn't occur anymore because of my fix for the `smallest_line_segment`, even though I didn't touch the code for handling colinearity.

smartavionics:
Here's the project from #1151 sliced by your current PR, note how there are still some flats on the curves...

Screenshot_2020-03-09_13-15-20

And here is what I get using my PR that has the additional fix for the co-linear wall collapse, it's smoother...

Screenshot_2020-03-09_13-18-58

BagelOrb:
Yes, but is the smoothness good? The whole goal of `simplify()` is to reduce the smoothness.

Are the constraints of `simplify` violated?

Could you please share that model to me somehow?

smartavionics:
> Yes, but is the smoothness good? The whole goal of simplify() is to reduce the smoothness.

Sure, but those flats are mighty flat, some would say ugly.

Here's the project file I used...

https://www.dropbox.com/s/b7ru4pk6gzx5uhl/MaxRes%20bug%20demo%2C%20%28AutoDropModelsOff%29%20ZeroHorizontalExpansion%20.3mf?dl=0

max-res and max-dev set to 0.05, no other changes.

BagelOrb:
I agree that the current behavior doesn't make sense. At max res = 0.1 the output is smoother than at max res = 0.05.

Perhaps I should give up the search for an automation test which shows the problem and just fix it without an automated test.

smartavionics:
> Perhaps I should give up the search for an automation test which shows the problem and just fix it without an automated test.

Don't agonise, just use my fix (43fb28f) or write something yourself that does similar.

BagelOrb:
Yeah. Why torture myself?!

BTW you forgot to initialize them.

BagelOrb:
Then now I am convinced we can close this PR in favor of my PR.

I am mostly afraid of your commit 'Fix point indexing'. It changes quite a lot. It seems you misunderstood the original intent of the implementation there...

Is it okay if I close this one in favor of mine?

smartavionics:
> BTW you forgot to initialize them.

I think you will find it was not possible for them to be referenced before they were assigned.

smartavionics:
> I am mostly afraid of your commit 'Fix point indexing'. It changes quite a lot. It seems you misunderstood the original intent of the implementation there...

Maybe, maybe not.

> Is it okay if I close this one in favor of mine?

Sure, but it would be polite of you to add some acknowledgement of the work I have done on this over the many months it has been outstanding. Otherwise, if people just look at your commits, my efforts will be completely invisible.

BTW, I haven't yet tested if your PR fixes #6215, I need to do that.

smartavionics:
Yes, https://github.com/Ultimaker/Cura/issues/6215, looks OK now with your PR.

smartavionics:
This has done its job, closing...

BagelOrb:
Thanks for all the insights and discussion!

After I made what I thought were my final commits yesterday I noticed the CI failed on another automated test. It turned out the colinear collapsing didn't really work.

I've spent some more time on it and realized with a lot of testing that the function can be simplified a lot further. It might be interesting for you to look at my current implementation.

smartavionics:
Sorry Tim, you've blown it.

Your code from yesterday appeared to work OK, your latest version has resurrected https://github.com/Ultimaker/Cura/issues/6215.

Screenshot_2020-03-10_09-01-37

BagelOrb:
Not anymore. I made a little mistake. I've just run all tests and it now finally seems to work perfectly!

BagelOrb:
Had to force push again. Sorry for that.

Ghostkeeper:
We generally let the person who wrote the code also test the code since he has a better understanding of where the code is fragile and where things have been going wrong during development. Of course, the component tester, the quality engineers AND the system integration testers still do their rounds of testing, some of which will also add automated tests.

PR #1115, opened 2019-08-22
Use min x/y distance when layer below protrudes beyond current layer (wall is sloped).

smartavionics:
This helps support adhere to sloping areas of the model as shown here...

Screenshot_2019-08-02_09-06-02

Fixes https://github.com/Ultimaker/Cura/issues/6123.

See also https://github.com/smartavionics/Cura/issues/12.


AbeFM:
I'll take a look at that. I was doing some more geometrical parts recently and found a lot of missing support material where I expected to have some.

I should get a specific version, or just the most current?

Unrelated: Is there a "don't support bridges" equivalent? I have some designs with lots of printable bridges that end up generating supports that I could block manually if I must.

AbeFM:
Does the new code work on ANYTHING that isn't perfectly vertical, or the mirror of the support angle, or the "bridge" angle.....?

smartavionics:
>I'll take a look at that. I was doing some more geometrical parts recently and found a lot of missing support material where I expected to have some.

> I should get a specific version, or just the most current?

This PR just contains the same code as my cura release which I think you have already tried and reported that it worked OK.

> Unrelated: Is there a "don't support bridges" equivalent? I have some designs with lots of printable bridges that end up generating supports that I could block manually if I must.

No, bridges and support are mutually exclusive. You can use bridges or support but if you use both together it will use support in preference to creating a bridge.

smartavionics:
> Does the new code work on ANYTHING that isn't perfectly vertical, or the mirror of the support angle, or the "bridge" angle.....?

It requires the layer below to protrude more than 1/10 of the layer height. So a slight tilt will not cause the min x/y distance to be used.

AbeFM:
Oh, you're using support line width - since I use different nozzles for support interface as support infill and object in general, this is probably a big upgrade... Or why I get weird issues sometimes. :-)

Anything in particular to keep an eye out for?

AbeFM:
Very happy to see this roll in - it was a major shortcoming and my many, unending thanks for SA for developing this and the rest of you for merging it in!!!

I've been using this pretty heavily and I find it solid.

PR #1118, opened 2019-08-23
Avoid prime tower brim duplication

smartavionics:
When build plate adhesion was brim and the prime tower brim was enabled, the brim was added twice. Now, when build plate adhesion is brim, enabling the prime tower brim doesn't change anything but when build plate adhesion is none or skirt, the prime tower brim setting does add the brim prime tower.


Ghostkeeper:
I don't know if you've taken this into account as I didn't look into the code yet, but one of our requirements for this feature is that we can have a brim for the normal model but a different-sized brim for the prime tower, so both can be enabled. Then it's just a matter of what the unit of the setting is: Is it an added width on top of the brim width, or should it configure the full width of the brim including the ordinary brim?

smartavionics:
Well, right now it will generate two brims of the same size on top of each other around the prime tower. This PR stops it doing that.

smartavionics:
I will look into adding to this PR to make the behaviour as you state above.

smartavionics:
At the moment, there is no setting for the width of the prime tower brim. It just uses the same width as the regular brim. You would need to add a separate prime tower brim width setting.

smartavionics:
I have simplified this PR whilst retaining its benefit and it will be easy to make the prime tower brim width different in the future if you decide to do that.

PR #1122, opened 2019-08-28
Ignore slight kinks in almost straight lines when z-seam preference is sharpest corner.

smartavionics:
Now, the corner angle has to be at least +/- 0.05 radians to be considered.

Prompted by the issue raised in https://community.ultimaker.com/topic/29326-cura-line-startz-seam-problem/.

The model there was getting some bad z-seam locations when z-seam alignment was sharpest corner and the corner preference was hide seam. The model only has convex corners but there are some rogue points in one of the long sides that produce a very slight concave outline and so they were being picked up as the best location for the z-seam.

### DIFF
--- a/src/pathOrderOptimizer.cpp
+++ b/src/pathOrderOptimizer.cpp
@@ -142,7 +142,12 @@ int PathOrderOptimizer::getClosestPointInPolygon(Point prev_point, int poly_idx)
         const Point& p2 = poly[(point_idx + 1) % poly.size()];
         // when type is SHARPEST_CORNER, actual distance is ignored, we use a fixed distance and decision is based on curvature only
         float dist_score = (config.type == EZSeamType::SHARPEST_CORNER && config.corner_pref != EZSeamCornerPrefType::Z_SEAM_CORNER_PREF_NONE)? 10000 : vSize2(p1 - prev_point);
-        const float corner_angle = LinearAlg2D::getAngleLeft(p0, p1, p2) / M_PI; // 0 -> 2
+        float corner_angle = LinearAlg2D::getAngleLeft(p0, p1, p2) / M_PI; // 0 -> 2
+        if (std::abs(corner_angle - 1) < 0.05)
+        {
+            // ignore slight kinks in almost straight lines
+            corner_angle = 1;
+        }
         float corner_shift;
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
I don't think this is something that we want. This is precisely the reason why Smart Hiding was made. Instead, set Seam Corner Preference to Smart Hiding and it'll do the same thing. A bit stronger of an effect, too. The point is to not put the seam on a flat surface, not to fight individual cases like what's being done here.

smartavionics:
You don't understand the issue. Using smart hiding doesn't help at all here. It will just chose the concave corner that is only present due to arithmetic inexactness (or whatever). Here you can see that even with smart hiding in use, the z-seam still appears along the "straight" edge.

Screenshot_2019-09-03_11-25-46

Ghostkeeper:
Then that looks like a bug with the Smart Hiding feature, no?

smartavionics:
> Then that looks like a bug with the Smart Hiding feature, no?

Perhaps, I have not looked at the smart hiding implementation and I never use it. Sorry.

smartavionics:
Perhaps the smart hiding code should look more like this:
```
            case EZSeamCornerPrefType::Z_SEAM_CORNER_PREF_WEIGHTED:
                //More curve is better score (reduced distance), but slightly in favour of concave curves.
                if (corner_angle < 1)
                {
                    dist_score -= (1 - corner_angle) * corner_shift;
                }
                else
                {
                    dist_score -= 1.1f * (corner_angle - 1) * corner_shift;
                }
                break;
```

That would work for the OP's model.

Ghostkeeper:
I had made an alternative implementation that we'll evaluate. It needs some care though because it would change existing profiles without a version upgrade.

thanhtacles:
Hi, was this issue ever resolved? I went so far as to artificially introduce a sharp corner into my model and it STILL won't target that corner
image 

Actual model, see the sharp triangle hole 
image

smartavionics:
Some changes were made by UM with regard to this but I can't remember what. Personally, I very rarely use the sharpest corner mode.

thanhtacles:
My deepest apologies smart, but I actually just realized I had a break in this model. The code was in fact working as intended. But thank you so much for your very punctual reply. I appreciate it

Issue #984, opened 2019-01-29
[3.6.0] GCC9 build failures: OpenMP data sharing


hroncok:
**Application Version**
3.6.0

**Platform**
Linux (Fedora 30) with GCC 9.0.1.

**Qt**
not relevant

**PyQt**
not relevant

**Display Driver**
not relevant

**Steps to Reproduce**
build CuraEngine 3.6.0 with gcc 9.

**Actual Results**

```
[ 78%] Building CXX object CMakeFiles/_CuraEngine.dir/src/utils/gettime.cpp.o
/usr/bin/c++  -DARCUS -DVERSION=\"3.5.1\" -I/builddir/build/BUILD/CuraEngine-3.5.1/./stb-e6afb9cbae4064da8c3e69af3ff5c4629579c1d2 -I/usr/include/polyclipping -I/builddir/build/BUILD/CuraEngine-3.5.1  -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -fPIC -O2 -DNDEBUG -Wall -fopenmp   -std=gnu++11 -o CMakeFiles/_CuraEngine.dir/src/utils/gettime.cpp.o -c /builddir/build/BUILD/CuraEngine-3.5.1/src/utils/gettime.cpp
BUILDSTDERR: /builddir/build/BUILD/CuraEngine-3.5.1/src/layerPart.cpp: In function 'void cura::createLayerParts(cura::SliceMeshStorage&, cura::Slicer*, bool, bool)':
BUILDSTDERR: /builddir/build/BUILD/CuraEngine-3.5.1/src/layerPart.cpp:50:46: error: 'total_layers' not specified in enclosing 'parallel'
BUILDSTDERR:    50 |     for (unsigned int layer_nr = 0; layer_nr < total_layers; layer_nr++)
BUILDSTDERR:       |                                     ~~~~~~~~~^~~~~~~~~~~~~~
[...]

PR #1124, opened 2019-08-30
Fix compilation with clang80/90

kevans91:
clang 8.0 and later, at the very least, will not allow va_start inside a
critical section like this, otherwise it will issue the following:

error: 'va_start' cannot be used in a captured statement

Move vararg manipulation out of the critical sections to fix this, as there
is little value in keeping them inside when the critical section is
primarily to synchronize output.

Mentioned in #984 but not the primary issue; rediscovered on FreeBSD.

### DIFF
--- a/src/utils/logoutput.cpp
+++ b/src/utils/logoutput.cpp
@@ -24,72 +24,72 @@ void enableProgressLogging()
 
 void logError(const char* fmt, ...)
 {
+    va_list args;
+    va_start(args, fmt);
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Thanks. It still compiles for me and the log entries still work. So if this fixes things for you, fine with me. And indeed the varargs stuff can be taken out of the critical section.

kevans91:
Thanks!

PR #1136, opened 2019-09-18
Implement variable width spiralized walls.

smartavionics:
Traditionally, models used for spiralization have been solid and a fixed width wall has
been generated that follows the model's outside profile. Now, the models can be hollow and
the spiralized wall line width will vary within limits set by spiralize_min_line_width and
spiralize_max_line_width which are specified as percentages of the nominal wall line width.


Ghostkeeper:
What will it adjust the line width to? Why would you want the walls to be thicker in some places?

Ghostkeeper:
Ah, nevermind. There is more information in the front-end PRs: https://github.com/Ultimaker/Cura/pull/6371

src/FffGcodeWriter.cpp
```diff
@@ -1602,8 +1602,113 @@ void FffGcodeWriter::processSpiralizedWall(const SliceDataStorage& storage, Laye
...
+        {
+            const Point& this_point = wall_outline[n];
+            const Point& prev_point = wall_outline[(n + wall_outline.size() - 1) % wall_outline.size()];
+            const Point& next_point = wall_outline[(n + 1) % wall_outline.size()];
+
+            Point bisector;
+            double abs_sine = 1;
+            coord_t line_width = default_line_width;
+
+            get_line_width(prev_point, this_point, next_point, line_width, bisector, abs_sine);
```
nallath: I think it's a bit cleaner to use a return statement (instead of an output parameter, since they are harder to understand what is being modified and what isn't.

src/FffGcodeWriter.cpp
```diff
@@ -1602,8 +1602,113 @@ void FffGcodeWriter::processSpiralizedWall(const SliceDataStorage& storage, Laye
...
+    std::vector<Point> shifts(wall_outline.size()); // per vertex shift to compensate for line width changes
+
+    const coord_t default_line_width = mesh.settings.get<coord_t>("wall_line_width_0");
+    const double min_line_width = mesh.settings.get<Ratio>("spiralize_min_line_width");
+    const double max_line_width = mesh.settings.get<Ratio>("spiralize_max_line_width");
+
+    if (min_line_width != 1.0 || max_line_width != 1.0)
+    {
+        // wall line width can vary
+
```
nallath: Why not use the already existing wall overlap compensation? That would save some re-implementation.

src/FffGcodeWriter.cpp
```diff
@@ -1602,8 +1602,113 @@ void FffGcodeWriter::processSpiralizedWall(const SliceDataStorage& storage, Laye
...
+        auto get_line_width = [&](const Point& prev_point, const Point& this_point, const Point& next_point, coord_t& line_width, Point& bisector, double& abs_sine)
+        {
+            const double corner_rads = LinearAlg2D::getAngleLeft(prev_point, this_point, next_point);
+
+            bisector = rotate(normal(next_point - this_point, default_line_width * 5), corner_rads / 2);
+            abs_sine = std::abs(std::sin(corner_rads / 2));
+
+            Polygons lines;
+            lines.addLine(this_point - bisector, this_point + bisector);
+#if 0
```
nallath: Debug code; should probably be removed.

src/FffGcodeWriter.cpp
smartavionics: Its actually producing 3 values: line_width, bisector and abs_sine.

src/FffGcodeWriter.cpp
smartavionics: The existing overlap compensation is horribly broken and should be avoided at all costs. Plus, I don't think it would be applicable here anyway.

src/FffGcodeWriter.cpp
smartavionics: It's very useful to have that but as it offends you I will remove it.

src/FffGcodeWriter.cpp
nallath: In that case, it's not just getting the line_width, so I guess we need to find a better name. A function like "get_line_width" to me feels like it returns a width (since thats what it's naming implies)

src/FffGcodeWriter.cpp
nallath: Horribly broken? What makes you say that? It's something that is done for pretty much every print. 
Might be me mistaken about it, but you're doing rather similar things, but a bit more rough (eg; checking where a line is intersecting with the model and then change its width).

smartavionics: Believe me, it's horribly broken.

Ghostkeeper: I believe you. In that case we should fix this and use the fixed overlap compensation algorithm instead for this feature as well. Of course, it's unmaintainable to just keep adding new implementations every time a bug occurs in one of the existing algorithms.

smartavionics: Arguably, the overlap compensation needs to be completely junked. Cura needs a new strategy for printing walls of variable widths. Just doing a sticking plaster job on the overlap compensation is not worth the effort.

Look, I don't care whether you use this PR or not but I'm not going to rewrite it to make use of the overlap compensation.

ninovanhooff:
Hi there, passing by an open PR. Seems to me like this change could be beneficial but want it to be implemented elsewhere. Closing. Feel free to re-open in case this is refactored.

smartavionics:
I don't think you people understand this stuff at all. I will keep this in my builds, your loss.

Ghostkeeper:
Sorry man, but adding re-implementation upon re-implementation is not a good way to maintain software over the course of multiple years. We'd rather fix or replace the old implementation than to have multiple implementations around where some parts use one implementation and some parts use another.

smartavionics:
Good, as every time someone whines about the terrible print quality due to the bad wall overlap compensation and the horrific gap filling and thin wall printing that Cura does, I tell them that you guys are going to fix it all real soon now with a new implementation of walls done right. I look forward to it.

PR #1138, opened 2019-09-20
Add MESH:NONMESH comment before adding finalizing gcodes to separate them from preceding mesh.

smartavionics:


Ghostkeeper:
Shouldn't this be rather at the end of the mesh plan? This would break if we'd put anything in between.

smartavionics:
I went for the simplest possible solution that solved the user's problem.

> This would break if we'd put anything in between.

Well, in that case you would have to move the NONMESH line to before that additional gcode. Hardly a big deal.

Ghostkeeper:
This is a hotfix, rather than an actual fix. Sorry, it makes the application hard to maintain like this. We need to have proper separation of concern.

It's weird that this would be a bug though. We only write the mesh comments here:

https://github.com/Ultimaker/CuraEngine/blob/673ce81c5bab7231644e4fe96fb51cc7b9cf9a8e/src/FffGcodeWriter.cpp#L1298

(And a few lines higher where it sets the mesh name to the current mesh's name.)

This seems to be a problem with the implementation of the setMesh function rather than with the end g-code function. Perhaps the location where we put the `MESH:` comment in is not appropriate. We currently write it when writing the g-code for a path, but we also have temperature inserts and such that are not normal paths. It indicates with comments there that the author wasn't really sure if that was the best position for it.

PR #1150, opened 2019-10-16
Don't do xor nor union when in surface mode

BagelOrb:
In surface mode we don't want to do Union overlapping volumes, but we also don;t want to perform a XOR on overlapping volumes, which is the implicit alternative when Union Overlapping Volumes is disabled.

Before with Union overlapping volumes:
image

Before without Union overlapping volumes:
image

After this PR without Union overlapping volumes:
image


rburema:
Devs: See internal ticket CURA-7080.

PR #1152, opened 2019-10-22
Simplify getClosestOnLineSegment().

smartavionics:
Now does 1 divide rather than 4 multiplies, 2 square roots and 2 divides.

### DIFF
--- a/src/utils/linearAlg2D.h
+++ b/src/utils/linearAlg2D.h
@@ -119,8 +119,7 @@ public:
         }
         else
         {
-            Point ret = p0 + projected_x / vSize(direction) * direction  / vSize(direction);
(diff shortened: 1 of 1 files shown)


src/utils/linearAlg2D.h
```diff
@@ -119,8 +119,7 @@ class LinearAlg2D
         }
         else
         {
-            Point ret = p0 + projected_x / vSize(direction) * direction  / vSize(direction);
-            return ret;
+            return p0 + projected_x * direction / x_p1;
```
Ghostkeeper: Took me a minute to understand why it's the same! I'll add a bit of documentation here I think...

PR #1154, opened 2019-10-28
I've observed that walls get printed before infill.

allyourcode:
Not sure if this is right, but I'm pretty sure that it is less wrong.


Ghostkeeper:
This is already documented down below the list: There is the "Infill Before Walls" setting which allows you to configure that. Since the setting is enabled by default, I'd decided to document it like that as well. I think that is better.

allyourcode:
The behavior that I've always seen is that walls get printed first. I didn't even know that there was a setting to swap the order. I'm surprised anyone would want that.

Ghostkeeper:
My settings guide has an article about the trade-off made with that setting: https://github.com/Ghostkeeper/SettingsGuide/blob/master/resources/articles/infill_before_walls.md

PR #1155, opened 2019-10-29
Fix Z hops when retracting

Ghostkeeper:
Quite a big change. I hope I've documented everything well.

So currently, Z hops seem to be mostly broken. There are no Z hops to speak of except when switching extruders. Most profile authors seem to have accepted this but it's still printing okay. This change is intended to fix those again.

I've also restructured the travel move function in order to align it more with what is actually happening. This removes a bit of state, though some state still remains in the `bypass_combing` variable. However the big part here is that it's now aligned with what the decision tree is doing. I've documented the decision tree very well, both in the code documentation and with a diagram in the `docs` folder. Please see Github's rich diff in order to read the decision tree diagram.

This change comes with a change to the Cura profiles as well, which disables the Z hops for all profiles. The idea behind this is that it hasn't been working anyway so disabling the Z hop ensures that the behaviour is the same as what it used to be. That change is made in these two repositories:
* Cura: https://github.com/Ultimaker/Cura/pull/6595
* cura-private-data: https://github.com/Ultimaker/cura-private-data/pull/42 (not publicly accessible)

One additional behavioural change is that the "Z Hop for Extruder Switch" setting can now be enabled separately even if the normal Z hop setting is disabled.

Implements CURA-6860.


Ghostkeeper:
Superseded by https://github.com/Ultimaker/CuraEngine/pull/1185.

PR #1162, opened 2019-12-04
Fix issue with retraction_min_travel_distance.

smartavionics:
1 - removed erroneous update of last_planned_position.

2 - added upstream logic (https://github.com/Ultimaker/CuraEngine/pull/1108)

3 - force a new path to be started otherwise paths that have different retraction status can be merged.


Ghostkeeper:
Developers, see issue CURA-6675 (also mentioned in the code; thanks Mark!)

PR #1163, opened 2019-12-04
Fix arithmetic overflow in shorterThen().

smartavionics:
It was failing for lengths > approx 46mm!

### DIFF
--- a/src/utils/IntPoint.h
+++ b/src/utils/IntPoint.h
@@ -85,7 +85,7 @@ INLINE float vSize2f(const Point& p0)
     return float(p0.X)*float(p0.X)+float(p0.Y)*float(p0.Y);
 }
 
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Wow, good find! Could've resulted in some pretty weird and hard-to-trace behaviour.

smartavionics:
> Wow, good find! Could've resulted in some pretty weird and hard-to-trace behaviour.

It did.

I didn't bother to change the spelling mistake (should be shorterThan()) as I think it's rather quaint as it is!

Ghostkeeper:
Yeah the spelling mistake is our gift from @daid and we're very thankful for it :stuck_out_tongue: We never bothered to change it either.

daid:
You should be more surprised that there are so few spelling mistakes :-)

PR #1167, opened 2019-12-11
Add support for Truncated Octahedron fill

spotrh:
This code was forward ported by me from the Lulzbot fork (based on 3.6.0). The original author is Victor Larchenko (@victor9999).

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -1571,7 +1571,7 @@ bool FffGcodeWriter::processSingleLayerInfill(const SliceDataStorage& storage, L
         }
         else
         {
-            gcode_layer.addLinesByOptimizer(infill_lines, mesh_config.infill_config[0], (pattern == EFillMethod::ZIG_ZAG) ? SpaceFillType::PolyLines : SpaceFillType::Lines, enable_travel_optimization
+            gcode_layer.addLinesByOptimizer(infill_lines, mesh_config.infill_config[0], (pattern == EFillMethod::ZIG_ZAG || pattern == EFillMethod::TRUNCATED_OCTAHEDRON) ? SpaceFillType::PolyLines : SpaceFillType::Lines, enable_travel_optimization
                 , /* wipe_dist = */ 0, /*float_ratio = */ 1.0, near_start_location);
         }
     }
--- a/src/infill.cpp
+++ b/src/infill.cpp
@@ -19,6 +19,10 @@
 #include "utils/polygonUtils.h"
 #include "utils/UnionFind.h"
 
+#define SQRT2MUL(x) ((30547*(x))/21600)
+#define OCTSLEN(x) ((43200*(x))/73747)
+#define OCTDLEN(x) ((21600*(x))/30547)
+
 /*!
  * Function which returns the scanline_idx for a given x coordinate
  *
@@ -153,6 +157,10 @@ void Infill::_generate(Polygons& result_polygons, Polygons& result_lines, const 
     case EFillMethod::GYROID:
         generateGyroidInfill(result_lines);
         break;
+    case EFillMethod::TRUNCATED_OCTAHEDRON:
+        generateTroctInfill(result_polygons, fill_angle);
+        generateTroctInfill(result_polygons, fill_angle + 90);
+        break;
     default:
         logError("Fill pattern has unknown value.\n");
         break;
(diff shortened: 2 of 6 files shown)


Ghostkeeper:
There also needs to be a change to the front-end to allow selecting this infill pattern?

spotrh:
> There also needs to be a change to the front-end to allow selecting this infill pattern?

Yes, you're right. I wanted to make sure there were no issues here first, but if you want me to go ahead and open a PR on that side, I will.

Ghostkeeper:
We haven't done a code review yet. But the idea in itself I think is fine, would be nice to add.

Ghostkeeper:
There are a few issues with this infill pattern:
* What would be a reason why a user would activate this pattern? Does it have a certain strength that no other pattern has?
* This pattern has a lot of overextrusion. There are two places where it overextrudes: The diagonal lines as well as the horizontal flats are each drawn twice: once for each of the orthogonal orientations. At least for the horizontal flats we should omit that from one of the directions. The other one may not be so easy to omit. We may need to make travel moves in there but that's not nice for getting a consistent flow, so it's probably best to put the two lines adjacent to each other and only have them overlap in a single point rather than the length of a line.
* The infill density should still be adjusted. Since the pattern has two sets of lines, which are (I assume) spaced on average by the Infill Line Distance, you're going to get a similar density to Grid. However there are also the completely filled flats, which will bring the density up. Maybe we can choose to ignore that, but that'll also need to be filtered out in our [function test](https://github.com/Ultimaker/CuraEngine/blob/e904e260716d072d6c766dc73fbc18f7fb224fce/tests/InfillTest.cpp#L28-L41) for that property.

BagelOrb:
I guess the maker of this PR is not going to fix it seeing as he just ported the code from Cura LulzBot edition.

I see no real mechanical value of using such a honeycomb pattern. I suggest we close this PR and not merge the infill pattern.

spotrh:
Apologies for the delay in response. The person who would be the best qualified to answer the questions posed would be the original author, @gringer. 

I found the original pull request, which goes into more detail as to why this infill was created:
https://github.com/alephobjects/CuraEngine/pull/2

My purpose in opening this PR was to ensure that no useful code was lost with the demise of cura-lulzbot. If you (the maintainers of Cura) feel that this is not useful code, then so be it.

BagelOrb:
See also https://github.com/Ultimaker/CuraEngine/pull/491

BagelOrb:
We already reviewed the code once. There were some issues with the code..

BagelOrb:
We should at least check what's in https://github.com/alephobjects/CuraEngine and not in our master.

At some point they switched from github to some privately hosted thing. Too bad we don't have the latest version of their repo, now that their website is down.

gringer:
I haven't worked on this code for Cura because I couldn't get compiled code installing properly on my Debian system; it was just too much effort for something that no one other than me was interested in using (from what I can tell from Youtube videos).

Now that there's a gyroid infill, it would be better to slightly modify *that* than to try to modify the existing broken truncated octahedron infill to work better. I updated the pattern for Slic3r to change direction only at the square bits on the top (where the gap distance is smallest), and that seems to work a lot better:

https://twitter.com/gringene_bio/status/879212511836360704

Gyroid is essentially the same pattern, it just uses a sine wave (or possibly semicircular wave) instead of the trapezoid wave used for the truncated octahedron infill.

gringer:
With regards to the comments / questions:
> What would be a reason why a user would activate this pattern?

As far as I'm aware, the truncated octahedron is the only non-planar self-tessellating 3D space-filling polyhedron, which makes it mathematically interesting.

> Does it have a certain strength that no other pattern has?

It is near-optimal in terms of the amount of face/edge material required to encapsulate a given volume:

https://en.wikipedia.org/wiki/Bitruncated_cubic_honeycomb

I expect that the hexagonal and triangular shapes it is composed of gives it reasonable strength in most directions. Because it's triangles and straight lines, it might even be stronger than gyroid, but I haven't done any comparison tests on that.

BagelOrb:
Interesting point that it is the only tesselating regular polyhedron. I didn't realize that yet.

We do provide an infill pattern composed of octahedra and tetrahedra, called Octet.

The ratio of volume w.r.t. surface area and edge length is interesting, but I don't think it can be related to mechanical properties in that way. I think the mechanical properties depend on other geometric features.

> it might even be stronger than gyroid

Strength is not unique material property. The gyroid is typically not stiff, but conversely it might have a relatively high yield strength. It also has a rather linear stress-strain curve and a small Poissons ratio. I wouldn't expect this honeycomb to exceed the gyroid in any of these properties, but that is based on total guesswork.

It could definitely be an interesting pattern, but without the mechanical tests or simulations which shows it has extraordinary properties I am hesitant to accept this PR.

Ghostkeeper:
Yeah, given the current flaws in the implementation and the age of the code, and given the shaky use case for this (other than mathematical interest) we'll close this PR and not merge this. Sorry!

PR #1169, opened 2019-12-24
CURA-6571 / CURA-6724 Fix support (interface) holes

rburema:


smartavionics:
Hi, before you merge this, please consider https://github.com/Ultimaker/CuraEngine/pull/1156 first. Thanks.

ninovanhooff:
I checked both approaches and think this this one is more clean. I'd prefere a second pair of eyes on this to confirm.

smartavionics:
I don't really know what you mean by "more clean". Note that this PR doesn't do the right thing if a polygon that contains holes is removed and the holes are not removed. Admittedly, that's not very likely to happen but why not get it right?

rburema:
@ninovanhooff I know mine _looks_ a bit smaller (and thus more maintainable) on first glance, but I don't think I can solve the problem that @smartavionics indicates without the verbosity in roughly the same ballpark.

However, before we merge that one: Though I don't forsee a problem (because it mostly leaves any non-buggy results alone) we should still test for slowdowns/memory-increase in @smartavionics code, just to be sure.

(The main problem here is that polygons aren't guaranteed to be hierarchical in clipperlib, I think. You just have 'paths' all over the place. But  that's a whole other thing.)

P.S. If we want to be _completely_ sure, we could also merge this first, then apply the older one on top after. But that might be overkill given the size of the solution(s).

smartavionics:
BTW, another difference between these PRs is that this one defaults to removing the holes (i.e. the same behaviour as before) but my PR defaults to not removing the holes because that fixes the issue that I was trying to solve (holes in infill being removed when they should not have been).  I now think that both of our PRs would benefit by removing the default value for that argument so that all calls to removeSmallAreas() would need to explicitly specify whether they require holes removing or not.

rburema:
> ). I now think that both of our PRs would benefit by removing the default value for that argument so that all calls to removeSmallAreas() would need to explicitly specify whether they require holes removing or not.

Good point.

smartavionics:
Having reviewed all the usages of that method, I agree that a default of false (as used in my PR) is appropriate. BTW, Happy New Year!

rburema:
(I was waiting for the merge to wish you this, but it's the 3rd already, so...)
Have a great 2020 as well!

nallath:
Closing this in favor of https://github.com/Ultimaker/CuraEngine/pull/1156

PR #1172, opened 2020-01-04
Implement fuzzy skin outside only option.

smartavionics:
As discussed in https://github.com/Ultimaker/Cura/issues/2550.


fieldOfView:
It looks... so simple!

smartavionics:
> It looks... so simple!

Occam's razor and all that jazz. It appears to satisfy the people asking for it so may as well submit it for consideration.

Ghostkeeper:
Developers, see issue CURA-7158.

Ghostkeeper:
Works well! Thanks for the feature.

PR #1174, opened 2020-01-05
Optimized the slicing part

marcus-bigrep:
changed the main loop from faces/layers to layers/faces
added some prechecks
parallization

### DIFF
--- a/src/slicer.cpp
+++ b/src/slicer.cpp
@@ -787,72 +787,59 @@ void SlicerLayer::makePolygons(const Mesh* mesh)
     polygons.removeDegenerateVerts(); // remove verts connected to overlapping line segments
 }
 
-Slicer::Slicer(Mesh* mesh, const coord_t thickness, const size_t slice_layer_count, bool use_variable_layer_heights, std::vector<AdaptiveLayer>* adaptive_layers)
-: mesh(mesh)
+Slicer::Slicer(Mesh* mesh, const coord_t thickness, const size_t slice_layer_count, 
+               bool use_variable_layer_heights, std::vector<AdaptiveLayer>* adaptive_layers)
 {
     const SlicingTolerance slicing_tolerance = mesh->settings.get<SlicingTolerance>("slicing_tolerance");
+    const coord_t initial_layer_thickness = 
+        Application::getInstance().current_slice->scene.current_mesh_group->settings.get<coord_t>("layer_height_0");
 
     assert(slice_layer_count > 0);
 
     TimeKeeper slice_timer;
 
-    layers.resize(slice_layer_count);
+    layers = 
+        buildLayersWithHeight(slice_layer_count, slicing_tolerance, initial_layer_thickness, thickness, 
+            use_variable_layer_heights, adaptive_layers);
 
-    // set (and initialize compensation for) initial layer, depending on slicing mode
-    const coord_t initial_layer_thickness = Application::getInstance().current_slice->scene.current_mesh_group->settings.get<coord_t>("layer_height_0");
-    layers[0].z = std::max(0LL, initial_layer_thickness - thickness);
-    coord_t adjusted_layer_offset = initial_layer_thickness;
-    if (use_variable_layer_heights)
-    {
-        layers[0].z = adaptive_layers->at(0).z_position;
-    }
-    else if (slicing_tolerance == SlicingTolerance::MIDDLE)
-    {
-        layers[0].z = initial_layer_thickness / 2;
-        adjusted_layer_offset = initial_layer_thickness + (thickness / 2);
-    }
 
-    // define all layer z positions (depending on slicing mode, see above)
-    for (unsigned int layer_nr = 1; layer_nr < slice_layer_count; layer_nr++)
+    std::vector<std::pair<int32_t, int32_t>> zbbox = buildZHeightsForFaces(*mesh);
+
+
+    buildSegments(*mesh, zbbox, layers);
+
+    log("slice of mesh took %.3f seconds\n", slice_timer.restart());
+
+    makePolygons(*mesh, slicing_tolerance, layers);
+    log("slice make polygons took %.3f seconds\n", slice_timer.restart());
+}
+
+void Slicer::buildSegments(const Mesh &mesh, const std::vector<std::pair<int32_t, int32_t>> &zbbox,
+    std::vector<SlicerLayer>& layers) 
+{
+    // OpenMP
+#pragma omp parallel for default(none) shared(mesh, zbbox, layers)
+    for (int layer_nr = 0; layer_nr < layers.size(); layer_nr++)
     {
-        if (use_variable_layer_heights)
+        int32_t z = layers[layer_nr].z;
+        layers[layer_nr].segments.reserve(100);
+
+        // loop over all mesh faces
+        for (unsigned int mesh_idx = 0; mesh_idx < mesh.faces.size(); mesh_idx++)
         {
-            layers[layer_nr].z = adaptive_layers->at(layer_nr).z_position;
-        }
-        else
-        {
-            layers[layer_nr].z = adjusted_layer_offset + (thickness * (layer_nr - 1));
-        }
-    }
+            if ((z < zbbox[mesh_idx].first) || (z > zbbox[mesh_idx].second))
+                 continue;
 
-    // loop over all mesh faces
-    for (unsigned int mesh_idx = 0; mesh_idx < mesh->faces.size(); mesh_idx++)
-    {
-        // get all vertices per face
-        const MeshFace& face = mesh->faces[mesh_idx];
-        const MeshVertex& v0 = mesh->vertices[face.vertex_index[0]];
-        const MeshVertex& v1 = mesh->vertices[face.vertex_index[1]];
-        const MeshVertex& v2 = mesh->vertices[face.vertex_index[2]];
+            // get all vertices per face
+            const MeshFace& face = mesh.faces[mesh_idx];
+            const MeshVertex& v0 = mesh.vertices[face.vertex_index[0]];
+            const MeshVertex& v1 = mesh.vertices[face.vertex_index[1]];
+            const MeshVertex& v2 = mesh.vertices[face.vertex_index[2]];
 
-        // get all vertices represented as 3D point
-        Point3 p0 = v0.p;
-        Point3 p1 = v1.p;
-        Point3 p2 = v2.p;
-
-        // find the minimum and maximum z point
-        int32_t minZ = p0.z;
-        int32_t maxZ = p0.z;
-        if (p1.z < minZ) minZ = p1.z;
-        if (p2.z < minZ) minZ = p2.z;
-        if (p1.z > maxZ) maxZ = p1.z;
-        if (p2.z > maxZ) maxZ = p2.z;
-
-        // calculate all intersections between a layer plane and a triangle
-        for (unsigned int layer_nr = 0; layer_nr < layers.size(); layer_nr++)
-        {
-            int32_t z = layers.at(layer_nr).z;
-
-            if (z < minZ) continue;
+            // get all vertices represented as 3D point
+            Point3 p0 = v0.p;
+            Point3 p1 = v1.p;
+            Point3 p2 = v2.p;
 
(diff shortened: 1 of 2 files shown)


Ghostkeeper:
We've made the main loop explicitly as faces -> layers as an optimisation. One face spans multiple layers, and it's clear which layers it spans; just look at the Z coordinates of all its vertices. So you essentially go through all faces once (which are many) and for each face you iterate through all its layers, which are often like 2 or 3 (or even 0 layers if the triangle is horizontal). The other way around you'd need to go through all layers once and for each layer go through all faces, which is much more work.

The other changes sound good to me, although parallelisation is maybe not desired since the current solution really only takes a fraction of a second for any normal print. Prechecks are nice of course and reducing pointer look-up should give a bit of a performance boost.

Do you have performance measurements to back up your claim that it's been improved?

I haven't done an actual code review yet, mind.

marcus-bigrep:
For the samples, which have a significant percentage of slicing, I have (new/old):
- Just slicing part: 0.37
- Entire run: ‭0.91

marcus-bigrep:
Sorry, I was referring to my own samples. We have a pretty big printer (1 cubic meter).
10% of my samples have a significant percentage in the slicing part. 
These 10% of samples have about a 10% reduction in entire slicing time.
(BTW, my average is 3.5% instead of 1.85% which is yours).

src/slicer.cpp
```diff
@@ -787,72 +787,59 @@ void SlicerLayer::makePolygons(const Mesh* mesh)
...
+}
 
-        // calculate all intersections between a layer plane and a triangle
-        for (unsigned int layer_nr = 0; layer_nr < layers.size(); layer_nr++)
+void Slicer::buildSegments(const Mesh &mesh, const std::vector<std::pair<int32_t, int32_t>> &zbbox,
+    std::vector<SlicerLayer>& layers) 
+{
+    // OpenMP
+#pragma omp parallel for default(none) shared(mesh, zbbox, layers)
+    for (int layer_nr = 0; layer_nr < layers.size(); layer_nr++)
```
Ghostkeeper: Comparing integers of different signedness (`Wsign-compare` in GCC).

src/slicer.cpp
```diff
@@ -787,72 +787,59 @@ void SlicerLayer::makePolygons(const Mesh* mesh)
...
+    {
+        int32_t z = layers[layer_nr].z;
+        layers[layer_nr].segments.reserve(100);
+
+        // loop over all mesh faces
+        for (unsigned int mesh_idx = 0; mesh_idx < mesh.faces.size(); mesh_idx++)
         {
-            int32_t z = layers.at(layer_nr).z;
+            if ((z < zbbox[mesh_idx].first) || (z > zbbox[mesh_idx].second))
+                 continue;
```
Ghostkeeper: So you've now reversed this for loop from `faces -> layers` to `layers -> faces`. This requires you to build this `zbbox` mapping though in order to limit the layers that it's iterating over for each face. I don't see the point of this change; when the original loop nesting was used, it could keep the minimum and maximum Z coordinates local in the loop so it doesn't need to track them in a separate mapping.

src/slicer.cpp
Ghostkeeper: I think the best option here is a `LayerIndex` because it's very semantic, but then you'd need a cast to tell the compiler that you're working with layer indices here because that vector doesn't return `LayerInd`ices.

src/slicer.cpp
```diff
@@ -962,27 +987,50 @@ Slicer::Slicer(Mesh* mesh, const coord_t thickness, const size_t slice_layer_cou
...
+        Point3 p2 = v2.p;
+
+        // find the minimum and maximum z point		
+        int32_t minZ = p0.z;
+        if (p1.z < minZ) minZ = p1.z;
+        if (p2.z < minZ) minZ = p2.z;
+
+        int32_t maxZ = p0.z;
+        if (p1.z > maxZ) maxZ = p1.z;
+        if (p2.z > maxZ) maxZ = p2.z;
```
Ghostkeeper: Code style: Brackets on new line.

src/slicer.cpp
```diff
@@ -962,27 +987,50 @@ Slicer::Slicer(Mesh* mesh, const coord_t thickness, const size_t slice_layer_cou
...
+
+        zHeights.emplace_back(std::make_pair(minZ, maxZ));
+    }
+
+    return zHeights;
 }
 
-SlicerSegment Slicer::project2D(const Point3& p0, const Point3& p1, const Point3& p2, const coord_t z) const
```
Ghostkeeper: Why did you remove this `const`?

src/slicer.h
```diff
@@ -504,7 +508,13 @@ class Slicer
...
      * \return The Y coordinate of the point to find.
      */
-    coord_t interpolate(const coord_t x, const coord_t x0, const coord_t x1, const coord_t y0, const coord_t y1) const;
+    static coord_t interpolate(const coord_t x, const coord_t x0, const coord_t x1, const coord_t y0, const coord_t y1)
+    {
+        const coord_t dx_01 = x1 - x0;
+        coord_t num = (y1 - y0) * (x - x0);
+        num += num > 0 ? dx_01 / 2 : -dx_01 / 2; // add in offset to round result
+        return y0 + num / dx_01;
+    }
```
Ghostkeeper: Code style: No implementation in headers.

src/slicer.h
```diff
@@ -518,9 +528,23 @@ class Slicer
...
+
+    /*! Create the polygons in layers. */
+    static void makePolygons(Mesh &mesh, SlicingTolerance slicing_tolerance, std::vector<SlicerLayer>& layers);
+
+    /* Creates a vector of layers and set their z value. */
+    static std::vector<SlicerLayer> buildLayersWithHeight(size_t slice_layer_count, SlicingTolerance slicing_tolerance,
+        coord_t initial_layer_thickness, coord_t thickness, bool use_variable_layer_heights,
+        const std::vector<AdaptiveLayer> *adaptive_layers);
+
+    /* Creates the segments and write them into the layers. */
```
Ghostkeeper: Could you write these documentation entries in Doxygen's style?

src/slicer.cpp
marcus-bigrep: Enabling parallelization is the main point about reversing the loop.

src/slicer.cpp
marcus-bigrep: Do you a style guide where I can look into?

src/slicer.cpp
marcus-bigrep: I replaced "const" by "static". This is slightly more efficient.

src/slicer.cpp
marcus-bigrep: I am afraid, OpenMP is not compatible with LayerIndex (at lease for MSVC).
BTW, what do you think about the new standard "std::for_each(std::execution::par,...)"?"

src/slicer.cpp
marcus-bigrep: Fixed

src/slicer.h
marcus-bigrep: fixed

src/slicer.cpp
Ghostkeeper: Why couldn't the parallelisation work over the original faces loop, then?

src/slicer.cpp
Ghostkeeper: The code style guidelines are here: https://github.com/Ultimaker/Meta

src/slicer.cpp
Ghostkeeper: We compile CuraEngine with GCC though, not MSVC. Nice to hear though that MSVC is still working (for the rest then). Alternatively you can use int and cast the size to an int, of course. I can't guarantee that we won't give a for loop a LayerIndex in the future though, since we don't test (or support) MSVC.

I think that `std::for_each` with parallel execution is a very nice alternative, if it works for us. It'd also work with MacOS, which is as of now still single-threaded. I hope it works fine together with OpenMP so that it can be transitioned softly rather than having to make a hard atomic transition.
Currently we're still compiling with C++11 though. Hopefully soon that will change to C++17.

src/slicer.cpp
marcus-bigrep: Because you need write access on the layers. Thus it does not make much sense to parallize the faces loop. Because then the layer part must be "critical". 
In particular, the layer loop shall be "heavy" to boost up parallisation. Thus I have reversed the loops. Otherwise you parallize just over a few layers.

src/slicer.cpp
marcus-bigrep: Apart from GCodeLayerThreader, CuraEngine parallization is just about loops and 
these loops are executed "sequentially". So the migration ought to be quite smooth. Maybe even GCodeLayerThreader could be converted to a simple loop...
Is there a specific time line to upgrade to C++17?

Ghostkeeper: Not really since we don't generally work with timelines and promises to fix something before a certain deadline. It was recently brought up again and moved up a bit in our backlog. Most likely we'd plan this in after the 4.5 branch has split off to reduce risk for that release (though personally I won't think there's any risk at all).

marcus-bigrep: Thanks. I am looking forward!

marcus-bigrep:
Do you need more information?

Ghostkeeper:
No, sorry, other things got in between for a bit. We'll review the latest changes again.

src/slicer.cpp
Ghostkeeper: Ah yes, I see. That's a good reason.

Ghostkeeper:
The slice result is still the same as far as I've found. I also found a performance improvement of 78% on average of the slicing stage. Good job!

jdumas:
Hi. I was reading the [documentation](https://github.com/Ultimaker/CuraEngine/blob/master/docs/slicing.md) that says you iterate over faces/layers, and was surprised to find the implementation iterates over layers/faces instead. Have you tried slicing *really* large meshes (+10M faces at least)? Unless you have a lot of cores, a serial implementation of the loop over faces/layers should be faster still. Even then, it is possible to make it parallel using `tbb::concurrent_vector<>` for multi-threaded segment insertions. I found that I can slice a mesh with 67M faces in about 3s on a laptop using that approach, which is really fast.

marcus-bigrep:
@jdumas Thanks for the remark. I'll adjust the docu.
I did benchmarks with a lot of big stls and it is much faster. The timings have been discussed in this pull request in detail. In particular, we are more than 2x faster for the slicing part (actually a lot more because most time is consumed in the stitching now which is part of the duration measurement).
We build a precheck structure which we test against. This is our way to integrate the optimization idea of the original loop order.
tbb::concurrent_vector<> : I have not tested this data structure here but I worked with it before and it has some overhead. I am pretty sure that our approach is better. BTW, some discussions about the ideas are here as well: https://github.com/Ultimaker/CuraEngine/pull/1174#discussion_r368581986

jdumas:
Well, I am not convinced that this approach is faster, but I would be willing to do some comparisons when I have some time. Could you point me out to the relevant performance tests that you have run so that I may try it out? It may take me some time before I get down to it, but it'd be interesting to compare.

marcus-bigrep:
Do it!
I am afraid, but "my" stls are under NDA and I cannot share them. Though I guess that they are not special in terms of this algorithm. So I suggest: try something out with your stls and if numbers are good, I can run it on my stls and share the numbers.

Ghostkeeper:
I've been trying this with rather simple shapes generated using OpenSCAD. In particular these two from my folder of stress tests:

* https://github.com/Ghostkeeper/SlicerTestModels/blob/master/stress_test/cylinder_hi.scad
* https://github.com/Ghostkeeper/SlicerTestModels/blob/master/stress_test/sphere.scad

And this one where I manually increased the resolution:

* https://github.com/Ghostkeeper/SlicerTestModels/blob/master/polytest/support_strength_infill.scad

Be careful with these though. They can easily generate files upwards of 2GB. Especially the last one has a tendency to crash OpenSCAD.

Ghostkeeper:
I've merged Marcus' updates to the docs here: https://github.com/Ultimaker/CuraEngine/pull/1212

Hopefully that makes it more clear.

PR #1175, opened 2020-01-06
Small fix for sequential printing

marcus-bigrep:


Ghostkeeper:
What does this fix? Is there a crash? How does that bug reproduce?

marcus-bigrep:
[UMS5_sequential_bug.zip](https://github.com/Ultimaker/CuraEngine/files/4060174/UMS5_sequential_bug.zip)

this crashes with the cura 4.4.1

Ghostkeeper:
Ah, yeah, I see it happening. Good find! The fix makes sense then, since your model contains one mesh group with multiple meshes in it.

PR #1177, opened 2020-01-07
4.4 - improved prime tower generation, no extra tool changes, no missing prime tower bands

mkudzia84:
Change on how the prime tower is generated in order to avoid skipping of printing of prime tower bands of unused extruders. Does not increase the number of tool changes while guaranteeing the prime  tower will be coherent and no printing 'in the air'

New strategy is :
- on tool change, slicer checks if previous extruder primed into the tower; if not add gcode to prime the tower with prevExtruder prior to the tool change (new behaviour)
- after the tool change, prime into the tower (current behaviour)
- at the end of layer processing, check which extruders haven't been primed in this layer; for each extruder that hasn't been primed, generate gcode for priming using currently active extruder

In this scenario - in example, 
```
layer #10, T0, T1 is active, T2, T3 is not active
- layer starts with T1 printing (T1 was last extruder in layer #9)
- T1->T0 is triggered
- [new behaviour] check if T1 primed in layer #10, it hasn't so generate gcode for priming into T1 prime tower band
- change T1->T0
- [current behaviour] check if T0 is primed, is not so generate gcode for priming into T0 prime tower band
- print with T0
...
...
- [new behaviour] end of layer processing, check which extruders haven't been primed -> [T2, T3] - for T2, T3, generate gcode for printing into the T2, T3 tower prime bands using current extruder settings/curren extruder [T0]
```
This guarantees all of the requested prime tower bands will be printed
No new tool changes are added

I have tested this in various configurations for my E3D toolchanger equiped with 4 toolheads - all was ok.

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -966,6 +966,7 @@ LayerPlan& FffGcodeWriter::processLayer(const SliceDataStorage& storage, LayerIn
     { // add prime tower if it hasn't already been added
         int prev_extruder = gcode_layer.getExtruder(); // most likely the same extruder as we are extruding with now
         addPrimeTower(storage, gcode_layer, prev_extruder);
+		addPrimeTowerFillTheGap(storage, gcode_layer);
     }
 
     if (!disable_path_optimisation)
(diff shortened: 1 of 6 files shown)


Ghostkeeper:
The problem that this PR is trying to solve are not a bug. See the discussion over in the linked ticket: https://github.com/Ultimaker/Cura/issues/6909

Ghostkeeper:
Could you add a boolean setting to fdmprinter.def.json that enables/disables this change? It doesn't apply to all materials and we'd like to keep the behaviour as much the same as possible.

Secondly, please merge this into master rather than 4.4. We probably won't release 4.4 any more.

wawanbreton:
Hi, any news on this PR ? It seems to be abandoned.. However I'd like to see it integrated, as I like to use dual print :) @mkudzia84 would you like me to keep working on this ?

PR #1178, opened 2020-01-09
Added options to only fill wall gaps and wall gaps and skin gaps but not infill gaps.

smartavionics:
Engine PR for https://github.com/Ultimaker/Cura/pull/6944.


smartavionics:
I goofed, this doesn't work right (yet).

PR #1187, opened 2020-01-22
Wall Collapse

BagelOrb:
When a 3D model has thin/narrow parts the walls might not fit exactly. When the model is just wider than an even number of walls there is a space left: underfill; when the model is narrower than an even number of walls there is overfill.

What Cura currently does is to fill the underfill areas using some dense lines pattern, which is then combined into a single extrusion line with adaptive width. (Note that this combining algorithm fails quite often, currently!)
What Cura currently does to deal with the overfill is to compensate the overlap and extrude less on the second pass.

However, the compensation approach is limited. If the overlap gets above 75% of the line width, the second pass is reduces to less than 25% flow, which makes the extrusion unreliable.

Instead we can simply avoid such overfill, by performing a mathematical closure operation on the walls. If walls are too close to each other they collapse and are removed. The resulting gap is then handled as overfill and is filled using the Gaps Between Walls settings.

This PR makes the above possible. I would recommend setting the default value to something like half the line width, were it not for the buggy perimeter gap combine algorithm, which currently produces quite horrible output. See images below.

If you set the collapse distance to the line width, ideally the (currently buggy) combining code should produce an extrusion line up to twice the normal width, which is quite wide. Therefore setting it to 50% would be better, since it would only require lines of 150% the normal size. 

Because the current perimeter gap combining code is so buggy I left the default at 0, which means the default Cura behavior remains unchanged: The inter-perimeter gaps are filled using lines up to 100% the wall line width.

set to zero: (normal)
image

Set to half line width:
image

set to line width:
image

### DIFF
--- a/src/WallsComputation.cpp
+++ b/src/WallsComputation.cpp
@@ -40,6 +40,9 @@ void WallsComputation::generateInsets(SliceLayerPart* part)
         part->print_outline = part->outline;
         return;
     }
+    
+    const coord_t wall_collapse_0 = settings.get<coord_t>("wall_collapse_0");
+    const coord_t wall_collapse_x = settings.get<coord_t>("wall_collapse_x");
 
     const coord_t wall_0_inset = settings.get<coord_t>("wall_0_inset");
     coord_t line_width_0 = settings.get<coord_t>("wall_line_width_0");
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
We currently have a setting called Minimum Wall Flow for this problem. Do you think this approach is better than Minimum Wall Flow?

BagelOrb:
Yes. Imagine a model which is 0.401 mm thick. It would be better to lay down a single line than to go over it twice and enforce the second time to not extrude. The minimum wall flow introduces unneeded travels.

Ghostkeeper:
Instead, with this feature it would not print the wall at all though.

BagelOrb:
That's because print thin walls is not enabled.

Ghostkeeper:
We're going to reject this pull request. I'm sorry. Minimum Wall Flow indeed introduces unneeded travels, but so does Print Thin Walls and Fill Gaps Between Walls to then print the part of the wall that got missing. Maybe Print Thin Walls will be more efficient with the travel moves since it can optimize the thin parts all together, but maybe not because it does a second pass over everything then.

But the real reason is that the use case of this feature would get quickly superseded by [Arachne](https://github.com/Ultimaker/CuraEngine/pull/1210), a result of us being so slow to process these... relative to the effort put into the Arachne algorithms.

BagelOrb:
Yeah exactly. It's superseded by Arachne.

Still has quite a way to go, though.

PR #1190, opened 2020-01-29
Simplify each inset when generating concentric infill.

smartavionics:
When generating concentric infill, simplify each inset to stop glitches propogating. 

Fixes https://github.com/Ultimaker/Cura/issues/7040.

### DIFF
--- a/src/infill.cpp
+++ b/src/infill.cpp
@@ -278,6 +278,7 @@ void Infill::generateConcentricInfill(Polygons& first_concentric_wall, Polygons&
     while (prev_inset->size() > 0)
     {
         new_inset = prev_inset->offset(-inset_value);
+        new_inset.simplify();
         result.add(new_inset);
         if (perimeter_gaps)
         {


rburema:
Also cherry-picked it to the new 4.5 branch.

BagelOrb:
I just saw this pop up in the commits.

The way it is currently implemented means we accumulate the errors introduced by simplify inward. It might be better to calculate the offsets accurately and then apply the simplification afterwards.

It's a bit of a discussion-point though. Cura used to calculate each offset as offsets from the outline in order to approximate the circular parts in the offsets better. However, at some point that changed due to the reasoning that accumulating the errors means that lines fit better to *each other*. I am now inclined to think the former was better, as the actual path of the print head will be more rounded than the computed path; the actual path is actually closer to the unsimplified path even though we have simplified it.

smartavionics:
> The way it is currently implemented means we accumulate the errors introduced by simplify inward. It might be better to calculate the offsets accurately and then apply the simplification afterwards.

It would probably help if you were not using the buggy simplify. See https://github.com/Ultimaker/CuraEngine/pull/1114 for fixes for a couple of problems with simplify().

BagelOrb:
Sure, but simplify should introduce geometrical errors by definition. In the commit from this PR those geometrical errors are propagated to further insets.

smartavionics:
> In the commit from this PR those geometrical errors are propagated to further insets.

Indeed, but using simplify removes the geometrical errors introduced by doing each offset. If we do as you are suggesting which is to do all the offsets first and then go back and simplify each of the new polygons then all it will do is produce a slightly smoother version of a mangled polygon., If you look at the kind of cumulative errors that occur (e.g. see above issue), simplifying one of those wonky polygons is going to make almost zero difference.

BagelOrb:
Yeah so I guess I am proposing to solve the original issue like this:

old code:
```
Polygons prev = part.outline;
while (insetting)
{
prev = prev.offset(line_width).simplify();
concentric_lines.add(prev);
}
```

new pseudo-code:
```
coord_t offset_val = 0;
while (insetting)
{
offset_val += line_width;
concentric_lines.add(outline.offset(line_width).simplify());
}
```

BagelOrb:
Never mind. It's merged already. It's not a big enough issue to validate making a new PR for....

smartavionics:
Yes, but I think your suggested change is good (apart from the obvious mistake), as it will indeed stop errors propagating.

Should be more like:

```
coord_t offset_val = 0;
while (insetting)
{
offset_val -= line_width;
concentric_lines.add(outline.offset(offset_val).simplify());
}
```

BagelOrb:
Depends on the application. For walls computation you would do `-=`, for brim computation you would use `+=`. Concentric toolpaths are generated in quite some locations...

PR #1195, opened 2020-01-31
Adding a Dockerfile that builds CuraEngine

centerorbit:
To build, cd to the root of the repo and run `docker build .`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,38 @@
+FROM ubuntu:latest
+
+RUN apt-get update && apt-get install -y \
+  autoconf \
+  cmake \
+  g++ \
+  git \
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Sorry that this took so long.

We've decided to reject this pull request. We are sticking with our CMake build system in order to keep the number of build systems to support to a minimum.

centerorbit:
@Ghostkeeper thanks for the consideration. It's unfortunate that you won't include the Dockerfile, as it's a nice way to provide a sterile build environment. If you take a look at it's contents, it's more or less self-documenting the steps needed for the CMake build.

All the same, I'll leave this branch here in case someone else finds it useful.

Issue #466, opened 2017-02-20
Dual Extrusion  Temperature Control Broken on Smoothieware


ghost:
On smoothieware `M104 TN Sxxx` causes tool N to be selected which makes the idle tool temperature changes activate the tool and begins printing with it. I checked on the smoothieware irc channel and the developers indicated that there is no clear specification and potentially CNC gcode relies on `M104 TN` switching the tool. The fix is simply to reactivate the active tool after setting the idle tool temperature.

This makes dual extrusion fail when using Cura 2.x master as the tool switches whenever a preheat or cooldown is injected.

Attached are the original gcode and my munged gcode (sed -e 's/M104 T0\(.*\)/M104 T0\1\nT1\n/' -e 's/M104 T1\(.*\)/M104 T1\1\nT0\n/')  in case they are useful in any way.

[WSDMP_squirtle_dual_body.gcode.txt](https://github.com/Ultimaker/CuraEngine/files/788341/WSDMP_squirtle_dual_body.gcode.txt)

[squirtle_fixed.gcode.txt](https://github.com/Ultimaker/CuraEngine/files/788342/squirtle_fixed.gcode.txt)

BagelOrb:
You could write a post processing script which does exactly that.
You can write those pretty easily.
Go to Cura/plugins/postprocessingplugin and add a script.

I don't think we will do this for all gcode cura generates, because it
might be the case that the UM3 moves to the switching bay on that T0
command.
I therefore don't agree with the smoothieboard developers.

I will also accept pull requests which only change the CuraEngine behaviour
for smoothie firmware.

Op 20 feb. 2017 20:45 schreef "Clinton Ebadi" <notifications@github.com>:

unknownlamer:
I will do a survey of current 3d printer and cnc firmwares and if it makes sense attempt to make the case for changing behavior to the smoothie developers. If not, I'll submit a PR to add a smoothie variant of gcode to generate the tool changes to avoid affecting printers with other firmwares.

unknownlamer:
Ahoy,

[...]

PR #1197, opened 2020-02-04
Add "Always Write Active Tool" machine setting to support Smoothie firmware

grk3010:
Added a new Firmware flavor "Smoothie" to support printers running Smoothieware firmware. This particularly addresses issue #466 and avoids the need to accomplish the task through post-processing scripts.


Ghostkeeper:
This can be considered as a way to be consistent with our previous system with g-code flavours, but it's more flexible to build this as a separate machine setting: Should temperature commands always specify the tool or not? That method has a few advantages:
* Easier to understand what the settings do.
* You can combine this behaviour with other quirks that we already have a solution for, such as firmware retraction.

Do you know how to convert this to a separate machine setting, or would you need some help?

grk3010:
I'll take a look at adding it as a machine setting and see if I can figure it out. I'm fairly experienced with C++/Qt/qml, but have limited experience with python. Are you phasing out the G-code flavor system and replacing it with separate machine settings?

grk3010:
> This can be considered as a way to be consistent with our previous system with g-code flavours, but it's more flexible to build this as a separate machine setting: Should temperature commands always specify the tool or not? That method has a few advantages:
> 
> * Easier to understand what the settings do.
> * You can combine this behaviour with other quirks that we already have a solution for, such as firmware retraction.
> 
> Do you know how to convert this to a separate machine setting, or would you need some help?

Just to clarify, the setting that I need would be to re-activate the active tool after changing the temperature of the inactive tool. Smoothie firmware will conduct a toolchange whenever it sees a "T" command so
T0
M109 T0 230
M104 T1 200

would leave T1 as the active tool instead of T0. The setting that I was changing for the Smoothie firmware is to add a tool call to the active tool after the M104 command for the inactive tool. I'm not sure what you would call this if implemented as a machine setting.

Ghostkeeper:
> Are you phasing out the G-code flavor system and replacing it with separate machine settings?

Yeah we're trying to. We're not in a hurry to remove the current g-code flavour system but we wouldn't like to add on to it.

I understand your explanation. That helps.

> The setting that I was changing for the Smoothie firmware is to add a tool call to the active tool after the M104 command for the inactive tool. I'm not sure what you would call this if implemented as a machine setting.

How about something more descriptive rather than imperative, like: "Inactive Extruder Commands Switch Tools"? Most machine settings are worded to describe what the printer is, rather than instruct how Cura should talk to the printer.

grk3010:
> > Are you phasing out the G-code flavor system and replacing it with separate machine settings?
> 
> Yeah we're trying to. We're not in a hurry to remove the current g-code flavour system but we wouldn't like to add on to it.
> 
> I understand your explanation. That helps.
> 
> > The setting that I was changing for the Smoothie firmware is to add a tool call to the active tool after the M104 command for the inactive tool. I'm not sure what you would call this if implemented as a machine setting.
> 
> How about something more descriptive rather than imperative, like: "Inactive Extruder Commands Switch Tools"? Most machine settings are worded to describe what the printer is, rather than instruct how Cura should talk to the printer.

Could you point me in the right direction as to what I would need to do to add this as a machine setting? Would that be implemented in Cura Engine or the python/qml front end?

Ghostkeeper:
The machine settings are in [this list](https://github.com/Ultimaker/Cura/blob/d7102729f17a3286e76ea29a408bc82f8dcba255/resources/definitions/fdmprinter.def.json#L34) in fdmprinter.def.json. These are not shown to the user unless fieldOfView's Printer Settings plug-in is installed, but the printer definition files can adjust them. You would need to add a boolean setting there, similar to [Firmware Retract](https://github.com/Ultimaker/Cura/blob/d7102729f17a3286e76ea29a408bc82f8dcba255/resources/definitions/fdmprinter.def.json#L379-L389), which also used to be part of the g-code flavours at some point. Please prefix the setting ID with `machine_` as to categorise them when they are listed in a profile as well. So something like `machine_temp_commands_require_tparam` or whatever seems logical to you.

In CuraEngine, instead of adding a check for the g-code flavour, you'd need to add a check for your new setting. You can obtain the setting value with something like `mesh_group->settings.get<bool>("machine_temp_commands_require_tparam")`.

mobilityapplab:
Ok, I've implemented this feature by adding a parameter "machine_always_write_active_tool" in fdmprinter.def.json instead of adding the firmware flavor. 

I've tested and it seems to work . I have a bit of cleanup work to do on a baseline smoothie machine definition that uses these settings, then i'll update the cura pull request as well

@Ghostkeeper, thanks for the help on this!

Ghostkeeper:
This part looks good, but can't be merged until https://github.com/Ultimaker/Cura/pull/7067 is merged.

PR #1200, opened 2020-02-13
Flow problem

BagelOrb:
Single-liner.

The feature specific flow was not applied to the actual gcode, while it was used for the layer view.

I have no idea why the code looked like this, but it seems to me to be an actual issue.

We should do 1 test print to see whether the feature specific flow settings used to not influence the actual flow and whether it is now fixed.


BagelOrb:
Closed because this PR was made unto 4.5, not master...

BagelOrb:
real PR: https://github.com/Ultimaker/CuraEngine/pull/1201

PR #1201, opened 2020-02-13
Flow problem

BagelOrb:
Single-liner.

The feature specific flow was not applied to the actual gcode, while it was used for the layer view.

I have no idea why the code looked like this, but it seems to me to be an actual issue.

We should do 1 test print to see whether the feature specific flow settings used to not influence the actual flow and whether it is now fixed.

### DIFF
--- a/src/pathPlanning/GCodePath.cpp
+++ b/src/pathPlanning/GCodePath.cpp
@@ -31,7 +31,7 @@ bool GCodePath::isTravelPath() const
 
 double GCodePath::getExtrusionMM3perMM() const
 {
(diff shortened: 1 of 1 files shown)


BagelOrb:
Hmm. It seems there is no problem. I can verify that the current gcode generation does make a difference in both visualization and amount of used material.

So there must be a documentation issue.

BagelOrb:
Jesus I'm stupid.

Smh.

BagelOrb:
The flow is simply applied within `getExtrusionMM3perMM`

PR #1205, opened 2020-02-17
feat: minimum supporting area size

BagelOrb:
Can be used to make point overhang more reliable, by supporting more of the downward hanging geometry.

Problem:
When the geometry has some downward pointing part, it may happen that the overhang is quit small.
When printing such a part on top of a tiny bit of support the printed part starts to wobble during printing, which eventually leads to spaghetti being printed.
It is not often that a model has such downward pointing geometry (perhaps 1% of all models), but *when* a model has such downward pointing geometry the print almost always fails - or at least shows defects.

Solution:
The solution we propose here is to consider geometry with a too small area as not being supported at all - regardless of the overhang angle. A layer part with an area smaller than the setting value is handled as if the overhang angle is 0*.

The fix of this PR fixes most of the pointy overhang problems using one simple setting. It's not a full proof solution, but it's very simple code, and it is very effective.

-------

*Results*

Without the setting:
image

With setting set to 10 mm^2:
image
closeup:
image

### DIFF
--- a/src/support.cpp
+++ b/src/support.cpp
@@ -1248,12 +1248,14 @@ std::pair<Polygons, Polygons> AreaSupport::computeBasicAndFullOverhang(const Sli
     Polygons supportLayer_supportee = mesh.layers[layer_idx].getOutlines();
     constexpr bool no_support = false;
     constexpr bool no_prime_tower = false;
-    Polygons supportLayer_supporter = storage.getLayerOutlines(layer_idx-1, no_support, no_prime_tower);
+    const double min_supporting_model_area = mesh.settings.get<double>("min_supporting_model_area");
+    Polygons supportLayer_supporter = storage.getLayerOutlines(layer_idx-1, no_support, no_prime_tower, false, false, min_supporting_model_area);
 
     const coord_t layer_height = mesh.settings.get<coord_t>("layer_height");
     const AngleRadians support_angle = mesh.settings.get<AngleRadians>("support_angle");
     const double tan_angle = tan(support_angle) - 0.01;  //The X/Y component of the support angle. 0.01 to make 90 degrees work too.
     const coord_t max_dist_from_lower_layer = tan_angle * layer_height; //Maximum horizontal distance that can be bridged.
+
     Polygons supportLayer_supported =  supportLayer_supporter.offset(max_dist_from_lower_layer);
     Polygons basic_overhang = supportLayer_supportee.difference(supportLayer_supported);
 
--- a/src/sliceDataStorage.cpp
+++ b/src/sliceDataStorage.cpp
@@ -54,17 +54,21 @@ SliceLayer::~SliceLayer()
 {
 }
 
-Polygons SliceLayer::getOutlines(bool external_polys_only) const
+Polygons SliceLayer::getOutlines(bool external_polys_only, double min_part_area) const
 {
     Polygons ret;
-    getOutlines(ret, external_polys_only);
+    getOutlines(ret, external_polys_only, min_part_area);
     return ret;
 }
 
-void SliceLayer::getOutlines(Polygons& result, bool external_polys_only) const
+void SliceLayer::getOutlines(Polygons& result, bool external_polys_only, double min_part_area) const
 {
     for (const SliceLayerPart& part : parts)
     {
+        if (min_part_area > 0.0 && INT2MM2(part.outline.area()) < min_part_area)
+        {
+            continue;
+        }
         if (external_polys_only)
         {
             result.add(part.outline.outerPolygon());
@@ -382,7 +386,7 @@ SliceDataStorage::SliceDataStorage()
     machine_size.include(machine_max);
 }
 
-Polygons SliceDataStorage::getLayerOutlines(const LayerIndex layer_nr, const bool include_support, const bool include_prime_tower, const bool external_polys_only, const bool for_brim) const
+Polygons SliceDataStorage::getLayerOutlines(const LayerIndex layer_nr, const bool include_support, const bool include_prime_tower, const bool external_polys_only, const bool for_brim, double min_part_area) const
 {
     if (layer_nr < 0 && layer_nr < -static_cast<LayerIndex>(Raft::getFillerLayerCount()))
     { // when processing raft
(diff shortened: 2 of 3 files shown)


BagelOrb:
If the support expansion is enabled, this will also support small horizontal walls:
image
But not if they are connected to larger parts.

Although this is an unintended side effect it is not an *undesired* side effect, it might actually be good for printing more reliably.

BagelOrb:
I suggest we set the default value to approx 10 mm^2, but we should at least perform *some* tests to verify that this resolves the problem.

BagelOrb:
Here's my testing model:
[support_tree.stl.zip](https://github.com/Ultimaker/CuraEngine/files/4213114/support_tree.stl.zip)

src/support.cpp
```diff
@@ -1248,12 +1248,14 @@ std::pair<Polygons, Polygons> AreaSupport::computeBasicAndFullOverhang(const Sli
     Polygons supportLayer_supportee = mesh.layers[layer_idx].getOutlines();
     constexpr bool no_support = false;
     constexpr bool no_prime_tower = false;
-    Polygons supportLayer_supporter = storage.getLayerOutlines(layer_idx-1, no_support, no_prime_tower);
+    const double min_supporting_model_area = mesh.settings.get<double>("min_supporting_model_area");
+    Polygons supportLayer_supporter = storage.getLayerOutlines(layer_idx-1, no_support, no_prime_tower, false, false, min_supporting_model_area);
```
Ghostkeeper, line 1252: This has two boolean traps.

BagelOrb:
> "The disadvantage is that pieces that don't become wider towards the top won't get supported at all any more. " (from JIRA)

I don't think this is true. According to what the OP (aka past tense me) said, parts that don't get wider to the top should get support when this feature is enabled; on the contrary, even parts that get *less wide* toward the top get support.

I don't recall the implementation or the testing exactly so I might be wrong. Are you sure such an object wouldn't get support?

Ghostkeeper:
I'll test that. It was just theory from a glance at the code at that point, so the claim is weak.

I've found one concern with this issue:
image

This setting causes a lot of support to be printed along the sides. I've reduced the X/Y/Z distances to 0 and added a wall around support to make it easier to see where it generates support exactly. It seems that the crenellations at the top of the tower are being supported now by towers. Maybe we need to move this filter to a different stage in the support generation?

Ghostkeeper:
You're right, it does support that case apparently. Why though? Doesn't it remove the tiny model part since it's less than my 20mm² area?

image

BagelOrb:
Indeed "minimum support area" means that smaller support areas will become larger, not that they will be removed.

Support should never automatically be removed. Overhang is overhang.

BagelOrb:
> This setting causes a lot of support to be printed along the sides

That's an unintentional side effect. I don't think it's possible to prevent this without doing some extensive geometrical analysis, but that would be way beyond scope.

BagelOrb:
crennelations? I'm not sure what you are talking about. It looks to me at a glance that theres some internal geometry perhaps or maybe you set the minimum area really large?

Ghostkeeper:
> crennelations? I'm not sure what you are talking about.

Sorry, I meant the [battlements](https://en.wikipedia.org/wiki/Crenellation), the small parts on the top of the tower that protect people from arrows.

I set the minimum area to be larger than the area of those battlements, 20mm². I don't know if we consider this to be particularly large (it would be a square of 4.5mm wide, which would be on the safe side for support I guess). But I didn't find the limits of how big the battlements actually are, and you could craft pieces that are smaller.

BagelOrb:
20mm2 (=4x5) is a bit much. I would think that a minimum area of 2x2 mm would already be good enough for most applications.

Ghostkeeper:
True. But a piece having 2x2mm bumps on the top is not uncommon either.

BagelOrb:
Hmm ok. Yeah solving that would require quite a bit more machinery.

I guess we would be forced to leave this feature off by default then.

The problem only occurs when the small bumps on top are at the edge of the base, like the battlements are on the sides of the tower.

I'm not sure which problem is bigger: extra support on such sides, or too small suport for downward facing points.

Ghostkeeper:
We've discussed this in our Cura & Cookies meeting and decided that we'd prefer not to merge this in with a known bug like this. Indeed we could just not set it in any profile, like we do with Enable Support, but we're also afraid that nobody will be using it then.

This problem also occurs when multiple small bumps are close together, as you can see in the crenellations at the smaller balcony on the side of the tower. All in all, it seems to be quite common with organic models.

We've also discussed a solution to this problem. We could apply the X/Y/Z distance also before the towers are generated. This makes this setting only effective if it's set to be larger than the minimum tower diameter, which is fair enough, but it does have the problem that you'd then need to re-apply the X/Y/Z distance again on the towers. To that end we've set a new issue in our backlog to refactor the X/Y/Z distance algorithm to a separate function so that we can more easily call that multiple times, which would also help for general maintenance of that code anyway. See CURA-7481. This feature is now blocked by that refactor.

Ghostkeeper:
We discussed this again in Cura & Cookies and we'd like to keep the main idea of this implementation like this, but make an adjustment to a different part of support that would remove the remaining problem: We'll apply the X/Y distance before propagating the support down. To this end, we need to move the propagation down later. This causes the support to disappear in the case where the model gets wider towards the bottom, preventing the remaining bug.

This might have side effects. In particular, we should be careful to keep the overall shape of normal supports the same. It could be that this makes the support disappear entirely at certain angles because the X/Y distance causes the lower layer to erase the existing support. We don't know if this is can be done yet with some combination of polygon operations so we'll have to experiment a bit there.

BagelOrb:
> It could be that this makes the support disappear entirely at certain angles because the X/Y distance causes the lower layer to erase the existing support.

This sounds pretty dangerous. Wouldn't that be an inherent problem with your proposed solution? Suppose you have an overhang angle of 45*, and 0.1mm layer height, then an overhang of 45* will produce overhanging strips of 0.1mm wide, so with a Z distance of zero, an XY distance more than 0.1mm would already remove the support completely.

BagelOrb:
The proposed feature has too many faults and we don't have any idea how to circumvent those, so I will retract this PR.

Issue #1151, opened 2019-10-20
VeryFine MaxResolution (<0.1mm) results in jagged artifacts and worse slicing


daT4v1s:
**Application Version**

Cura 4.3.0 , Official Build

**Platform**

Linux Ubuntu 18.04 LTS , Ultimaker 2+

**Display Driver**

Intel Graphics

**Steps to Reproduce**

set max resolution to LessThan 0.1 mm
slice print
view tool-path result in PreviewPane

**Actual Results**

the lines look tangles,and coarser
i thought a lower number meant a finer result, 
but if was simply coarser the artifacts are expected to be regular, but are jagged here
and also there are occasional spurious blotches

is this due to code being emitted being finder than what the printer can represent?
(heard of 0.04mm but haven't verified this)

also the printing times changed irregularly 
 for other prints when tweaking this value 
(more than 10 minutes at some values for a 2-4 hours print)

**Expected results**

lines look smoother when using a finer resolution,
print paths follow the curve, remain smooth,
no spurious blobs around the print

**Additional Information**

this is still the case after disabling/resetting various settings, except for the resolution related

various resolution setting sweep, starting from 0.01mm (supposedly finest)
notice the random blob @ 0.03mm
![animated](https://user-images.githubusercontent.com/47403824/67154114-0f8e3300-f2c4-11e9-99c7-454964f0d0fb.png)

- this became more noticeable when using a finer mesh tessellation (relevant of not?)

daT4v1s:
oops - maybe I should have filed this 
in https://www.github.com/Ultimaker/Cura/issues
how shall i move this?

Ghostkeeper:
That's okay. Maybe you can provide a project file that reproduces the issue? That's via File -> Save... (**Not** the save button in the bottom left.) I've tried this with a high resolution cylinder but wasn't able to get that effect.

One thing you could try as a test is to disable wall overlap compensation.

daT4v1s:
the .3mf cura-file is 34mb, (>10MB attachment limit)
the artifacts are still there despite turning off OverlapCompensation 
![image](https://user-images.githubusercontent.com/47403824/67363661-e4157d80-f53b-11e9-87e5-00b32e91b762.png)

Ghostkeeper:
You could upload it elsewhere (WeTransfer?) and post a link here.

I tried to reproduce it by loading a 700MB sphere with 1cm diameter into Cura, the highest resolution model I have. But the surface was smooth.

[...]

PR #1214, opened 2020-03-07
Fix simplify() and test it extensively

BagelOrb:
Better tests for the simplify() function.

It is a very important function and the tests were flawed.

EDIT: 
The tests proved that something was wrong with the function, and something was wrong with earlier tests.

Therefore I had to rewrite and fix simplify()

Thanks to @smartavionics for laying down the ground work for this PR.


BagelOrb:
New circle test:
Starts out really smooth in the East and becomes every rougher in the clockwise direction.
Before simplify: black, after: red
image

BagelOrb:
Fixed spiral test:
added far away geometry to force the algorithm to start at a specific location with simplification
black: original polygon, red: FAILING test
The inner point of the spiral has to be in the output.

image

BagelOrb:
Sinusoidal test:
Checks whether sinusoidal errors get simplified.
Black: original, red: simplified
Note that there is an error because the long segments on the right is simplified, but this test doesn't check for that.
image

BagelOrb:
Extreme high poly test.
Segments are approx 3 micron
image

BagelOrb:
This PR now also includes a fix for the changed long line segment problems shown in figure 2 and 3 of this thread.

BagelOrb:
Zig zag test which captures the intent of the original spiral test.
This test proved the implementation was faulty. It used to simplify half of the bends it should have before I fixed it.
image

BagelOrb:
here's the file which shows the resolution bug which I wasn't able to reproduce in an automated test:
Bug:
image
Solved:
image

[max_res_bug.3mf.txt.7z.004.part.zip](https://github.com/Ultimaker/CuraEngine/files/4308081/max_res_bug.3mf.txt.7z.004.part.zip)
[max_res_bug.3mf.txt.7z.003.part.zip](https://github.com/Ultimaker/CuraEngine/files/4308082/max_res_bug.3mf.txt.7z.003.part.zip)
[max_res_bug.3mf.txt.7z.001.part.zip](https://github.com/Ultimaker/CuraEngine/files/4308083/max_res_bug.3mf.txt.7z.001.part.zip)
[max_res_bug.3mf.txt.7z.002.part.zip](https://github.com/Ultimaker/CuraEngine/files/4308084/max_res_bug.3mf.txt.7z.002.part.zip)

BagelOrb:
I've completed all my tests and verified that all automated tests succeed and that the 4 linked issues are solved by this PR as well.

JIRA ticket CURA-7243 (no github issue)
Ultimaker/Cura#6604 (JIRA: CURA-6955)
https://github.com/Ultimaker/CuraEngine/issues/1151
https://github.com/Ultimaker/Cura#6215

tests/utils/PolygonTest.cpp
```diff
@@ -305,18 +326,263 @@ TEST_F(PolygonTest, simplifyLimitedError)
...
-    //2A = b * h
-    //2A / b = h
-    const coord_t height = 4 * area / diagonal_length; //Error of the first vertex we want to keep, so we must set the limit to something slightly lower than this.
-    spiral_polygons.simplify(999999999, height - 10);
-
-    EXPECT_THAT(spiral.size(), testing::AllOf(testing::Ge(11 - 5), testing::Le(11 - 4))) << "Should merge segments of length 1000 through 1400 and (optionally) first with last.";
+    sine_polygons.simplify(length / 2, 2 * deviation);
+    if (visualize)
+    {
+        SVG svg("output/simplifySineLimitedError.svg", AABB(sine_before));
```
nallath: There is a lot of code duplication with writing these svg files. I think we should convert it into a re-usable function

src/utils/polygon.h
```diff
@@ -451,6 +451,13 @@ class PolygonRef : public ConstPolygonRef
      * 
      * Removes verts which are connected to line segments which are both too small.
      * Removes verts which detour from a direct line from the previous and next vert by a too small amount.
+     *
+     * Criteria:
+     * 1. Never remove a vertex if either of the connceted segments is larger than \p smallest_line_segment
+     * 2. Never remove a vertex if the distance between that vertex and the final resulting polygon would be higher than \p allowed_error_distance
+     * 3. Simplify uses a heuristic and doesn't neccesarily remove all removable vertices under the above criteria.
+     * 4. But simplify may never violate these criteria.
+     * 5. Unless the segments or the distance is smaller than the rounding error of 5 micron
```
nallath: This confuses me. Does it refer to 4? (if so, it should be a part of 4) If not, I don't understand what it does refer to.

tests/utils/PolygonTest.cpp
```diff
@@ -305,18 +326,263 @@ TEST_F(PolygonTest, simplifyLimitedError)
...
+    zigzag_polygons.simplify(non_simplifiable_bound, max_height);
+
+    if (visualize)
+    {
+        SVG svg("output/simplifyIncreasingLimitedError.svg", AABB(zigzag_before));
+        svg.writePolygon(zigzag_before);
+        svg.nextLayer();
+        svg.writePolygon(zigzag, SVG::Color::RED);
+    }
+    EXPECT_THAT(zigzag.size(), testing::Eq(zigzag_before.size() - simplifiable_bend_count)) << "Should simplify bends with height 100 up to 500";
```
nallath: That seems a bit light of a test, doesn't it? It's now only checking if the number of polygons is correct. Should there also not be some sort of test that actually checks if the error isn't too great?

tests/utils/PolygonTest.cpp
BagelOrb: Well. It's not like there's any complexity in the code. Nothing can go wrong. It would maybe be a bit cleaner.. Sure,

nallath:
Rounded corners do get worse though (it's more noticeable if you switch between the two images in image viewer):

Before:
ClosupWithoutChanges

After: 
ClosupWithChanges

You can see it at the entire model as well:
Before
WithoutChangesToEngine
After
WithChangesToEngine

tests/utils/PolygonTest.cpp
nallath: Yeah, it's a bit nitpicking, but I like my code to be clean :)

src/utils/polygon.h
BagelOrb: Yes it refers to 4.

BagelOrb:
> Rounded corners do get worse though

Yes. That's the whole goal of simplify().

tests/utils/PolygonTest.cpp
BagelOrb: No. It's not checking `zigzag_polygons`, it's checking the one polygon `zigzag`.

Actually this test case is testing the most important thing atm. It only succeeds if exactly the number of bends which we expect to be dsimplified out actually do. No more and no less.

tests/utils/PolygonTest.cpp
BagelOrb: The conceptual load of reading the current code is no more, or perhaps even less than reading a new function which fits on one line. I don't think the physical length of the code should be a criterion to decide on code-cleanliness.

I think it is nitpicking. Let's not discuss. I would be okay with you fixing this, but I am overburdened atm.

BagelOrb:
Did you check the frontend PR for this PR??

The new simplify simplifies twice as hard as the old one, so I made a frontend PR to update all simplify settings to be halved. This also includes an upgrade system to update old profiles to new profiles with half the value.

Ghostkeeper:
I've tested the basic functionality with a variety of organic and non-organic-but-rounded shapes and found no immediately visible concerns. The maximum resolution seems to be held in honour unless the maximum deviation is in concern. The maximum deviation seems to always be held in honour except for a minimum of 5 micrometres. Perhaps we want to set the minimum for Maximum Deviation in the front-end to 0.005mm?

I have a cylinder with 100,000 sides with a radius of 10. ([this one](https://github.com/Ghostkeeper/SlicerTestModels/blob/master/stress_test/cylinder_hi.scad)). That results in side lengths of 0.000314mm, or approximately 0.3 micrometres, so basically just multiple vertices on every integer coordinate possible along the circle. I sliced this using a maximum resolution and maximum deviation of 0.005mm, so that the exception for "pretty much straight" doesn't get in the way. Let's take an inset of 0.175mm into account for the outer wall line width. You'd expect then that a maximum deviation of 0.005mm can lead to side lengths of up to `2 * sin(acos((9.825 - 0.005) / 9.825)) * 9.825 = 0.626817mm`

Here is a snippet of g-code that I took from around the middle of the height of the cylinder (since each face is also divided diagonally into 2 triangles).

```
G1 X122.701 Y114.697 E88.99405
G1 X122.292 Y115.03 E88.99653
G1 X121.845 Y115.355 E88.99913
G1 X121.408 Y115.634 E89.00157
G1 X120.938 Y115.899 E89.00411
G1 X120.497 Y116.118 E89.00642
```

I'm seeing distances here in the ballpark of 0.5mm, **which is a bit lower than I'd expect** but in the correct order of magnitude.

I've also tested whether seams get removed from straight bits, which seems to work fine.

I've also sliced this model with new and old (skin and infill removed to reduce noise from things that aren't simplified):
image
The new algorithm has a deviation of 0.025mm rather than 0.05mm, but the resolution is the same. The old file was 12.0MB. The new g-code file is 10.5MB. This suggests that **in this case the simplify algorithm removed a bit more**, even with the lower deviation setting.

It's not entirely immune against leaving in vertices in a straight line. Here's an example where it chose the wrong vertices to remove. It chose the diagonals of a flat face. Parameters are increased to make it visible:
image
However the criterion that it needs to remove the best vertex is not in the scope of this algorithm so I think this should be accepted.

Bug CURA-7243 seems to be fixed.
Bug CURA-6729 / https://github.com/Ultimaker/Cura/issues/6215 seems to be fixed.
Bug CURA-6933 / https://github.com/Ultimaker/CuraEngine/issues/1151 seems to be fixed.
Bug CURA-6955 / https://github.com/Ultimaker/Cura/issues/6604 seems to be fixed.

In general it's pretty hard to find faults with this.

BagelOrb:
Did you guys look at the frontend changes? https://github.com/Ultimaker/Cura/pull/7247

Im not sure whether it's easy to port these changes to a new version upgrade system for 4.7 instead of 4.6

Ghostkeeper:
Yes, I did. Like I said in my comparison:

> The new algorithm has a deviation of 0.025mm rather than 0.05mm, but the resolution is the same.

I didn't apply the version upgrade in my tests, but I did use the new defaults to compare.

To change the front-end for 4.7, we'd need to change the setting version there to 13 and rename the plug-in.

BagelOrb:
Ah ok. Good!

jornada812:
I upgraded to 4.8 and re-sliced my model. It looks like the simplification changes were not made as planned by the developer. It got worse https://yadi.sk/i/oyUH92ART1Pteg

nallath:
> I upgraded to 4.8 and re-sliced my model. It looks like the simplification changes were not made as planned by the developer. It got worse https://yadi.sk/i/oyUH92ART1Pteg

That looks an awefull lot like coasting. Try disabling that. If that doesn't work, please create a new issue.

PR #1218, opened 2020-03-20
CURA-7284 support-brim vs. skirt ordering

rburema:
Skirt should always be printed before support-brim (when the same extruder is used).

### DIFF
--- a/src/SkirtBrim.cpp
+++ b/src/SkirtBrim.cpp
@@ -152,7 +152,8 @@ void SkirtBrim::generate(SliceDataStorage& storage, Polygons first_layer_outline
     const ExtruderTrain& support_infill_extruder = scene.current_mesh_group->settings.get<ExtruderTrain&>("support_infill_extruder_nr");
     if (allow_helpers && support_infill_extruder.settings.get<bool>("support_brim_enable"))
     {
-        generateSupportBrim(storage);
+        const bool merge_with_model_skirtbrim = !is_skirt;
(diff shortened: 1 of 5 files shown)


src/SkirtBrim.cpp
```diff
@@ -84,7 +84,7 @@ void SkirtBrim::getFirstLayerOutline(SliceDataStorage& storage, const size_t pri
     }
 }
 
-coord_t SkirtBrim::generatePrimarySkirtBrimLines(const coord_t start_distance, size_t primary_line_count, const coord_t primary_extruder_minimal_length, const Polygons& first_layer_outline, Polygons& skirt_brim_primary_extruder)
+coord_t SkirtBrim::generatePrimarySkirtBrimLines(const coord_t start_distance, size_t& primary_line_count, const coord_t primary_extruder_minimal_length, const Polygons& first_layer_outline, Polygons& skirt_brim_primary_extruder)
```
ninovanhooff: Assuming the Added & indicates that primarly_line_count lives in the heap. Is this a bug fix? Otherwise I would expect additional changes that refactor this variable to the heap

rburema: Passing by reference (or pointer, for that matter) doesn't always indicate that something lives on the heap. You can have references (and pointers) to a memory location to the stack as well.

(Though I'm not sure heap and stack are used very much as terms any more, they've become inaccurate/legacy.)

PR #1219, opened 2020-03-21
Aggressively smooth polygons when multiplying infill lines.

smartavionics:
This removes a lot of jaggies which looks nicer and reduces print time.

This makes a great different when using gyroid pattern. Here's a couple of images that show the difference. Notice the big change in predicted print time

Before this PR...
Screenshot_2020-03-21_13-51-56

With this PR...
Screenshot_2020-03-21_13-51-25

I haven't seen any detrimental effect on the other infill patterns with this change.


Ghostkeeper:
This seems to be fighting a symptom of the problem though. What causes those jags in the first place?

PR #1220, opened 2020-03-21
Improve ordering of inner walls to reduce length of travels immediately before outer wall.

smartavionics:
Before, it was possible that if some of the inner walls did not pass near to the outer wall z-seam
location, it could end up doing a long travel move immediately before the outer wall was
started even though there was an inner wall right next to the z-seam location.
Now, as long as there is a level 1 wall next to the z-seam location, it will print that
immediately before the outer wall so any underextrusion following a long travel move should
occur in the level 1 wall rather than the outer wall.

Here's a project that was improved by this PR.

[CCR10_combo_mount_ela_rev2.zip](https://github.com/Ultimaker/CuraEngine/files/4363170/CCR10_combo_mount_ela_rev2.zip)


rburema:
Working on this as internal ticket CURA-7415

rburema:
Once again, thanks for your work!

PR #1227, opened 2020-03-30
Remove nozzle length and filament park distance settings

Ghostkeeper:
They were unused.
I don't really need to remove them from that test since CuraEngine will ignore any settings it doesn't know. But it won't hurt.

Contributes to issue CURA-7328.


Ghostkeeper:
Seems that the test failed because it couldn't publish a Debian package to the Rundeck environment for building. It got a 404 error. I don't know what's up with that, but I think it's a glitch since 3 hours ago a different commit was passing this.

konskarm:
> Seems that the test failed because it couldn't publish a Debian package to the Rundeck environment for building. It got a 404 error. I don't know what's up with that, but I think it's a glitch since 3 hours ago a different commit was passing this.

I rerun the tests and it worked fine. It must have been an unlucky incident.

PR #1228, opened 2020-03-31
CURA-7284 fix skin-brim ordering also for prime-tower brim

rburema:


src/FffPolygonGenerator.cpp
```diff
@@ -1029,21 +1029,28 @@ void FffPolygonGenerator::processPlatformAdhesion(SliceDataStorage& storage)
...
-    //If brim for prime tower is used, add the brim for prime tower separately.
-    //Since FffGCodeWriter assumes that the outermost contour is last, add this first. Keep it ordered!
-    bool should_brim_prime_tower = storage.primeTower.enabled && mesh_group_settings.get<bool>("prime_tower_brim_enable");
-    if (should_brim_prime_tower)
+
+    EPlatformAdhesion adhesion_type = mesh_group_settings.get<EPlatformAdhesion>("adhesion_type");
+
+    // The switch statement is not used here, since if brim for the prime tower is needed, it should go after any skirt, but before any brim.
+    if (adhesion_type == EPlatformAdhesion::SKIRT)
+    {
```
ninovanhooff:
```suggestion
    if (adhesion_type == EPlatformAdhesion::SKIRT)
    {
        // If the prime tower needs a brim, tower brim should go after any skirt, but before any model brim.
```

Had to read this comment 5 times. I hope my suggestion is more clear :-)

src/FffPolygonGenerator.cpp
```diff
@@ -1029,21 +1029,28 @@ void FffPolygonGenerator::processPlatformAdhesion(SliceDataStorage& storage)
...
+        primary_line_count = train.settings.get<size_t>("skirt_line_count");
+        SkirtBrim::getFirstLayerOutline(storage, primary_line_count, true, first_layer_outline);
+        SkirtBrim::generate(storage, first_layer_outline, train.settings.get<coord_t>("skirt_gap"), primary_line_count);
+    }
+
+    // Generate any brim for the prime tower, should happen _after_ any skirt, but _before_ any other brim (since FffGCodeWriter assumes that the outermost contour is last).
```
ninovanhooff: Hmm my suggestion more or less duplicates this. I would also be okay with leaving just this, as it's easy to understand.

src/FffPolygonGenerator.cpp
rburema: Feel free to change this!

src/FffPolygonGenerator.cpp
rburema: Up to you!

PR #1231, opened 2020-04-06
Start skirt at point nearest to current position + print skirt from outer line inwards.

smartavionics:
It always used to print the skirt from the outer line inwards but recently that had
changed so I have reinstated the original behaviour.

These images show how with the PR the skirt is started at the point nearest the current location (0, 0)

Without this PR...

Screenshot_2020-04-06_11-20-48

With this PR...

Screenshot_2020-04-06_11-21-27

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -1028,18 +1028,25 @@ void FffGcodeWriter::processSkirtBrim(const SliceDataStorage& storage, LayerPlan
         start_close_to = gcode_layer.getLastPlannedPositionOrStartingPosition();
     }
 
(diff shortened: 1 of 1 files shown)


ashleysommer:
Probably need to investigate why this behavior changed between 4.5.0 and 4.6.0, and determine if changing it back will break something else?

rburema:
Hi @ashleysommer, Good point, but: I made [the fix](https://github.com/Ultimaker/Cura/issues/4829#issuecomment-608056362) that (unfortunately) changed this, and I can see @smartavionics has made this fix on top of my earlier one(s), rather than replace it.

rburema:
Working on this as internal ticket CURA-7414

Issue #1233, opened 2020-04-10
Print time incorrect; minutes are always zero


marciot:
**Application Version**

CuraEngine, master branch

**Platform**

Linux, CuraEngine command line compiled with gcc (ENABLE_ARCUS off)

**Qt**

N/A

**PyQt**

N/A

**Display Driver**

N/A

**Steps to Reproduce**

1. Slice a file that takes more than an hour to print.
2. Inspect the console log, compare the "Print time (s)" to "Print time (hr|min|s)"
3. Notice these do not match

**Actual Results**

```
Print time (s): 87813
Print time (hr|min|s): 24h 0m 23s
```

**Expected results**

```
Print time (s): 87813
Print time (hr|min|s): 24h 23m 33s
```

**Additional Information**

It seems the minutes are being shown as seconds, and the minutes are always zero in "Print time (hr|min|s)"

For the example above, here is the time computation:

```
(24h * 60 * 60) + (23m * 60) + 33s =
86400s + 1380s + 33s =
87813s
```

rburema:
Since I just merged your PR, this issue can be closed.

PR #1235, opened 2020-04-10
Fix incorrect display of print time (issue #1233)

marciot:
This fixes the incorrect display of the print time (#1233). This is because the `sprintf` "%d" specifier cannot be used for `int64_t`.

Before:

```
Print time (s): 87813
Print time (hr|min|s): 24h 0m 23s
```

After:

```
Print time (s): 87813
Print time (hr|min|s): 24h 23m 33s
```

### DIFF
--- a/src/gcodeExport.cpp
+++ b/src/gcodeExport.cpp
@@ -1364,7 +1364,7 @@ void GCodeExport::finalize(const char* endCode)
     int64_t print_time = getSumTotalPrintTimes();
     int mat_0 = getTotalFilamentUsed(0);
     log("Print time (s): %d\n", print_time);
-    log("Print time (hr|min|s): %dh %dm %ds\n", print_time / 60 / 60, (print_time / 60) % 60, print_time % 60);
+    log("Print time (hr|min|s): %dh %dm %ds\n", int(print_time / 60 / 60), int((print_time / 60) % 60), int(print_time % 60));
     log("Filament (mm^3): %d\n", mat_0);
     for(int n=1; n<MAX_EXTRUDERS; n++)
         if (getTotalFilamentUsed(n) > 0)


rburema:
Thanks for looking into this yourself!

However, a more appropriate way to fix this would probably be to use [`%lld`](https://en.cppreference.com/w/cpp/io/c/fprintf) instead though, (not `%ld` as standard `long`s are only guaranteed to hold 32 bits, unlike `long long`s, see [here](https://en.cppreference.com/w/cpp/language/types#Properties)).

marciot:
> Thanks for looking into this yourself!
> 
> However, a more appropriate way to fix this would probably be to use [`%lld`](https://en.cppreference.com/w/cpp/io/c/fprintf) instead though, (not `%ld` as standard `long`s are only guaranteed to hold 32 bits, unlike `long long`s, see [here](https://en.cppreference.com/w/cpp/language/types#Properties)).

Yes, there are many ways to fix this. I saw the use of a 64-bit integer and my thought was, why? A 32-bit integer can represent 4,294,967,295 seconds, which works out to 139 years. So, why use a 64-bit integer? Do we really expect a print longer than 140 years? :)

Personally, I would replace the `int_64t` with a `int_32t` and use "%ld", but I chose to fix in the easiest way possible, Feel free to adjust it to taste!

rburema:
> but I chose to fix in the easiest way possible, Feel free to adjust it to taste!

Fair enough. I mostly brought it up in case you didn't know. I think it's acceptable anyway :-)

Ghostkeeper:
We can't use %lld since it's not supported in our Windows compiler, iirc. We've had this go back and forth a few times in the past. I expect it'd be documented in the commit messages of the blame there.

rburema:
I think they might've fixed that ... let me check VS2019 for a bit.
Yup, compiles & works just fine now!
... but the version now merged also supports older windows compilers then, which is a bonus.

PR #1241, opened 2020-04-23
CURA-7368 remove debug code

rburema:
- compiles & tests ok on windows (but not in build env. because boost-include is missing)
- remove most of debug code and some asserts


src/BeadingStrategy.cpp
```diff
@@ -9,29 +9,8 @@ namespace arachne
...
-    {
-        coord_t calculated_thickness = transition_thicknesses[transition_idx];
-        coord_t supposed_thickness = strategy->transition_thickness(transition_idx);
-        assert(std::abs(calculated_thickness - supposed_thickness) < 10);
-        if (std::abs(calculated_thickness - supposed_thickness) >= 10)
-            ret = true;
-    }
-#endif // DEBUG
-    return ret;
+    // LA-TODO: can this work check
```
nallath, line 12: This should be removed or resolved.

src/BeadingStrategy.cpp
```diff
@@ -51,7 +30,7 @@ float BeadingStrategy::getTransitionAnchorPos(coord_t lower_bead_count) const
     return 1.0 - float(transition_point - lower_optimum) / float(upper_optimum - lower_optimum);
 }
 
-virtual std::vector<coord_t> BeadingStrategy::getNonlinearThicknesses(coord_t lower_bead_count) const
+/*virtual*/ std::vector<coord_t> BeadingStrategy::getNonlinearThicknesses(coord_t lower_bead_count) const
```
nallath, line 33: This should be removed.

src/SkeletalTrapezoidation.cpp
```diff
@@ -635,8 +569,7 @@ void SkeletalTrapezoidation::collapseSmallEdges(coord_t snap_dist)
                 {
                     std::cerr << edge_from_3->from->p << " - " << edge_from_3->to->p << '\n';
                 }
-                assert(++count < 100);
-                if (count > 1000) break;
+                if (++count > 1000) break;
```
nallath, line 572:
```suggestion
                if (++count > 1000) 
                {
                    break;
                }
```

src/SkeletalTrapezoidation.cpp
```diff
@@ -1331,27 +1194,20 @@ void SkeletalTrapezoidation::generateTransition(edge_t& edge, coord_t mid_pos, c
     float mid_rest = transition_mid_position * inner_bead_width_ratio_after_transition;
     float end_rest = inner_bead_width_ratio_after_transition;
 
-    
-        debugCheckGraphCompleteness();
-        debugCheckGraphConsistency();
-
     { // lower bead count transition end
```
nallath, line 1197: Not sure if that's for this review, but the {} aren't really needed here, right?

src/SkeletalTrapezoidation.h
```diff
@@ -22,6 +22,7 @@
 
 namespace arachne
 {
+    using namespace cura;
 
 extern bool generate_MAT_STL;
```
nallath, line 27: Isn't this also debug code?

src/mainArachne.cpp
```diff
@@ -43,8 +43,8 @@
 
 #include "TestGeometry/VariableWidthGcodeTester.h"
 
-using arachne::Point;
-
+//using arachne::Point;
```
nallath, line 46: Remove commented out code

src/SkeletalTrapezoidation.h
rburema: No, this is needed for compiling on windows. Unless we want to change every line that calls into the cura namespace from `function(...` to `cura::function`,

src/SkeletalTrapezoidation.cpp
rburema: No, but they're sometimes useful for documentation (and can actually help with scoping, but that isn't needed here).

src/SkeletalTrapezoidation.h
Ghostkeeper: I suppose eventually we'll want to remove the `arachne` namespace altogether, to keep things simple?

For now it's necessary though because a lot of classes and functions are literally copy-pasted from the main CuraEngine code and so they have the same names.

PR #1242, opened 2020-04-23
Fix edge case in moveInside

DerGenaue:
This fixes an edge case in the moveInside method:
The comparison `ax_length >= ab_length` MUST be the same for the points IN the loop and the check BEFORE the loop.

The edge case causing the issue:
Px is projected exactly onto the last line of the polygon => `projected_p_beyond_prev_segment` is initialized to `false`.
At the end of the loop, this last line is checked; but because in the loop, `ax_length` was divided by `ab_length`, this rounded up to the next integer, which means that inside the loop, Px is projected ONTO the last vertex (`projected_p_beyond_prev_segment = true`) and the loop ends without ever checking the last vertex.
This should resolve that issue.

I did not run proper tests on this yet, so please check carefully that it doesn't introduce any other bugs.
I'm planning to rewrite the moveInside in a more readable and less duplicated way, anyway; but that will take more time

### DIFF
--- a/src/utils/polygonUtils.cpp
+++ b/src/utils/polygonUtils.cpp
@@ -238,7 +238,8 @@ unsigned int PolygonUtils::moveInside(const Polygons& polygons, Point& from, int
         bool projected_p_beyond_prev_segment = dot(p1 - p0, from - p0) >= vSize2(p1 - p0);
         for(const Point& p2 : poly)
         {   
-            // X = A + Normal( B - A ) * ((( B - A ) dot ( P - A )) / VSize( A - B ));
+            // X = A + Normal(B-A) * (((B-A) dot (P-A)) / VSize(B-A));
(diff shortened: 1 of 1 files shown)


rburema:
Pretty sure `x/y` does not give the same ratio as `x^2/y^2`.

Can you add a project or test where this'll help?

Also it's probably best to label this 'draft' until you have some tests done.

DerGenaue:
jup that is correct;
but check the code precisely:
before it was `(x/y)/y` ; now it is `x/(y*y)`
Naming might be a bit off

rburema:
I see. Yes that they're both called '...2' when one indicates power and the other indicates ... not-power, threw me for a loop.

DerGenaue:
You think that's a better naming?

rburema:
Much better. In my investigations it seems that this is only ever off by a single micron in practice.

Thanks for the fix!

I see what you mean by 'less duplicated' now as well. We seem to have 3 copies of the same method doing nearly identical things!

PR #1248, opened 2020-05-01
Now cubicsubdiv infill pattern can be rotated.

smartavionics:
Only uses first element from infill_angles.

This PR also fixes the regression that was introduced when the type of infill_angle was changed to AngleRadians.


smartavionics:
So this PR fixes https://github.com/Ultimaker/Cura/issues/7635.

Ghostkeeper:
Should be merged together with https://github.com/Ultimaker/Cura/pull/7640.

rburema:
Thank you once again!

PR #1250, opened 2020-05-04
Alternate wall direction

BagelOrb:
Alternate printing directions of walls with each inset and each layer in order to homogenize directional stress buildup.


smartavionics:
I would have thought that this could cause a lot of problems as the reversed polygons will look like holes and have -ve area. This PR is going to need some good testing to ensure that it has no adverse effects anywhere.

BagelOrb:
I made the PR so that some colleague of mine can do some testing before it actually gets merged.

It's not gonna be interesting for all materials, so it's just an option and disabled by default.

It will just be interesting for materials where the stress buildup has more significant effect than the tiny hole introduced by the 180* turn to print the next inset. For some materials the stress buildup is huge.

BagelOrb:
We are also getting external requests for this feature. His reasoning [here](https://github.com/prusa3d/PrusaSlicer/issues/4263#issuecomment-630649212) is solid and is approximately the same as the reason why we developed it (ask Johan).

Please consider merging this PR.

Ghostkeeper:
I agree with Smartavionics here that this is quite a dangerous implementation. Writing code, I normally assume that positive shapes are winding counter-clockwise and negative shapes are winding clockwise. So does the `ClipperLib::Area` function, which is used in a dozen places, and the `ClipperLib::Orientation` function similarly.

As a result, this implementation breaks all features which check for area or orientation, such as Minimum Tower-Supported Diameter, bridging, Fuzzy Skin Outside Only, and some utility functions like `ensureInsideOrOutside`. To test something easy and tangible going wrong, try enabling Fuzzy Skin Outside Only and and Alternate Wall Direction.

I think a better implementation would be located at the `FffGcodeWriter` stage, by iterating in reverse over the wall line segments. That would keep the actual parts in the correct orientation, but only print the walls in the correct order.

BagelOrb:
It's not changing the direction of original part, only those of the insets.

My implementation does seem problematic for Fuzzy skin on outside only, but I find it harder to believe it impacts support towers or bridging.

How about I perform the `polygon.reverse()` just before the call to `addPolygonsByOptimizer` ? Would that be okay?

BagelOrb:
Thanks for pointing out those weaknesses. I had underestimated the problem that Mark tried to point out. I'm sorry for that.

BagelOrb:
I've looked at the code and it's rather awkward to change the inset polygons in FffGcodeWriter because the whole SliceMeshStorage is const at that point.

Copying the insets and altering the copy is really cumbersome, because it would introduce changes all the way into InsetOrderOptimizer. That means a lot of changes and therefore very bug-prone.

Instead I opted to alter the insets at the very end of `processDerivedWallsSkinInfill`, which is called almost at the end of `FffPolygonGenerator`. However, `generateSupportInfillFeatures` is still called after it, but I assume that the gradual support etc doesn't depend on the walls; `generateOverhangAreas` is called way before `processDerivedWallsSkinInfill`...

Please see the new implementation.

With coasting enabled you can see in a single screenshot that it works (at least on the outer wall):
image

BagelOrb:
Somehow something went wrong during the last rebase just before the last push. Now it's fixed.

maruhe:
what happened to this feature?
I'd be happy to use it

jsourice:
> what happened to this feature?
> I'd be happy to use it

It's natively available on ideamaker now, for ceramic and metal filament that need sintering it's crucial.

BagelOrb:
This feature wasn't merged because it had to do with the walls, while the Arachne version of Cura would interfere with that, so it was decided to postpone merging this in favor of reimplementing it after Arachne would become a part of mainline Cura.

The fact that it has taken more than a year to merge Arachne means it might have been better to merge this feature after all, but that's a bit late now. Now it's more efficient to reimplement the feature after Arachne is merged and not bother with this version at all.

BagelOrb:
This feature has been reimplemented after Arachne was merged. This PR is stale and superfluous now.

rburema:
It turned out to be easier to just start again 'post'-Arachne. See: https://github.com/Ultimaker/CuraEngine/pull/1541 and https://github.com/Ultimaker/CuraEngine/pull/1545

Note: The first one seems 'closed', but the relevant commits have been cherry-picked. The last one is the same feature, but for support.

Will be in 5.0, but it might be that you need to make material profiles to adjust this setting (not enabled for frontend by default).

PR #1251, opened 2020-05-04
Cura 7318 ironing offsets with half a skin line width

konskarm:
Ironing is using the infill generator for the ironing path. The infill generator does not know that, so it generates "infill" taking into consideration the outer walls. As a result, despite the ironing inset at 0, the zigzag pattern was never reaching the outer wall while the concentric pattern was exceeding it. This PR fixes that by compensating for the outline_offsets before using the infill generator to generate the ironing path.

### DIFF
--- a/src/TopSurface.cpp
+++ b/src/TopSurface.cpp
@@ -37,7 +37,6 @@ bool TopSurface::ironing(const SliceMeshStorage& mesh, const GCodePathConfig& li
     const bool zig_zaggify_infill = pattern == EFillMethod::ZIG_ZAG;
     constexpr bool connect_polygons = false; // midway connections can make the surface less smooth
     const coord_t line_spacing = mesh.settings.get<coord_t>("ironing_line_spacing");
-    const coord_t outline_offset = -mesh.settings.get<coord_t>("ironing_inset");
     const coord_t line_width = line_config.getLineWidth();
     const std::vector<AngleDegrees>& top_most_skin_angles = (mesh.settings.get<size_t>("roofing_layer_count") > 0) ? mesh.roofing_angles : mesh.skin_angles;
     assert(top_most_skin_angles.size() > 0);
@@ -45,6 +44,16 @@ bool TopSurface::ironing(const SliceMeshStorage& mesh, const GCodePathConfig& li
     constexpr coord_t infill_overlap = 0;
     constexpr int infill_multiplier = 1;
     constexpr coord_t shift = 0;
+
+    coord_t ironing_inset = -mesh.settings.get<coord_t>("ironing_inset");
+    if (pattern == EFillMethod::ZIG_ZAG && ironing_inset == 0)
+    {
+        //Compensate for the outline_offset decrease that takes place when infill is generated for the zigzag pattern
+        const float width_scale = (float)mesh.settings.get<coord_t>("layer_height") / mesh.settings.get<coord_t>("infill_sparse_thickness");
+        ironing_inset += width_scale * line_width / 2;
+    }
+    const coord_t outline_offset = ironing_inset;
+
     Infill infill_generator(pattern, zig_zaggify_infill, connect_polygons, areas, outline_offset, line_width, line_spacing, infill_overlap, infill_multiplier, direction, layer.z - 10, shift);
     Polygons ironing_polygons;
     Polygons ironing_lines;
--- a/src/TopSurface.cpp
+++ b/src/TopSurface.cpp
@@ -52,6 +52,11 @@ bool TopSurface::ironing(const SliceMeshStorage& mesh, const GCodePathConfig& li
         const float width_scale = (float)mesh.settings.get<coord_t>("layer_height") / mesh.settings.get<coord_t>("infill_sparse_thickness");
         ironing_inset += width_scale * line_width / 2;
     }
+    else if (pattern == EFillMethod::CONCENTRIC)
+    {
+        //Counteract the outline_offset increase that takes place when infill is generated for the concentric pattern
+        ironing_inset -= line_width / 2;
+    }
     const coord_t outline_offset = ironing_inset;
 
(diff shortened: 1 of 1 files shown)


src/TopSurface.cpp
```diff
@@ -37,14 +37,28 @@ bool TopSurface::ironing(const SliceMeshStorage& mesh, const GCodePathConfig& li
...
     constexpr coord_t infill_overlap = 0;
     constexpr int infill_multiplier = 1;
     constexpr coord_t shift = 0;
+
+    coord_t ironing_inset = -mesh.settings.get<coord_t>("ironing_inset");
+    if (pattern == EFillMethod::ZIG_ZAG && ironing_inset == 0)
+    {
+        //Compensate for the outline_offset decrease that takes place when using the infill generator to generate ironing with the zigzag pattern
+        const float width_scale = (float)mesh.settings.get<coord_t>("layer_height") / mesh.settings.get<coord_t>("infill_sparse_thickness");
+        ironing_inset += width_scale * line_width / 2;
```
Ghostkeeper, line 53: Somehow this doesn't seem to be working correctly. I tested this with a 1cm cube centered around the origin, and with the Ironing Inset at 0, and I got the following g-code to iron it:

```
;TYPE:SKIN
G1 F1500 E111.37361
G1 F1200 X4.999 Y-4.899 E111.37369
G1 X4.999 Y-4.758 E111.37378
G1 X4.758 Y-4.999 E111.374
G1 X4.616 Y-4.999 E111.37409
G1 X4.999 Y-4.616 E111.37443
G1 X4.999 Y-4.475 E111.37451
G1 X4.475 Y-4.999 E111.37498
G1 X4.333 Y-4.999 E111.37507
G1 X4.999 Y-4.333 E111.37566
G1 X4.999 Y-4.192 E111.37575
G1 X4.192 Y-4.999 E111.37646
```
[Zipped project file](https://github.com/Ultimaker/CuraEngine/files/4616309/iron_around_centre.zip)

You can see that it goes to X4.999 and Y-4.999. Bar a rounding error of 1 micron, this is exactly to the border with the centreline of the ironing line.

I'd expect that it goes to the edge of the top surface with the edge of the ironing line. In that project file the Skin Line Width is 0.4mm and the ironing flow is 10% of that, so it should keep 0.04mm away from the border.

That is, unless we consider this the new expected behaviour.

src/TopSurface.cpp
```diff
@@ -37,14 +37,28 @@ bool TopSurface::ironing(const SliceMeshStorage& mesh, const GCodePathConfig& li
...
     const AngleDegrees direction = top_most_skin_angles[layer.getLayerNr() % top_most_skin_angles.size()] + AngleDegrees(90.0); //Always perpendicular to the skin lines.
     constexpr coord_t infill_overlap = 0;
     constexpr int infill_multiplier = 1;
     constexpr coord_t shift = 0;
+
+    coord_t ironing_inset = -mesh.settings.get<coord_t>("ironing_inset");
+    if (pattern == EFillMethod::ZIG_ZAG && ironing_inset == 0)
+    {
+        //Compensate for the outline_offset decrease that takes place when using the infill generator to generate ironing with the zigzag pattern
+        const float width_scale = (float)mesh.settings.get<coord_t>("layer_height") / mesh.settings.get<coord_t>("infill_sparse_thickness");
```
Ghostkeeper, line 52: Infill and the layer height have nothing to do with this, right? This is using the skin line width.

Also, use the Ratio type for this variable.

src/TopSurface.cpp
```diff
@@ -37,14 +37,28 @@ bool TopSurface::ironing(const SliceMeshStorage& mesh, const GCodePathConfig& li
...
+        ironing_inset += width_scale * line_width / 2;
+    }
+    else if (pattern == EFillMethod::CONCENTRIC)
+    {
+        //Counteract the outline_offset increase that takes place when using the infill generator to generate ironing with the concentric pattern
+        ironing_inset -= line_width / 2;
```
Ghostkeeper, line 58: This inset also doesn't seem to be working properly. I get this g-code with a 1cm cube centred around the origin, towards the very end:

```
G0 F7200 X4.9 Y4.9
G1 F1200 X-4.9 Y4.9 E111.96965
G1 X-4.9 Y-4.9 E111.97579
G1 X4.9 Y-4.9 E111.98194
G1 X4.9 Y4.9 E111.98808
;TIME_ELAPSED:670.436642
```
This is the last and outermost contour of ironing. It's offset from the original shape by 0.1mm here, which is way more than the `0.4mm / 2 * 10% = 0.02mm` that I'd expect.

To reproduce, use the same project file as mentioned before, but change the ironing pattern to concentric.

src/TopSurface.cpp
konskarm: The infill generator uses the width_scale when getting the offset for the zigzag pattern (line 101-105 in TopSurface.cpp https://github.com/Ultimaker/CuraEngine/blob/master/src/infill.cpp#L103). When ironing is done using the infill generator, we must counteract this offset in order to get exactly an ironing inset of 0. This is why width scale is used here.

src/TopSurface.cpp
konskarm: Fixed in https://github.com/Ultimaker/CuraEngine/pull/1251/commits/b49ddf7c5c73c76c5a6aac788cb8acafeef60794 by moving the ironing line by ironing_line_width/2 inwards.

src/TopSurface.cpp
konskarm: I did not take into account the ironing flow. Fixed in https://github.com/Ultimaker/CuraEngine/pull/1251/commits/b49ddf7c5c73c76c5a6aac788cb8acafeef60794 by moving the ironing line by ironing_line_width/2 inwards.

src/TopSurface.cpp
Ghostkeeper: Ah, strange that it does that. The Infill Layer Thickness shouldn't be applicable to skin at all, I'd think. Nor to the parts of infill that are thinner than the Infill Layer Thickness. Fine then, but we might be looking at a bug here or at least some needlessly complex code. But if the infill generator is at fault it could be hard to fix now.

PR #1210, opened 2020-03-04
Use Arachne to vary line width

Ghostkeeper:
For now, this is just a WIP comparison to be able to offer review comments on libArachne.

Do not merge!

image

| 4.8 behaviour | Arachne behaviour |
|--------------------|---------------------------|
| alt text | alt text |

See CURA-7235.

PR #1252, opened 2020-05-08
Add closing operation on inner walls.

rburema:
This will prevent rounding errors from merging holes with each other too early. We might want to skip this and just wait until we get libArachne properly merged, which will _also_ solve this problem.
CURA-7424

### DIFF
--- a/src/WallsComputation.cpp
+++ b/src/WallsComputation.cpp
@@ -96,6 +96,9 @@ void WallsComputation::generateInsets(SliceLayerPart* part)
             }
         }
 
+        // Closing operation and epsilon to prevent rounding errors from merging hole polygons to early:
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
I don't really know if this would be a good solution. I think you had the same feeling. My first instinct is that this could break things by accidentally fusing walls together. However if I think hard about it I can't think of a case where you wouldn't want that to happen.

Ghostkeeper:
After discussion with the team, we've decided to close this PR in favour of #1210 . This could erase very small holes in the model, although they would still be fused together by virtue of being so close together. In general, all developers had a bad feeling about this although we couldn't logically give a counterexample of why this PR is bad. We think #1210 is likely a more reliable way to fix this.

PR #1255, opened 2020-05-19
CURA-7448 refactor temporary maps into fields

rburema:
SkeletalTrapezoidation used some structures that where determined should reside in SkeletalTrapezoidationEdge instead.

Uses smart pointers to make a weak reference in the edge, so that it's all cleaned when it should be, and it's semantically clear that the edge doesn't necessarily always have those fields.

### DIFF
--- a/src/SkeletalTrapezoidation.h
+++ b/src/SkeletalTrapezoidation.h
@@ -56,7 +56,7 @@ class SkeletalTrapezoidation
     using TransitionEnd = SkeletalTrapezoidationEdge::TransitionEnd;
 
     template<typename T>
-    using temp_data_t = std::vector<std::shared_ptr<T>>;
+    using ptr_vector_t = std::vector<std::shared_ptr<T>>;
 
     float transitioning_angle; //!< How pointy a region should be before we apply the method. Equals 180* - limit_bisector_angle
     coord_t discretization_step_size; //!< approximate size of segments when parabolic VD edges get discretized (and vertex-vertex edges)
@@ -198,7 +198,7 @@ protected:
      */
     bool filterUnmarkedRegions(edge_t* to_edge, coord_t bead_count, coord_t traveled_dist, coord_t max_dist);
 
-    void generateTransitionMids(temp_data_t<std::list<TransitionMiddle>>& edge_transitions);
+    void generateTransitionMids(ptr_vector_t<std::list<TransitionMiddle>>& edge_transitions);
 
     void filterTransitionMids();
 
@@ -216,7 +216,7 @@ protected:
 
     bool filterEndOfMarkingTransition(edge_t* edge_to_start, coord_t traveled_dist, coord_t max_dist, coord_t replacing_bead_count);
 
-    void generateTransitionEnds(temp_data_t<std::list<TransitionEnd>>& edge_transition_ends);
+    void generateTransitionEnds(ptr_vector_t<std::list<TransitionEnd>>& edge_transition_ends);
 
     /*!
      * Also set the rest values at nodes in between the transition ends
@@ -228,7 +228,7 @@ protected:
     /*!
      * \param edge_to_transition_mids From the upward halfedges to their transitions mids
      */
-    void generateTransition(edge_t& edge, coord_t mid_R, coord_t transition_lower_bead_count, temp_data_t<std::list<TransitionEnd>>& edge_transition_ends);
+    void generateTransition(edge_t& edge, coord_t mid_R, coord_t transition_lower_bead_count, ptr_vector_t<std::list<TransitionEnd>>& edge_transition_ends);
 
     /*!
      * \p start_rest and \p end_rest refer to gap distances at the start and end pos in terms of ratios w.r.t. the inner bead width at the high end of the transition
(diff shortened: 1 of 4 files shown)


src/SkeletalTrapezoidationEdge.h
```diff
@@ -39,11 +70,53 @@ class SkeletalTrapezoidationEdge
...
+
+    bool hasExtrusionJunctions(bool ignore_empty = false) const
+    {
+        return extrusion_juntions_.use_count() > 0 && (ignore_empty || ! extrusion_juntions_.lock()->empty());
+    }
+    void initExtrusionJunctions(std::shared_ptr<std::vector<ExtrusionJunction>> storage)
+    {
+        extrusion_juntions_ = storage;
+    }
+    std::shared_ptr<std::vector<ExtrusionJunction>> extrusion_juntions()
```
nallath, line 108: Shouldn't this be extrusionJunctions() or getExtrusionJunctions() (Or well at least extrusion_junctions, because I think the missing c should be added :P)

rburema: Probably, and yes.

src/SkeletalTrapezoidation.h
```diff
@@ -229,27 +210,27 @@ class SkeletalTrapezoidation
...
+    void generateTransitionEnds(temp_data_t<std::list<TransitionEnd>>& edge_transition_ends);
 
     /*!
      * Also set the rest values at nodes in between the transition ends
      */
-    void applyTransitions(std::unordered_map<edge_t*, std::list<TransitionEnd>>& edge_to_transition_ends);
+    void applyTransitions();
 
-    void filterMarkedLocalOptima();
+    //void filterMarkedLocalOptima(); // < -- function missing?
```
nallath, line 226: If it's missing, just remove it altogether, right?

src/SkeletalTrapezoidation.h
```diff
@@ -51,6 +52,11 @@ class SkeletalTrapezoidation
     using edge_t = HalfEdge<SkeletalTrapezoidationJoint, SkeletalTrapezoidationEdge>;
     using node_t = HalfEdgeNode<SkeletalTrapezoidationJoint, SkeletalTrapezoidationEdge>;
     using Beading = BeadingStrategy::Beading;
+    using TransitionMiddle = SkeletalTrapezoidationEdge::TransitionMiddle;
+    using TransitionEnd = SkeletalTrapezoidationEdge::TransitionEnd;
+
+    template<typename T>
+    using temp_data_t = std::vector<std::shared_ptr<T>>;
```
nallath, line 59: I'm not a big fan of this name. Sure, in all cases where you use this vector of pointers, you use it as a temporary data storage, but that's not what this structure is (It's just how you happen to use it, right?). 

That being said, I'm also not sure how we should name this. Perhaps vector_t?

src/SkeletalTrapezoidationEdge.h
```diff
@@ -39,11 +70,53 @@ class SkeletalTrapezoidationEdge
...
+
+    bool hasTransitionEnds(bool ignore_empty = false) const
+    {
+        return transition_ends_.use_count() > 0 && (ignore_empty || ! transition_ends_.lock()->empty());
+    }
+    void initTransitionEnds(std::shared_ptr<std::list<TransitionEnd>> storage)
+    {
+        transition_ends_ = storage;
+    }
+    std::shared_ptr<std::list<TransitionEnd>> transition_ends()
```
nallath, line 95: It's technically a function, so it should be transitionEnds (or getTransitionEnds).

src/SkeletalTrapezoidationEdge.h
```diff
@@ -39,11 +70,53 @@ class SkeletalTrapezoidationEdge
...
+    {
+        return extrusion_juntions_.lock();
+    }
+
 private:
     int_least8_t is_marked; //! whether the edge is significant; whether the source segments have a sharp angle; -1 is unknown
-};
-
 
+    std::weak_ptr<std::list<TransitionMiddle>> transitions_;
```
nallath, line 116: Nitpicking, but we usually use _transitions to mark something as protected (or well, it's because of python). But should we in the case of C++ even make a distinction? I don't think we do that in the rest of the code as far as I know.

src/SkeletalTrapezoidationEdge.h
```diff
@@ -39,11 +70,53 @@ class SkeletalTrapezoidationEdge
     {
         return is_marked >= 0;
     }
+
+    bool hasTransitions(bool ignore_empty = false) const
+    {
+        return transitions_.use_count() > 0 && (ignore_empty || ! transitions_.lock()->empty());
+    }
+    void initTransitions(std::shared_ptr<std::list<TransitionMiddle>> storage)
```
nallath, line 78: Nitpicking; it's not really initializing them, it's more of a setter (if there is a specific reason for it to be an init, I would like to see that in the documentation)

src/SkeletalTrapezoidationEdge.h
```diff
@@ -1,9 +1,14 @@
-//Copyright (c) 2019 Ultimaker B.V.
+//Copyright (c) 2020 Ultimaker B.V.
 
 
 #ifndef VORONOI_QUADRILATERALIZATION_EDGE_H
 #define VORONOI_QUADRILATERALIZATION_EDGE_H
```
nallath, line 5: Hmm, the header guard has a different name than the file name.

src/SkeletalTrapezoidation.h
rburema: Ah I see, if we're going for a _type_ type, we can probably make it `ptr_vector_t` then.

src/SkeletalTrapezoidationEdge.h
rburema: I used this because the function names would conflict (that is also why the function names are this way ... because I would've liked to use them as fields), not an issue anymore when I rename the functions to what you suggested.

I never know which side the `_` goes on anymore, since I've been forced to use two styles that used opposite endings a number of times :-p

src/SkeletalTrapezoidationEdge.h
rburema: True. I just needed an initializer, but I guess the implementation is broad enough to be a setter. I'll change it.

src/SkeletalTrapezoidationEdge.h
rburema: I should be less laser focused and do some more scouting during tickets sometimes... :-)

Issue #1265, opened 2020-06-05
zig_zaggify_infill creates 1/2 linewidth gap between infill and perimeters


richfelker:
Without and with `zig_zaggify_infill`, all other settings same:

![Screenshot_2020-06-05 G-Code Analyser - Analyse your 3D printing G-Code to provide accurate information such as print time ](https://user-images.githubusercontent.com/10458007/83836281-7db42880-a6e2-11ea-9b3e-22f7b8f5a8d0.png)
![Screenshot_2020-06-05 G-Code Analyser - Analyse your 3D printing G-Code to provide accurate information such as print time  (1)](https://user-images.githubusercontent.com/10458007/83836283-7ee55580-a6e2-11ea-86d9-106b0f599ed2.png)

Adding 1/2 line width (0.4/2 = 0.2) to the `infill_overlap_mm` suffices to work around this, but I think it's a bug that this is needed.

smartavionics:
Hi, please provide a project file (File -> Save, zip or rename, attach) that illustrates this issue. Thanks.

richfelker:
I'm not using the Cura gui, just CuraEngine. I'll try to make up a self-contained example (stl file and minimal json CuraEngine config file) and post it here but the problem does not seem to be hard to reproduce. It was happening all over the place and was the source of all my prints coming out britle wherever they had flat vertical walls (no support from the infill).

richfelker:
Also, I checked whether there are any Python-expression values in `fdmprinter.def.json` that depend on the value of `zig_zaggify_infill` to derive other settings, where failure to duplicate that derivation might be the cause, but didn't find any. So it doesn't seem likely that this issue is specific to command line use of CuraEngine.

richfelker:
@smartavionics: This issue is not present in your fork, and likely isn't present in upstream master either. I suspect it was a regression that was already fixed that just happens to be present in the version I had built, and I'm trying to ascertain exactly which revision that is. I'll update or close this as appropriate once I find out.

richfelker:
Bug is present in e904e260 and current master, but was not present in some earlier build I have that I can't identify the exact commit it's from.

richfelker:
This is reproducible with 4.6.1 tag and the above mentioned versions, using nothing but:

```
CuraEngine slice -j fdmprinter.def.json -s layer_height=0.2 \
 -s zig_zaggify_infill=true -l cube.stl -o cube.gcode
```

[...]

PR #1267, opened 2020-06-06
Don't reduce the infill outline offset unduly when infill_sparse_thickness < layer_height.

smartavionics:
Does it even make sense for infill_sparse_thickness to be < layer_height? I don't know but
let's not get caught out if it is.

As discussed in https://github.com/Ultimaker/CuraEngine/issues/1265


Ghostkeeper:
I think this is the wrong fix. See my comment here: https://github.com/Ultimaker/CuraEngine/issues/1265#issuecomment-640481271

Otherwise, this would also not give the correct results if the infill layer thickness is a fractional thickness higher than 1, e.g. 1.5x the layer height. You'd get the lines to overlap again. You'd need to use `round_divide` like how the actual infill layer thickness is calculated to get it to align (incorrectly) with the edge of the infill area.

smartavionics:
No longer required....

PR #1268, opened 2020-06-07
Shrink layer's polygons and then expand the parts they produced to avoid parts splitting later.

smartavionics:
Otherwise, if a polygon narrows to an outer line width or less wide, it would be split
later when the insets are created thus producing a part with multiple outlines.

Fixes the issue discussed in https://community.ultimaker.com/topic/32796-starting-points-out-of-control


smartavionics:
Closing this PR because although it does fix the above mentioned issue, it also causes problems as it removes thin walls.

PR #1269, opened 2020-06-07
Fix merging of skin areas on different layers for skin edge support.

smartavionics:
It now leaves a tiny gap between the areas when they are combined. Before, without the
gap, areas on different layers could merge together thus losing some skin edge support.

Fixes https://github.com/Ultimaker/Cura/issues/7792


smartavionics:
Just noticed that this PR can produce excess skin support with some prints so I'm doing more work on it.

smartavionics:
I think this is worth looking at now.

rburema:
Thanks again for all these fixes!

We thought this change was a bit too large to put into 4.7, so unfortunately people that don't have your build(s) will have to wait a few months for this one (4.8 probably), but at least there's some progress.

(Why can I still close this when it's already merged? Just because I didn't do it online for this one?)

smartavionics:
Understood, no problem.

PR #1273, opened 2020-06-10
CURA-5990 Standard up! Enable use of C++17

rburema:


src/utils/LazyInitialization.h
```diff
@@ -92,18 +91,18 @@ class LazyInitialization : public std::optional<T>
      */
     T& operator*()
     {
-        if (!std::optional<T>::instance)
+        if (!std::optional<T>::has_value())
         {
-            std::optional<T>::instance = constructor();
+            std::optional<T>::operator=(*constructor());
```
BagelOrb, line 96: This is performing a copy of the object returned by the constructor.

Please refer to my PR, i.e. the branch called `cpp_17`

BagelOrb (dismissed):
Perhaps a bit bold of me to use the 'Request changes' option of github given that I am not in the Cura team, but please look at my comment.

Your code performs an unneeded copy right after construction of the object, which in one instance causes the object to live both on the heap and on the stack (assuming that the LazyObject was allocated on the stack).

src/utils/LazyInitialization.h
BagelOrb: https://github.com/Ultimaker/CuraEngine/pull/1398

Specifically, see the solution at line https://github.com/Ultimaker/CuraEngine/pull/1398/files#diff-097df74eec44f37f38cd9e42cf9e09928c7500d0887d8e367c042797445ef53fR136

rburema: (This can now progress again, though it's not sure when we'll pick it up.)

Your link leads to this highlighted line:

> LazyInitialization<std::unique_ptr<LocToLineGrid>, Comb*, const coord_t> outside_loc_to_line;

(snipped comment)

Instead, I assume you mean this?

> T new_object = constructor();
> std::optional<T>::operator=(std::move(new_object));

In this case, however, there still is a copy, just in the first line.

BagelOrb: No, that's not what I meant. The link was correct.

You shouldn't perform the copy of the object; you should prevent the copy from happening. However, an optional object *is* the object, so using `optional<object>` has to perform the copy. So the question is how we can prevent the copy from happening.

Instead of using `optional<object> = copy` you should do `optional<object*> = original`. That way you prevent the copy.

My PR does refactor all those cases where object copies would otherwise have taken place.

BagelOrb: No I mean that we should prevent the copy, not by changing the copy constructor, but be preventing it to be called in the first place.

So instead of an `optional<grid>` we have an `optional<grid*>` so that we only copy the pointer, not the underlying data.

rburema: We're _finally_ moving ahead with this!

Remarks resolved by 'manual cherry-pick' in: https://github.com/Ultimaker/CuraEngine/pull/1273/commits/ef72c2a161eaaebe9e353c712d8df1adb32f02c4

Ghostkeeper (approved):

BagelOrb:
🎉

PR #1275, opened 2020-06-15
Fix remaining review issues

Ghostkeeper:
This fixes all remaining review issues from the original review of libArachne (before it's integrated in the CuraEngine code). It's mostly documentation, but also a bunch of constness, re-using of code, cleaning up, re-using code, and a few fixes.

Fixes issue CURA-7373.


src/utils/polygon.cpp
```diff
@@ -105,21 +106,24 @@ void Polygons::makeConvex()
...
+            const Point& c = poly[(i + 1) % poly.size()];
+            if (LinearAlg2D::pointIsLeftOfLine(b, a, c) < 0)
             {
                 poly.remove(i);
             }
             else
             {
+                convexified.path->push_back(b);
                 a = b;
                 i++;
```
nallath, line 123: Should this still happen?

Ghostkeeper: Oh, maybe I made a mistake with the commit here. The entire case should be turned around (since the `if` condition is turned around). I thought I'd deleted the `poly.remove(i)` as well, but maybe I stashed stuff in between or something. This is definitely not correct.

PR #1278, opened 2020-06-24
Fix read STL file error

WinstonMao:
The ASCII value of a character may be decimal 26 (0x1A, \SUB, SUBSTITUTE) in a binary STL file.Opening a file in “r” mode may return an error

### DIFF
--- a/src/MeshGroup.cpp
+++ b/src/MeshGroup.cpp
@@ -195,7 +195,7 @@ bool loadMeshSTL_binary(Mesh* mesh, const char* filename, const FMatrix3x3& matr
 
 bool loadMeshSTL(Mesh* mesh, const char* filename, const FMatrix3x3& matrix)
 {
-    FILE* f = fopen(filename, "r");
+    FILE* f = fopen(filename, "rb");
     if (f == nullptr)
     {
         return false;


WinstonMao:
[error_file.zip](https://github.com/Ultimaker/CuraEngine/files/4823442/error_file.zip)

ninovanhooff:
Internal reference: CURA-7577

ninovanhooff:
Opened and sliced the provided file on Ubuntu 18 both from source and in the 4.6.1 appimage. No error encountered.
What operating system and version of cura did you use?

ninovanhooff:
Another question is what will happen to non-binary stl files with this fix applied

WinstonMao:
> Opened and sliced the provided file on Ubuntu 18 both from source and in the 4.6.1 appimage. No error encountered.
> What operating system and version of cura did you use?
> W
> Two of our developers on windows also encountered no problems
This problem occurs when using CuraEngine in commandline mode.When you slice it with Cura,the error file is not read by CuraEngine .Actually, Cura reads the stl file and sends its triangle data to curaengine with Arcus comunication.

And non-binary stl files with this fix applied will also work too.

WinstonMao:
> Another question is what will happen to non-binary stl files with this fix applied

This problem occurs when using CuraEngine in commandline mode.When you slice it with Cura,the error file is not read by CuraEngine .Actually, Cura reads the stl file and sends its triangle data to curaengine with Arcus comunication.

And non-binary stl files with this fix applied will also work too.

PR #1280, opened 2020-06-29
Cura 7199 remove spaghetti infill

jellespijker:
Removed spaghetti infill from the engine

See also https://github.com/Ultimaker/Cura/pull/7995


src/FffPolygonGenerator.cpp
```diff
@@ -722,50 +721,42 @@ void FffPolygonGenerator::processInfillMesh(SliceDataStorage& storage, const siz
...
-                mesh.cross_fill_provider = new SierpinskiFillProvider(mesh.bounding_box, mesh.settings.get<coord_t>("infill_line_distance"), mesh.settings.get<coord_t>("infill_line_width"), cross_subdisivion_spec_image_file);
-            }
-            else
-            {
-                if (cross_subdisivion_spec_image_file != "" && cross_subdisivion_spec_image_file != " ")
-                {
-                    logError("Cannot find density image \'%s\'.", cross_subdisivion_spec_image_file.c_str());
-                }
-                mesh.cross_fill_provider = new SierpinskiFillProvider(mesh.bounding_box, mesh.settings.get<coord_t>("infill_line_distance"), mesh.settings.get<coord_t>("infill_line_width"));
+                logError("Cannot find density image \'%s\'.", cross_subdisivion_spec_image_file.c_str());
```
ninovanhooff, line 757: Guessing this is a typo: subdivision ?
I guess this would be a good opportunity to fix those in a separate commit

jellespijker: Done! also fixed a Clang-tidy warning about checking against empty instead of an empty object

PR #1282, opened 2020-07-02
New Tree Support Implementation

ThomasRahm:
# Tree Support V2
I reimplemented the tree support, as the current implementation does not meet my expectations.   
[I have a fork](https://github.com/ThomasRahm/CuraEngine) where a compiled version of a Cura with this Tree Support modification can be downloaded.

# How the new tree support looks:
New tree support

## Feedback I have gotten recently 

_"I just want to thank you for these. With your stock settings I am getting amazing results . As someone coming from Prusa/superslicer , and new to any version of Cura, i'm shocked that supports can come off this cleanly , and leave no trace on the print."_[<sup>1</sup>](https://www.reddit.com/r/3Dprinting/comments/p7n4x4/comment/iot9pov/?utm_source=share&utm_medium=web2x&context=3)

_"I've been having a bunch of issues with the Main Cura tree supports not properly getting all of my overhangs, I just installed your fork and it looks like a vast improvement! "_[<sup>2</sup>](https://www.reddit.com/r/3Dprinting/comments/p7n4x4/i_created_a_new_tree_support_implementation_for/?sort=new)

_"The tree supports in Cura create a massive trunk that can surround whole parts in some cases, I always wanted a tool to cut archs out of those trunks to reduce the amount of filament wasted, or an arch-shaped blocker that would achieve the same result, but your tree supports seem to achieve a similar result"_[<sup>3</sup>](https://www.reddit.com/r/3Dprinting/comments/p7n4x4/comment/ikq49ho/?utm_source=share&utm_medium=web2x&context=3)

_"What you've done here is awesome work, and I thank you for it."_[<sup>4</sup>](https://www.reddit.com/r/3Dprinting/comments/p7n4x4/comment/iinkio8/?utm_source=share&utm_medium=web2x&context=3)

_"I want to show to all of you the good work of this mod, in my opinion it is a good support way."_[<sup>5</sup>](https://community.ultimaker.com/topic/40567-cura-tree-support-mod-ts2/)

_"I'm still using this as my daily and tbh I think it's a shame it's getting slept on by not just Cura but the likes of prusaslicer/superslicer/etc. It's so much better than the stock Cura tree supports"_[<sup>6</sup>](https://github.com/ThomasRahm/CuraEngine/issues/3#issuecomment-1214519257)

## Currently known issues:
* Slicing with cross support and 0% support density causes the slice progress to hang. This is not an issue of my implementation and to my understanding also occurs in the official version. [Bug report of this issue](https://github.com/Ultimaker/CuraEngine/issues/1412).  
* There are issues when `Minimum Support X/Y Distance` is larger than `Support X/Y Distance`. This can trigger an error message if the version displaying them is used. I will not fix this one, except by maybe disallowing it in the UI, as I think the assumption that a minimum is not larger than the expected one is reasonable.  

# Settings:

## Recommended settings

**To see the tree support specific settings set the `Setting Visibility` to `all`.**

I recommend enabling roofs and setting a reasonable `Minimum Support Roof Area` (I use 10 mm²). My implementation will make sure to correctly support the lines generated by the support roof. If no roof is possible the branches will try to avoid moving for as many layer as the roof is thick. They will generate as regular tree support branches, as i have serious doubts about the stability of roofs in branches.  
Do not use concentric floor. It will generate as intended (by my understanding at least), but I suspect it will not print (it will be obvious when you see it) when support infill is low.  
**Set `Minimum support area` to 0**, higher values may cause very pointy overhangs to not be supported (as intended by my current understanding of this feature)

## New Tree Support Settings

The changes for the `fdmprinter.def.json` can be found [here.](https://github.com/ThomasRahm/Cura/blob/TS2/resources/definitions/fdmprinter.def.json)

<details>
<summary>Tree Support Preferred Branch Angle vs Tree Support Maximum Branch Angle</summary>

`Tree Support Maximum Branch Angle` is ONLY used when `Tree Support Preferred Branch Angle` is not enough to avoid the model. This causes branches to only move with `Tree Support Preferred Branch Angle` to merge with other branches. 

Idea behind this is that `Tree Support Preferred Branch Angle` is a very stable angle, but if some parts could not be supported with the smaller angle the larger angle can be used to reach these parts.  

Slow vs Fast branch angle  

</details>

<details>
<summary>Tree Support Diameter Increase To Model</summary>

Branches that can reach the buildplate and branches that cannot are able to merge. This setting limits how much the diameter of the branch connecting with the model can be increased by such merges.  

This setting is intended to specify how important it is that connections with the model are avoided (the lower this setting, the less contact area with the model).

Small Diameter Increase To Model:  
Small Diameter increase to model

Large Diameter Increase To Model:  
Large Diameter increase to model

</details>

<details>
<summary>Tree Support Minimum Height To Model</summary>

If only a few layers space are left between the position where the branch begins and where it ends, removing the support is impossible. This setting removes all branches that have a height lower than `Tree Support Minimum Height To Model` to prevent this. Only branches that rest on the model are removed.

Small Minimum Height To Model:  
Small Diameter increase to model

Large Minimum Height To Model:  
Large Diameter increase to model

</details>

<details>
<summary>Tree Support Initial Layer Diameter</summary>

Branches try to ensure to reach this diameter at layer 0.

Small Initial Layer Diameter:  
Initial Layer Diameter

Large Initial Layer Diameter:  
Initial Layer Diameter

</details>

<details>
<summary>Tree Support Branch Density</summary>

A percentage of space that has to be filled by the top most layer of a branch aka the layer that supports the overhang. When a roof is supported the overhang is the line of the roof. When no support roof is used it behaves similarly to the supporting (top most) layer when using regular support. High values (depends on your other settings) of >35% can cause the resulting support areas to overlap. This causes cura to merge them to a single area, as cura assumes every support area is supported by support areas below. If a low support infill is used this assumption is wrong. There is nothing reasonable I can do to prevent this, so in this case adjust your support infill accordingly.

Simulated regular support pattern:   
Simulate regular support pattern

</details>

<details>
<summary>Tree Support Tip Diameter</summary>

Defines the diameter of the top most support area of a branch.
Using a large diameter without having a reasonable large support density would obviously not print well.

<details>
<summary>Can replace tips with roof!</summary>

Happens automatically when `Minimum Support Roof Area` is smaller than a circle with diameter `Tree Support Tip Diameter` and roofs are enabled.

Large Tip Diameter with minimum roof area being small:  
Tip radius large interface 

Large Tip Diameter with minimum roof area being large:  
Tip Radius large2

</details>

Small Tip Diameter:  
Small Tip Diameter

Large Tip Diameter:  
Large Tip Diameter

</details>

<details>
<summary>Support Interface Priority</summary>

This setting defines how support roof and regular support (e.g. from other branches) interact.  

Support preferred(Support Area Overwrites Interface Area):  
Support Area Overwrites Interface Area  

Interface preferred(Interface Area Overwrites Support Area):  
Interface Area Overwrites Support Area  

Support lines preferred(Support Lines Overwrite Interface Area):  
Support Lines Overwrite Interface Area  

Interface Lines preferred(Interface Lines Overwrite Support Area. This one will not be what you want when Support Wall Count > 0):  
Interface Lines Overwrite Support Area  

Both Overlap (This causes filament be extruded twice at the same location (Once the support and once the roof). This may cause issues when printing. Use with care!):  
Both Overlap  

</details>

<details>
<summary>Tree Support Limit Branch Reach and Tree Support Optimal Branch Range</summary>

Limit how far each branch should travel from the point it supports. This can make the support more sturdy, but will increase the amount of branches. Branches can violate this value to reach their destination.

Tree Support Limit Branch Reach Disabled:  
Tree Support Limit Branch Reach Disabled
Tree Support Limit Branch Reach Enabled with a small distance:  
Tree Support Limit Branch Reach Enabled with a small distance

</details>

<details>
<summary>Tree Support Rest Preference</summary>

The preferred placement of the support structures. Note that the buildplate is assumed to be also a flat surface <sup>(If not i think you have bigger issues than this setting)!</sup>  
Only visible when the support can rest on the model.

Buildplate:  
Buildplate
On any flat surface:  
On any flat surface

</details>

`Tree Support Branch Diameter` and `Tree Support Branch Diameter Angle` behave like in the current tree support implementation, but `Tree Support Branch Diameter Angle` is ignored when a diameter increase would invalidate a branch, as a small branch has a higher chance of supporting the model than no branch at all.

# What does it do better:
<details>
<summary>Uses significant less filament.</summary>

Also makes the branches of my implementation much easier to remove from the model.  

Current tree support:  
Current tree support

My tree support:  
New tree support

Comparison of material usage:  
Comparison

Disclaimer: This is only the material used for support structures, not the whole model. The measuring unit is mm of 1.75mm diameter filament, as displayed in Cura. Measurement was done in the 4.7.1 version, and exact numbers could differ slightly. 

</details>

<details>
<summary>Can handle large branch angles.</summary>

Current tree support:  
Current tree support

My tree support:  
New tree support

</details>

<details>
<summary>Supports Roofs more consistently.</summary>

The images show the bottom-most roof layer and a few layers of the branches below. In the current tree support some lines of the roof have a large distance from one point where support is below it to the next. My implementation ensures consistent spacing.

Current tree support:  
Current tree support

My tree support:  
New tree support

</details>

<details>
<summary>Branches avoid support blocker.</summary>

Branches avoid support blocker

</details>

<details>
<summary>Setting applicable now on a per-mesh basis.</summary>
All tree support related settings are now settable per mesh and support related settings are correctly used by the tree support. This may increase slicing time drastically for settings that are not only relevant on the top most support layers.  

So to avoid slicing times being longer by factor 10 try not to slice 10 different models with different branch radii (Slicing 2 models with different settings should not take much longer than slicing 2 times).  
  
Trees with settings that are incompatible with each other can not merge and will avoid each other. 

Few settings differences allow merges:  
Different interface settings can merge

</details>

<details>
<summary>More settings of the regular support are now supported by the tree support</summary>

The newly supported settings are: `Support Distance Priority`, `Support Horizontal Expansion`, `Support Interface Horizontal Expansion`, `Minimum Support Roof Area` (Some of these were enabled in the UI, but did not do anything if I understand the code correctly) and `Minimum Support Area` (though I would recommend leaving this setting at 0).  
</details>

<details>
<summary>Slices faster on multicore cpus.</summary>

Measured was the time it takes to generate the tree support for the test model (hobgoblin).  

Slice performance graph

Settings used for benchmarking:  
[Profile for current tree support](https://gist.github.com/ThomasRahm/770a93576a7e4f68aa21d94545a474d2/raw/Test_Live.curaprofile)  
[Profile for my tree support implementation](https://gist.github.com/ThomasRahm/770a93576a7e4f68aa21d94545a474d2/raw/Test_Me.curaprofile)  

Benchmarking was done in the 4.10 Version.  

</details>

## Further attributes this implementation has:

If a model is sliced twice on the same machine, the same trees will result. If this is somehow not the case something went very wrong. (Trees generated on systems with a different amount of threads could theoretically differ).  
Every point that can be supported with the given Branch Angle limitations, will be supported when the support can only rest on the build plate. This means some other settings may be paused to ensure such behavior. For example: `Tree Support Branch Diameter Angle` will be paused if a radius increase would cause a branch to become invalid.

# How the algorithm generating the trees works

This explains the idea behind the algorithm. As such it is simplified a fair bit to avoid explaining fixes for some edge cases.
<details>
    <summary>0. The model</summary>
    
The model

</details>

<details>

<summary>1. Calculate avoidance</summary>
An avoidance is an area that has to be avoided to achieve certain given results. For example an area in which a branch would be unable to reach the buildplate has to be avoided, if the support is not allowed to rest on the model.
This idea is identical to the current tree support. My implementation just uses a lot more avoidances(slow, fast, slow to model, fast to model, fast without holes, fast without holes to model), and precalculates them in parallel. 
Avoidances are calculated from the bottom up.

Avoidance

</details>

<details>
<summary>2. Generate influence areas</summary>
An influence area is an area,in which every point of the represented current branch, can reach all support points it has to support. Is is similar to an avoidance, but other that the avoidance, it represents possible center point areas, rather than excluding impossible ones. If two influence areas intersect, only the intersection will be used going forward. This causes branches to merge.
Influence areas are calculated from the top downwards, as they grow from the overhang downwards.

Influence areas

Influence areas 2

</details>

<details>
<summary>3. Generate tree path</summary>
Using the influence areas, the center points of the circles that build the branches are set.
This happens from the bottom up.

Tree path

</details>

<details>
<summary>4. Draw Tree</summary>
The circles, which center-point was calculated in 3. are drawn. Here also happens a fair bit of post processing to avoid edge-cases by the fact that the drawn radius in the tip of a branch is always increased (for stability reasons), even if a normal circle would not fit. 

Resulting tree

</details>

# Possible future improvements:
Automatically increase branch height when the angle is large.  

# Further Information
I did not make the mini [hobgoblin](https://www.thingiverse.com/thing:2841750) I used to test. I used it because it cloak makes it very hard to slice in a way that the cloak is supported while not having a connection of the branch to the model.  
All screen-shots of the current tree support were made using version 4.7.1.  
Thank you to [DerGenaue](https://github.com/DerGenaue) for providing the images explaining the algorithm and creating the graphs out of the performance data.


Ghostkeeper:
A couple of comments while you're working on this, to address the questions you've posed and some concerns:
* Code style and naming conventions for C++ is a combination of the [Generic Code Conventions](https://github.com/Ultimaker/Meta/blob/master/general/generic_code_conventions.md) of Ultimaker and the [exceptions/expansions of that geared towards C++](https://github.com/Ultimaker/Meta/blob/master/languages/c%2B%2B_code_conventions.md).
* New settings have to be added to `fdmprinter.def.json` in the front-end. [Here](https://github.com/Ultimaker/Cura/pull/2790/files) are the changes that originally introduced tree support, as an example. If you're on Windows you can edit that file in your Program Files to test. For Linux and MacOS that file will be packed inside the application file, which is a little harder to modify sometimes. It's also pretty easy for us to fix if you need some help.
* This is a *big* pull request. We're talking several thousand lines of code here, a few new features as to customisation on how the branches are generated, and refactoring. This goes against some of the [etiquette](https://readwrite.com/2014/07/02/github-pull-request-etiquette/) on how pull requests are presented. It's not necessarily a problem, but you have to understand that this is seriously hard to plan into a schedule. It also imposes serious risk for us: There are likely going to be new bugs in this code (unavoidable) and once we need to debug this, we are going to be less familiar with this code. This makes it harder to approve of a pull request.
* The collision volumes were originally implemented with the same technique as what you're doing now: Multi-threaded generation of the volumes at different radii. It's still supposed to be multi-threaded, but it now lazily generates them in order to reduce the total amount of work, rather than pre-generating them. We did performance tests when this was implemented to compare performance, using different sorts of computers (few cores vs. many cores, old memory and new memory sticks) and different sorts of models (tall/small, organic/artificial) and found improvements for the most common cases.
* We've learned to avoid creating settings for every little thing that could be adjusted. If nobody is going to adjust a setting, you might as well not have the setting. This makes a feature less overwhelming for someone who is not so involved in Cura or in your feature.
* Normally we prefer to have only one implementation of a feature. This is something that our users sometimes don't like (e.g. when we replaced the Prime Tower implementation), but it's important for the developers to be able to maintain an application with as many features as Cura. We're 6 programmers working with 2.5 million lines of code here, so we try to prevent having to maintain 5 different implementations of Tree Support that we added over time, so to say. For that we must be sure that the new implementation is capable of filling all the use cases of the old implementation. If some use cases are still possible but less than ideal (e.g. increased printing time) that may be acceptable.
* For Cura 4.7, Tree Support is getting particular attention to try to get it out of the Experimental section. You might see some merge collisions and it might even alleviate your concerns with it.
* That problem you found with the bridge settings adjusting the line directions is known and currently accepted. It's hard to ensure that the line directions of support are always perpendicular because that involves breaking up the support into zones with different line directions which is not beneficial to the strength of the supports or the quality of the overhangs either. I'd say that your fix is not proper then (as you describe it) because it would cause it to go wrong in different cases then.

We didn't review your code yet though. This is only a reaction to the description you posted on the pull request.

ThomasRahm:
Thank you for your feedback.

Regarding the bullet points:

1. Thank you very much, i will change my code over time as soon as i have time for it.

2. Thanks, i did not know it was this easy :)
I added a preliminary settings file to my description above.

3. I understand. One of my main reasons to open the draft was that a friend wanted to be able to share the resulting binary with a friend, and i did not feel comfortable sharing it without publishing the source.


4. I suspect the difference is that i only calculate the collision volumes up to the highest required layer, so collision volumes with a high radius may only be calculated for a a few layer, instead of for the whole model.
I am sure though that currently the lazy generation does not happen multi-threaded as i believe the comment:`\warning This class is NOT currently thread-safe and should not be accessed in OpenMP blocks` in TreeSupport.h.


5. I agree and have reduced the amount of available settings.


6. I did include the old implementation only because i believe every change breaks someones work-flow, and i did try to avoid that. 
My implementation aims to behave superior in every way and be a complete replacement. As such i now removed the old tree support.

7. I am happy to hear that tree support gets extra attention in 4.7, and will try to ensure my implementation remains superior ;)
I do not expect to be done with the documentation before its release as my exams are at the end of the month.

8. I feared as much. This means i will have to check whether a skin area is a bridge, and what the line orientation will be. I will have to investigate how to get to this information.

Ghostkeeper:
> I did include the old implementation only because i believe every change breaks someones work-flow, and i did try to avoid that.

True. It's a concession we've needed to make. Our position on it has so far been that those people can stick with the old Cura version if they liked the workflow that one provided, rather than trying to pack all dozens of Cura versions in one release.

Ghostkeeper:
On point 8, honestly it's probably easier on yourself and on the code base if you drop that feature! It's not in the expected behaviour now, and while the result will be better in the short term, if that makes the code harder to maintain in the long term it's also going to hurt. So you're welcome to try fixing it, but keep in mind that code complexity is something we'll look at too.

ThomasRahm:
Hi, i would like to release a version with my tree support implementation for others to be able to beta test it. 
For this i would just upload a zip with a complete compiled cura with the treesupport replaced to the release section of my github fork. 
I already modified the version from 4.9 to 4.9_TS2 to avoid issues with cached definitions. 
Is there some issue with me doing it this way?

Ghostkeeper:
No, I expect that would work fine. Maybe choose carefully which operating system(s) you'll release on.

I wonder how you'll know that your Cura version is being used, though? If you get no bug reports, is that a sign that it's bug free, or that nobody found your version?

Ghostkeeper:
I should also say that reviewing and testing 5000 lines of C++ code is no small task. Might take quite a while to get that in, especially considering there is not going to be any push from our managers to merge this, I expect. It kind of hinges on us being able to explain well why this change is necessary.

ThomasRahm:
Thanks for the response :)
I currently plan to post a link to my treesupport version on reddit. I hope that the post will get enough attention, so that i can get feedback and bug reports to evaluate the situation in addition to enabling my treesupport to show its worth. I plan on releasing for windows only, as im going to assume that linux users are more likely able to compile it themselves. I dont know yet when i will have enough time available to be able to all this, as i want to be able to fix potential bugs fast when they come up, though i hope to find time within a month.

chrisvanderpennen:
Hey, I've been using the fork for a few months and I'm very happy with the results, particularly for printing miniatures - to the point where I'm keeping it installed alongside the main Cura release specifically for minis.

I find the current tree algorithm tends to produce thick walls of merged support branches, that are difficult to separate and remove without damaging the model.  This new algorithm instead converges quickly to single thick branches, which can very easily be clipped and separated.

ThomasRahm:
Sorry for the late response, git-hub did not send me an email notification, for some reason i still have to figure out.
To my knowledge i did no changes in the Polygons class.  
I currently only changed `TreeSupport.h` and`TreeSupport.cpp`  (and after the 5.0 update `algorithms.h` as i need a `#pragma omp for nowait` replacement and a few lines in `support.cpp` to keep emulated support lines from splitting into two).
Any other changes should come from me merging the different release branches.  
If you see any commit changing some file, other than these, which commit is not in master please tell me so i can get rid of it, as it's inclusion was accidental!

ThomasRahm:
Hi,
When should i change this pull request from draft to regular?
I think in the current state my implementation is better than the current tree support, but there are still some further improvements possible regarding pointy overhangs, which fail more often than i would like, and branches getting too thin when moving very fast (>65°).

casperlamboo:
Hi @ThomasRahm,

Thank you for you contribution. We are very much interested in including your improved tree support in our 5.3 release. As soon as the merge conflicts are resolved the [WIP] can be removed and we'll start reviewing.

ThomasRahm:
Ok, i resolved the merge conflict.

While reviewing please consider that all `todo LayerIndex` are that the cura::parallel_for should be used with the template type LayerIndex, but that does not work as this function uses distance, which is not defined on LayerIndex and i did not want to change cura::parallel_for. So as a workaround i used coord_t as type instead.

casperlamboo:
Thanks for resolving the merge conflicts! I'm planning on starting the code review this week. In the mean time, can the `[WIP]` status be removed from the PR?

ThomasRahm:
Please note that this PR also contains #1671.

rburema:
@ThomasRahm Since we think this is a very valuable improvement on the one hand, and a big change on the other, that, after discussion with the team we've decided that it would be best to release this as a 'spotlight' beta based on 5.2.x to both promote this and have some extra beta-testers during the holidays even before the normal 5.3-beta period starts (probably sometime next year).

(I've taken over the review from @casperlamboo for the moment.)

Our question to you is, would you be OK/comfortable if we put this contribution in the spotlight like that to the community?

ThomasRahm:
I just fixed a bug that can prevent some support from generating for overhangs (with a smallish overhang angle e.g. 60°), that is noticeable when the support distance priority is `z_overrides_xy` (otherwise the support could be further down for overhangs smaller than 90°)

The change in `TreeSupport.cpp` to the `extra_outset` is to prevent differences in supported areas between a tip diameter of one line width compared to two line width tip diameter.

Ps: 
In the branch where you do the formatting i think found [an error in the file TreeSupport.cpp line 299 (getEvaluatePointForNextLayerFunction)](https://github.com/Ultimaker/CuraEngine/commit/2772aa8f8bed32e82d558915c883b510ee35d601?w=1#diff-a3f5326030c5fbfe74b5e0a9c231058002aa5327b647bbff9c10646286126784R299). (The link does not automatically go to the selected line) 

The function will be returned, as such capturing `current_layer` and `xy_overrides` by reference will cause issues, as when the function is executed, the references will no longer be valid.

rburema:
Thanks!

Good timing, I was just about to point you to that branch so you know what's going on when I saw this come in (... or, well, 6 hours ago).

I'll switch it back to by-copy (instead of by-reference) for the bug I introduced.

I'll try to cherry-pick the bugfix for the other bug as well to that branch (We're basing the alpha on 5.2.x, so people don't see all the half-finished 5.3 stuff we're still debugging.)

ThomasRahm:
I wanted to ask what why the `support_tree_bp_diameter` [was disabled in the UI](https://github.com/Ultimaker/Cura/pull/14098/files?diff=split&w=1#diff-835d0333a273b72a90298b1d2ccecee98f4e9633e2518d2f56e28ee39b515b99R4689).
The other two that were disabled were settings that my tree support does not use (`support_tree_branch_distance` and `support_tree_collision_resolution`)

I personally use `support_tree_bp_diameter` a lot to improve adhesion with the buildplate (with enabled support brim) as especially for small models, having a small branch radius, but still having a large connection area with the buildplate can be beneficial.

nallath:
I just talked to our print process engineers, and they indicate it was disabled because they didn't notice any change when increasing/decreasing the setting. So I assume that something in the refactor might have broken it.

ThomasRahm:
Hm, you are correct, something broke though this was not caused by your re-factor...
Here is a comparison of what it should do:
[Small Initial Layer Diameter](https://gist.githubusercontent.com/ThomasRahm/770a93576a7e4f68aa21d94545a474d2/raw/Inital%2520Layer%2520Diameter%2520Small.PNG)
[Large Initial Layer Diameter](https://gist.githubusercontent.com/ThomasRahm/770a93576a7e4f68aa21d94545a474d2/raw/Inital%2520Layer%2520Diameter.PNG)

Thanks for the bug report, i will get back to you on this.

rburema:
Though I _do_ see it change something (even) from the current source -- I'm not sure now on what build PP&M based their conclusions :-/

normal_initial_diameter
big_initial_diameter

- I propose we re-enable it in any case. In the worst case we have a setting in the xmas-alpha that does nothing after all.

ThomasRahm:
Update: The setting works when `support type` is `Touching buildplate` or the `rest preference` is `Buildplate`, as when the `rest preference` is `On any flat surface` the branches do not know if the will rest on the buildplate or on the model and because of that the radius of the branch will not be increased (to not needlessly increase the contact area with the model).

jellespijker:
@rijkvanmanen and @pkuiper-ultimaker FYI

ThomasRahm:
The issue is that `On any flat surface` is implemented by never using the regular avoidance, just the avoidance to model (as logically a branch can further down choose to rest on the model instead), and the initial layer diameter is no increased when a branch is not known to rest on the buildplate . 

Possible solutions are:
1. Changing avoidance to model to not include the buildplate and then calculate avoidances by unioning the to model avoidance without the radius increase with the to buildplate avoidance with the larger radius. Downside is that these hybrid avoidances cant be pre-calculated, so there would definitely be a performance impact and this would take quite a bit of work. 
2. Just increase the Collision Radius if possible, and when it turned out that it rests on the buildplate set a flag to also increase the radius. Easy to implement, but has the downside being that branches may go to buildplate as they can no longer rest on the model because of the increased collision radius.

I implemented the second option. 
When the initial layer diameter is reasonable the downside may not be obvious, but there can be an argument made to document this behavior somewhere (E.g. in the description of rest preference: A large initial layer diameter can cause branches to rest on the buildplate (instead of the model))

P.S. I noticed that there could be cases where the fix could cause issues when a branch that can go to the buildplate, which radius was already increased, merges with a branch that rests on the model, as the resulting branch will rest on the model causing a possible jump in radius. I will solve this at some point, but it may be better to not include the the above mentioned fix in the spotlight beta as it will probably be after Christmas by then.

rburema:
@ThomasRahm The spotlight will be an alpha even, so there's still time to fix some bugs! [The 5.3.x release itinerary is now: 5.3._-xmas-tree-alpha (-> merge that into an actual 5.3 branch) -> 5.3-beta -> 5.3.0 final]

Something slightly different: @rijkvanmanen & I had an investigation into an unrelated (support) issue, and (as a side effect) it turns out the default mitre parameter for the clipperlib 'offset'-operation, `jtMitre` gives terrible results (can be asymmetrical for a symmetric model even). That could be the cause of at least some of the 'rounding errors' you're seeing. We where originally OK with that for CuraEngine, since it results in cheaper operations, but it seems its wrong (or at the very least gives counter-intuitive results) in many instances. In our (unrelated to tree) experiment yesterday, we had good results with `jtRound` (though no clue yet how much slower that will make the code overall -- or if we could also use `jtSquare` instead).

ThomasRahm:
I already use `jtRound` in most places. `jtRound` is a bit slower, but not by too much. Most performance impact comes from the fact that the resulting Polygon has much more vertices, so calling simplify from time to time is essential. In `increaseAreas` there is even a line that effectively uses a `jtRound` offset twice, as a single round offset still had issues for very small influence areas (with few vertices). There may be an argument made to change the ones that don't use `jtRound` to `jtSquare` or `jtRound`. All offsets directly related to influence areas or avoidances should in my opinion stay `jtRound`, as non-accurate offsets can cause disappearing branches (in other words, cause an influence area to disappear).

One reason why the `jtMiter` may not work as expected for you is that the default miter limit of `1.2` is actually a miter limit of `2.0`, as clipper has an internal minimum for it.

I think the reason for me encountering so many rounding errors is that sometimes influence areas can become very small (i definitely encountered areas smaller than 50 square-microns consisting of only a hand-full of vertices in the past, but hope i mellowed down the tendencies for that small influence areas) causing me to run into some rounding errors that are part of clipper and some that are unavoidable as sometimes the difference between an influence area that works fine and one that causes errors can be few microns in width.

Furthermore, in the known issues you listed "Support Interface are not always working corretly". Can i get an example for this ?

Also i have seen the commit https://github.com/Ultimaker/CuraEngine/commit/3c4180381b8ecf503c299cf65b442a2ccd58726 and i think i see the reason. When the rest preference is `on any flat surface` the regular avoidance is not precalculated, as it will not be needed. So if it is then requested before the loop, it then will be calculated on the fly, which may take a while if it works correctly (The calculating of avoidances when needed is not very well tested, as it is just a fallback for when an error occurs). If you have a case where it actually deadlocks (and not just takes minutes instead of seconds), i also would like an example, as this should not be happening.

rburema:
Thanks!

I should have remembered you already mostly using `jtRound`. Rijk also pointed out the 1.2 thing as well. I'm not sure that's the problem though. I suspect they put it that low because the higher parameter was already causing the sort of problems we're still seeing.

---

About the known issues:

> Furthermore, in the known issues you listed "Support Interface are not always working corretly". Can i get an example for this ?

This is what QA came back with when testing:

> When Interface priority is set to preferred, there is missing interface for some branches. 
> 
> The interface should follow the model, not the support.
> 
> image-20221220-131549

---

About the last minute bug fix:

> Also i have seen the commit https://github.com/Ultimaker/CuraEngine/commit/3c4180381b8ecf503c299cf65b442a2ccd58726 and i think i see the reason.
> [...]
> If you have a case where it actually deadlocks (and not just takes minutes instead of seconds), i also would like an example, as this should not be happening.

Ah, yes, that makes sense. The model that took long enough to convince me to write it down as 'deadlock' (even though I didn't do a thorough investigation yet, so it isn't confirmed -- we where trying to get the alpha out yesterday when this came up as a last-minute bug), is this:
[possible_deadlock.zip](https://github.com/Ultimaker/CuraEngine/files/10294082/possible_deadlock.zip)

rburema:
P.S. Coming to think of it, it probably isn't a deadlock, at least not in all threads, as it was still spamming the logs full of warnings.

ThomasRahm:
Which `preferred`? The settings for interface priority are `Support preferred`, `Interface preferred`, `Support lines preferred`, `Interface lines preferred` and `Both overlap`.
I do not generate support roof for overhangs smaller than `minimum support roof area`. If the `tip diameter` is large enough or `minimum support roof area` is small enough that the tip area would be larger than the `minimum support roof area` I replace the first few layers of a branch with interface (more similar to how the old tree support worked with roof). This behavior to replace tips with interface could have been a explicit setting, but I wanted to avoid adding another checkbox so I made it dependent on tip diameter and minimum roof area. If this causes confusion, it may be better as an explicit setting.
 
In short, if the `interface priority` here is set to `Support preferred` and the `minimum support roof area` is larger than `pi*((tip diameter/2)^2)` the behavior visible in the screenshot is exactly what I would expect. Also I am not sure what is meant with "The interface should follow the model, not the support.", so it is possible I don't see what the expected behavior here is.
Can you ask QA if the behavior with `minimum support roof area` = 0 and `tip diameter` = `1.5` or `branch diameter` is closer to what was expected?

rburema:
I'll be off work until Tuesday (and the coming weekend is not one for tinkering for me), but I'll pass it on!

rijkvanmanen:
While trying to reproduce the issues we found with the support interface, I saw that quite a few of the issues are resolved in the alpha release, so that is great! But this one is still causing issues:

I enabled the support interface for this simple geometry, but no interface is generated: 
image

I found that after increasing the support horizontal expansion to one linewidth, I do get a support interface (but also a few extra ugly branches):
image

ThomasRahm:
What is your `minimum support roof area` setting?
I only generate roof for an overhang-area (on one layer) if it is larger than `minimum support roof area`. Support horizontal expansion enlarges the overhang area, causing it to become large enough to generate roof.
If the `minimum support roof area` < pi*(`Tree support tip diameter`/2)^2 the top of the tip will be replaced with roof. 
From where I'm at this seems like a settings issue, because if roof would be generated here, there would be at least two areas (top-most and bottom-most) for which roof would be generated that are smaller than `minimum support roof area`, which is (by my understanding) exactly what `minimum support roof area` tries to prevent.

ThomasRahm:
The `support horizontal expansion` does not directly affect the roof, but `support roof horizontal expansion` is by default (in the UI) `support horizontal expansion` and as the UI does not allow increasing `support roof horizontal expansion` without increasing `support horizontal expansion`. I tend to adjust the latter in the UI, sorry i was a bit unclear here. The restriction to `support roof horizontal expansion` could be removed for tree support though.
 
I now understand what you mean with follow the model and why it is relevant.
I calculate roof for overhang on a layer by layer basis and not in a certain order, so when generating roof, i do not have any knowledge if there will be roof in the layers above or below. More problematic is that it depends on the pattern whether even a usable interface will generate in a thin area. I am of the opinion for an acceptable solution, there would need to be a way to store not only interface areas, but also lines (or a way to ensure a thin area results in a predictable line, like a hypothetical support interface wall line count option that could be re-used on an area-by-area basis). 
If you change the interface pattern [in your example that generates a nice roof](https://user-images.githubusercontent.com/37628975/209951402-f2109482-24dd-4ff0-bcb9-4ea40fadf7ac.png) to triangles, the result would probably not be what you wanted. Seems to me like the line orientation seemed to accidentally aligned with the overhang, this is often not the case though. If the area to be supported would be curved (or rotated so that the interface lines no longer align with the overhang), the line interface results would not result in a this nice outcome for example.

Quite a bit of the code generating the interface does not work in a way currently that it is easy to make behavior similar to regular interface. E.g. Overhangs that are placed further down because of `xy overrides z` are currently never interface. It should be possible to make the interface more similar to the interface of regular support, but i think it may require quite a bit of reworking of `generateInitialAreas`, for which i currently do not have time for.

In short -unless i missed something-, to solve this in a satisfying way i would have need to redo interface areas in general and how they are stored (for an option to let certain areas have walls) or add another variable for interface lines, and write the corresponding code in `FFFGcodeWriter`, which i tried to avoid.

ThomasRahm:
Hi, i wanted to know if it would be possible to get responses to the questions noted [here](https://github.com/ThomasRahm/CuraEngine/pull/7), as i would like to merge the changes into the branch that this pull request is based on.
Also i have now seen two threads of people preferring the old support blocker behavior, while most seem to prefer the new one. Is there an easy way to add a dropdown to the support blocker when tree supports are enabled and give the support modifier mesh a corresponding flag? I did not look to far into the UI.

I also did look into the issue about support resting on support blocker a bit.
It is a bit more complicated than i expected. I have a few potential ideas, but with my exam time coming up i will not have much time to try to implement them before mid February.

jellespijker:
@rburema do you have time to answer @ThomasRahm ?

btw Thomas if you got any other questions and think you might benefit from a live chat, feel free to ping me on LostInTech Discord channel.

jellespijker:
Nevermind Remco, I heard from @DerGenaue that the questions have been answered already and the PR has been merged

rburema:
@jellespijker I did have time :-)

ThomasRahm:
I fixed the support resting on support blocker issue.

grafik
(The support blockers are one directly above the build-plate and one directly above the base-plate.)

This commit also improves performance quite a bit by now using also the deviation setting when simplifying in TreeModelVolumes.
One random test without support blocker on a 15° rotated 300% hobgoblin.
24s => 10s  
Before: `Total time used creating Tree support for the currently grouped meshes: 24099.117000000002 ms. Different subtasks:
Calculating Avoidance: 7752.14 ms Creating inital influence areas: 145.714 ms Influence area creation: 14665.388 ms Placement of Points in InfluenceAreas: 33.694 ms Drawing result as support 1502.18 ms`
After: `Total time used creating Tree support for the currently grouped meshes: 9925.063 ms. Different subtasks:
Calculating Avoidance: 2252.962 ms Creating inital influence areas: 185.448 ms Influence area creation: 5576.292 ms Placement of Points in InfluenceAreas: 67.812 ms Drawing result as support 1842.548 ms`

I am currently using 
```
Polygons next_latest_avoidance = simplifier.polygon(latest_avoidance);
latest_avoidance = next_latest_avoidance.unionPolygons(latest_avoidance);    
```
to simplify in a way that the polygon can only get larger by simplifying. If there were an option to ensure the resulting polygon only gets larger by simplifying, it could improve performance further.

jellespijker:
Looks like @ThomasRahm fixed CURA-10161 (resting on support blockers) :+1: , we still need to review/component test and QA it ofcourse

PR #1283, opened 2020-07-03
When using a shared heater, preserve other extruders' waited_for_temperature state.

smartavionics:
It was always resetting the other extruders' waited_for_temperature state when the current
temperature change wasn't going to wait but we only want to do that when the new
temperature is greater than the current temperature.

Discussed in https://community.ultimaker.com/topic/33196-eliminating-m109-calls-when-changing-toolsshared-heater/


rburema:
Thanks for the fix! It was deemed low risk enough that we'll try to get it in 4.7 -- it'll probably be in the beta as well.

PR #1295, opened 2020-07-19
Add gcode comments for skipped retractions

Sophist-UK:
I have not added a comment for the last condition that skips retraction because I have insufficient knowledge to know what is happening.

This PR is in support of issue https://github.com/Ultimaker/Cura/issues/8069 whereby we are finding it difficult to understand why a retraction is not happening sometimes on minimum layer lift.

### DIFF
--- a/src/gcodeExport.cpp
+++ b/src/gcodeExport.cpp
@@ -890,6 +890,7 @@ void GCodeExport::writeRetraction(const RetractionConfig& config, bool force, bo
     double retraction_diff_e_amount = old_retraction_e_amount - new_retraction_e_amount;
     if (std::abs(retraction_diff_e_amount) < 0.000001)
     {
+        writeComment("Retraction skipped - too small");
         return;
     }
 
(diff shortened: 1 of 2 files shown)


src/gcodeExport.cpp
```diff
@@ -904,11 +905,13 @@ void GCodeExport::writeRetraction(const RetractionConfig& config, bool force, bo
         }
         if (!force && config.retraction_count_max <= 0)
         {
+            writeComment("Retraction cancelled - retraction max <= 0");
             return;
         }
         if (!force && extruded_volume_at_previous_n_retractions.size() == config.retraction_count_max
             && current_extruded_volume < extruded_volume_at_previous_n_retractions.back() + config.retraction_extrusion_window * extr_attr.filament_area) 
         {
+            writeComment("Retraction cancelled - retraction max exceeded");
```
Sophist-UK, line 914: I would like to add some more detail about how far back the n'th retraction was, but my c++ is not up to it.

Sophist-UK:
I have no idea why CI/CD failed.

Sophist-UK:
Ah - it was a unit test. I will fix.

Sophist-UK:
Ok - I can't create a development environment on my laptop, and I there is no hint what the test string now is to match it. If anyone can help get the test working that would be appreciated.

Sophist-UK:
P.S. I note in https://github.com/Ultimaker/Cura/issues/8069 that @Ghostkeeper believes that this PR may not be merged because comments are discouraged in the Gcode.

I can understand this in general - people should be able to read the Gcode and match it to what is happening - but I believe that a comment  **is** appropriate when Gcode that might be expected is cancelled / omitted.

PR #1296, opened 2020-07-20
Fix initial layer thickness when empty initial layers are removed.

smartavionics:
Fix for https://github.com/Ultimaker/Cura/issues/7787.

### DIFF
--- a/src/FffPolygonGenerator.cpp
+++ b/src/FffPolygonGenerator.cpp
@@ -828,7 +828,7 @@ bool FffPolygonGenerator::isEmptyLayer(SliceDataStorage& storage, const unsigned
 
 void FffPolygonGenerator::removeEmptyFirstLayers(SliceDataStorage& storage, size_t& total_layers)
 {
-    int n_empty_first_layers = 0;
+    size_t n_empty_first_layers = 0;
     for (size_t layer_idx = 0; layer_idx < total_layers; layer_idx++)
     {
         if (isEmptyLayer(storage, layer_idx))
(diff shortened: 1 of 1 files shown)


nallath:
This doesn't quite give the correct results (Although i did use an exaggurated situation)

1. Set initial layer height to 4mm 
2. Print a 20x20x20 mm cube. 
3. Number of layers = 161 (1 * 4mm + 160 * 0.1mm) -> 20mm height
4. Set Z of object to 5
5. Slice again, nr of layers : 200 -> (1* 4mm + 199 *0.1) -> 23.9mm high

The result is somewhat better since now at least the first layer actually starts at z 0.

smartavionics:
I just tested this with a 10x10x10 cube, layer height 0.1 and initial layer height of 2 and the overall object height did not change according to the layer view...

Without this PR I see...

Screenshot_2020-07-21_13-20-42

With this PR I see...

Screenshot_2020-07-21_13-21-18

smartavionics:
The object was placed at z = 5.

smartavionics:
Here's another view of what you get before this PR is applied, this time with the skirt enabled so you can see that although the empty layers have been removed, there's a gap because the thickness of the first layer is 0.1 rather than 2mm and this is what leads to the underextrusion as described in the issue.

Screenshot_2020-07-21_13-30-09

smartavionics:
This PR is only concerned with ensuring that the first layer gets the correct thickness so that the amount extruded is right. Yes, there could be other bugs here related to the number of layers.

nallath:
Fair enough. I'll add the other issue to the backlog.

PR #1298, opened 2020-07-23
CURA-7614: Fix missing skin infill in roofing layers

konskarm:
The no_air_above depends on the insets so it has to be recalculated for the roofing layers
before regenerating the skin insets (when taking into account the extra skin wall count).

Fixes issue https://github.com/Ultimaker/CuraEngine/commit/f0b56479c97aad5b6e105e9f2df751186e7b4997#r40756573
introduced by https://github.com/Ultimaker/CuraEngine/pull/1287.

CURA-7614

### DIFF
--- a/src/skin.cpp
+++ b/src/skin.cpp
@@ -473,57 +473,65 @@ void SkinInfillAreaComputation::generateInfill(SliceLayerPart& part, const Polyg
 void SkinInfillAreaComputation::generateRoofing(SliceLayerPart& part)
 {
     const size_t roofing_layer_count = mesh.settings.get<size_t>("roofing_layer_count");
-    const size_t wall_idx = std::min(size_t(2), mesh.settings.get<size_t>("wall_line_count"));
 
     for (SkinPart& skin_part : part.skin_parts)
     {
         Polygons roofing;
         if (roofing_layer_count > 0)
         {
-            Polygons no_air_above = getWalls(part, layer_nr + roofing_layer_count, wall_idx);
-            if (!no_small_gaps_heuristic)
-            {
-                for (int layer_nr_above = layer_nr + 1; layer_nr_above < layer_nr + roofing_layer_count; layer_nr_above++)
-                {
-                    Polygons outlines_above = getWalls(part, layer_nr_above, wall_idx);
-                    no_air_above = no_air_above.intersection(outlines_above);
-                }
-            }
-            if (layer_nr > 0)
-            {
-                // if the skin has air below it then cutting it into regions could cause a region
-                // to be wholely or partly above air and it may not be printable so restrict
-                // the regions that have air above (the visible regions) to not include any area that
-                // has air below (fixes https://github.com/Ultimaker/Cura/issues/2656)
-
-                // set air_below to the skin area for the current layer that has air below it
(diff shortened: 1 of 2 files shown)


src/skin.cpp
```diff
@@ -473,57 +473,97 @@ void SkinInfillAreaComputation::generateInfill(SliceLayerPart& part, const Polyg
...
             skin_part.inner_infill = skin_part.inner_infill.intersection(no_air_above);
 
-            // Insets have previously NOT been generated for any layer if the top/bottom pattern is concentric.
-            // In this case, we still want to generate insets for the roofing layers based on the extra skin wall count.
-            if (skin_part.roofing_fill.size() > 0
-                && (layer_nr > 0 && mesh.settings.get<EFillMethod>("roofing_pattern") != EFillMethod::CONCENTRIC))
+            // Insets are NOT generated for any layer if the top/bottom pattern is concentric.
+            // In this case, we still want to generate insets for the roofing layers based on the extra skin wall count,
+            // if the roofing pattern is not concentric.
+            if (skin_part.roofing_fill.size() > 0
```
jellespijker, line 489: Check for empty do not compare against size

According to Clang documentation

readability-container-size-empty

Checks whether a call to the size() method can be replaced with a call to empty().

The emptiness of a container should be checked using the empty() method instead of the size() method. It is not guaranteed that size() is a constant-time function, and it is generally more efficient and also shows clearer intent to use empty(). Furthermore some containers may implement the empty() method but not implement the size() method. Using empty() whenever possible makes it easier to switch to another container in the future.

src/skin.cpp
```diff
@@ -473,57 +473,97 @@ void SkinInfillAreaComputation::generateInfill(SliceLayerPart& part, const Polyg
...
-                generateInnerSkinInfill(skin_part);
-                skin_part.roofing_fill = skin_part.inner_infill.difference(no_air_above);
-                skin_part.inner_infill = skin_part.inner_infill.intersection(no_air_above);
-                skin_part.inner_infill = skin_part.inner_infill.offset(-skin_inset_count * skin_line_width);
+                regenerateRoofingFillAndInnerInfill(part, skin_part);
             }
+            // On the contrary, unwanted insets are generated for roofing layers because of the non-concentric top/bottom pattern.
+            // In such cases we want to clear the skin insets first and then regenerate the proper roofing fill and inner infill
+            // in the concentric roofing_pattern.
+            else if (skin_part.roofing_fill.size() > 0
```
jellespijker, line 502: Check against empty not size. See comment above

src/skin.h
```diff
@@ -127,13 +127,30 @@ class SkinInfillAreaComputation
     void generateInfill(SliceLayerPart& part, const Polygons& skin);
 
     /*!
-     * Calculate the areas which are 'directly' under air,
-     * remove them from the \ref SkinPart::inner_infill and save them in the \ref SkinPart::roofing_fill of the \p part
+     * Remove the areas which are 'directly' under air from the \ref SkinPart::inner_infill and 
+     * save them in the \ref SkinPart::roofing_fill of the \p part.
      * 
      * \param[in,out] part Where to get the sSkinParts to get the outline info from and to store the roofing areas
```
jellespijker, line 133: typo, extra sSkinParts

jellespijker: the typo is located on three positions line 133, line 140, line 148 probably done with a search and replace

PR #1299, opened 2020-07-24
Remove Wall Overlap Compensation in arachne

konskarm:
Arachne handles the wall overlap compensation so there is no need for this setting anymore.

CURA-7559


src/LayerPlan.cpp
```diff
@@ -535,21 +535,21 @@ void LayerPlan::addExtrusionMove(Point p, const GCodePathConfig& config, SpaceFi
...
-        const Ratio flow = (wall_overlap_computation) ? flow_ratio * wall_overlap_computation->getFlow(p0, p1) : flow_ratio;
+        const Ratio flow = flow_ratio;
         addExtrusionMove(p1, config, SpaceFillType::Polygons, flow, spiralize);
         p0 = p1;
     }
     if (polygon.size() > 2)
     {
         const Point& p1 = polygon[start_idx];
-        const Ratio flow = (wall_overlap_computation) ? flow_ratio * wall_overlap_computation->getFlow(p0, p1) : flow_ratio;
+        const Ratio flow = flow_ratio;
```
konskarm, line 552: If there is a way to get the flow in arachne, it should replace the flow ratio. At the moment I am leaving it as the default (which is 1.0).

src/InsetOrderOptimizer.cpp
```diff
@@ -570,60 +565,23 @@ bool InsetOrderOptimizer::processInsetsWithOptimizedOrdering()
...
-        // use a slightly reduced line width so that compensation only occurs between insets at the same level (and not between insets in adjacent levels)
-        wall_overlapper_x = new WallOverlapComputation(wall_x_polys, mesh_config.insetX_config.getLineWidth() - 1);
-    }
-
     // create a vector of vectors containing all the inset polys
     inset_polys.clear();
 
-    // if overlap compensation is enabled, use the polys that have been tweaked by the
-    // overlap computers, otherwise use the original, un-compensated, polys
+    // Use the original, un-compensated, polys
```
Ghostkeeper, line 571: Not really relevant any more as there is no compensation to speak of.

src/InsetOrderOptimizer.cpp
```diff
@@ -570,60 +565,23 @@ bool InsetOrderOptimizer::processInsetsWithOptimizedOrdering()
...
-        if (wall_overlapper_0)
-        {
-            inset_polys[0].push_back(wall_0_polys[poly_idx]);
-        }
-        else
-        {
-            inset_polys[0].push_back(part.insets[0][poly_idx]);
-        }
+        inset_polys[0].push_back(part.insets[0][poly_idx]);
     }
```
Ghostkeeper, line 577: Can't we now just replace the entire `for` loop?

```
inset_polys.push_back(part.insets[0]);
```

src/InsetOrderOptimizer.cpp
```diff
@@ -570,60 +565,23 @@ bool InsetOrderOptimizer::processInsetsWithOptimizedOrdering()
...
-            {
-                inset_polys[inset_level].push_back(wall_x_polys[wall_x_polys_index++]);
-            }
-            else
-            {
-                inset_polys[inset_level].push_back(part.insets[inset_level][poly_idx]);
-            }
+            inset_polys[inset_level].push_back(part.insets[inset_level][poly_idx]);
         }
     }
```
Ghostkeeper, line 586: And here the same. Maybe we can even move the outer wall case in with the rest of this `for` loop, letting `inset_level` start at 0. We don't have a difference any more between `wall_overlapper_x` and `wall_overlapper_0`, so these are now the same.

Ghostkeeper: With some luck, if the polygons are not being modified here, we could also just use `part.insets` and leave out the entire `inset_polys` variable. That would prevent a copy of the data.

src/LayerPlan.cpp
Ghostkeeper, line 552: You don't need the `flow` variable; you can just use `flow_ratio` instead.

src/LayerPlan.cpp
```diff
@@ -535,21 +535,21 @@ void LayerPlan::addExtrusionMove(Point p, const GCodePathConfig& config, SpaceFi
...
-void LayerPlan::addPolygon(ConstPolygonRef polygon, int start_idx, const GCodePathConfig& config, WallOverlapComputation* wall_overlap_computation, coord_t wall_0_wipe_dist, bool spiralize, const Ratio& flow_ratio, bool always_retract)
+void LayerPlan::addPolygon(ConstPolygonRef polygon, int start_idx, const GCodePathConfig& config, coord_t wall_0_wipe_dist, bool spiralize, const Ratio& flow_ratio, bool always_retract)
 {
     Point p0 = polygon[start_idx];
     addTravel(p0, always_retract);
     for (unsigned int point_idx = 1; point_idx < polygon.size(); point_idx++)
     {
         Point p1 = polygon[(start_idx + point_idx) % polygon.size()];
-        const Ratio flow = (wall_overlap_computation) ? flow_ratio * wall_overlap_computation->getFlow(p0, p1) : flow_ratio;
+        const Ratio flow = flow_ratio;
```
Ghostkeeper, line 545: You don't need the `flow` variable. You can just use `flow_ratio` instead.

src/InsetOrderOptimizer.cpp
konskarm: I merged the for loops into one. The inset_polys is modified along the way, so I don't think it's a good idea to replace it with the part.insets.

PR #1300, opened 2020-07-27
Arachne Path Optimizer

Ghostkeeper:
This is a rework of the `PathOrderOptimizer` to function with libArachne.

The origin of this work stems from the fact that libArachne can produce a mix of polygons and polylines which need to be optimised together. You can get a mix of polylines and polygons, which you'd like to print all at once while minimising the travel moves between them. Previously, CuraEngine had a `PathOrderOptimizer` which only handled polygons, and a `LineOrderOptimizer` which only handled line segments (`Polygon` instances with 2 vertices). This has been refactored into a single class called `PathOrderOptimizer` which handles both.

The `PathOrderOptimizer` has its interface changed significantly. Notable changes are:
* The optimizer is now a template class. This is necessary because it now needs to deal with a vector of `ExtrusionJunction`s as input, which doesn't contain a `Polygon` or a reference to a `Polygon` anywhere. As such, the optimiser needs to know how to get the vertex coordinates from this vector. There is now a template-specialised function `getVertexData` which needs to be implemented for every type of input data, which gets the vertex data from the input so that the optimiser can work with the input.
* Previously the optimiser would return a list of indices to the input array and a list of starting vertices. Because the input can now be a combination of polylines and polygons and potentially come from different input lists, this no longer works. Instead, the optimiser contains a field called `paths` to provide the output, in order. These `Path` instances contain the original input, as well as whether they were polylines or polygons, its starting vertex and the direction to print them in. This allows the optimiser to change polylines into polygons if the endpoints are close together too, which allows placing the seam correctly in a polyline which actually closed on itself.

**Notes for a reviewer:**
This will appear more daunting than it actually is, due to the many changed lines of code. Part of that is due to refactoring and documentation. There are 3 main algorithms that need to be reviewed, which by themselves are not so large:
* `PathOrderOptimizer::optimize`, the main optimization function that chooses the order in which paths get printed.
* `PathOrderOptimizer::findStartLocation`, which finds the position of the seam on a polygon (or the closest endpoint for a polyline).
* `InsetOrderOptimizer::processInsetsIndexedOrdering`, which used to just categorise the paths by inset #, but now needs to actually optimise the order within each of these categories.

There is also a minor algorithm `getCombedDistance` which is largely the same as the original algorithm, but factored out to a separate function. The rest of the code changes are either for users of the class to work with the new interface of this optimizer, or to clean up.

I've cleaned up and moved some code as well:
* The `pathOrderOptimizer.h` and `pathOrderOptimizer.cpp` files have been renamed to have a capital P, as the code style requires.
* The `ZSeamConfig` struct has been moved to a separate file, as the code style requires.
* The `GcodeWriter` class has been deleted. This was a copy of the actual `GCodeWriter` (with capital C), modified for use with libArachne in the research phase and not used in our code. This work required me to refactor that (or face compilation errors) and I didn't want to do that for code that was going to be deleted anyway.
* The `OrderOptimizer` class has been deleted. This was also only used during research and is now completely replaced by the `PathOrderOptimizer`.
* As mentioned before, the `PathOrderOptimizer` and `LineOrderOptimizer` have been merged into one class, so there is now no more `LineOrderOptimizer`.

**Notes for a tester:**
Some pointers for how to test this:
* The order of printing walls is optimised.
* The order of printing skin parts is optimised.
* The order of printing infill parts is optimised.
* The order of printing support parts is optimised.
* The order of printing support interface parts is optimised.
* The order of printing gap filling is optimised.
* The whole Z seam algorithm has been refactored. Test thoroughly!
* Bugfix when compiled in debug mode: No more assertion failure if the Z Seam Position is precisely on one of the vertices. I don't think this caused actual problems in release builds though.
* Bugfix: When the seam is set to "User Specified" but the seam corner preference is "Inner" or "Outer", it should now listen to that preference and place the seam in the nearest inner or outer corner, if there is any. If there is no inner or outer corner (as preferred) it chooses the nearest suboptimal vertex.
* It should still listen to the Outer Before Inner Walls setting.
* If Infill Travel Optimization is enabled, it should print the infill in a better order if it needs to perform combing moves in order to print the infill or skin.
* Please check if all walls are closed that should be closed (CURA-7564). But keep in mind that for inner walls, libArachne doesn't seem to always close them properly (CURA-7597). At least check for the outer walls, extra skin contours, extra infill contours, concentric infill, etc.

The following is considered out of scope:
* Optimize Wall Printing order (CURA-7557).
* libArachne not always closing inner walls (CURA-7597).
* Tests failing on CI server (CURA-7565). They should not fail locally as long as you've compiled CuraEngine completely.

Implements CURA-7558 and CURA-7564.


src/utils/polygonUtils.cpp
```diff
@@ -1255,6 +1255,10 @@ bool PolygonUtils::polygonCollidesWithLineSegment(const Polygons& polys, const P
 
 bool PolygonUtils::polygonCollidesWithLineSegment(const Polygons& polys, const Point& startPoint, const Point& endPoint)
 {
+    if(endPoint == startPoint)
+    {
+        return false; //Zero-length line segments never collide.
```
rburema, line 1260: Not even when equal to one of the polygons' points?

src/PathOrderOptimizer.cpp
```diff
@@ -0,0 +1,61 @@
...
+template<>
+ConstPolygonRef PathOrderOptimizer<const SupportInfillPart*>::getVertexData(const SupportInfillPart* path)
+{
+    return path->outline.outerPolygon();
+}
+
+template<>
+ConstPolygonRef PathOrderOptimizer<const std::vector<ExtrusionJunction>*>::getVertexData(const std::vector<ExtrusionJunction>* path)
+{
+	cached_vertices.emplace_back();
```
rburema, line 52: 2 extra spaces, alignment off.

src/PathOrderOptimizer.h
```diff
@@ -0,0 +1,591 @@
...
+            available_candidates.reserve(nearby_candidates.size());
+            for(const size_t candidate : nearby_candidates)
+            {
+                if(picked[candidate])
+                {
+                    continue; //Not a valid candidate.
+                }
+                available_candidates.push_back(candidate);
+            }
+            if(available_candidates.empty()) //We may need to broaden our search through all candidates then.
```
rburema, line 269: At some point we might want to consider a kd-tree that can do proper 'n nearest neighbour' searches.

src/InsetOrderOptimizer.cpp
```diff
@@ -254,14 +253,14 @@ void InsetOrderOptimizer::processHoleInsets()
     if (optimize_backwards)
     {
         // reverse the optimized order so we end up as near to the outline z-seam as possible
-        std::reverse(order_optimizer.polyOrder.begin(), order_optimizer.polyOrder.end());
+        //std::reverse(order_optimizer.paths.begin(), order_optimizer.paths.end());
```
rburema, line 256: Commented out code.

src/InsetOrderOptimizer.cpp
```diff
@@ -503,30 +501,34 @@ void InsetOrderOptimizer::processOuterWallInsets(const bool include_outer, const
...
+            for(ConstPolygonRef poly : part_inner_walls)
+            {
+                orderOptimizer.addPolygon(poly);
+            }
             orderOptimizer.optimize();
             if (!outer_inset_first)
             {
                 // reverse the optimized order so we end up as near to the outline z-seam as possible
-                std::reverse(orderOptimizer.polyOrder.begin(), orderOptimizer.polyOrder.end());
+                //std::reverse(orderOptimizer.paths.begin(), orderOptimizer.paths.end());
```
rburema, line 524: Commented out code.

src/utils/polygonUtils.cpp
Ghostkeeper: That's an edge case. It would be up to the user of the function whether that would need to trigger a collision.

In the `PathOrderOptimizer` and the `Comb` classes it is not desired, because a zero-length line segment would then trigger a combing calculation which finds the intersections between the line segment and a polygon. To find the intersections, it'll try to rotate the polygon and line segment such that the line segment is aligned with the X axis. Since the line segment has zero length, this calculation yields a transformation matrix which is broken (it includes some `infinity()` floats).

The only other place where this is used is in the infill calculations to see when to break up infill lines and to join them. Zero-length infill lines aren't really relevant there since they don't get printed.

So thus far the other interpretation of the edge case is not necessary anywhere.

src/PathOrderOptimizer.cpp
Ghostkeeper: Fixed. It was actually a tab instead of a space.

My IDE was doing that sometimes because my default is configured to indent with tabs for personal projects. I had made an exception for CuraEngine, but only for C++. Sometimes it would randomly detect a file to be C instead of C++ and then it would indent. I've now configured the IDE with another exception for C and header files. Why it would ever think that this template-heavy code is not C++ is a mystery to me, but it shouldn't happen again.

src/PathOrderOptimizer.h
Ghostkeeper: Yeah. Out of scope for now though, I'd say. It's not very slow yet.

src/InsetOrderOptimizer.cpp
Ghostkeeper: Done.

I had to comment it out because there is no std::swap for `Path`.

src/InsetOrderOptimizer.cpp
Ghostkeeper: Done.

src/utils/polygonUtils.cpp
rburema: >  It would be up to the user of the function whether that would need to trigger a collision.

Document in header?

src/PathOrderOptimizer.h
rburema: Fair enough  : - )

PR #1301, opened 2020-07-28
Iron the topmost bottom layer when spiralizing a solid model.

smartavionics:
See https://community.ultimaker.com/topic/33489-ironing-does-not-work-with-spiralize-outer-contour-enabled/

### DIFF
--- a/src/TopSurface.cpp
+++ b/src/TopSurface.cpp
@@ -27,7 +27,7 @@ void TopSurface::setAreasFromMeshAndLayerNumber(SliceMeshStorage& mesh, size_t l
     {
         // when spiralizing, the model is often solid so it's no good trying to determine if there is air above or not
         // in this situation, just iron the topmost of the bottom layers
-        if (layer_number == mesh.settings.get<size_t>("bottom_layers") - 1)
+        if (layer_number == mesh.settings.get<size_t>("initial_bottom_layers") - 1)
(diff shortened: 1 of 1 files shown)


src/TopSurface.cpp
```diff
@@ -23,7 +23,19 @@ void TopSurface::setAreasFromMeshAndLayerNumber(SliceMeshStorage& mesh, size_t l
         mesh_above = mesh.layers[layer_number + 1].getOutlines();
     } //If this is the top-most layer, mesh_above stays empty.
 
-    areas = mesh.layers[layer_number].getOutlines().difference(mesh_above);
+    if (mesh.settings.get<bool>("magic_spiralize"))
+    {
+        // when spiralizing, the model is often solid so it's no good trying to determine if there is air above or not
+        // in this situation, just iron the topmost of the bottom layers
+        if (layer_number == mesh.settings.get<size_t>("bottom_layers") - 1)
```
Ghostkeeper, line 30: This should be the `initial_bottom_layers` setting, to coincide with other places where Spiralize is being used.
```suggestion
        if (layer_number == mesh.settings.get<size_t>("initial_bottom_layers") - 1)
```

smartavionics: Ah, yes, good catch, fix coming up...

PR #1303, opened 2020-08-12
Fix outer wall on inside.

rburema:
Using a meta-strategy (actually a convenience class that seems to combine both the meta and underlying strat) to ignore walls beyond the count. This enables us to use the whole shape instaed of first cutting out a tube/donut and _then_ applying lib-arachne. Besides the 'outer wall also in inside' this will fix a few other problems, like the small gaps we sometimes saw in the 'middle' wall and maybe the problems we have with getting support walls to be a proper ammount.

CURA-7553

### DIFF
--- a/src/WallsComputation.cpp
+++ b/src/WallsComputation.cpp
@@ -9,7 +9,7 @@
 #include "utils/polygonUtils.h"
 
 // libArachne
-#include "BeadingStrategy/DistributedBeadingStrategy.h" // TODO?: Might want to pull in the meta-strategies at some point.
+#include "BeadingStrategy/LimitedDistributedBeadingStrategy.h" // TODO?: Might want to pull in the meta-strategies at some point.
 #include "SkeletalTrapezoidation.h"
 
 namespace cura {
@@ -129,15 +129,15 @@ void WallsComputation::generateInsets(SliceLayerPart* part)
 
     constexpr float transitioning_angle = 0.5;
 
-    Polygons tubeshape = part->outline.tubeShape(line_width_0 + line_width_x * (inset_count - 1), 0).offset(-10).offset(10);
-    tubeshape.simplify(50, 50);
-    tubeshape.removeColinearEdges(0.03);
-    tubeshape.fixSelfIntersections();
-    tubeshape.removeSmallAreas(INT2MM(line_width_0/2) * INT2MM(line_width_0/2), false); // TODO: complete guess as to when arachne starts breaking, but it doesn't function well when an area is really small apearantly?
-    if (tubeshape.area() > 0)
+    Polygons prepared_outline = part->outline.offset(-10).offset(10);
+    prepared_outline.simplify(50, 50);
+    prepared_outline.removeColinearEdges(0.03);
+    prepared_outline.fixSelfIntersections();
+    prepared_outline.removeSmallAreas(INT2MM(line_width_0/2) * INT2MM(line_width_0/2), false); // TODO: complete guess as to when arachne starts breaking, but it doesn't function well when an area is really small apearantly?
+    if (prepared_outline.area() > 0)
     {
-        const DistributedBeadingStrategy beading_strat(line_width_0, line_width_0 * 2, transitioning_angle);  // TODO: deal with beading-strats & (their) magic parameters
-        SkeletalTrapezoidation wall_maker(tubeshape, beading_strat, beading_strat.transitioning_angle);
+        const LimitedDistributedBeadingStrategy beading_strat(line_width_0, line_width_0 * 2, inset_count * 2, transitioning_angle);  // TODO: deal with beading-strats & (their) magic parameters
+        SkeletalTrapezoidation wall_maker(prepared_outline, beading_strat, beading_strat.transitioning_angle);
         wall_maker.generateToolpaths(part->wall_toolpaths);
     }
 }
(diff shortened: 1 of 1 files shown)


src/WallsComputation.cpp
```diff
@@ -129,15 +129,15 @@ void WallsComputation::generateInsets(SliceLayerPart* part)
...
-    tubeshape.removeColinearEdges(0.03);
-    tubeshape.fixSelfIntersections();
-    tubeshape.removeSmallAreas(INT2MM(line_width_0/2) * INT2MM(line_width_0/2), false); // TODO: complete guess as to when arachne starts breaking, but it doesn't function well when an area is really small apearantly?
-    if (tubeshape.area() > 0)
+    Polygons prepared_outline = part->outline.offset(-10).offset(10);
+    prepared_outline.simplify(50, 50);
+    prepared_outline.removeColinearEdges(0.03);
+    prepared_outline.fixSelfIntersections();
+    prepared_outline.removeSmallAreas(INT2MM(line_width_0/2) * INT2MM(line_width_0/2), false); // TODO: complete guess as to when arachne starts breaking, but it doesn't function well when an area is really small apearantly?
+    if (prepared_outline.area() > 0)
```
jellespijker, line 137: why not check against `!prepared_outline.empty()` it stands to reason that after all simplifications there are at least three polygons present, which aren't self intersecting and therefore should have an area greater then 0.
This safes computation time, since the area() needs to be calculated while empty just checks well for empty ;-)

src/WallsComputation.cpp
```diff
@@ -129,15 +129,15 @@ void WallsComputation::generateInsets(SliceLayerPart* part)
...
-    tubeshape.simplify(50, 50);
-    tubeshape.removeColinearEdges(0.03);
-    tubeshape.fixSelfIntersections();
-    tubeshape.removeSmallAreas(INT2MM(line_width_0/2) * INT2MM(line_width_0/2), false); // TODO: complete guess as to when arachne starts breaking, but it doesn't function well when an area is really small apearantly?
-    if (tubeshape.area() > 0)
+    Polygons prepared_outline = part->outline.offset(-10).offset(10);
+    prepared_outline.simplify(50, 50);
+    prepared_outline.removeColinearEdges(0.03);
+    prepared_outline.fixSelfIntersections();
+    prepared_outline.removeSmallAreas(INT2MM(line_width_0/2) * INT2MM(line_width_0/2), false); // TODO: complete guess as to when arachne starts breaking, but it doesn't function well when an area is really small apearantly?
```
jellespijker, line 136: I see `line_width_0/2' being calculated multiple times, why not calculate it once and store it in a const var?

src/WallsComputation.cpp
```diff
@@ -129,15 +129,15 @@ void WallsComputation::generateInsets(SliceLayerPart* part)
...
     constexpr float transitioning_angle = 0.5;
 
-    Polygons tubeshape = part->outline.tubeShape(line_width_0 + line_width_x * (inset_count - 1), 0).offset(-10).offset(10);
-    tubeshape.simplify(50, 50);
-    tubeshape.removeColinearEdges(0.03);
-    tubeshape.fixSelfIntersections();
-    tubeshape.removeSmallAreas(INT2MM(line_width_0/2) * INT2MM(line_width_0/2), false); // TODO: complete guess as to when arachne starts breaking, but it doesn't function well when an area is really small apearantly?
-    if (tubeshape.area() > 0)
+    Polygons prepared_outline = part->outline.offset(-10).offset(10);
+    prepared_outline.simplify(50, 50);
```
jellespijker, line 133: Maybe use `constexpr` for the magic numbers in the `simplify` and `removeColinearEdges`.
These numbers will also be used by other libArachne wall generations, such as supportwalls. Maybe defined them once for the whole of CuraEngine?

jellespijker: same goes btw for the division factor used by the `LimitedDistributedBeadingStrategy(... line_width_0 * 2, inset_count * 2,..)` I would like to propose to make these `constexpr` for the whole of CuraEngine

rburema: Yeah agreed. I didn't pay much attention to this since they're likely to change (the functions themselves), but might as well make the parameters `constexpr` in case those functions _do_ stay where they are.

src/WallsComputation.cpp
rburema: Because I copied it over from somewhere else I think. Shame, I'm usually the first to remove such double calcs!

src/WallsComputation.cpp
rburema: I'm not sure, but I thought empty is not the same as area 0 in all cases (consider a degenerate polygon of 2 lines in a cycle for instance). You can also have negative area for holes, etc.

src/WallsComputation.cpp
rburema: Except the beading strategy parameters should maybe be up to the user ... Kostas is working on that as well.

konskarm: I've made the beading parameters user-defined:
image

src/WallsComputation.cpp
rburema: Just checked, area and empty don't do the same thing. I think this one's better, since we won't have potential infinitesimal extrusions this way (won't fix).

jellespijker: I agree with you assessment here.

src/WallsComputation.cpp
jellespijker: @konskarm looks good. This remark is unrelated to this ticket, but did you make that reusable? Such that it could also be used for support and other walls?

PR #1305, opened 2020-08-13
CURA-7555: Add libArachne parameters in the front-end

konskarm:
Currently Cura provides the necessary parameters only for computing the walls of the Shell. We may need to add similar parameters for support walls.

Cura equivalent: https://github.com/Ultimaker/Cura/pull/8189

CURA-7555


src/WallsComputation.cpp
```diff
@@ -125,16 +125,29 @@ void WallsComputation::generateInsets(SliceLayerPart* part)
...
     // Call on libArachne:
-    // TODO: Using line_width_0 here _even_ though it's also making the other walls! (that is, while line_width_x is adhered to when creating the area, it's ignored otherwise)
 
-    constexpr float transitioning_angle = 0.5;
+    const coord_t bead_width = line_width_0; // TODO: for now use only the outer wall line width
+    auto strategy_type       = settings.get<StrategyType>("beading_strategy_type");
+    auto transition_length   = settings.get<coord_t>     ("transition_length");
+    auto transitioning_angle = settings.get<AngleRadians>("transitioning_angle");
+
+    coord_t *min_bead_width = nullptr, *min_feature_size = nullptr;
```
jellespijker, line 134: Don't forget to clean these up

Ghostkeeper: There is no need for these to be pointers. You will get a setting value for it anyway, even if the setting is hidden in the front-end, so you can just set them to be whatever the setting value is. And if it needs to sometimes get a value and sometimes not, using the `optional` class would be a better choice anyway, because it gets stored in the stack then and doesn't need memory management.

If they are no longer pointers you don't need to use the `new` keyword for them any more. And you won't need to delete them any more either. Much easier! And imperceptibly faster!

For the BeadingStrategy we still need a pointer of some kind though, because we're using virtual overloading there, with its inheritance. A `shared_pointer` or a `weak_pointer` would be better, but since we're not using those elsewhere in our code either a normal pointer is acceptable.

konskarm: Currently the way the makeStrategy function is implemented, requires for both of these to be nullptrs in order to **NOT** apply the widening meta-strategy. If any of them has a value, it will apply it. Also we are not necessarily getting a value for those from the frontend unless it is explicitly requested to enable the widening beading. If this is enabled, then indeed the nullptrs are overridden by the front-end values and the widening gets enabled.
If I convert them to non-pointer variables, then I need to instead send an additional boolean that will tell the makeStrategy to apply or not the widening meta-scheme. I didn't go for this option but maybe I should to avoid the hassle of pointers.

konskarm: For now I am freeing the pointers, but feel free to go for the `optional` if you think the pointers should be avoided.

src/WallsComputation.cpp
```diff
@@ -125,16 +125,24 @@ void WallsComputation::generateInsets(SliceLayerPart* part)
     }
 
     // Call on libArachne:
-    // TODO: Using line_width_0 here _even_ though it's also making the other walls! (that is, while line_width_x is adhered to when creating the area, it's ignored otherwise)
 
-    constexpr float transitioning_angle = 0.5;
+    const coord_t bead_width = line_width_0; // TODO: for now use only the outer wall line width
+    auto strategy_type = settings.get<StrategyType>("beading_strategy_type");
+
+    coord_t *min_bead_width = nullptr, *min_feature_size = nullptr; // TODO: clean the pointers when we are done
```
jellespijker, line 132: `*` is part of the type `coord_t* min_bead_width`

I will change these

jellespijker (approved):

src/WallsComputation.cpp
Ghostkeeper: Instead of passing numerous setting values around to functions, we've recently preferred to pass the relevant settings objects around. This hasn't been possible everywhere (which is how we still have those functions like the `Infill` constructor with dozens of parameters) but it prevents the need for things like additional boolean parameters. When the function needs to check for such a `wideningEnabled` boolean, it can just check the setting value for itself. Keeps the implementation better contained to that function, too.

PR #1306, opened 2020-08-17
Workaround loss of first layer polygons when slicing tolerance is not middle.

smartavionics:
Without this PR, slicing anything with slicing tolerance not set to middle will result in the first layer polygons being empty. When the slicing mode is exclusive, that will make the first layer be removed. This PR stops that lossage.

### DIFF
--- a/src/slicer.cpp
+++ b/src/slicer.cpp
@@ -916,7 +916,8 @@ std::vector<SlicerLayer> Slicer::buildLayersWithHeight(size_t slice_layer_count,
     layers_res.resize(slice_layer_count);
 
     // set (and initialize compensation for) initial layer, depending on slicing mode
-    layers_res[0].z = std::max(0LL, initial_layer_thickness - thickness);
+    // slice first layer at height 1 uM not 0, otherwise it is empty
+    layers_res[0].z = std::max(1LL, initial_layer_thickness - thickness);
     coord_t adjusted_layer_offset = initial_layer_thickness;
     if (use_variable_layer_heights)
     {
@@ -927,7 +928,6 @@ std::vector<SlicerLayer> Slicer::buildLayersWithHeight(size_t slice_layer_count,
         layers_res[0].z = initial_layer_thickness / 2;
         adjusted_layer_offset = initial_layer_thickness + (thickness / 2);
     }
-
     // define all layer z positions (depending on slicing mode, see above)
     for (unsigned int layer_nr = 1; layer_nr < slice_layer_count; layer_nr++)
     {


Ghostkeeper:
Arguably you'll want to lose a bit on the top and bottom side of the print, since it's intended to provide more tolerance... Right?

smartavionics:
> Arguably you'll want to lose a bit on the top and bottom side of the print, since it's intended to provide more tolerance... Right?

Arguably, not. If I slice (using exclusive mode) a cylinder or some other shape that doesn't vary in cross section I don't expect to lose the top and bottom layers while in the horizontal, the outlines are not altered. Which is what happens at the moment.

smartavionics:
> Arguably you'll want to lose a bit on the top and bottom side of the print, since it's intended to provide more tolerance... Right?

I think you will find that it loses the first layer even when the mode is inclusive.

While I'm here I can also mention that the exclusive mode is asymmetric inasmuch that it replaces each layer's polygons with the intersection of the current layer and the layer above. Why doesn't it also take into account the layer below the current layer? As it is now, inverting an object on the build plate would change the shape of the object's polygons, is that really desirable behaviour?

Ghostkeeper:
> Arguably, not. If I slice (using exclusive mode) a cylinder or some other shape that doesn't vary in cross section I don't expect to lose the top and bottom layers while in the horizontal, the outlines are not altered. Which is what happens at the moment.

Every model varies in cross section. It varies at the very bottom and the very top, even if you just have a cube. I don't see why the very bottom or the very top would need to work differently. Layer height rounding aside, I'd expect that these two cylinders would have the same height:
image
One is a simple cylinder, the other is the same cylinder but with a bar through it. You could consider the second one to have a "sloped" side but that really doesn't matter. What matters is that they should both fit inside a 1cm slot with some tolerance. Making exceptions for the first and last layers would create inconsistencies then.

My findings when testing this are different from yours though. I have this model to test with: https://github.com/Ghostkeeper/SlicerTestModels/blob/master/text/layer_numbers.scad . It generates 101 layers (0 through 100). If you want to test with mirroring the model, you should set the initial layer height to be equal to the normal layer height. What I get with tolerance to "Middle" is 101 layers as you'd expect, 0 through 100:
image

When I set it to "exclusive", the first layer is a combination of 0 and 1:
image
So that means that it took a slice at Z 0 and found the "0" shape there, and a slice at Z0.1 and found the "1" shape there.

However at the top the result is not what I'd expect. The last layer is a combination of 98 and 99:
image
In the log it states that layer 99 (0-indexed) is empty. That means it took a slice at Z9.8 and Z9.9 and took the intersection of "98" and "99". And then it probably took a slice at Z9.9 and Z10.0 but didn't find the "100" at Z10.0, and then discarded that layer as empty. I would also expect there to be 100 layers (being 100 borders between the 101 filled layers) but it produced only 99.

What your pull request is really about is the rounding errors. If a triangle has an edge that is exactly on the height of a cross section, should it be taken into account with the cross section? Your result is wrong differently than mine, but both are wrong. If we'd expect the slicing to be symmetric regardless of whether a model is turned upside down (aside from initial layer height or adaptive layers) then we'd need to fix how it deals with rounding errors. Currently with 0.1mm layer height and a cube of 10mm, with Middle tolerance, you get 100 layers. But with a 10.05mm cube, you still get 100 layers. For a symmetric slice, that would need to be 101 layers then, with the 101st layer being at Z10.05 exactly.

The place to fix that would probably be either [with the original layer count calculation](https://github.com/Ultimaker/CuraEngine/blob/8971cfef7d91c5ca808dc76c58ce8e9a1adee24a/src/FffPolygonGenerator.cpp#L134) or [where it decides which triangle belongs to each layer](https://github.com/Ultimaker/CuraEngine/blob/8971cfef7d91c5ca808dc76c58ce8e9a1adee24a/src/slicer.cpp#L851-L898). Not with a 1 micrometre offset on just the first layer. Because that's going wrong for all layers then.

smartavionics:
Thanks for putting the effort into that reply. For me, this is all academic as I only ever use middle. However, other people seem to expect that using tolerances other than middle should not result in layers going missing. Even if you don't care for these PRs at least the issues they hoped to address have been highlighted.

PR #1307, opened 2020-08-17
Don't throwaway last layer when slicing tolerance is exclusive.

smartavionics:
I can't see the benefit in throwing away the final layer when the slicing tolerance is exclusive so this PR retains the final layer. Perhaps the final layer should actually be a duplicate of the previous layer, i'm not sure about that.

Without this PR, a whole layer is missing...

Screenshot_2020-08-17_11-46-03

With this PR, the layer is not missing but the outline is slight wrong compared to the previous layers....

Screenshot_2020-08-17_11-46-44

### DIFF
--- a/src/slicer.cpp
+++ b/src/slicer.cpp
@@ -968,7 +968,6 @@ void Slicer::makePolygons(Mesh& mesh, SlicingTolerance slicing_tolerance, std::v
         {
             layers[layer_nr].polygons = layers[layer_nr].polygons.intersection(layers[layer_nr + 1].polygons);
         }
-        layers.back().polygons.clear();
         break;
     case SlicingTolerance::MIDDLE:
     default:


Ghostkeeper:
> I can't see the benefit in throwing away the final layer when the slicing tolerance is exclusive

The benefit is that you'll have more tolerance between parts that need to fit together. That's more or less the intended use case of the setting.

Sophist-UK:
I have not examined the code, but my experience with similar issues is that this issue may be caused by a loop boundary condition with a < rather than a <= resulting in one less loop - and one less layer - that it should have had.

I also noted that with the Layer x of y gcode post-processor enabled, the last layer printed on a print today was 163 of 164 - layer 164 of 164 was not printed. This may be a genuine and significant bug.

Sophist-UK:
Here is a another picture showing a missing layer.
image

Sophist-UK:
Ok - I have now looked at the code, and I can see why Mark has proposed this change, and why it is not perfect.

To fix the issue shown whereby the top layer is NOT inside the profile, will not be easy. The problem here is that the Cura Engine does not create the polygons in the first place according to this setting, but instead applies a kludge later to create this effect. So to fix it, I think you probably need to re-engineer the polygon creation to apply this setting at source rather than as a later bolt on using the layer above.

Ghostkeeper:
Fixed in a different way here: https://github.com/Ultimaker/CuraEngine/pull/1313

Sophist-UK:
I am not commenting on the effectiveness of #1313, however as far as I can see, that does not fix this problem because that code does not take into account this setting nor does it remove this code which results in one less layer than there should be.

Ghostkeeper:
This line of code is done on purpose and shouldn't be removed.
The Exclusive tolerance takes the intersection between 2 cross sections to create the layer data for one layer. As a result, there will always be 1 layer less than the number of cross sections. That's also how it's implemented: You can see that the for loop just above it stops one short of `layers_ref.size()`. The last layer is not processed there. The last layer is the intersection between the last cross section and the layer above the total height of the model, which is always empty. With Inclusive, the last cross section is simply left as-is.

#1313 fixes the case where there was one layer too few if the last cross section was made exactly on the height of the last cross section. In that case the last layer used to be empty, but this will no longer be the case. That's why I believe it fixes this case, although I don't have that model that the issue was reported with originally. As far as I can see the number of layers is correct now.

Sophist-UK:
I don't think that #1313 can fix this issue of losing a layer since the code in this cancelled PR explicitly removes a layer.

To fix this properly, the Slicing Tolerance setting needs to be applied at the time that the polygon is created (which is actually the correct method) rather than as a post-creation fix-up (which is a kludge).

Ghostkeeper:
Like I said, the code indeed explicitly removes a layer which is *correct*.

It first creates N cross sections, which are then combined into N-1 layers. That last cross section is used two lines of code higher when it combined the last two cross sections into one layer. It still needs to create the cross section.

Sophist-UK:
Sorry - but whilst it is correct because it has no way of determining the profile of the top layer, I don't believe it is correct as it leaves the model 1 layer short on the z-axis i.e. smaller than the model by 1 layer. See the pictures above.

As I said previously, I believe that the correct algorithm would be to determine the profile edge position when you create the profile, not apply a kludge later.

Ghostkeeper:
> As I said previously, I believe that the correct algorithm would be to determine the profile edge position when you create the profile, not apply a kludge later.

Good. Because https://github.com/Ultimaker/CuraEngine/pull/1313 did adjust the edge cases when the cross sectional line segments are created, not any later. I think you might be confusing the pull requests here?

> I don't believe it is correct as it leaves the model 1 layer short on the z-axis i.e. smaller than the model by 1 layer. See the pictures above.

What do you think is the purpose of setting Slicing Tolerance to "Exclusive", if not to create more vertical tolerance? Yes, it's intended to create more vertical tolerance in every case, not just when the surface is sloped. That's a requirement, otherwise pieces may not fit together well for people that use the feature with models that have a flat top side.

Sophist-UK:
I do not believe that #1313 does as I am suggesting for two reasons:

1. If it did it would reference the Slicing Tolerance setting and create polygons differently depending on the setting; and

2. If it did, the code that this PR refers to would be redundant and have been deleted as part of #1313.

Sophist-UK:
Since the extrusion is not oblong in cross-section, and since the model's sides are generally not absolutely vertical, there are two reasons why the extrusion cannot be printed **horizontally** precisely to the model's surface. This is IMO what the Slicing Tolerance setting is trying to tweak.

The accuracy of the vertical z-axis print is generally more accurate because the extrusion has the edges of the nozzle exactly at the height specified which limit the height of the extrusion. **AFAIK, Cura already ensures that the top of the top layer does not exceed the model height, so there is no reason to lose another layer. (See pictures above provided by SmartAvionics).**

*****In other words, horizontal and vertical constraints need to be handled differently.*****

I believe that the purpose of `exclusive` is to make the print fit as closely as possible horizontally to the model, whilst staying strictly exactly-on or within the model's horizontal boundaries (e.g. in case it needs to fit inside something else).

I believe that the purpose of `inclusive` is to make the print fit as closely as possible to the model whilst ensuring that the extrusion is strictly exactly-on or outside the model's boundaries (e.g. in case you are sanding down the surface to make it perfectly smooth and you want the model to be exact once the sanding is done).

`inclusive` and `exclusive` require some guesswork on the horizontal shape of the extrusion which will depend on a number of factors IMO including but not limited to: filament type, nozzle temperature, nozzle shape, nozzle size, line width, layer height, extrusion speed, acceleration and jerk (which might affect the nozzle pressure), retraction prime, coasting, overlapping extrusions etc. and whether the extrusion is being laid down next to other previous extrusions. Because there are so many factors, it is probably not possible for Cura to know exactly what the filament cross section is. This it can only be an approximate guess about how the extrusion might extend beyond or be inside the line width which is what Cura calculates nozzle position with.

So `inclusive` and `exclusive` require some approximations of horizontal extrusion cross-section to adjust either the boundary of the polygons or the nozzle position when printing outer-walls so that the polygon / line-width edge is entirely inside / outside the model surface (assuming the extrusion is perfectly rectangular in cross-section) plus a further adjustment for how the extrusion isn't perfectly rectangular. The purpose of `middle` is IMO to not use such approximations, and fit the edge of the polygons and the edge of the line width to the exact position of the model wall half-way up the layer height, and let the other factors impact this without optimisations.

An additional set of rules need to apply for the top surface (which should probably have it's own "Vertical Slicing Tolerance" setting - with the existing setting renamed "Horizontal Slicing Tolerance"). This might also have settings: Inclusive, Middle, Exclusive. `Exclusive` would work as at present - slice an integral number of layers so that the top of the top layer was exactly-on or inside the model. `Middle` would slice an additional fixed height layer wherever any part of the `Exclusive` top layer was more than half of the layer-height inside the model. `Inclusive` would slice an additional fixed height layer wherever any part of the `Exclusive` top layer was inside the model. 

An additional option `Thin Top Surface` might also be possible, whereby you define a smaller layer height to be used for additional top layers, then combined with any of the above modes, in order that the tolerance can be finer than the standard layer height.

(An additional option `Variable` might be possible for `Vertical Slicing Tolerance` - but equally it might be technically impossible. The idea would be to print an additional variable height layer wherever the `Exclusive` top layer was inside the model, varying the layer height in order to match the difference. This may not be technically possible, because e.g. you might print a .2mm high extra extrusion on one area, and then a .1mm high extrusion immediately next door, and the sides of the nozzle would then impact the previous extrusion. However you might be able to print this last variable height layer by ordering the segments in increasing height but I suspect that this would lead to a very high number of short extrusions and travels and retractions likely resulting in poor surface quality.)

I hope this helps prompt discussion about this whole area.

Ghostkeeper:
The algorithm in `::makePolygons` that this PR is about uses an extra cross section as intermediary stage. Since it needs to combine two cross sections as part of its workings, it needs to throw away one extra cross section at the end after using it to create the last layer. That's what this line is about. Now you may think you'd want to keep that cross section and turn it into a complete layer, but:
* That layer will always be the same as the layer below unless the model is <2 layers thick at the top.
* Not deleting the layer will violate the requirements you stated here (which indeed coincide with my expected behaviour):
  > I believe that the purpose of exclusive is to make the print fit as closely as possible horizontally to the model, whilst staying strictly exactly-on or within the model's horizontal boundaries (e.g. in case it needs to fit inside something else).

To demonstrate, here is a screenshot of before this PR (i.e. current master):
Master
And here is a screenshot of after this PR:
mb-preserve-final-exclusive-layer

If the final layer is not deleted, the layer exceeds the volume of the model.

I'd also argue that every effect the Exclusive tolerance has will also be shown in lower layers than just the top, so making a fix for just the top layer will never be a good solution.

> If it did it would reference the Slicing Tolerance setting and create polygons differently depending on the setting

The reason why adjusting the edge cases has a big effect specifically on Exclusive, even though Exclusive is not mentioned in the code there, is because Exclusive is always hitting one specific edge case: It's slicing at Z=0mm. Previously, if a triangle had two vertices at Z=0mm and one vertex higher, its lower edge would not get included with the cross section at Z=0mm. Because typically no vertex goes below 0mm either, it would always make the first layer completely empty, and then the entire model would get moved down. With our recent adjustments those triangles will be counted as part of the cross section at Z=0mm and so the first layer is no longer always empty, and the model won't be shifted downwards and the vertical size is correct again.
Inclusive was also hitting that edge case, only Inclusive by nature then still makes a layer consisting of just the cross section at Z=0.1mm (or whatever the initial layer height is).

Note that the edge case is still the same if a model has an overhanging flat surface at an exact multiple of any layer height. It doesn't occur just at Z=0mm.

Now you also speak about modelling bead shapes that are not rectangular. Cura currently always models a line as a rectangular box which is pretty far from the truth sometimes. Although how far from the truth that is depends greatly on the context. Not just the material and environment, but also whether or not there is other material besides it. After all, Cura instructs the printer to extrude precisely the amount of volume required for the rectangular box. So if you're printing lines side by side, e.g. for skin, it should fill the space precisely. Reality is different, but the skin lines certainly have a different shape than a loose skirt line on the build plate.
Maybe the discussion about modelling different bead shapes than rectangular is not something we should have in a closed pull request though. It gets really complicated really quickly then.

Sophist-UK:
@Ghostkeeper I can see why you think this fixes the problem - since it prevents a missing layer i.e. effectively producing an extra layer so that the last layer is either exactly-on or above the model - and with this the code to delete the top layer when Slicing Tolerance is `Exclusive` now makes sense. You have tried to explain why the extra layer that is now created has the same horizontal profile as the one below, but I haven't understood it - intuitively it seems to me that it should be possible to make this profile fit the model in the same way as lower layers, because if our minds can envision this, then there should be a way of encoding it in an algorithm, but you appear to be saying that this is not the case.

Thinking about extrusion shape, the flow of melted filament seems to me to be reasonably complex to determine. The nozzle extrudes a circular cross section, so even in the simplest case where line-width = nozzle-width = layer-height and there are extrusions both sides the correct distance apart, it seems to me that the flow pressure is used to squish the filament into a rectangular (or square shape) against the adjacent extrusions and so how the extrusion flows is unclear. Then when you take away a constraining wall or change the line-width or layer-height, the flow and shape of the extrusion is less easy to predict. If there were results available of scientific measurements of this, that might allow Cura to make a better approximation, but I think you are correct that without these Cura should assume a rectangular cross-section, and users can tweak the material flow values to try to compensate for non-rectangular shapes (e.g. by reducing material flow to reduce edge distortion on e.g. outer walls, or increasing material flow (possibly with outer wall offset) to compensate for extra width on the outside of the outer wall and so less pressure and non-rectangular profile and lower adhesion against the inner wall. But it would really be quite useful to have at least some rules-of-thumb on how to use these settings to improve the adhesion between outer and inner walls whilst trying to make the outer wall skin as rectangular and accurate to the model as possible. (I might ask Stephan of CNC Kitchen if he might experiment on this.)

Finally, I am also confused about why the shape of the slicing pictures are different i.e. yours with square ends and @smartavionics with triangular ends. Is this something from Mark's own modded version of Cura or some other cause?

Sophist-UK:
@Ghostkeeper It occurs to me with a little more thought that the most common printing situation (leaving aside sparse infill where we are much less concerned about extrusion shape and adhesion to horizontally adjacent extrusions) is orinting with an extrusion below and to one side, and no extrusion on the other side.

And it seems to me that in this case, a slight offset of the nozzle towards the side which has an existing extrusion might increase the pressure which pushes the filament into a perfect fit to the existing wall, thus both increasing the inter-wall adhesion and avoiding gaps which will result in filament being pushed in the other direction making the other side of the extrusion be pushed outside the rectangular box.

On the other hand, I expect there to be some friction between the existing layer below and the bottom of the current layer being pushed outwards which would cause the outside profile of the extrusion to become convex, so the more you push the filament sideways as it is extruded the greater the convexity and distortion outside the rectangular model.

I will see if CNC Kitchen are interested in testing this theory.

Ghostkeeper:
It's very good and useful to be critical about such things, so it's good to have brought up the discussion. But from what I can see it's no longer an issue with Exclusive. There may of course still be other edge cases that we haven't seen yet, anyway, so it's good to have people think along.

> You have tried to explain why the extra layer that is now created has the same horizontal profile as the one below, but I haven't understood it

It's not the same if the top side of the model contains parts that are vertically very thin, but if the top layer is strictly smaller than the layer below then it would always be the same. I'll try to explain in different words and with an example:
`Exclusive` creates a cross section at the top side of the layer and a cross section at the bottom side of the layer, and then takes the intersection of those two polygons as the final layer. For example, using 0.1mm layers and slicing a model that's 5.33mm tall, the last layer would range from Z=5.2mm to Z=5.3mm. If the cross section at Z=5.3mm is strictly smaller than the cross section at Z=5.2mm, then the resulting intersection would be exactly the cross section at Z=5.3mm. By not throwing away the last cross section (at Z=5.3mm), you're then repeating that layer. But then in the span of height from Z=5.3mm to Z=5.4mm.

> Finally, I am also confused about why the shape of the slicing pictures are different i.e. yours with square ends and @smartavionics with triangular ends. Is this something from Mark's own modded version of Cura or some other cause?

No, I just reduced the Outer Wall Line Width to 0.01mm to make layer view show the lines closer to what CuraEngine actually models, when viewed from the side. A trick to clarify the screenshot better. :smile: 

Predicting the shape of the extruded line is indeed very complex. Ultimaker has done some research on this but I'm not at liberty to share that. There are also published papers about it (at least, I've seen some of BagelOrb's colleagues from the Delft University present those). Anecdotally, from what I've seen, the cross section of lines is normally like a rounded rectangle but with slightly larger rounding radius at the top side. You can see this best by freezing a print, then breaking the brittle PLA in two (e.g. with a hammer) and then using a magnifying glass or something on the breaking plane.

PR #1309, opened 2020-08-18
Cura 7551 fly on the wall

jellespijker:
CURA-7551


src/FffGcodeWriter.cpp
```diff
@@ -2379,159 +2380,186 @@ bool FffGcodeWriter::addSupportToGCode(const SliceDataStorage& storage, LayerPla
...
+            constexpr coord_t offset_from_outline = 0;
+            constexpr bool use_endpieces = true;
+            constexpr Polygons* perimeter_gaps = nullptr;
+            const Point infill_origin;
+            const Polygons in_outline = area_offset > 0 ? area_with_offset : area;
+
+            constexpr coord_t pocket_size = 0;
+            constexpr coord_t wall_line_count = 0;
+
+            Infill infill_comp(config.pattern,
```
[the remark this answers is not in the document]
jellespijker: Valid question. I used the naming scheme in the master branch ( https://github.com/Ultimaker/CuraEngine/blob/603554e1f80b100cbb50e5d694ac7ed96da8e334/src/FffGcodeWriter.cpp#L698 ).

I think it is short for computation, There are 12 usages in FffGcodeWriter.cpp of this type, most are named `infill_comp` except in the addSupport(Bottoms/Roofs)ToGCode here they're called bottom/roof_computation.
I will leave the name `infill_comp` but I will add a comment that here the infill computations are done

src/LayerPlan.h
```diff
@@ -587,6 +588,7 @@ class LayerPlan : public NoCopy
      */
     void addWall(ConstPolygonRef wall, int start_idx, const SliceMeshStorage& mesh, const GCodePathConfig& non_bridge_config, const GCodePathConfig& bridge_config, coord_t wall_0_wipe_dist, float flow_ratio, bool always_retract);
     void addWall(const std::vector<ExtrusionJunction>& wall, int start_idx, const SliceMeshStorage& mesh, const GCodePathConfig& non_bridge_config, const GCodePathConfig& bridge_config, coord_t wall_0_wipe_dist, float flow_ratio, bool always_retract, const bool is_closed = false, const bool is_reversed = false);
+    void addWall(const std::vector<ExtrusionJunction>& wall, int start_idx, const GCodePathConfig& path_config);
```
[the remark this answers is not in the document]
jellespijker: Yeah I got lazy, and I saw that the top addWall function also reused the old documentation. As punishment I will add documentation for both overloads ;-)

src/FffGcodeWriter.cpp
```diff
@@ -2379,159 +2380,187 @@ bool FffGcodeWriter::addSupportToGCode(const SliceDataStorage& storage, LayerPla
     return support_added;
 }
 
-
-bool FffGcodeWriter::processSupportInfill(const SliceDataStorage& storage, LayerPlan& gcode_layer) const
+bool FffGcodeWriter::processSupport(const SliceDataStorage& storage, LayerPlan& gcode_layer) const
```
Ghostkeeper, line 2383: This function doesn't process the support roof or support bottom, which is also part of support. So I think this rename is confusing!

src/FffGcodeWriter.cpp
```diff
@@ -2379,159 +2380,187 @@ bool FffGcodeWriter::addSupportToGCode(const SliceDataStorage& storage, LayerPla
...
+    // Cluster the used support settings for easy handling and maintainability
+    SupportConfig config
+    {
+        first_layer,
+        lower_layers,
+        extruder_nr,
+        SupportConfig::LineDistance(first_layer, infill_extruder),
+        infill_extruder.settings.get<coord_t>("infill_overlap_mm"),
+        SupportConfig::InfillAngle(lower_layers, gcode_layer.getLayerNr(), storage.support),
+        1, // there is no frontend setting for this (yet)
```
Ghostkeeper, line 2420: This is a "boolean trap" in integer form. I wouldn't know from reading what this `1` signifies.

src/FffGcodeWriter.cpp
```diff
@@ -2379,159 +2380,187 @@ bool FffGcodeWriter::addSupportToGCode(const SliceDataStorage& storage, LayerPla
...
     }
-    island_order_optimizer.optimize();
 
-    //Print the thicker infill lines first. (double or more layer thickness, infill combined with previous layers)
-    for(const PathOrderOptimizer<const SupportInfillPart*>::Path& path : island_order_optimizer.paths)
+    for(const auto& path : wall_toolpaths)
     {
-        const SupportInfillPart& part = *path.vertices;
+        if(path.empty())
+            continue;
```
Ghostkeeper, line 2458: Code style: Brackets around indentation.
```suggestion
        {
            continue;
        }
```

src/FffGcodeWriter.cpp
```diff
@@ -2379,159 +2380,187 @@ bool FffGcodeWriter::addSupportToGCode(const SliceDataStorage& storage, LayerPla
...
 
-            if (support_lines.size() > 0 || support_polygons.size() > 0)
+            constexpr coord_t offset_from_outline = 0;
+            constexpr bool use_endpieces = true;
+            constexpr Polygons* perimeter_gaps = nullptr;
+            const Point infill_origin;
+            const Polygons in_outline = area_offset > 0 ? area_with_offset : area;
+
+            constexpr coord_t pocket_size = 0;
+            constexpr coord_t wall_line_count = 0;
```
Ghostkeeper, line 2517: Previously we were using the `Infill` class with the appropriate `wall_line_count` to implement the correct number of walls around the support. The point of this was that most places where the infill patterns get used can also have extra walls around it: support infill, skin, and normal infill. If there needs to be a change, you'd then only need to implement that change in one spot.

You've set `wall_line_count` to 0 however. Why did you implement the extra wall here in `FffGcodeWriter` and not in the `Infill` class, where the implementation could get re-used for the skin and infill? I think this is going to introduce needless code duplication and more work to implement the other "extra wall" settings, and also add on to the `FffGcodeWriter` more logic which is not related to the orchestrating role of the class (which is already a problem for many other things in the class, I know). It'll also make the `wall_line_count` parameter unused code.

src/FffGcodeWriter.cpp
```diff
@@ -2379,159 +2380,187 @@ bool FffGcodeWriter::addSupportToGCode(const SliceDataStorage& storage, LayerPla
...
-            Polygons support_lines;
-            for (unsigned int density_idx = part.infill_area_per_combine_per_density.size() - 1; (int)density_idx >= 0; density_idx--)
-            {
-                if (combine_idx >= part.infill_area_per_combine_per_density[density_idx].size())
-                {
-                    continue;
-                }
+    // process sub-areas in this support infill area with different densities
+    if (config.line_distance <= 0 || part.infill_area_per_combine_per_density.empty())
+        return added_something;
```
Ghostkeeper, line 2480: Code style: brackets around indented code block.
```suggestion
    {
        return added_something;
    }
```

src/FffGcodeWriter.cpp
```diff
@@ -2379,159 +2380,187 @@ bool FffGcodeWriter::addSupportToGCode(const SliceDataStorage& storage, LayerPla
...
+            {
+                line_distance_here /= 2;
             }
 
-            if (support_lines.size() > 0 || support_polygons.size() > 0)
+            constexpr coord_t offset_from_outline = 0;
+            constexpr bool use_endpieces = true;
+            constexpr Polygons* perimeter_gaps = nullptr;
+            const Point infill_origin;
+            const Polygons in_outline = area_offset > 0 ? area_with_offset : area;
```
Ghostkeeper, line 2514: This makes a copy of the polygon data. Use a reference instead.

src/LayerPlan.h
```diff
@@ -577,17 +578,39 @@ class LayerPlan : public NoCopy
...
      * \param is_reversed Whether to print this wall in reverse direction.
      */
-    void addWall(ConstPolygonRef wall, int start_idx, const SliceMeshStorage& mesh, const GCodePathConfig& non_bridge_config, const GCodePathConfig& bridge_config, coord_t wall_0_wipe_dist, float flow_ratio, bool always_retract);
     void addWall(const std::vector<ExtrusionJunction>& wall, int start_idx, const SliceMeshStorage& mesh, const GCodePathConfig& non_bridge_config, const GCodePathConfig& bridge_config, coord_t wall_0_wipe_dist, float flow_ratio, bool always_retract, const bool is_closed = false, const bool is_reversed = false);
 
+    /*!
+     * Add a wall to the g-code starting
+     * \param wall he wall as ExtrusionJunctions
+     * \param path_config The config with which to print the wall lines
+     */
```
Ghostkeeper, line 611: Documentation seems to be incomplete here.

src/LayerPlan.cpp
```diff
@@ -1010,6 +1010,21 @@ void LayerPlan::addWall(const std::vector<ExtrusionJunction>& wall, int start_id
     }
 }
 
+void LayerPlan::addWall(const std::vector<ExtrusionJunction>& wall, const GCodePathConfig& path_config)
+{
+    ExtrusionJunction junction{*wall.begin()};
+    constexpr bool force_retract = false;
+    addTravel(junction.p, force_retract);
```
Ghostkeeper, line 1017: Would be good to document what this travel is doing.

src/LayerPlan.cpp
```diff
@@ -1010,6 +1010,21 @@ void LayerPlan::addWall(const std::vector<ExtrusionJunction>& wall, int start_id
     }
 }
 
+void LayerPlan::addWall(const std::vector<ExtrusionJunction>& wall, const GCodePathConfig& path_config)
+{
+    ExtrusionJunction junction{*wall.begin()};
```
Ghostkeeper, line 1015: Would segfault if the wall is an empty vector.

src/SupportInfillPart.cpp
```diff
@@ -14,27 +14,34 @@ SupportInfillPart::SupportInfillPart(const PolygonsPart& outline, coord_t suppor
...
+        constexpr float max_colinear_angle = 0.03;
+        constexpr bool remove_holes = false;
+        const double small_area_length = INT2MM(static_cast<double>(support_line_width) / 2);
+
+        prepared_outline = outline.offset(-epsilon_offset).offset(epsilon_offset);
+        prepared_outline.simplify(smallest_segment, allowed_distance);
+        prepared_outline.removeColinearEdges(max_colinear_angle);
+        prepared_outline.fixSelfIntersections();
+        prepared_outline.removeSmallAreas(small_area_length * small_area_length, remove_holes);
+        infill_area = outline.offset(-static_cast<coord_t>(small_area_length) - static_cast<coord_t>(support_line_width) * (inset_count_to_generate));
```
Ghostkeeper, line 43: There's a lot of magic here. Could you document what each of these steps are doing and why they are necessary? And why are those hard-coded parameters chosen?

Like, is it necessary for the Voronoi diagram to be generated or something? Maybe it would be good to have a re-usable function for that, or add it to the routine that actually generates the Voronoi.

src/settings/SupportConfig.cpp
```diff
@@ -0,0 +1,75 @@
...
+//CuraEngine is released under the terms of the AGPLv3 or higher.
+
+#include "SupportConfig.h"
+
+#include "types/Ratio.h"
+
+namespace cura
+{
+
+bool SupportConfig::LowerLayer(const size_t& layer_no)
```
Ghostkeeper, line 11: Code style: Methods use lower camel case.
```suggestion
bool SupportConfig::lowerLayer(const size_t& layer_no)
```
This also goes for the rest of the methods in this class.

Ghostkeeper, line 11: The reference to this parameter is not needed for objects that are smaller than the size of a reference. Also goes for some parameters in the other methods of this class (like the `bool`s).

src/settings/SupportConfig.cpp
```diff
@@ -0,0 +1,75 @@
...
+AngleDegrees SupportConfig::InfillAngle(const bool& lower_layers, const size_t& layer_no, const SupportStorage& storage)
+{
+    if (!lower_layers)
+    {
+        return storage.support_infill_angles.at(layer_no % storage.support_infill_angles.size());
+    }
+
+    // handles negative layer numbers
+    const auto divisor = storage.support_infill_angles_layer_0.size();
+    const auto index = ((layer_no % divisor) + divisor) % divisor;
```
Ghostkeeper, line 72: I dislike this use of `auto`. The `index` is a `LayerIndex` and that's important to know.

src/support.cpp
```diff
@@ -114,7 +116,7 @@ void AreaSupport::splitGlobalSupportAreasIntoSupportInfillParts(SliceDataStorage
 
 void AreaSupport::generateSupportInfillFeatures(SliceDataStorage& storage)
 {
-    AreaSupport::prepareInsetsAndInfillAreasForForSupportInfillParts(storage);
+    AreaSupport::prepareSupport(storage);
```
Ghostkeeper, line 119: Might as well be renamed to `prepare` then.
But maybe it's better to specify what it's preparing. In this case it seems to already actually generate the walls, through `wall_maker.generateToolpaths`?

src/support.h
```diff
@@ -4,6 +4,9 @@
 #ifndef SUPPORT_H
 #define SUPPORT_H
 
+#include <memory>
```
Ghostkeeper, line 7: Why is this necessary? You don't seem to use it here.

src/support.h
```diff
@@ -4,6 +4,9 @@
...
+#include "BeadingStrategy/BeadingStrategy.h"
```
Ghostkeeper, line 8: Move this include to the .cpp file, where it's actually being used.

src/support.cpp
```diff
@@ -406,44 +413,18 @@ void AreaSupport::combineSupportInfillLayers(SliceDataStorage& storage)
...
-        }
-    }
-}
-
 void AreaSupport::cleanup(SliceDataStorage& storage)
 {
-    const coord_t support_line_width = Application::getInstance().current_slice->scene.current_mesh_group->settings.get<coord_t>("support_line_width");
-    for (unsigned int layer_nr = 0; layer_nr < storage.support.supportLayers.size(); layer_nr++)
+    const auto support_line_width = Application::getInstance().current_slice->scene.current_mesh_group->settings.get<coord_t>("support_line_width");
+    for (auto& layer : storage.support.supportLayers)
```
Ghostkeeper, line 419: I dislike these uses of `auto` too. It's a bit of a matter of opinion, but it's useful to know that these are physical sizes (`coord_t`) and layer indexes (`LayerIndex`). And it's also useful for the compiler to break on it if the type happens to change, for instance the type of `supportLayers`.

src/FffGcodeWriter.cpp
jellespijker: I thought a got them all, this one slipped through. Will change it.

src/LayerPlan.cpp
jellespijker: In theory all empty walls, and therefore empty vectors, are all ready delete and filtered out, before it reaches this point.
What is your opinion about adding an assert, it will make it easier to debug, but won't slow down a release.

src/SupportInfillPart.cpp
jellespijker: Agreed with the use of a reusable function, for now these parameters and steps are copied how @rburema handles the walls, they need to be fine-tuned at a later time. The current values are "dikke duim" values that seem to work. I could add `// todo: magic number` So they are easier to find when we will fine-tune those parameters.

src/settings/SupportConfig.cpp
jellespijker: True. I will change that and therefore also remove the `const` since it will be passed by copy and the `const` will be redundant then

src/settings/SupportConfig.cpp
jellespijker: I personally like the use of `auto` letting the compiler decide the actual type and it will make the code more generic. That being said, the current code is still in C++11 and I have no objections to adhere to an older (not saying better or worse here) code style.
I will change it.

src/support.cpp
jellespijker: We differ on style ;-) and that is okay it would be a boring world otherwise :-) 
Since my fondness of `auto` deviates from most of the code currently used I will adapt. Maybe only use it for lambda's and the like from now on.

src/support.h
jellespijker: Remnant of an old test, will remove it

src/support.h
jellespijker: Remnant of an old test, will move it to the .cpp

src/LayerPlan.cpp
Ghostkeeper: An assertion is fine, yeah. As long as it has a good error message alongside it.

src/support.cpp
Ghostkeeper: The code base uses it for iterators and lambdas sometimes, to prevent the iterator's type repeating the container's type and repeating the types of the parameters in the lambda's type. But for the rest the code base is statically typed.

rburema: This set of my nerd alert: `auto` is _not_ dynamic typing! It still gets resolved at compile time, and is therefore still static :-p

PR #1313, opened 2020-08-28
Add documentation for and fix the edge cases encountered in slicing

Ghostkeeper:
There's a lot of thought in these if statements but it had no documentation at all. And it turns out there was something slightly wrong about it too in how it deals with edge cases.

Fixes Ultimaker/Cura#8240 and replaces #1306.


Ghostkeeper:
The only real change here is with the `=` symbols in the comparisons. The rest is documentation.

PR #1316, opened 2020-09-02
Shrinkage compensation

Ghostkeeper:
This is the CuraEngine component of the shrinkage compensation story.

In this implementation, it compensates for the inherent shrinkage of the plastic when the plastic cools down from the printing temperature to room temperature. The mesh is scaled as one of the very first steps then, transforming it with respect to the centre of the bounding box of the mesh group.

Implements CURA-7118.


CMakeLists.txt
```diff
@@ -196,6 +196,7 @@ set(engine_SRCS # Except main.cpp.
     src/utils/AABB.cpp
     src/utils/AABB3D.cpp
     src/utils/Date.cpp
+	src/utils/FMatrix4x3.cpp
```
nallath, line 199:
```suggestion
    src/utils/FMatrix4x3.cpp
```

src/MeshGroup.h
```diff
@@ -30,6 +30,15 @@ class MeshGroup : public NoCopy
...
     void finalize();
+
+	/*!
+	 * Scale the entire mesh group.
+	 *
+	 * The mesh group is scaled around the bottom center of its bounding box. So
+	 * that's the center in the X and Y directions, but Z=0. This simulates the
+	 * shrinkage while sticking to the build plate.
+	 */
+	void scale(const Ratio factor);
```
nallath, line 41: I think it would be better to name it "scaleFromBottom" or something like that. Scale by a factor to me tends to mean that it's doing the scale from the center (the documentation does say it though, so it's not really a big remark).

src/utils/FMatrix4x3.h
```diff
@@ -0,0 +1,63 @@
...
+/*!
+ * A 4x3 affine transformation matrix.
+ *
+ * This matrix behaves as if it's a 4x4 transformation matrix, but the bottom
+ * row is always identity.
+ */
+class FMatrix4x3
+{
+public:
+	/*!
```
nallath, line 23: This is suddenly indented with tabs. The other code of CuraEngine seems to use spaces.

nallath (changes requested):

CMakeLists.txt
Ghostkeeper: Ah, yeah. I've set my default in my IDE to indent with tabs and then made an exception for CuraEngine to indent .cpp and .h files with spaces, but that doesn't apply to CMakeLists.txt. Sorry!

src/MeshGroup.h
Ghostkeeper: Agreed! Is done now.

src/utils/FMatrix4x3.h
Ghostkeeper: Fixed all of them. I need to take another hard look at my IDE because this keeps going wrong. Even restarting the IDE sometimes fixes it so it's definitely buggy...

nallath (approved):

PR #1317, opened 2020-09-04
Fix a crash where stair-step code was called when not stair-stepping.

rburema:
JIRA ticket pending.


Ghostkeeper (approved):
No idea why the CI is breaking on publishing the package.

PR #1321, opened 2020-09-10
Fix simplify leaving way too small line segments. 

nallath:
https://github.com/Ultimaker/Cura/issues/8321

I've run some tests with this and it seems to resolve the issue reported.

**MezzerSealWithFix**
| Line length | Num segments |
| 1 | 60544 |
| 0.5 | 6256 |
| 0.1 | 686 |
| 0.05 | 32 |
| 0.01 | 32 |
| 0.005 | 32 |
| 0.001 | 32 |

**MezzerSealWithoutFix**
| Line length | Num segments |
| 1 | 92846 |
| 0.5 | 33547 |
| 0.1 | 5732 |
| 0.05 | 2359 |
| 0.01 | 719 |
| 0.005 | 579 |
| 0.001 | 0 |

**BenchyWithFix**
| Line length | Num segments |
| 1 | 120345 |
| 0.5 | 46781 |
| 0.1 | 2046 |
| 0.05 | 834 |
| 0.01 | 237 |
| 0.005 | 124 |
| 0.001 | 31 |

**BenchyWithoutFix**
| Line length | Num segments |
| 1 | 124375 |
| 0.5 | 50275 |
| 0.1 | 4015 |
| 0.05 | 2429 |
| 0.01 | 1596 |
| 0.005 | 1390 |
| 0.001 | 27 |

Benchy before:
BenchyBowBeforeFix

Benchy after:
BenchyBowAfterFix

CURA-7709

### DIFF
--- a/src/utils/polygon.cpp
+++ b/src/utils/polygon.cpp
@@ -295,17 +295,17 @@ Polygons ConstPolygonRef::offset(int distance, ClipperLib::JoinType join_type, d
     return ret;
 }
 
-bool lineLineIntersection(Point A, Point B, Point C, Point D, Point* output)
+bool lineLineIntersection(Point a, Point b, Point c, Point d, Point& output)
 {
     // Line AB represented as a1x + b1y = c1
-    double a1 = B.Y - A.Y;
-    double b1 = A.X - B.X;
-    double c1 = a1*(A.X) + b1*(A.Y);
+    double a1 = b.Y - a.Y;
+    double b1 = a.X - b.X;
+    double c1 = a1*(a.X) + b1*(a.Y);
 
     // Line CD represented as a2x + b2y = c2
-    double a2 = D.Y - C.Y;
-    double b2 = C.X - D.X;
-    double c2 = a2*(C.X)+ b2*(C.Y);
+    double a2 = d.Y - c.Y;
+    double b2 = c.X - d.X;
+    double c2 = a2*(c.X)+ b2*(c.Y);
 
     double determinant = a1 * b2 - a2 * b1;
 
@@ -316,8 +316,8 @@ bool lineLineIntersection(Point A, Point B, Point C, Point D, Point* output)
     }
     else
     {
-        output->X = (b2 * c1 - b1 * c2) / determinant;
-        output->Y = (a1 * c2 - a2 * c1) / determinant;
+        output.X = (b2 * c1 - b1 * c2) / determinant;
+        output.Y = (a1 * c2 - a2 * c1) / determinant;
         return true;
     }
 }
@@ -413,7 +413,7 @@ void PolygonRef::simplify(const coord_t smallest_line_segment_squared, const coo
                 // We should instead move this point to a location where both edges are kept and then remove the previous point that we wanted to keep.
                 // By taking the intersection of these two lines, we get a point that perseves the direction (so it makes the corner a bit more pointy)
                 Point intersection_point;
-                bool has_intersection = lineLineIntersection(previous_previous, previous, current, next, &intersection_point);
+                bool has_intersection = lineLineIntersection(previous_previous, previous, current, next, intersection_point);
                 
                 if (!has_intersection || LinearAlg2D::getDist2FromLine(intersection_point, previous, current) > allowed_error_distance_squared)
                 {
--- a/src/utils/polygon.cpp
+++ b/src/utils/polygon.cpp
@@ -418,7 +418,7 @@ void PolygonRef::simplify(const coord_t smallest_line_segment_squared, const coo
                 if(has_intersection)
                 {
                     // Find out if it is a degenerate intersection.
-                    if(vSize2(current - intersection_point) > allowed_error_distance_squared)
+                    if(LinearAlg2D::getDist2FromLine(intersection_point, previous, current) > allowed_error_distance_squared)
                     {
                         // The intersection is way to far away. Drop it. 
                         continue;
--- a/src/utils/polygon.cpp
+++ b/src/utils/polygon.cpp
@@ -435,10 +435,6 @@ void PolygonRef::simplify(const coord_t smallest_line_segment_squared, const coo
                     {
                         new_path.pop_back();
                     }
-                    else
-                    {
-                        continue;
-                    }
                 }
             }
             else
--- a/src/utils/polygon.cpp
+++ b/src/utils/polygon.cpp
@@ -295,32 +295,7 @@ Polygons ConstPolygonRef::offset(int distance, ClipperLib::JoinType join_type, d
     return ret;
 }
 
-bool lineLineIntersection(Point a, Point b, Point c, Point d, Point& output)
-{
-    // Line AB represented as a1x + b1y = c1
-    double a1 = b.Y - a.Y;
-    double b1 = a.X - b.X;
-    double c1 = a1*(a.X) + b1*(a.Y);
 
-    // Line CD represented as a2x + b2y = c2
-    double a2 = d.Y - c.Y;
-    double b2 = c.X - d.X;
-    double c2 = a2*(c.X)+ b2*(c.Y);
-
-    double determinant = a1 * b2 - a2 * b1;
-
-    if (determinant == 0)
-    {
-        // The lines are parallel
-        return false;
-    }
-    else
-    {
-        output.X = (b2 * c1 - b1 * c2) / determinant;
-        output.Y = (a1 * c2 - a2 * c1) / determinant;
-        return true;
-    }
-}
 
 
 void PolygonRef::simplify(const coord_t smallest_line_segment_squared, const coord_t allowed_error_distance_squared)
@@ -413,7 +388,7 @@ void PolygonRef::simplify(const coord_t smallest_line_segment_squared, const coo
                 // We should instead move this point to a location where both edges are kept and then remove the previous point that we wanted to keep.
                 // By taking the intersection of these two lines, we get a point that perseves the direction (so it makes the corner a bit more pointy)
                 Point intersection_point;
-                bool has_intersection = lineLineIntersection(previous_previous, previous, current, next, intersection_point);
+                bool has_intersection = LinearAlg2D::lineLineIntersection(previous_previous, previous, current, next, intersection_point);
                 
                 if (!has_intersection || LinearAlg2D::getDist2FromLine(intersection_point, previous, current) > allowed_error_distance_squared)
                 {
--- a/src/utils/polygon.cpp
+++ b/src/utils/polygon.cpp
@@ -405,7 +405,7 @@ void PolygonRef::simplify(const coord_t smallest_line_segment_squared, const coo
         }
   
         if (length2 < smallest_line_segment_squared
-            && height_2 <= allowed_error_distance_squared) // removing the vertex doesn't introduce too much error.) // removing the vertex doesn't introduce too much error.
+            && height_2 <= allowed_error_distance_squared) // removing the vertex doesn't introduce too much error.)
         {
             if(next_length2 > smallest_line_segment_squared)
             {
--- a/src/utils/polygon.cpp
+++ b/src/utils/polygon.cpp
@@ -354,9 +354,8 @@ void PolygonRef::simplify(const coord_t smallest_line_segment_squared, const coo
         const coord_t removed_area_next = current.X * next.Y - current.Y * next.X; // Twice the Shoelace formula for area of polygon per line segment.
         const coord_t negative_area_closing = next.X * previous.Y - next.Y * previous.X; // area between the origin and the short-cutting segment
         accumulated_area_removed += removed_area_next;
-        
+
         const coord_t length2 = vSize2(current - previous);
-        const coord_t next_length2 = vSize2(current - next);
 
         const coord_t area_removed_so_far = accumulated_area_removed + negative_area_closing; // close the shortcut area polygon
         const coord_t base_length_2 = vSize2(next - previous);
@@ -378,10 +377,11 @@ void PolygonRef::simplify(const coord_t smallest_line_segment_squared, const coo
         {
             continue;
         }
-  
+
         if (length2 < smallest_line_segment_squared
             && height_2 <= allowed_error_distance_squared) // removing the vertex doesn't introduce too much error.)
         {
+            const coord_t next_length2 = vSize2(current - next);
             if (next_length2 > smallest_line_segment_squared)
             {
                 // Special case; The next line is long. If we were to remove this, it could happen that we get quite noticeable artifacts.
@@ -424,7 +424,6 @@ void PolygonRef::simplify(const coord_t smallest_line_segment_squared, const coo
         previous = current; //Note that "previous" is only updated if we don't remove the vertex.
         new_path.push_back(current);
     }
-    
 
     *path = new_path;
 }
--- a/src/utils/polygon.cpp
+++ b/src/utils/polygon.cpp
@@ -407,7 +407,7 @@ void PolygonRef::simplify(const coord_t smallest_line_segment_squared, const coo
         if (length2 < smallest_line_segment_squared
             && height_2 <= allowed_error_distance_squared) // removing the vertex doesn't introduce too much error.)
         {
-            if(next_length2 > smallest_line_segment_squared)
+            if (next_length2 > smallest_line_segment_squared)
             {
                 // Special case; The next line is long. If we were to remove this, it could happen that we get quite noticeable artifacts.
                 // We should instead move this point to a location where both edges are kept and then remove the previous point that we wanted to keep.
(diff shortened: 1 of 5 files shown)


src/utils/polygon.cpp
```diff
@@ -372,7 +371,6 @@ void PolygonRef::simplify(const coord_t smallest_line_segment_squared, const coo
         if ((height_2 <= 1 //Almost exactly colinear (barring rounding errors).
             && LinearAlg2D::getDist2FromLine(current, previous, next) <= 1) // make sure that height_2 is not small because of cancellation of positive and negative areas
             || (length2 < smallest_line_segment_squared
-                && next_length2 < smallest_line_segment_squared // Segments are small
```
Ghostkeeper, line 375: Hmm, I wonder if this is a safe solution.

This line is there because removing a segment may have small deviation but a large effect if that causes a very long line segment to shift one of its vertices. Some ASCII art to illustrate:

```
  ________________________________________________>
 /
|
```

These are 3 line segments: One vertical, one diagonal and one horizontal. The winding order is along the arrow, clockwise in this illustration. The algorithm may decide to remove the diagonal line segment, which makes the long horizontal line shift:

```
                                                     ________
_______...............---------------''''''''''''''''
|
```

Those shifts are visible in the print.

I'll perform some testing if this is actually happening before putting it back though. The effects of changes here are sometimes unpredictable.

BagelOrb:
The currently proposed code change would violate the criteria described in the documentation of this function:
"Never remove a vertex if either of the connceted segments is larger than \p smallest_line_segment"

I see that where this function is called the two parameters smallest_line_segment and allowed_error_distance are both set to meshfix_maximum_resolution, while in the default parameters of the function signature the former is twice the latter.

Perhaps we should increase the smallest_line_segment to twice the meshfix_maximum_resolution.

Or perhaps we should simply bump up the Max Resolution setting for these printers. Note that the old setting values for that setting were tweaked for a broken implementation of simplify() which removed too much, so it makes sense that now that it is working again these printer definitions will have a Max Resolution setting which doesn't remove enough.

BagelOrb:
Please keep me in the loop.

src/utils/polygon.cpp
```diff
@@ -370,21 +399,63 @@ void PolygonRef::simplify(const coord_t smallest_line_segment_squared, const coo
...
-                && height_2 <= allowed_error_distance_squared) // removing the vertex doesn't introduce too much error.
-        )
+            && LinearAlg2D::getDist2FromLine(current, previous, next) <= 1)) // make sure that height_2 is not small because of cancellation of positive and negative areas
         {
-            continue; //Remove the vertex.
+            continue;
         }
+  
+        if (length2 < smallest_line_segment_squared
+            && height_2 <= allowed_error_distance_squared) // removing the vertex doesn't introduce too much error.) // removing the vertex doesn't introduce too much error.
```
BagelOrb, line 408: comment is duplicated

src/utils/polygon.cpp
```diff
@@ -370,21 +399,63 @@ void PolygonRef::simplify(const coord_t smallest_line_segment_squared, const coo
...
+        {
+            if(next_length2 > smallest_line_segment_squared)
```
BagelOrb, line 410: space after if

src/utils/polygon.cpp
```diff
@@ -370,21 +399,63 @@ void PolygonRef::simplify(const coord_t smallest_line_segment_squared, const coo
...
+            {
+                // Special case; The next line is long. If we were to remove this, it could happen that we get quite noticeable artifacts.
+                // We should instead move this point to a location where both edges are kept and then remove the previous point that we wanted to keep.
+                // By taking the intersection of these two lines, we get a point that perseves the direction (so it makes the corner a bit more pointy)
+                Point intersection_point;
+                bool has_intersection = lineLineIntersection(previous_previous, previous, current, next, &intersection_point);
 
-        //Don't remove the vertex.
+                if(has_intersection)
```
BagelOrb, line 418: The cases where there is no intersection or where it is far away are of the same kind. I suggest to simplify the code as follows:
```
if (!has_intersection || vSize2(current - intersection_point) > allowed_error_distance_squared)
{
  continue; // We coulnd't find a new place to combine the previous and current vertex, so remove the current vertex.
}
else
{
  // New point seems like a valid one.
  current = intersection_point;
... etc.
```

src/utils/polygon.cpp
```diff
@@ -370,21 +399,63 @@ void PolygonRef::simplify(const coord_t smallest_line_segment_squared, const coo
...
+                    current = intersection_point;
+                    
+                    // If there was a previous point added, remove it.
+                    if(new_path.size() > 0)
+                    {
+                        new_path.pop_back();
+                    }
+                    else
+                    {
+                        continue;
```
BagelOrb, line 437: Wait, I really don't think this should be a continue. Why alter current if you're gonna throw it away right after?
I think the continue should be omitted, so that the intersection_point gets added as the first thing on the output poly. Then at the end the algorithm should already omit the very last point, correct?

In any case this line of code needs a comment to justify / explain it.

src/utils/polygon.cpp
```diff
@@ -370,21 +399,63 @@ void PolygonRef::simplify(const coord_t smallest_line_segment_squared, const coo
...
-            continue; //Remove the vertex.
+            continue;
         }
+  
+        if (length2 < smallest_line_segment_squared
+            && height_2 <= allowed_error_distance_squared) // removing the vertex doesn't introduce too much error.) // removing the vertex doesn't introduce too much error.
+        {
+            if(next_length2 > smallest_line_segment_squared)
+            {
+                // Special case; The next line is long. If we were to remove this, it could happen that we get quite noticeable artifacts.
```
BagelOrb, line 412: The documentation of the function should be updated to reflect this functionality

src/utils/polygon.cpp
```diff
@@ -370,21 +399,63 @@ void PolygonRef::simplify(const coord_t smallest_line_segment_squared, const coo
...
+                bool has_intersection = lineLineIntersection(previous_previous, previous, current, next, &intersection_point);
 
-        //Don't remove the vertex.
+                if(has_intersection)
+                {
+                    // Find out if it is a degenerate intersection.
+                    if(vSize2(current - intersection_point) > allowed_error_distance_squared)
+                    {
+                        // The intersection is way to far away. Drop it. 
+                        continue;
```
BagelOrb, line 424: I'm not sure this should be a `continue`. Note that this code is in the case where the next segment is too long to be able to remove the vertex according to criterium 1. 

image
black = original; red = simplified according to this line.

src/utils/polygon.cpp
BagelOrb: Note that this may violate the allowed error distance constraint, because here we only checked on `height_2 <= allowed_error_distance_squared` , but we didn't check on the hieght_2 of the previous vert.

image

src/utils/polygon.cpp
```diff
@@ -370,21 +399,63 @@ void PolygonRef::simplify(const coord_t smallest_line_segment_squared, const coo
...
+                // We should instead move this point to a location where both edges are kept and then remove the previous point that we wanted to keep.
+                // By taking the intersection of these two lines, we get a point that perseves the direction (so it makes the corner a bit more pointy)
+                Point intersection_point;
+                bool has_intersection = lineLineIntersection(previous_previous, previous, current, next, &intersection_point);
 
-        //Don't remove the vertex.
+                if(has_intersection)
+                {
+                    // Find out if it is a degenerate intersection.
+                    if(vSize2(current - intersection_point) > allowed_error_distance_squared)
```
BagelOrb, line 421: Note that this ignores the height constraint for the special case. 
image

It should read  `if (LinearAlg2D::getDist2FromLine(intersection_point, previous, current) > allowed_error_distance_squared)`

src/utils/polygon.cpp
BagelOrb: (The if condition should be altered, as suggested below)

src/utils/polygon.cpp
nallath: The distance is only too long if the lines are almost paralel. At that point, it will create an intersection that is very far away, which results in weird ass spikes being put there.

BagelOrb: There are 3 alternatives: either we A) remove the current vertex, or B) we remove the current and last vertex and use the intersection point, or we C) leave both in. In this if-case you opt for A, but I am saying we should opt for C. You misinterpreted my comment to mean that we should opt for B.

BagelOrb: Possible 4th alternative: cut a long segment so that the shortcut is as long as the allowed error distance (or smallest segment length ?)
image

BagelOrb:
The function documentation should probably change, because the following would be unsimplifiable according to spec:
```
""""""""""""""""""""""""""""""""i,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
```

src/utils/polygon.cpp
nallath: Current solution (A):
CurrentSolution
Leave both in (C):
KeepBoth

Option C is retaining some of the weird points. I also think that my previous assessment is incorrect; It happens to be wrong if both of the lines are almost in the same direction; It's probably because of a rounding error that the found point is very far away.

src/utils/polygon.cpp
```diff
@@ -295,6 +295,34 @@ Polygons ConstPolygonRef::offset(int distance, ClipperLib::JoinType join_type, d
     return ret;
 }
 
+bool lineLineIntersection(Point A, Point B, Point C, Point D, Point* output)
```
Ghostkeeper, line 298: We should probably make this part of a namespace class like `LinearAlg2D`. And use lowercase variable names.

Ghostkeeper: Use a reference for the output, rather than a pointer. You don't need the reference to change to a different address.

src/utils/polygon.cpp
Ghostkeeper: > Wait, I really don't think this should be a continue. Why alter current if you're gonna throw it away right after?
I think the continue should be omitted, so that the intersection_point gets added as the first thing on the output poly. Then at the end the algorithm should already omit the very last point, correct?

Agreed! As it stands now, `current` would get skipped instead of `previous` in this case, and you might alter a very long line segment between `current` and `next`.

Ghostkeeper: > Note that this may violate the allowed error distance constraint, because here we only checked on height_2 <= allowed_error_distance_squared , but we didn't check on the hieght_2 of the previous vert.

No, because we move `current` to the intersection between `prevprev -> prev` and `current -> next`. Forgive my crude edit of your image:
simplify_intersection
So that ignored height should be 0.

Ghostkeeper (changes requested):

src/utils/polygon.cpp
nallath: As discussed with @Ghostkeeper, we've opted to go for option E.

It's basically option C, but when the line segment is below 5 mu, we still discard it.

src/utils/polygon.h
```diff
@@ -449,15 +449,19 @@ class PolygonRef : public ConstPolygonRef
...
-     * 1. Never remove a vertex if either of the connceted segments is larger than \p smallest_line_segment
-     * 2. Never remove a vertex if the distance between that vertex and the final resulting polygon would be higher than \p allowed_error_distance
-     * 3. Simplify uses a heuristic and doesn't neccesarily remove all removable vertices under the above criteria.
-     * 4. But simplify may never violate these criteria.
-     * 5. Unless the segments or the distance is smaller than the rounding error of 5 micron
+     * Note that the simplify is a best effort algorithm. It does not guarantee that no lines below the provided smallest_line_segment_squared are left.
+     *
+     * The following example (Two very long line segments (" & , respectively) that are connected by a very small line segment (i) is unsimplifable by this
+     * function, even though the actual area change of removing line segment i is very small. The reason for this is that in the case of long lines, even a small
+     * deviation from it's original direction is very noticeable in the final result, especially if the polygons above make a slightly different choice.
```
BagelOrb, line 461: I think this new documentation made the criteria a bit more fuzzy. We should formulate precise criteria which `simplify` may never violate.

Let's add the criterion:
`The direction of segments longer than \p smallest_line_segment always remains unaltered (but their end points may change if it is connected to a small segment)`

src/utils/polygon.h
```diff
@@ -449,15 +449,19 @@ class PolygonRef : public ConstPolygonRef
     /*! 
      * Removes consecutive line segments with same orientation and changes this polygon.
      * 
-     * Removes verts which are connected to line segments which are both too small.
-     * Removes verts which detour from a direct line from the previous and next vert by a too small amount.
+     * 1. Removes verts which are connected to line segments which are too small.
+     * 2. Removes verts which detour from a direct line from the previous and next vert by a too small amount.
+     * 3. Moves a vert when a small line segment is connected to a much longer one. in order to maintain the outline of the object.
```
BagelOrb, line 454: I think you should call it 'merging the 2 verts' rather than 'moving one vert'.

src/utils/polygon.cpp
```diff
@@ -370,21 +399,58 @@ void PolygonRef::simplify(const coord_t smallest_line_segment_squared, const coo
...
+                // We should instead move this point to a location where both edges are kept and then remove the previous point that we wanted to keep.
+                // By taking the intersection of these two lines, we get a point that perseves the direction (so it makes the corner a bit more pointy)
+                Point intersection_point;
+                bool has_intersection = lineLineIntersection(previous_previous, previous, current, next, &intersection_point);
+                
+                if (!has_intersection || LinearAlg2D::getDist2FromLine(intersection_point, previous, current) > allowed_error_distance_squared)
+                {
+                    if(length2 < 25)
+                    {
+                        // We're allowed to always delete segments of less than 5 micron.
```
BagelOrb, line 422: Wait no this is dangerous! The whole idea of this function was to avoid escalating segment removal. The default argument `allowed_error_distance` is already 5mu; this whole function was built to be able to remove 5mu segments, but do it in a safe way.

The continue here should be removed.

src/utils/polygon.cpp
BagelOrb: Agreed. The `continue` problem is solved, so this is not an issue any more.

src/utils/polygon.cpp
BagelOrb: This whole function was built to remove line segments below 5mu, but to do it in a safe way. If you choose option E then we lose the safety this function should guarantee. We might introduce escalation problems again where a lot of consecutive 5mu line segments get removed and introduce a too big error.

BagelOrb (changes requested):
I'm not sure whether it's safe to throw away 5mu segments just like that.

src/utils/polygon.h
nallath: It's not _quite_ merging them.  I would interpret merging as averaging the two points.

src/utils/polygon.cpp
nallath: It only does it if the intersection fails. The intersection only fails if the direction of the two lines is almost exactly the same. 

So it's dangerous if you don't take it into context.

BagelOrb: The intersection only fails if the direction of the previous and next line is almost exactly the same, but not the same as the current line segment, so it's not like the collinearity means that this case can be ignored.

> The alternative is that I simply always delete the point, regardless if the length is <5mu, which is the solution I had before this.

The alternative is option C or E, not option B. That is: either leave the small segment in, or move either of the verts of the current segment along the long segment, such that the current segment becomes as long as `smallest_line_segment`.

> If we don't delete these tiny points, we still get these very tiny line segments in some spots, which wreak havoc on a print.

That may be the case for some printers, but I'd like to challenge the idea that this is inherently a problem for any firmware. Only an abundance of tiny line segments should wreak havoc, but as long as they are connected to a long segment it shouldn't cause buffer underruns and so it shouldn't be a problem.

I don't think `simplify` should be made to guarantee that no small line segments exist. The function is called all over the place in the engine; if it's a gcode problem then maybe we need a separate filtering step at the gcode generation phase. For example, alter `LayerPlan::addExtrusionMove` to only do anything if the line segment is long enough.

BagelOrb: Hmm. Actually...

We are in the case where `if (next_length2 > smallest_line_segment_squared)` so there is no escalation problem.

Still I think it's better to adhere to the criterion that no segment longer than `shortest_line_segment` may have it's direction altered, which is possible with option C or D.

src/utils/polygon.h
BagelOrb: Yeah, but it's even less moving one point. Maybe just 'dissolving the middle segment'?

src/utils/polygon.cpp
BagelOrb: In a live talk we decided that the effort and code complexity wasnt worth it.

rburema (approved):

PR #1323, opened 2020-09-11
Cura 7626 infill walls

jellespijker:
This PR introduces some using aliases for ExtrusionJunction and ExtrusionLine Containers. I also created a class WallToolPaths in order to minimize code duplication. These changes didn't take into account @Ghostkeeper lattest fixes for the near intersect.

The following methods are not yet supported:

~~Multiply lines (see PR #1326 for fix on master)
Concentric infill
Cross 3D infill
Randomize infill start
ZSeam (if applicable)
Infill Travel optimization~~
I created this draft PR to get your opinion on the logic. I still have to get over it with a fine comb for code style. But I thought it prudent make you aware of my current approach.


src/FffGcodeWriter.cpp
```diff
@@ -1613,58 +1617,76 @@ bool FffGcodeWriter::processSingleLayerInfill(const SliceDataStorage& storage, L
...
+                for (const PathJunctions& paths : bins)
+                {
+                    for (const LineJunctions& line : paths)
+                    {
+//                      Todo: do a proper translation of these commands to the addWall
+//                      gcode_layer.addTravel(PolygonUtils::findNearestVert(gcode_layer.getLastPlannedPositionOrStartingPosition(), infill_polygons).p(), force_comb_retract);
+//                      gcode_layer.addPolygonsByOptimizer(infill_polygons, mesh_config.infill_config[0], ZSeamConfig(), 0, false, 1.0_r, false, false, near_start_location);
+                        gcode_layer.addWall(line, mesh_config.infill_config[0], false);
+                    }
+                }
```
Ghostkeeper, line 1677: It would make more sense to let it add the polygons by optimizer again (but then in a way that's compatible with these variable width paths).
The logic that dictates that lines should be printed in indexed order should be contained in the `PathOrderOptimizer` which chooses between indexed ordering and the to-be-implemented optimized ordering by itself. Having logic regarding these "bins" outside of that class would be bad design. Let that class convert the toolpaths to these bins if it needs to, and otherwise not.

src/WallToolPaths.cpp
```diff
@@ -0,0 +1,125 @@
...
+    constexpr float transitioning_angle = 0.5;
+
+    Polygons prepared_outline = outline.offset(-epsilon_offset).offset(epsilon_offset);
+    prepared_outline.simplify(smallest_segment, allowed_distance);
+    PolygonUtils::fixSelfIntersections(epsilon_offset, prepared_outline);
+    prepared_outline.removeDegenerateVerts();
+    prepared_outline.removeColinearEdges();
+    // TODO: complete guess as to when arachne starts breaking, but it doesn't  function well when an area is really
+    // small apparently?
+    prepared_outline.removeSmallAreas(small_area_length * small_area_length, false);
```
Ghostkeeper, line 42: Was this code copied from somewhere?

src/WallToolPaths.cpp
```diff
@@ -0,0 +1,125 @@
...
+    }
+    return inner_contour;
+}
+
+const Polygons& WallToolPaths::getOutline() const
+{
+    return outline;
+}
+
+const BinJunctions& WallToolPaths::getBinJunctions()
```
Ghostkeeper, line 88: (Again, I think the whole logic with `BinJunctions` should be kept within the indexed-ordering function of the `PathOrderOptimizer` class.)

src/infill.cpp
```diff
@@ -93,9 +114,94 @@ void Infill::generate(Polygons& result_polygons, Polygons& result_lines, const S
     }
 }
 
+void Infill::_generate(Polygons& result_lines, const SierpinskiFillProvider* cross_fill_provider, const SliceMeshStorage* mesh)
```
Ghostkeeper, line 117: Why did you choose to move this part to a private function?

src/utils/polygon.h
```diff
@@ -991,9 +992,40 @@ class Polygons
         }
     }
 
-    void fixSelfIntersections()
+    void resize(const coord_t& numerator, const coord_t divisor)
```
Ghostkeeper, line 995: Why does this function take a numerator and divisor, rather than just a `const Ratio scale` or something? Seems a bit confusing.

src/utils/polygon.h
```diff
@@ -991,9 +992,40 @@ class Polygons
...
+            return;
+        }
+
+        Polygons& thiss = *this;
+        for (size_t p = 0; p < size(); p++)
+        {
+            for (Point& pt : thiss[p])
+            {
+                pt.X = (pt.X * numerator) / divisor;
+                pt.Y = (pt.Y * numerator) / divisor;
```
Ghostkeeper, line 1008: The `Point` class also supports the operator `*` which does the same.

src/utils/polygon.h
```diff
@@ -991,9 +992,40 @@ class Polygons
...
+            }
+        }
+    }
+
+    void translate(const Point vec)
```
Ghostkeeper, line 1013: Duplicates the `PolygonRef::translate` function.

src/utils/polygon.h
Ghostkeeper, line 995: Since the `Polygons` class generally behaves like a vector, it would perhaps be better to rename this to `scale` or something like that. The `resize` function of a vector changes the length of the vector.

Ghostkeeper (changes requested):

src/utils/ExtrusionLine.h
```diff
@@ -55,8 +55,9 @@ struct ExtrusionLine
     /*!
      * Export the included junctions as vector.
      */
-    void appendJunctionsTo(std::vector<ExtrusionJunction>& result) const;
+    void appendJunctionsTo(LineJunctions& result) const;
 };
 
+using ToolPaths = std::vector<std::list<ExtrusionLine>>; //<! The toolpaths generated by libArachne
```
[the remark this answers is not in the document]
jellespijker: VariableWidthPath is indeed a better type name I will adjust it.

src/FffGcodeWriter.cpp
```diff
@@ -1468,63 +1466,65 @@ bool FffGcodeWriter::processSingleLayerInfill(const SliceDataStorage& storage, L
     {
         return false;
     }
-    const coord_t infill_line_distance = mesh.settings.get<coord_t>("infill_line_distance");
```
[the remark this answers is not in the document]
jellespijker: Clang tidy warns against this, since the get specifies the type name in the template all ready.
Since auto isn't dynamically determined but known at compile time. I'm tempted to leave it as auto otherwise the code base will never be modernized.

But if you feel strongly about it I have no problems returning it to hard coded type.

src/FffGcodeWriter.cpp
```diff
@@ -1468,63 +1466,65 @@ bool FffGcodeWriter::processSingleLayerInfill(const SliceDataStorage& storage, L
...
+    const auto infill_overlap = mesh.settings.get<coord_t>("infill_overlap_mm");
+    const auto infill_multiplier = mesh.settings.get<size_t>("infill_multiplier");
+    const auto wall_line_count = mesh.settings.get<size_t>("infill_wall_line_count");
     AngleDegrees infill_angle = 45; //Original default. This will get updated to an element from mesh->infill_angles.
-    if (mesh.infill_angles.size() > 0)
+    if (!mesh.infill_angles.empty())
     {
         const size_t combined_infill_layers = std::max(unsigned(1), round_divide(mesh.settings.get<coord_t>("infill_sparse_thickness"), std::max(mesh.settings.get<coord_t>("layer_height"), coord_t(1))));
-        infill_angle = mesh.infill_angles.at((gcode_layer.getLayerNr() / combined_infill_layers) % mesh.infill_angles.size());
+        infill_angle = mesh.infill_angles.at((static_cast<unsigned long>(gcode_layer.getLayerNr()) / combined_infill_layers) % mesh.infill_angles.size());
```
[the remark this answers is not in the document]
jellespijker: Agreed will change it

src/FffGcodeWriter.cpp
```diff
@@ -1533,20 +1533,20 @@ bool FffGcodeWriter::processSingleLayerInfill(const SliceDataStorage& storage, L
...
             Polygons skin_above_upper;  // skin regions on the 2nd and subsequent layers above
 
             for (size_t i = 1; i <= skin_edge_support_layers; ++i)
             {
-                const size_t skin_layer_nr = gcode_layer.getLayerNr() + i;
+                const size_t skin_layer_nr = static_cast<unsigned long>(gcode_layer.getLayerNr()) + i;
                 if (skin_layer_nr < mesh.layers.size())
                 {
-                    for (const SliceLayerPart& part : mesh.layers[skin_layer_nr].parts)
+                    for (const SliceLayerPart& layerPart : mesh.layers[skin_layer_nr].parts)
```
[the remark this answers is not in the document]
jellespijker: Agreed will change this, I also updated my IDE so it will suggest the propper names next time

src/FffGcodeWriter.cpp
```diff
@@ -1613,58 +1617,76 @@ bool FffGcodeWriter::processSingleLayerInfill(const SliceDataStorage& storage, L
...
+    if (!infill_lines.empty() || walls_generated)
     {
         added_something = true;
         setExtruder_addPrime(storage, gcode_layer, extruder_nr);
         gcode_layer.setIsInside(true); // going to print stuff inside print object
         std::optional<Point> near_start_location;
         if (mesh.settings.get<bool>("infill_randomize_start_location"))
         {
-            srand(gcode_layer.getLayerNr());
+            srand(static_cast<unsigned int>(gcode_layer.getLayerNr()));
```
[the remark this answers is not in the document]
jellespijker: I believe so as well, will check once I'm behind a computer again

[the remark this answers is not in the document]
jellespijker: I wanted to ask your opinion about the use of rand(). This function has limited randomness it a better option would be to use the <random> library and functions instead.

See https://wiki.sei.cmu.edu/confluence/plugins/servlet/mobile?contentId=88046554#content/view/88046554

src/FffGcodeWriter.cpp
```diff
@@ -1613,58 +1617,76 @@ bool FffGcodeWriter::processSingleLayerInfill(const SliceDataStorage& storage, L
...
             if(!infill_lines.empty())
             {
-                near_start_location = infill_lines[rand() % infill_lines.size()][0];
+                near_start_location = infill_lines[static_cast<unsigned int>(rand()) % infill_lines.size()][0];
```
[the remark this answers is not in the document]
jellespijker: Agreed will change

src/utils/polygonUtils.cpp
```diff
@@ -1410,4 +1410,69 @@ Polygons PolygonUtils::connect(const Polygons& input)
     return ret;
 }
 
+/* Note: Also tries to solve for near-self intersections, when epsilon >= 1
+ */
+void PolygonUtils::fixSelfIntersections(const coord_t epsilon, Polygons& thiss)
```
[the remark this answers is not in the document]
jellespijker: Pulled it in with commit f644a74
@rburema moved to PolygonUtils

src/FffGcodeWriter.cpp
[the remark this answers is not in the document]
jellespijker: Changed the cast to `size_t` since `size_t == unsigned long`

src/infill.cpp
jellespijker: The original _generate( ... ) which I intend to overload was also private. Not sure if I'm gonna need this part in the final implementation.

src/utils/polygon.h
jellespijker: I'll have to look into this, since I'm not the original author. I think @rburema wrote it, maybe he can comment on this.

src/utils/polygon.h
jellespijker: I'll have to look into this, since I'm not the original author. I think @rburema wrote it, maybe he can comment on this.

src/utils/polygon.h
jellespijker: I'll have to look into this, since I'm not the original author. I think @rburema wrote it, maybe he can comment on this.

src/utils/polygon.h
jellespijker: I'll have to look into this, since I'm not the original author. I think @rburema wrote it, maybe he can comment on this.

src/FffGcodeWriter.cpp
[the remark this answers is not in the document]
Ghostkeeper: We should make one change at a time. If there's a bug with the uniform distribution it'll be hard to trace down. And we want to get things finished too.
Furthermore there are different levels of security for random number generators. The uniform distribution is not yet considered cryptographically secure, but does remove some modulo bias. But that also means it's more expensive to calculate. We don't need any security here since there is no attack vector. The worst that an attacker could do is cause the Z seams to align in a certain way.

src/WallToolPaths.cpp
jellespijker: @rburema his implementation

src/WallToolPaths.cpp
jellespijker: It was Moved it to InsetOrderOptimizer

src/FffGcodeWriter.cpp
```diff
@@ -1613,58 +1617,76 @@ bool FffGcodeWriter::processSingleLayerInfill(const SliceDataStorage& storage, L
...
             }
             else
             {
-                PolygonRef start_poly = infill_polygons[rand() % infill_polygons.size()];
-                near_start_location = start_poly[rand() % start_poly.size()];
+                // TODO: libArachne randomize infill walls
+                PolygonRef start_poly = infill_polygons[static_cast<unsigned int>(rand()) % infill_polygons.size()];
+                near_start_location = start_poly[static_cast<unsigned long>(rand()) % start_poly.size()];
```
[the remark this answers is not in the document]
jellespijker: Will create a ticket in Jira so this can be implemented at a later time

src/utils/polygon.h
jellespijker: Agreed will change

jellespijker: Maybe because of rounding errors. Why I tried your suggestion just now it didn't print any walls

src/utils/polygon.h
jellespijker: Changed

src/WallToolPaths.cpp
Ghostkeeper: Ah I see it was moved away from `WallsComputation` to this new WallToolPaths. Fine then, at least no code duplication.

src/WallToolPaths.h
```diff
@@ -0,0 +1,85 @@
...
+#include "utils/ExtrusionLine.h"
+#include "utils/polygon.h"
+
+namespace cura
+{
+class WallToolPaths
+{
+public:
+    /*!
+     * A class that creates the toolpaths given an outline, nominal bead width and maximum amount of walls
```
Ghostkeeper, line 20: Offline remark: Is this going to replace `WallsComputation` then?

src/infill.cpp
```diff
@@ -93,9 +123,94 @@ void Infill::generate(Polygons& result_polygons, Polygons& result_lines, const S
...
+            break;
+        case EFillMethod::CROSS:
+        case EFillMethod::CROSS_3D:
+            if (!cross_fill_provider)
+            {
+                logError("Cannot generate Cross infill without a cross fill provider!\n");
+                break;
+            }
+//            generateCrossInfill(*cross_fill_provider, result_polygons, result_lines);
+            // TODO: handle CROSS_3D
```
Ghostkeeper, line 188: Okay, so this breaks cross infill and concentric infill. I don't really see a reason why this should be taken out of scope of this ticket. We need to transfer the `result_polygons` to the new format here.

src/infill.cpp
```diff
@@ -408,16 +529,16 @@ void Infill::addLineInfill(Polygons& result, const PointMatrix& rotation_matrix,
...
-        if (n < 0)
-        {
-            return -1;
-        }
-        if (n > 0)
-        {
-            return 1;
-        }
-        return 0;
+      coord_t n = (*(coord_t*)a) - (*(coord_t*)b);
```
Ghostkeeper, line 542: We indent with 4 spaces, not 2.

src/infill.h
```diff
@@ -105,13 +106,15 @@ class Infill
      * \param mesh The mesh for which to generate infill (should only be used for non-helper objects)
      * \param[in] cross_fill_provider The cross fractal subdivision decision functor
      */
-    void generate(Polygons& result_polygons, Polygons& result_lines, const SierpinskiFillProvider* cross_fill_provider = nullptr, const SliceMeshStorage* mesh = nullptr);
+    [[deprecated]] void generate(Polygons& result_polygons, Polygons& result_lines, const SierpinskiFillProvider* cross_fill_provider = nullptr, const SliceMeshStorage* mesh = nullptr); // Todo: remove if no longer used
```
Ghostkeeper, line 111: Sorry, the `[[deprecated]]` is a C++14 feature.
```suggestion
    void generate(Polygons& result_polygons, Polygons& result_lines, const SierpinskiFillProvider* cross_fill_provider = nullptr, const SliceMeshStorage* mesh = nullptr); // Todo: remove if no longer used
```

src/infill.h
```diff
@@ -105,13 +106,15 @@ class Infill
...
+    void generate(VariableWidthPaths& toolpaths, Polygons& result_lines, const SierpinskiFillProvider* cross_fill_provider = nullptr, const SliceMeshStorage* mesh = nullptr);
 
 private:
     /*!
      * Generate the infill pattern without the infill_multiplier functionality
      */
-    void _generate(Polygons& result_polygons, Polygons& result_lines, const SierpinskiFillProvider* cross_fill_pattern = nullptr, const SliceMeshStorage* mesh = nullptr);
+    [[deprecated]] void _generate(Polygons& result_polygons, Polygons& result_lines, const SierpinskiFillProvider* cross_fill_pattern = nullptr, const SliceMeshStorage* mesh = nullptr); // Todo: remove if no longer used
```
Ghostkeeper, line 118: The `[[deprecated]]` is not allowed by C++11.
```suggestion
    void _generate(Polygons& result_polygons, Polygons& result_lines, const SierpinskiFillProvider* cross_fill_pattern = nullptr, const SliceMeshStorage* mesh = nullptr); // Todo: remove if no longer used
```

src/utils/polygon.h
Ghostkeeper: You must have duplicated this function from Rburema's original code then, since there is no `translate` function that you remove in a different location. The functionality is already present anyway in `PolygonRef` so it should be removed from here or from there, or one should call the other.

Ghostkeeper (changes requested):

src/WallToolPaths.h
jellespijker: Good question.
Maybe we should discuss this together with @rburema offline.
I will leave it be for now

src/infill.cpp
jellespijker: Implemented

src/utils/polygon.h
jellespijker: I could be wrong but from what I understand is that they're two different classes, I believe that this one is a translation for the collection of PolygonRef(s).

I'm gonna change it to:
```
    void translate(const Point vec)
    {
        if (vec.X == 0 && vec.Y == 0)
        {
            return;
        }

        for (PolygonRef poly : *this)
        {
            poly.translate(vec);
        }
    }
```

src/utils/polygon.h
jellespijker: Tried out the following code changes. But I'm getting some strange results. Created a patch see code block below

```
Index: src/settings/types/Ratio.h
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- src/settings/types/Ratio.h	(revision 9420b87291d9817c29ccbbfe149a20ad6028a9c6)
+++ src/settings/types/Ratio.h	(date 1600289522015)
@@ -24,6 +24,10 @@
      */
     constexpr Ratio(double value) : value(value) {};
 
+    template <typename T>
+    constexpr Ratio(const T& numerator, const T& divisor)
+        : value(static_cast<double>(numerator / divisor)) {};
+
     /*
      * \brief Casts the Ratio instance to a double.
      */
@@ -85,4 +89,4 @@
 
 }
 
-#endif //RATIO_H
\ No newline at end of file
+#endif //RATIO_H
Index: src/utils/polygonUtils.cpp
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- src/utils/polygonUtils.cpp	(revision 9420b87291d9817c29ccbbfe149a20ad6028a9c6)
+++ src/utils/polygonUtils.cpp	(date 1600289060211)
@@ -1421,6 +1421,8 @@
     }
 
     const coord_t half_epsilon = (epsilon + 1) / 2;
+    const Ratio ratio(static_cast<coord_t>(1), epsilon);
+    const Ratio ratio_inv(epsilon, static_cast<coord_t>(1));
 
     // Shrink (making _near_ self-intersections into _actual_ self-intersecrtions), fix, grow back to original size.
     // Do this repeatedly with different offsets, so points that are close together do actually merge.
@@ -1428,9 +1430,10 @@
     for (const Point& translate_vec : translate_vecs)
     {
         thiss.translate(translate_vec);
-        thiss.scale(1, epsilon);
+
+        thiss.scale(ratio);
         ClipperLib::SimplifyPolygons(thiss.paths);
-        thiss.scale(epsilon, 1);
+        thiss.scale(ratio_inv);
         thiss.translate(-translate_vec);
     }
 
Index: src/utils/polygon.h
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- src/utils/polygon.h	(revision 9420b87291d9817c29ccbbfe149a20ad6028a9c6)
+++ src/utils/polygon.h	(date 1600289367278)
@@ -16,8 +16,10 @@
 
 #include <initializer_list>
 
-#include "IntPoint.h"
+#include "logoutput.h"
 #include "../settings/types/AngleDegrees.h" //For angles between vertices.
+#include "../settings/types/Ratio.h"
+#include "IntPoint.h"
 
 #define CHECK_POLY_ACCESS
 #ifdef CHECK_POLY_ACCESS
@@ -992,9 +994,9 @@
         }
     }
 
-    void scale(const coord_t& numerator, const coord_t divisor)
+    void scale(const Ratio& ratio)
     {
-        if (numerator == 1 && divisor == 1)
+        if (ratio == 1.)
         {
             return;
         }
@@ -1004,7 +1006,10 @@
         {
             for (Point& pt : thiss[p])
             {
-                pt = pt * numerator / divisor;
+                log("prev point x %d, y %d\n",pt.X, pt.Y);
+                log("Ratio: %d\n", ratio.value);
+                pt = pt * ratio;
+                log("point x %d, y %d\n",pt.X, pt.Y);
             }
         }
     }
Index: tests/settings/SettingsTest.cpp
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- tests/settings/SettingsTest.cpp	(revision 9420b87291d9817c29ccbbfe149a20ad6028a9c6)
+++ tests/settings/SettingsTest.cpp	(date 1600288761207)
@@ -150,6 +150,14 @@
     EXPECT_DOUBLE_EQ(Ratio(0.01618), settings.get<Ratio>("test_setting")) << "With ratios, the input is interpreted in percentages.";
 }
 
+TEST_F(SettingsTest, CalculateRatio)
+{
+    const coord_t num = 10;
+    const coord_t div = 400;
+    Ratio ratio(num, div);
+    EXPECT_DOUBLE_EQ(0.025, ratio) << "With ratios, the input is interpreted in percentages.";
+}
+
 TEST_F(SettingsTest, AddSettingDuration)
 {
     settings.add("test_setting", "1234.5678");
```

jellespijker: Fixed it with some additions to the Ratio with an addition of a numerator, divisor constructor and equality operators.

The ratio should be static_cast<double> before multiplication with the point

jellespijker:
Wall's well that ends well

jellespijker:
Only thing that is still subject to debate is if we should implement it connect infill polygons that it connects to the walls generated with libArachne

src/FffGcodeWriter.cpp
```diff
@@ -1634,35 +1634,36 @@ bool FffGcodeWriter::processSingleLayerInfill(const SliceDataStorage& storage, L
...
-                    if (infill_not_below_skin.size())
+                    if (!infill_not_below_skin.empty())
                     {
                         constexpr Polygons* perimeter_gaps = nullptr;
                         constexpr bool connected_zigzags = false;
                         constexpr bool use_endpieces = false;
                         constexpr bool skip_some_zags = false;
                         constexpr int zag_skip_count = 0;
+                        constexpr coord_t outline_offset = 0;
+                        const size_t min_wall_line_count = std::max(static_cast<size_t>(1), wall_line_count);
```
rburema, line 1656: For literals, you can use `1L`, `1UL` or `1LL` and the like.

rburema (commented):
It's quite interesting to see your 'maximalist' approach (change all the things!) to my own minimalist one (change as little as possible, fixing it in one line is '''better'''). I think I should go your way more often, as I can now see that way of thinking might stifle my growth as a developer as well.

src/infill.cpp
```diff
@@ -402,10 +455,32 @@ void Infill::generateCrossInfill(const SierpinskiFillProvider& cross_fill_provid
...
+{
+    Polygons result_polygons;
+    if (mirror_offset)
+    {
+        const coord_t offset = infill_multiplier % 2 == 0 ? outline_offset : outline_offset + infill_line_width / 2 ;
+        result_polygons.add(outer_contour.offset(offset));
+        generateCrossInfill(cross_fill_provider, result_polygons, result_lines);
+        for (PolygonRef poly_line : result_polygons)
+        {
+            for (unsigned int point_idx = 1; point_idx < poly_line.size(); point_idx++)
```
rburema, line 468: Use `size_t` for comparison with `size()`.

src/infill.cpp
```diff
@@ -42,6 +43,58 @@ static inline int computeScanSegmentIdx(int x, int line_width)
...
+        result_lines.add(generated_result_lines);
+    }
+    if (connect_polygons)
+    {
+        // remove too small polygons
+        coord_t snap_distance = infill_line_width * 2; // polygons with a span of max 1 * nozzle_size are too small
+        auto it = std::remove_if(result_polygons.begin(), result_polygons.end(), [snap_distance](PolygonRef poly) { return poly.shorterThan(snap_distance); });
+        result_polygons.erase(it, result_polygons.end());
+
+        PolygonConnector connector(infill_line_width, infill_line_width * 3 / 2);
```
rburema, line 92: Maybe add some brackets? This made me look up the operator precedence again ;-)

jellespijker:
> It's quite interesting to see your 'maximalist' approach (change all the things!) to my own minimalist one (change as little as possible, fixing it in one line is '''better'''). I think I should go your way more often, as I can now see that way of thinking might stifle my growth as a developer as well.

I need to touch things before I actually know how it behaves, and modernizing some "old-style" code forces me to really think what is happening there. That and I have the clang-tidy and CppChecks warnings highlighted in bright neon bars and that brings out the OCDer in me:

> "Must fix warning!"
 "No leave it alone Jelle!"
"But it is a warning must fixxxxxxxxxxxxxxxxx now"

Ha I sound like a deranged lunatic now.

But I think I can benefit from the opposite as well. In that (and many other) sense(s) I have a lot to learn from you as well. So if you ever see me getting overboard again, for something that can be fixed in a one-liner don't hesitate to smack me on the back of the head and point out the obvious.

src/BeadingStrategy/BeadingStrategyFactory.cpp
```diff
@@ -33,7 +33,7 @@ std::string to_string(StrategyType type)
     }
 }
 
-BeadingStrategy* BeadingStrategyFactory::makeStrategy(StrategyType type, coord_t preferred_bead_width_outer, coord_t preferred_bead_width_inner, coord_t preferred_transition_length, float transitioning_angle, const coord_t* min_bead_width, const coord_t* min_feature_size, int max_bead_count)
+BeadingStrategy* BeadingStrategyFactory::makeStrategy(StrategyType type, coord_t preferred_bead_width_outer, coord_t preferred_bead_width_inner, coord_t preferred_transition_length, float transitioning_angle, const std::unique_ptr<coord_t>& min_bead_width, const std::unique_ptr<coord_t>& min_feature_size,  const coord_t max_bead_count)
```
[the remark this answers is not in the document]
jellespijker: One at a time ;-)
Every time I work in the neighborhood of one of those raw pointers, I check out how deeply ingrained it is and then decide to fix

src/LayerPlan.cpp
```diff
@@ -1010,6 +1010,22 @@ void LayerPlan::addWall(const std::vector<ExtrusionJunction>& wall, int start_id
     }
 }
 
+void LayerPlan::addWall(const LineJunctions& wall, const GCodePathConfig& path_config,
```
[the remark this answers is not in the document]
jellespijker: Agreed, will change name to `addInfillWall(....)`

src/WallToolPaths.cpp
```diff
@@ -0,0 +1,127 @@
...
+    , min_feature_size(widening_beading_enabled ? new coord_t(settings.get<coord_t>("min_feature_size")) : nullptr)
+    , small_area_length(INT2MM(static_cast<double>(bead_width_0) / 2))
+    , transition_length(transition_length_multiplier * bead_width_0)
+    , toolpaths_generated(false)
+{
+}
+
+const VariableWidthPaths& WallToolPaths::generate()
+{
+    assert(("inset count should be more then 0", inset_count > 0));
```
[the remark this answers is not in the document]
jellespijker: The instance for the WallToolPaths is only created if the `wall_line_count > 0`
https://github.com/Ultimaker/CuraEngine/blob/7a44ae598e33408546337b7591fa278ce2d17739/src/infill.cpp#L53

src/FffGcodeWriter.cpp
jellespijker: You're absolutely correct, I don't know where my mind was at the time. Will change those `static_cast`s

src/infill.cpp
jellespijker: Now you're getting the hang of the 'maximalist' approach, since I didn't touch this ;-)
But agreed I will change it :-)

src/infill.cpp
jellespijker: Now you're getting the hang of the 'maximalist' approach, since I didn't touch this. But I agree with your assessment.

This will however result at least two extra `static_cast`s since the bracket operator of PolygonRef takes an `unsigned int` Which is wrong imo. But changing that to `size_t` will be a huge change in the code. The compiler will trip over al ot of ambiguous warnings
`error: use of overloaded operator '[]' is ambiguous (with operand types 'cura::PolygonRef' and 'unsigned int')` because we do it wrong in a lot of places. This could also result in a bigger footprint since unsigned int is 32bit and size_t is 64bit

I will compare against `size_t` and static_cast to `unsigned int`

PR #1324, opened 2020-09-11
CURA-7676 re-enable outer- and inner-wall widths

rburema:
We'd previously disabled the inner and outer wall width distinction to get variable wall widths set up. This PR again makes a distinction between the 1st wall and the inner walls.

Note that for 'Distributed', this makes little sense and does in fact have little effect beyond getting the total width right. However, the InwardDistributed (which makes use of the aforementioned base or actually intermediate class), _can_ handle the distinction. Per discussion with the original implementor it seems that this last _should actually be the default_ (or perhaps the _only_ user visible non-meta strategy, next to perhaps CentralDeviation).

For this reason, and also because I removed the Naive (would've been usable for debugging if it itself wasn't buggy) and LimitedDistributed (superseded by applying the separate Limited strategy to the Distributed one), there's also a PR **in the frontend** ~~which I will make/add here as soon as I'm done with this one.~~ See here for the frontend PR: https://github.com/Ultimaker/Cura/pull/8370

Lastly, any TODO's remaining in code are outside of the scope of this ticket.


src/BeadingStrategy/CenterDeviationBeadingStrategy.cpp
```diff
@@ -11,58 +11,79 @@ CenterDeviationBeadingStrategy::Beading CenterDeviationBeadingStrategy::compute(
...
+        for (coord_t bead_idx = 1; bead_idx < bead_count / 2; bead_idx++)
         {
-            ret.bead_widths.emplace_back(optimal_width);
-            ret.toolpath_locations.emplace_back(optimal_width * (bead_idx * 2 + 1) / 2);
+            ret.bead_widths.emplace_back(optimal_width_inner);
+            ret.toolpath_locations.emplace_back(optimal_width_outer + optimal_width_inner * ((bead_idx - 1) * 2 + 1) / 2);
         }
+
+        // middle (if any):
+        const coord_t optimal_width_middle = bead_count == 1 ? optimal_width_outer : optimal_width_inner;
```
konskarm, line 31: In this case, the bead_count will always be > 1, so the optimal_width_middle is always the optimal_width_inner.

rburema: Ah yes, made the exception for 1 bead count after I made this. It can be removed in that case.

src/BeadingStrategy/CenterDeviationBeadingStrategy.cpp
```diff
@@ -11,58 +11,79 @@ CenterDeviationBeadingStrategy::Beading CenterDeviationBeadingStrategy::compute(
...
 }
 
 coord_t CenterDeviationBeadingStrategy::getOptimalBeadCount(coord_t thickness) const
 {
-    coord_t naive_count = (thickness / 2 + optimal_width / 2) / optimal_width * 2;
-    coord_t optimal_thickness = naive_count * optimal_width;
-    coord_t overfill = optimal_thickness - thickness;
+    coord_t thickness_left = thickness;
+    coord_t naive_count = 0;
+    naive_count += std::min(2LL, ((thickness / 2 + optimal_width_outer / 2) / optimal_width_outer) * 2);
```
konskarm, line 79: Why not initialize the naive_count immediately with this value?

rburema: There was an if statement around it before I thought to use the std::min I think. I'm fixing these two right now, let me know if you find more.

src/BeadingStrategy/DistributedBeadingStrategy.cpp
```diff
@@ -30,17 +30,27 @@ DistributedBeadingStrategy::Beading DistributedBeadingStrategy::compute(coord_t
...
+    const coord_t optimal_thickness = this->getOptimalThickness(lower_bead_count);
+    return optimal_thickness + (optimal_thickness < 2 ? optimal_width_outer : optimal_width_inner) / 2;
 }
 
 coord_t DistributedBeadingStrategy::getOptimalBeadCount(coord_t thickness) const
 {
-    return (thickness + optimal_width / 2) / optimal_width;
+    coord_t thickness_left = thickness;
+    coord_t count = 0;
+    count += std::min(2LL, (thickness + optimal_width_outer / 2) / optimal_width_outer);
```
konskarm, line 47: Initialize count in one line

rburema: Wait I did it again? Another commit then

konskarm (approved):

PR #1325, opened 2020-09-12
Code reviewing. Fix out of bounds array and lost checks for segments ended with mesh vertice.

skarasov:
Code reviewing.
Fix out of bounds array and lost checks for segments ended with mesh vertice.

### DIFF
--- a/src/slicer.cpp
+++ b/src/slicer.cpp
@@ -788,20 +788,20 @@ void SlicerLayer::makePolygons(const Mesh* mesh)
     polygons.removeDegenerateVerts(); // remove verts connected to overlapping line segments
 }
 
-Slicer::Slicer(Mesh* i_mesh, const coord_t thickness, const size_t slice_layer_count, 
+Slicer::Slicer(Mesh* i_mesh, const coord_t thickness, const size_t slice_layer_count,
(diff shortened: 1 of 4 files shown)


Ghostkeeper:
I didn't see this pull request yet and fixed the out of bounds array myself too, from your comment. However this review fixes a number of different small issues as well, so it would still be good to merge.

Only... You re-added a few if statements that I also removed from a different comment of yours. Indeed these if-statements can never be true and they are no longer necessary, so they shouldn't be added back. Could you remove them?

skarasov:
No. In #1313 you move ">=z" from one if-statement to another. That's why `if (p2.z == z)` became never true.
My point was - checks for segment ending on mesh vertice shoud also be moved. Not to completely remove them.
This request fix it.

Ghostkeeper:
Indeed it appears that `endVertex` now always remains on nullptr. I was confused by the Github code review page because it had left precisely the contents of an if-statement out in its diff view, showing the wrong if-statement above the changed code. Sorry. We need to check each of the if-statements to see if `endVertex` is set correctly if ending exactly on a 3D vertex. This variable is used for stitching. This PR is expected to fix that.

Ghostkeeper (approved):

Ghostkeeper:
Thanks for paying attention!

PR #1326, opened 2020-09-15
Fix infill multiplier not connected lines

jellespijker:
When Connect Infill Lines was unchecked the multiplier didn't get the
correct number of offsets. A global flag was introduced mirror_offset
which ensures that the extra infill line offsets are generated in the
correct direction.
This is now only done for infill multiplier values higher then 3.

### DIFF
--- a/src/infill.cpp
+++ b/src/infill.cpp
@@ -201,14 +201,19 @@ void Infill::multiplyInfill(Polygons& result_polygons, Polygons& result_lines)
         }
     }
     result.add(first_offset);
-    Polygons reference_polygons = first_offset;
-    for (size_t infill_line = 1; infill_line < infill_multiplier / 2; infill_line++) // 2 because we are making lines on both sides at the same time
+    if (infill_multiplier > 3)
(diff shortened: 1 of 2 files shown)


src/infill.cpp
```diff
@@ -201,14 +201,19 @@ void Infill::multiplyInfill(Polygons& result_polygons, Polygons& result_lines)
         }
     }
     result.add(first_offset);
-    Polygons reference_polygons = first_offset;
-    for (size_t infill_line = 1; infill_line < infill_multiplier / 2; infill_line++) // 2 because we are making lines on both sides at the same time
+    if (infill_multiplier > 3)
```
nallath, line 204: I'd like to see some documentation why infill_multiplyer > 3 is important here.

nallath (changes requested):

src/infill.cpp
jellespijker: will add it

nallath (approved):

PR #1330, opened 2020-09-24
[CURA-7676] fix take outer wall width into account on only one side, not both

rburema:
Already done at some places, but not completely followed through.

Turns out that in order to make this work completely (or at least the CURA-7676 internal ticket, I also had to fix the plain Distributed Case to _do_ take into account the outer wall after all.

### DIFF
--- a/src/BeadingStrategy/DistributedBeadingStrategy.cpp
+++ b/src/BeadingStrategy/DistributedBeadingStrategy.cpp
@@ -19,16 +19,19 @@ DistributedBeadingStrategy::Beading DistributedBeadingStrategy::compute(coord_t 
     }
     else if (bead_count > 1)
     {
-        // Outer wall, always the same size:
-        ret.bead_widths.emplace_back(optimal_width_outer);
-        ret.toolpath_locations.emplace_back(optimal_width_outer / 2);
+        const float widen_by = static_cast<float>(thickness) / (optimal_width_outer + optimal_width_inner * (bead_count));
+
+        // Outer wall:
+        const coord_t distributed_width_outer = static_cast<coord_t>(optimal_width_outer * widen_by);
+        ret.bead_widths.emplace_back(distributed_width_outer);
+        ret.toolpath_locations.emplace_back(distributed_width_outer / 2);
 
         // Evenly distributed inner walls:
-        const coord_t distributed_width_inner = std::max(optimal_width_inner / 2, (thickness - optimal_width_outer) / (bead_count - 1));
+        const coord_t distributed_width_inner = (thickness - distributed_width_outer) / (bead_count - 1);
         for (coord_t bead_idx = 1; bead_idx < bead_count; bead_idx++)
         {
             ret.bead_widths.emplace_back(distributed_width_inner);
-            ret.toolpath_locations.emplace_back((optimal_width_outer - distributed_width_inner / 2) + distributed_width_inner * bead_idx);
+            ret.toolpath_locations.emplace_back((distributed_width_outer - distributed_width_inner / 2) + distributed_width_inner * bead_idx);
         }
 
         ret.left_over = 0; // There should be nothing left over, as we've distributed the remaining space.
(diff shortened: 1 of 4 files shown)


src/BeadingStrategy/InwardDistributedBeadingStrategy.cpp
```diff
@@ -11,37 +11,49 @@ InwardDistributedBeadingStrategy::Beading InwardDistributedBeadingStrategy::comp
...
         
         auto getWeight = [middle, this](coord_t bead_idx)
         {
-            float dev_from_middle = bead_idx - middle;
+            float dev_from_middle = (bead_idx - 1) - middle;
             return std::max(0.0f, 1.0f - one_over_distribution_radius_squared * dev_from_middle * dev_from_middle);
         };
         
-        for (coord_t bead_idx = 0; bead_idx < bead_count; bead_idx++)
+        for (coord_t bead_idx = 1; bead_idx < bead_count; bead_idx++)
```
jellespijker, line 42: Maybe don't use a raw loop here but a std::accumulate,
That would als me that float total_weight can be made const

src/BeadingStrategy/DistributedBeadingStrategy.cpp
```diff
@@ -11,14 +11,30 @@ DistributedBeadingStrategy::Beading DistributedBeadingStrategy::compute(coord_t
...
+        ret.bead_widths.emplace_back(thickness);
+        ret.toolpath_locations.emplace_back(thickness / 2);
+        ret.left_over = 0;
+    }
+    else if (bead_count > 1)
+    {
+        const float widen_by = static_cast<float>(thickness) / (optimal_width_outer + optimal_width_inner * (bead_count));
+
+        // Outer wall:
+        const coord_t distributed_width_outer = static_cast<coord_t>(optimal_width_outer * widen_by);
```
jellespijker, line 25: const auto ...= static_cast<coord_t>

jellespijker (commented):
Looks good. A couple of small suggestions

src/BeadingStrategy/InwardDistributedBeadingStrategy.cpp
rburema: Except that no range is involved, and I'm not (lazily?) making a range out of `bead_idx`s just so I can use `accumulate`. Notice also that we start from 1, which we could rectify with `std::next`, but at this point we're already well past the complexity of a simple, readable for loop ;-)

I had a look through the other algorithms... It's a shame they _all_ work on sequences, really.

src/BeadingStrategy/DistributedBeadingStrategy.cpp
rburema: Fixed in 2a9aa3a82

src/BeadingStrategy/InwardDistributedBeadingStrategy.cpp
```diff
@@ -11,37 +11,49 @@ InwardDistributedBeadingStrategy::Beading InwardDistributedBeadingStrategy::comp
...
+        ret.bead_widths.emplace_back(thickness);
+        ret.toolpath_locations.emplace_back(thickness / 2);
+        ret.left_over = 0;
+    }
+    else if (bead_count > 1)
+    {
+        const float widen_by = static_cast<float>(thickness) / (optimal_width_outer + optimal_width_inner * (bead_count));
+
+        // Outer wall:
+        const coord_t distributed_width_outer = static_cast<coord_t>(optimal_width_outer * widen_by);
```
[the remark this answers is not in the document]
rburema: Fixed in 2a9aa3a82

src/BeadingStrategy/InwardDistributedBeadingStrategy.cpp
```diff
@@ -11,37 +11,49 @@ InwardDistributedBeadingStrategy::Beading InwardDistributedBeadingStrategy::comp
...
+        const coord_t to_be_divided = thickness - (distributed_width_outer + distributed_width_inner * (bead_count - 1));
+
         float total_weight = 0;
-        float middle = static_cast<float>(bead_count - 1) / 2;
+        const float middle = static_cast<float>(bead_count - 2) / 2;
         
         auto getWeight = [middle, this](coord_t bead_idx)
         {
-            float dev_from_middle = bead_idx - middle;
+            float dev_from_middle = (bead_idx - 1) - middle;
```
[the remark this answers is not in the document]
rburema: 'I see a variable and I want to paint it const'
Fixed in 2a9aa3a82

rburema:
> Looks good. A couple of small suggestions

Thanks! For the really simple things like `const` or `auto` please feel free just to make those commits as part of the review. (OTOH also please feel completely free to keep reviewing like that for any reason and leave it up to the original coder to deal with, I don't mind.)

src/BeadingStrategy/InwardDistributedBeadingStrategy.cpp
jellespijker: It sure is I can't wait for Ranges
```c++
auto p = bead | drop(1) | accumulate;
```

rburema: Oh my gosh... It's like Haskell and pipes from the Unix command line are happily married :-D
Please tell me that this will be lazily evaluated as well?

PR #1331, opened 2020-09-24
[CURA#5713] Add check of option retract before nozzle first move.

skarasov:
CURA#5713
For bowden extruders its unwanted to retract before first move!

Depends on CURA#8441.


Ghostkeeper:
See the discussion at https://github.com/Ultimaker/Cura/pull/8441

PR #1333, opened 2020-09-26
Code cleaning.

skarasov:
Code cleaning.
CLang adaptation.
Ref iterator in for loop in poligon.cpp.
Removed excessive ";".
Fix ret val of operator-- in LayerIndex.h.

### DIFF
--- a/libs/rapidjson/rapidjson.h
+++ b/libs/rapidjson/rapidjson.h
@@ -1,5 +1,5 @@
 // Tencent is pleased to support the open source community by making RapidJSON available.
-// 
+//
 // Copyright (C) 2015 THL A29 Limited, a Tencent company, and Milo Yip. All rights reserved.
 //
 // Licensed under the MIT License (the "License"); you may not use this file except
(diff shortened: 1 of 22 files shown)


src/layerPart.cpp
```diff
@@ -95,7 +95,7 @@ void createLayerParts(SliceMeshStorage& mesh, Slicer* slicer)
 
     // OpenMP compatibility fix for GCC <= 8 and GCC >= 9
     // See https://www.gnu.org/software/gcc/gcc-9/porting_to.html, section "OpenMP data sharing"
-#if defined(__GNUC__) && __GNUC__ <= 8
```
Ghostkeeper, line 98: If `__GNUC__` is defined, it should never also have `__clang__` defined, right?

skarasov: Maybe, but Clang 10 with MinGW traps here.

libs/rapidjson/rapidjson.h
```diff
@@ -1,15 +1,15 @@
 // Tencent is pleased to support the open source community by making RapidJSON available.
-//
```
rburema, line 2: 'rapidjson' is in the libs folder. The libs folder contains a file saying:
```
cat .\libs\DO_NOT_CHANGE_THESE
These libraries are extern, and while they're utilized by this project,
they are in fact not actually developed on this repository.
If changes are desired for these projects, please refer to their official repositories.
```

rburema:
I merged the appliccable parts of this PR in https://github.com/Ultimaker/CuraEngine/commit/5e6f3d45a2e4cac11f0423e72edc9619392fccde ... github is not registering this as merging this PR, so I'm closing it manually.

PR #1335, opened 2020-09-30
CURA-7630 Skin walls with libArachne

rburema:


jellespijker:
I have no comments on the code changes.
But I'm interested why you choose not to generate the walls with the Infill class?
The whole purpose of the rewriting the infill class was to reduce code duplication and make it easier to maintain. Did the current implementation of the Infill class not serve your (skins) purpose?

rburema:
@jellespijker To be completely honest, I have a hard time seeing how the infill-class is conceptually appropriate here.

(I should probably have brought that up during the review though, sorry.)

rburema:
@jellespijker I could work to use the infill class instead today (and maybe the 7676 double walls bug, if it does indeed come from there), since I think chances are bigger that I finish either of those (problems I'm already familiar with) rather than the bug (a few scattered hours to basically start from scratch), before I leave until Tuesday at the end of this day.

jellespijker:
I'm inclined to say leave it.
The code looks clean and neat as it is, and I think rewriting it into a forced use of the class might be to much effort

rburema:
Yeah ok, leave it for now ... will look better for the sprint as well ;-)

I mean it works now, we can pick it up later!

jellespijker:
> I mean it works now, we can pick it up later!

Tell me lies, tell me sweet little lies
....

We got enough work as it is ;-)

PR #1342, opened 2020-10-16
CURA-7769: Fix Z-Seam being placed halfway through a straight wall

konskarm:
This was caused by the fact that simplify would sometimes not remove a point from a 3-point straight line due to an overflow error. This in-turn lead to Z-Seam being able to select that point.

Overflow error in simplify:
If three points (a, p, b) form an almost straight line, simplify will remove the point p from the path. In order to detect that, simplify checks wheter the distance of p from ab is 0. When calculating that distance, an overflow error could occur, which would lead simplify to believe that the points are not actually in the same line, thus leaving p inside the path.

This is now fixed by checking first whether the error prone variable 'dott' exceeds the numerical limit of type coord_t (long long), i.e. if dott is greater than the square root of LLONG_MAX. If that's the case, then the function getDist2FromLine will calculate the square root of the necessary distance ax in order to avoid squaring dott.

CURA-7769

### DIFF
--- a/src/utils/LinearAlg2D.cpp
+++ b/src/utils/LinearAlg2D.cpp
@@ -9,8 +9,6 @@
 
 #include "IntPoint.h" // dot
 
-#define SQRT_LLONG_MAX_FLOOR 3037000499
-
 namespace cura 
 {
 
@@ -193,6 +191,8 @@ bool LinearAlg2D::lineSegmentsCollide(const Point& a_from_transformed, const Poi
 
 coord_t LinearAlg2D::getDist2FromLine(const Point& p, const Point& a, const Point& b)
 {
+    constexpr coord_t SQRT_LLONG_MAX_FLOOR = 3037000499;
+
     //  x.......a------------b
     //  :
     //  :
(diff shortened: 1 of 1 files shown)


src/utils/LinearAlg2D.cpp
```diff
@@ -9,6 +9,8 @@
 
 #include "IntPoint.h" // dot
 
+#define SQRT_LLONG_MAX_FLOOR 3037000499
```
rburema, line 12: Use constexpr.

Unfortunately sqrt is not a constexpr enabled function, and will not even be in C++20, so we can't do `constexpr size_t sqrt_long_max_floor = std::sqrt(std::numeric_limits<coord_t>::max());`, but we _can_ do `constexpr size_t sqrt_long_max_floor = 3037000499;`.

src/utils/LinearAlg2D.cpp
```diff
@@ -200,13 +202,23 @@ coord_t LinearAlg2D::getDist2FromLine(const Point& p, const Point& a, const Poin
     const Point vap = p - a;
     const coord_t ab_size2 = vSize2(vab);
     const coord_t ap_size2 = vSize2(vap);
+    coord_t ax_size2, px_size2;
```
rburema, line 205: This is more a matter of taste, but we only need `px_size2` outside if the scopes below. Just write `const coord_t ax_size2 = ...` twice below, and you're not really duplicating code and get rid of something that might be confusing here.

Defining multiple variables like this is also frowned upon a bit, because you
 a) leave them undefined,
 b) can do things like `int a, b[10], c;`.

rburema (commented):
I have a couple of small remarks, but I can put them in myself, so I'll just leave these as comments instead of a 'request changes'.

PR #1344, opened 2020-10-21
Remove intermediate vertexes in really long line segments

konskarm:
Due to rounding errors when calculating the sqrt(ab_size2), simplify was still leaving vertexes
that existed in very long line segments. To get around that issue, getDistFromLine is introduced
that calculates that distance following a different formula than the dot product. This way, it
avoids calculating the squares and square roots of enormous lines that would otherwise generate
overflow and rounding errors.

This getDistFromLine is now used inside simplify when deciding to remove this vertex, thus making
simplify less sensitive to overflows.

CURA-7769

### DIFF
--- a/src/utils/linearAlg2D.h
+++ b/src/utils/linearAlg2D.h
@@ -353,6 +353,19 @@ public:
     static coord_t getDist2FromLine(const Point& p, const Point& a, const Point& b);
 
     /*!
+     * Get the distance from a point \p p to the line on which \p a and \p b lie.
+     * It calculates that distance via the area of the triangle (pab): dist=2*Area(pab)/size(ab).
+     * This approach is less overflow-prone but more computationally-expensive compared to
+     * calculating the distance via the dot product.
+     *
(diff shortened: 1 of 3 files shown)


src/utils/linearAlg2D.h
```diff
@@ -352,6 +352,19 @@ class LinearAlg2D
...
+    /*!
+     * Get the distance from a point \p p to the line on which \p a and \p b lie.
+     * It calculates that distance via the area of the triangle (pab): dist=2*Area(pab)/size(ab).
+     * This approach is less overflow-prone but more computationally-expensive compared to
+     * calculating the distance via the dot product.
+     *
+     * \param p The point to measure the distance from.
+     * \param a One of the points through which the line goes.
+     * \param b One of the points through which the line goes.
+     * \return The distance between the point and the line, squared.
```
konskarm, line 364:
```suggestion
     * \return The distance between the point and the line.
```

PR #1345, opened 2020-10-23
CURA-7794 simplify hotfixes

rburema:
Some hotfixes we _might_ want to put into beta.

The first five commits (the 6th one fixes an oversight in the 3rd one) should probably be looked at as separate potential fixes.

There _still_ might be line-segments that are really small generated for some of these though, this should be investigated at a later stage.

To indicate the precariousness and that we should be extra careful, I'll designate this as a draft PR.

### DIFF
--- a/src/skin.cpp
+++ b/src/skin.cpp
@@ -387,7 +387,10 @@ void SkinInfillAreaComputation::generateSkinInsets(SkinPart& skin_part)
         }
 
         // optimize polygons: remove unnecessary verts
-        skin_part.insets[inset_idx].simplify();
+        const ExtruderTrain& train_wall = mesh.settings.get<ExtruderTrain&>(inset_idx == 0 ? "wall_0_extruder_nr" : "wall_x_extruder_nr");
+        const coord_t maximum_resolution = train_wall.settings.get<coord_t>("meshfix_maximum_resolution");
+        const coord_t maximum_deviation = train_wall.settings.get<coord_t>("meshfix_maximum_deviation");
+        skin_part.insets[inset_idx].simplify(maximum_resolution, maximum_deviation);
         if (skin_part.insets[inset_idx].size() < 1)
         {
             skin_part.insets.pop_back();
--- a/src/utils/polygon.cpp
+++ b/src/utils/polygon.cpp
@@ -355,6 +355,11 @@ void PolygonRef::simplify(const coord_t smallest_line_segment_squared, const coo
         accumulated_area_removed += removed_area_next;
 
         const coord_t length2 = vSize2(current - previous);
+        if (length2 < 25)
+        {
+            // We're allowed to always delete segments of less than 5 micron.
+            continue;
+        }
 
         const coord_t area_removed_so_far = accumulated_area_removed + negative_area_closing; // close the shortcut area polygon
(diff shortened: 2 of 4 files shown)


src/skin.cpp
```diff
@@ -387,7 +387,10 @@ void SkinInfillAreaComputation::generateSkinInsets(SkinPart& skin_part)
         }
 
         // optimize polygons: remove unnecessary verts
-        skin_part.insets[inset_idx].simplify();
+        const ExtruderTrain& train_wall = mesh.settings.get<ExtruderTrain&>(inset_idx == 0 ? "wall_0_extruder_nr" : "wall_x_extruder_nr");
+        const coord_t maximum_resolution = train_wall.settings.get<coord_t>("meshfix_maximum_resolution");
+        const coord_t maximum_deviation = train_wall.settings.get<coord_t>("meshfix_maximum_deviation");
+        skin_part.insets[inset_idx].simplify(maximum_resolution, maximum_deviation);
```
nallath, line 393: I can't believe we didn't do this before...

rburema: I know, right?

rburema: It's not _quite_ as bad as it looks, as it gets simplified with the right parameters elsewhere .... but I'm pretty sure there are alterations to those polygons in the meantime.

rburema: There's also a few instances I didn't get to, because they do raft, support, infill etc, but we might want to have a look later at those as well.

src/utils/polygon.cpp
```diff
@@ -355,6 +355,11 @@ void PolygonRef::simplify(const coord_t smallest_line_segment_squared, const coo
         accumulated_area_removed += removed_area_next;
 
         const coord_t length2 = vSize2(current - previous);
+        if (length2 < 25)
```
nallath, line 358: I initially thought that this would cause the long lines to become shifted again. To my suprise, this doesn't happen...

nallath:
Benchy (Before) UM3
| Line length | Num segments |
| 1 | 193674 |
| 0.5 | 81584 |
| 0.1 | 11479 |
| 0.05 | 7833 |
| 0.01 | 4558 |
| 0.005 | 4282 |
| 0.001 | 872 |

Benchy (After) UM3 
| Line length | Num segments |
| 1 | 179749 |
| 0.5 | 59354 |
| 0.1 | 9394 |
| 0.05 | 6837 |
| 0.01 | 4346 |
| 0.005 | 4125 |
| 0.001 | 869 |

Benchy (Before) CR10
| Line length | Num segments |
| 1 | 127642 |
| 0.5 | 108347 |
| 0.1 | 36203 |
| 0.05 | 13643 |
| 0.01 | 880 |
| 0.005 | 515 |
| 0.001 | 60 |

Benchy (After) CR10
| Line length | Num segments |
| 1 | 125283 |
| 0.5 | 105999 |
| 0.1 | 33834 |
| 0.05 | 9078 |
| 0.01 | 840 |
| 0.005 | 511 |
| 0.001 | 60 |

src/utils/polygon.cpp
rburema: No, because previously this was happening _only_ for the long lines.

PR #1346, opened 2020-10-23
Minor nits in getDistFromLine

caseorum:
These edits save a few operations.

### DIFF
--- a/src/utils/LinearAlg2D.cpp
+++ b/src/utils/LinearAlg2D.cpp
@@ -245,7 +245,7 @@ coord_t LinearAlg2D::getDistFromLine(const Point& p, const Point& a, const Point
     {
         return ap_size;
     }
-    const coord_t area_times_two = abs((b.Y - a.Y) * p.X - (b.X - a.X) * p.Y + b.X * a.Y - b.Y * a.X);
+    const coord_t area_times_two = abs((p.X - b.X) * (p.Y - a.Y) + (a.X - p.X) * (p.Y - b.Y)); // Shoelace formula, factored
     const coord_t px_size = area_times_two / ab_size;
     return px_size;
 }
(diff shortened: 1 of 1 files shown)


src/utils/LinearAlg2D.cpp
```diff
@@ -240,12 +240,11 @@ coord_t LinearAlg2D::getDistFromLine(const Point& p, const Point& a, const Point
...
     const Point vap = p - a;
     const double ab_size = vSize(vab);
-    const coord_t ap_size = vSize(vap);
     if(ab_size == 0) //Line of 0 length. Assume it's a line perpendicular to the direction to p.
     {
-        return ap_size;
+        return vSize(vap);
     }
-    const coord_t area_times_two = abs((b.Y - a.Y) * p.X - (b.X - a.X) * p.Y + b.X * a.Y - b.Y * a.X);
+    const coord_t area_times_two = abs((p.X - b.X) * (p.Y - a.Y) + (a.X - p.X) * (p.Y - b.Y));
```
Ghostkeeper, line 247: For reviewer: These are indeed equivalent: https://www.wolframalpha.com/input/?i=%28%28n+-+b%29+*+p+-+%28m+-+a%29+*+q+%2B+m+*+b+-+n+*+a%29+-+%28%28p+-+m%29+*+%28q+-+b%29+%2B+%28a+-+p%29+*+%28q+-+n%29%29

caseorum: Thank you @Ghostkeeper. I forgot about Wolfram Alpha.

This is the shoelace formula, I just factored it more to minimize multiplications, since these are int values. It can be rearranged many ways, but I don't think there is any that is particularly clear or pretty. The compiler doesn't care either.

@konskarm Awesome solution to the overflow. There's no real reason to use the squared one anymore.

Ghostkeeper: Yeah I figured that, but due to the variable names here, and the quantity of variables, it is completely opaque what exactly is calculated so I figured I'd check.

konskarm:
Thanks for the contribution @caseorum. It is indeed the shoelace formula, maybe it makes it clearer that way.

caseorum:
On second thought, I think maybe only a comment could make the source clear.

konskarm (approved):

konskarm:
I thought it would be fine to change the base branch of this PR to our 4.8 branch, but on a second thought I'll just merge this in master and cherry-pick the commits.

PR #1347, opened 2020-10-23
CURA-7682: Print thin walls with libArachne

konskarm:
This PR changes the following:

- Renames the widening_beading_enabled variable of WallToolpaths to print_thin_walls
- Removed the processing of the outline gaps, since this is a process that is already covered by libArachne
- Fixes a small issue where a local variable was shadowing an input variable

Merge along with https://github.com/Ultimaker/Cura/pull/8601.

CURA-7682


src/FffGcodeWriter.cpp
```diff
@@ -2112,7 +2063,7 @@ void FffGcodeWriter::processSkinInsets(const SliceDataStorage& storage, LayerPla
     // add skin walls aka skin perimeters
     if (extruder_nr == skin_extruder_nr)
     {
-        bool added_something = false;
+        added_something = false;
```
Ghostkeeper, line 2066: Oof, this might fix some unintended travels to 0,0 if this happens to be the only thing printed on a layer. Good find.

Ghostkeeper (approved):

PR #1352, opened 2020-11-04
Fill gaps between walls fills gaps between skin and infill

konskarm:
This was caused due to a change done in https://github.com/Ultimaker/CuraEngine/pull/1284. Even though the change was fixing the original issue, small gaps were being generated between the infill and the skin walls due to the offsetted skin (as the offset can create discrepancies). 

image

These gaps are the ones being filled by the small gap filling technique, even if there is no infill wall.

image

Upon further inspection of the original issue https://github.com/Ultimaker/Cura/issues/6571, it turns out that the infill-skin overlap was intentional and it was dependend on the inner wall line width, which is irrelevant. https://github.com/Ultimaker/CuraEngine/blob/e4acd35cd832fe170457b2ff7e10f671f65feb9c/src/skin.cpp#L61 
The bigger the inner wall width, the more the overlap. After giving it more thought, it seems that there is no reason for the infill and skin to overlap, since it is causing problems. Also there is no documentation as to why it was implemented like this in the first place.

Removing the infill-skin overlap fixes both the original problem https://github.com/Ultimaker/Cura/issues/6571 and the new issue https://github.com/Ultimaker/Cura/issues/8626.

CURA-7804

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -1658,7 +1658,7 @@ bool FffGcodeWriter::processSingleLayerInfill(const SliceDataStorage& storage, L
 
             if (infill_below_skin.size())
             {
-                if (infill_below_skin.offset(-tiny_infill_offset).size())
+                if (!infill_below_skin.offset(-tiny_infill_offset).empty())
                 {
                     // there is infill below skin, is there also infill that isn't below skin?
                     Polygons infill_not_below_skin = in_outline.difference(infill_below_skin);
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -1658,10 +1658,7 @@ bool FffGcodeWriter::processSingleLayerInfill(const SliceDataStorage& storage, L
 
             if (infill_below_skin.size())
             {
-                // need to take skin/infill overlap that was added in SkinInfillAreaComputation::generateInfill() into account
-                const coord_t infill_skin_overlap = mesh.settings.get<coord_t>((part.insets.size() > 1) ? "wall_line_width_x" : "wall_line_width_0") / 2;
-
-                if (infill_below_skin.offset(-(infill_skin_overlap + tiny_infill_offset)).size())
+                if (infill_below_skin.offset(-tiny_infill_offset).size())
                 {
                     // there is infill below skin, is there also infill that isn't below skin?
                     Polygons infill_not_below_skin = in_outline.difference(infill_below_skin);
--- a/src/skin.cpp
+++ b/src/skin.cpp
@@ -431,7 +431,7 @@ void SkinInfillAreaComputation::generateInfill(SliceLayerPart& part, const Polyg
     const size_t wall_line_count = mesh.settings.get<size_t>("wall_line_count");
     const coord_t infill_line_distance = mesh.settings.get<coord_t>("infill_line_distance");
 
-    coord_t offset_from_inner_wall = -infill_skin_overlap;
+    coord_t offset_from_inner_wall = -innermost_wall_line_width / 2;
     if (wall_line_count > 0)
     { // calculate offset_from_inner_wall
         coord_t extra_perimeter_offset = 0; // to align concentric polygons across layers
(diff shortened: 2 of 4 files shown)


src/skin.cpp
```diff
@@ -431,7 +416,7 @@ void SkinInfillAreaComputation::generateInfill(SliceLayerPart& part, const Polyg
     const size_t wall_line_count = mesh.settings.get<size_t>("wall_line_count");
     const coord_t infill_line_distance = mesh.settings.get<coord_t>("infill_line_distance");
 
-    coord_t offset_from_inner_wall = -infill_skin_overlap;
+    coord_t offset_from_inner_wall = 0;
```
jellespijker, line 419: Good practice imo to make a lambda for such a calculation. This way the `offset_from_inner_wall` can be made const, and const is good. Newer compilers will be able to optimize the hell out of that. It will also be clear to readers of this code in the future what the actual purpose is of this piece of code.
 
```c++
auto get_offset = [](const Settings& settings)
{
    const auto wall_line_count = settings.get<size_t>("wall_line_count");
    const auto wall_line_count = settings.get<size_t>("wall_line_count");
    if (wall_line_count > 0)
    { // calculate offset_from_inner_wall
        coord_t extra_perimeter_offset = 0; // to align concentric polygons across layers
        const auto fill_pattern = settings.get<EFillMethod>("infill_pattern");
        if (fill_pattern == EFillMethod::CONCENTRIC
            && infill_line_distance > settings.get<coord_t>("infill_line_width") * 2)
        {
            if (settings.get<bool>("alternate_extra_perimeter")
                && layer_nr % 2 == 0)
            { // compensate shifts otherwise caused by alternating an extra perimeter
                extra_perimeter_offset = -innermost_wall_line_width;
            }
            if (layer_nr == 0)
            { // compensate for shift caused by walls being expanded by the initial line width multiplier
                const auto normal_wall_line_width_0 = settings.get<coord_t>("wall_line_width_0");
                const auto normal_wall_line_width_x = settings.get<coord_t>("wall_line_width_x");
                const coord_t normal_walls_width = normal_wall_line_width_0 + (wall_line_count - 1) * normal_wall_line_width_x;
                const coord_t walls_width = normal_walls_width * settings.get<Ratio>("initial_layer_line_width_factor");
                extra_perimeter_offset += walls_width - normal_walls_width;
                while (extra_perimeter_offset > 0)
                {
                    extra_perimeter_offset -= infill_line_distance;
                }
            }
        }
        offset_from_inner_wall += extra_perimeter_offset - innermost_wall_line_width / 2;
    }
};

const coord_t offset_from_inner_wall = get_offset(mesh.settings);
...
```

src/skin.cpp
```diff
@@ -460,11 +445,10 @@ void SkinInfillAreaComputation::generateInfill(SliceLayerPart& part, const Polyg
         offset_from_inner_wall += extra_perimeter_offset - innermost_wall_line_width / 2;
     }
     Polygons infill = part.insets.back().offset(offset_from_inner_wall);
-
-    infill = infill.difference(skin.offset(infill_skin_overlap));
+    infill = infill.difference(skin);
```
jellespijker, line 448: this will create extra copying

why not:

```c++
part.infill_area = part.insets.back().offset(offset_from_inner_wall).difference(skin);
part.infill_area.removeSmallAreas(MIN_AREA_SIZE);
```

jellespijker (changes requested):

src/FffGcodeWriter.cpp
```diff
@@ -1658,10 +1658,7 @@ bool FffGcodeWriter::processSingleLayerInfill(const SliceDataStorage& storage, L
 
             if (infill_below_skin.size())
             {
-                // need to take skin/infill overlap that was added in SkinInfillAreaComputation::generateInfill() into account
-                const coord_t infill_skin_overlap = mesh.settings.get<coord_t>((part.insets.size() > 1) ? "wall_line_width_x" : "wall_line_width_0") / 2;
-
-                if (infill_below_skin.offset(-(infill_skin_overlap + tiny_infill_offset)).size())
+                if (infill_below_skin.offset(-tiny_infill_offset).size())
```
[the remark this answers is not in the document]
jellespijker: make sure you invert the bool btw

```c++
if (!infill_below_skin.offset(-tiny_infill_offset).empty())
```

jellespijker (approved):

rburema:
... was this ever merged from 4.8 to the main branch?

Ghostkeeper:
Doesn't look like it, no.

PR #1355, opened 2020-11-09
Fix polyline chaining

BagelOrb:
TL;DR: This PR solves a bug with the print order of open polylines (infill and surface mode), and it enhances the performance of the engine.

The old code used to have an arbitrary constraint that polylines should always be split up into short straight segments per polyline. This was because the Cura 2014 code used to use the same order optimizer for both lines and polygons, and only disambiguated between the two by checking whether the length of the poly was 2 or more than 2. In the meantime the order optimizer for lines and polygons has been pulled apart, so there is no reason to have a constraint on polylines that they should always be single line segments.

I've now refactored some of the algorithms dealing with polylines, so that we are able to directly move polylines into the result. This saves on half of the RAM used to store the points of the polylines and it saves on computation time, because the separated segments don't need to be rejoined anymore at the order optimizer stage.

Surface mode @ PR:
image
time elapsed: 1 second

surface mode @ current master :
image
image
time elapsed: 1 second

[UM3_open_mesh_surface_mode_x4.3mf.rename.zip](https://github.com/Ultimaker/CuraEngine/files/5510210/UM3_open_mesh_surface_mode_x4.3mf.rename.zip)

PS:
This PR used to include a fix for the infill wipe distance, but that would require quite a bit more work to get it to actually work bugless, so I've removed that from this PR. After merging this PR it should be considered whether we would want to make a separate function `Polygons PolygonUtils::chainPolylines(const Polygons& polylines)`, so that `LayerPlan::addLinesByOptimizer` can apply the wipe distance always at the end of a polyline, without risking that the end of a polyline is the start of the next polyline.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -217,6 +217,7 @@ set(engine_SRCS # Except main.cpp.
 
 # List of tests. For each test there must be a file tests/${NAME}.cpp.
 set(engine_TEST
+    ClipperTest
     GCodeExportTest
     InfillTest
     LayerPlanTest
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -209,6 +209,7 @@ set(engine_SRCS # Except main.cpp.
     src/utils/PolygonProximityLinker.cpp
     src/utils/polygonUtils.cpp
     src/utils/polygon.cpp
+    src/utils/PolylineStitcher.cpp
     src/utils/ProximityPointLink.cpp
     src/utils/SVG.cpp
     src/utils/socket.cpp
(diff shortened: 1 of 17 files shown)


BagelOrb:
This PR includes a unit test for the clipper bug. The test is failing because there is a bug in Clipper - not in my code.

BagelOrb:
Here are some results on slice times on some random poor Armadillo using a release build.

This branch:
8.81, 8.38, 8.82, 8.91, 8.99, 8.58, 8.82

Master:
9.75, 9.56, 9.72, 9.40, 9.39, 9.23, 9.01

Stddev: 0.2, diff: 0.7

Improvement: 7.2%

Honestly the improvement is a lot higher than I expected!

BagelOrb:
This PR is superseded by https://github.com/Ultimaker/CuraEngine/pull/1362

PR #1357, opened 2020-11-13
CURA-7597 Unclosed contours and similar missing middle inner walls fix.

rburema:
This bug was introduced when the distinction between outer and inner skin walls was (re) introduced (to libArachne). Initially, the approach was to update each individual base strategy with the concept of outer and inner walls. At best, this was a really error prone and unmaintainable mess, and at worst outright impossible.

Instead, the situation of the base-strategies is reverted to basically as before (with, for them, only one wall-width), and a new meta-strategy is introduced, that takes care of the outer and inner wall distinction.

( Note that his is made to be compatible with any zero-width signalling walls that someone might want to add at some point, even though it does its stuff _after_ LimitedBeadingstrategy :-) )

The idea is as follows: Initially, for the parent (base) beading strategy, set the 'optimum' width as a weighted average defined by 'what the total thickness of the fill would be if max-bead width is always reached' (so take the situation where the actually requested total wall width is reached as the ground truth we base all the rest on). The base beading-strategy will thus produce the 'right' number of walls but the wrong bead-widths (and locations) -- because it assumes the walls are some average. This is then rectified by the newly introduced meta-strategy, which redistributes the bead-widths. It takes a look at the factor of the thickness the outer and inner walls would have occupied in the ideal situation and applies that to the current situation. This keeps a) the original beadings simple to understand and debug b) the meanings and behaviour of the original beadings intact while still c) preserving the widths of the outer and inner walls the user (settings) has/have input at the very least in the case of 'full' walls (and does a decent attempt at at least preserving the relative relation between those two values with a lesser thickness).

Fixes internal ticket CURA-7597 and it's duplicates CURA-7766 and CURA-7823 .

### DIFF
--- a/src/BeadingStrategy/BeadingStrategy.h
+++ b/src/BeadingStrategy/BeadingStrategy.h
@@ -38,11 +38,10 @@ public:
         coord_t left_over; //! The distance not covered by any bead; gap area.
     };
 
-    coord_t optimal_width_outer; //! Optimal bead width for the outermost wall(s)
-    coord_t optimal_width_inner; //! Optimal bead width for the inner wall(s)
+    coord_t optimal_width; //! Optimal bead width, nominal width off the walls in 'ideal' circumstances.
     
     std::string name;
-    
+
     coord_t default_transition_length; //! The length of the region to smoothly transfer between bead counts
 
     /*!
@@ -51,9 +50,8 @@ public:
      */
     AngleRadians transitioning_angle;
 
-    BeadingStrategy(coord_t optimal_width_outer, coord_t optimal_width_inner, coord_t default_transition_length, float transitioning_angle = M_PI / 3)
-    : optimal_width_outer(optimal_width_outer)
-    , optimal_width_inner(optimal_width_inner)
+    BeadingStrategy(coord_t optimal_width, coord_t default_transition_length, float transitioning_angle = M_PI / 3)
+    : optimal_width(optimal_width)
     , default_transition_length(default_transition_length)
     , transitioning_angle(transitioning_angle)
     {
--- a/src/BeadingStrategy/BeadingStrategyFactory.cpp
+++ b/src/BeadingStrategy/BeadingStrategyFactory.cpp
@@ -3,6 +3,13 @@
 
 #include "BeadingStrategyFactory.h"
 
+#include "InwardDistributedBeadingStrategy.h"
+#include "LimitedBeadingStrategy.h"
+#include "CenterDeviationBeadingStrategy.h"
+#include "WideningBeadingStrategy.h"
+#include "DistributedBeadingStrategy.h"
+#include "RedistributeBeadingStrategy.h"
+
 namespace cura
 {
 
@@ -35,12 +42,17 @@ std::string to_string(StrategyType type)
 
 BeadingStrategy* BeadingStrategyFactory::makeStrategy(StrategyType type, coord_t preferred_bead_width_outer, coord_t preferred_bead_width_inner, coord_t preferred_transition_length, float transitioning_angle, const std::unique_ptr<coord_t>& min_bead_width, const std::unique_ptr<coord_t>& min_feature_size,  const coord_t max_bead_count)
 {
+    const coord_t preferred_weighted_average_wall_width =
+        max_bead_count > 2 ?
+        ((preferred_bead_width_outer * 2) + preferred_bead_width_inner * (max_bead_count - 2)) / max_bead_count :
+        (max_bead_count <= 0 ? preferred_bead_width_inner : preferred_bead_width_outer);
+
     BeadingStrategy* ret = nullptr;
     switch (type)
     {
-        case StrategyType::Center:             ret = new CenterDeviationBeadingStrategy(preferred_bead_width_outer, preferred_bead_width_inner, transitioning_angle);       break;
-        case StrategyType::Distributed:        ret = new DistributedBeadingStrategy(preferred_bead_width_outer, preferred_bead_width_inner, preferred_transition_length, transitioning_angle);     break;
-        case StrategyType::InwardDistributed:  ret = new InwardDistributedBeadingStrategy(preferred_bead_width_outer, preferred_bead_width_inner, preferred_transition_length, transitioning_angle, inward_distributed_center_size);  break;
+        case StrategyType::Center:             ret = new CenterDeviationBeadingStrategy(preferred_weighted_average_wall_width, transitioning_angle);       break;
+        case StrategyType::Distributed:        ret = new DistributedBeadingStrategy(preferred_weighted_average_wall_width, preferred_transition_length, transitioning_angle);     break;
+        case StrategyType::InwardDistributed:  ret = new InwardDistributedBeadingStrategy(preferred_weighted_average_wall_width, preferred_transition_length, transitioning_angle, inward_distributed_center_size);  break;
         default:
             logError("Cannot make strategy!\n");
             return nullptr;
@@ -57,6 +69,8 @@ BeadingStrategy* BeadingStrategyFactory::makeStrategy(StrategyType type, coord_t
     {
         logDebug("Applying the Limited Beading meta-strategy with maximum bead count = %d.", max_bead_count);
         ret = new LimitedBeadingStrategy(max_bead_count, ret);
+        logDebug("Applying the Redistribute meta-strategy with outer-wall width = %d, inner-wall width = %d", preferred_bead_width_outer, preferred_bead_width_inner);
+        ret = new RedistributeBeadingStrategy(preferred_bead_width_outer, preferred_bead_width_inner, ret);
     }
     return ret;
 }
--- a/src/BeadingStrategy/CenterDeviationBeadingStrategy.cpp
+++ b/src/BeadingStrategy/CenterDeviationBeadingStrategy.cpp
@@ -6,97 +6,75 @@
 namespace cura
 {
 
-CenterDeviationBeadingStrategy::Beading CenterDeviationBeadingStrategy::compute(coord_t thickness, coord_t bead_count) const
-{
-    Beading ret;
-
-    ret.total_thickness = thickness;
-    if (bead_count == 1)
+    CenterDeviationBeadingStrategy::Beading CenterDeviationBeadingStrategy::compute(coord_t thickness, coord_t bead_count) const
     {
-        ret.bead_widths.emplace_back(thickness);
-        ret.toolpath_locations.emplace_back(thickness / 2);
-        ret.left_over = 0;
-    }
-    else if (bead_count > 1)
-    {
-        // first half minus middle:
-        ret.bead_widths.emplace_back(optimal_width_outer);
-        ret.toolpath_locations.emplace_back(optimal_width_outer / 2);
-        for (coord_t bead_idx = 1; bead_idx < bead_count / 2; bead_idx++)
-        {
-            ret.bead_widths.emplace_back(optimal_width_inner);
-            ret.toolpath_locations.emplace_back(optimal_width_outer + optimal_width_inner * ((bead_idx - 1) * 2 + 1) / 2);
-        }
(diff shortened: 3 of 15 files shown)


src/BeadingStrategy/CenterDeviationBeadingStrategy.cpp
```diff
@@ -6,97 +6,75 @@
...
+        { // when we move away from the strategy which replaces two beads by a single one in the middle
+            return (lower_bead_count + 1) * optimal_width - overfill_bound;
+        }
     }
-    else
+
+    coord_t CenterDeviationBeadingStrategy::getOptimalBeadCount(coord_t thickness) const
     {
-        return naive_count;
+        const coord_t naive_count = (thickness / 2 + optimal_width / 2) / optimal_width * 2;
```
jellespijker, line 63: These expressions can be simplied to:

```c++
const coord_t optimal_thickness = thickness + optimal_width;
const coord_t naive_count = optimal_thickness  / optimal_width;
const coord_t overfill = optimal_thickness;
```
Maybe not do the last asignment of the `overfill` ;-) but replace the it further on in the codde with `optimal_thickness`

src/BeadingStrategy/CenterDeviationBeadingStrategy.cpp
```diff
@@ -6,97 +6,75 @@
...
+        const coord_t overfill = optimal_thickness - thickness;
+        if (overfill > overfill_bound)
+        {
+            return naive_count - 1;
+        }
+        else if (-overfill > underfill_bound)
+        {
+            return naive_count + 1;
+        }
+        else
```
jellespijker, line 74: remove last `else` and let the function return `naive_count`

src/BeadingStrategy/InwardDistributedBeadingStrategy.cpp
```diff
@@ -6,61 +6,49 @@
...
+            };
+
+            for (coord_t bead_idx = 0; bead_idx < bead_count; bead_idx++)
+            {
+                total_weight += getWeight(bead_idx);
+            }
+
+            for (coord_t bead_idx = 0; bead_idx < bead_count; bead_idx++)
+            {
+                coord_t width = optimal_width + to_be_divided * getWeight(bead_idx) / total_weight;
```
jellespijker, line 33: can be `const`

src/BeadingStrategy/RedistributeBeadingStrategy.cpp
```diff
@@ -0,0 +1,72 @@
...
+#include <algorithm>
+
+namespace cura
+{
+    BeadingStrategy::Beading RedistributeBeadingStrategy::compute(coord_t thickness, coord_t bead_count) const
+    {
+        Beading ret = parent->compute(thickness, bead_count);
+
+        // Actual count and thickness as represented by extant walls. Don't count any potential zero-width 'signalling' walls.
+        bead_count = std::count_if(ret.bead_widths.begin(), ret.bead_widths.end(), [](const coord_t& width) { return width > 0; });
```
jellespijker, line 15: brackets on newline

jellespijker, line 15: reference has no added value for simple types `[](const coord_t width)`

src/BeadingStrategy/RedistributeBeadingStrategy.h
```diff
@@ -0,0 +1,61 @@
...
+        RedistributeBeadingStrategy(coord_t optimal_width_outer, coord_t optimal_width_inner, BeadingStrategy* parent) :
+            BeadingStrategy(parent->optimal_width, parent->default_transition_length, parent->transitioning_angle),
+            parent(parent),
+            optimal_width_outer(optimal_width_outer),
+            optimal_width_inner(optimal_width_inner)
+        {
+            name = "RedistributeBeadingStrategy";
+        }
+
+        virtual ~RedistributeBeadingStrategy() override {}
```
jellespijker, line 42: `virtual` is redundant since allready declared `override` also why not use `= default`

src/BeadingStrategy/RedistributeBeadingStrategy.h
```diff
@@ -0,0 +1,61 @@
...
+
+        Beading compute(coord_t thickness, coord_t bead_count) const override;
+
+        coord_t getOptimalThickness(coord_t bead_count) const override { return parent->getOptimalThickness(bead_count);  }
+        coord_t getTransitionThickness(coord_t lower_bead_count) const override { return parent->getTransitionThickness(lower_bead_count); }
+        coord_t getOptimalBeadCount(coord_t thickness) const override { return parent->getOptimalBeadCount(thickness); }
+        coord_t getTransitioningLength(coord_t lower_bead_count) const override { return parent->getTransitioningLength(lower_bead_count); }
+        float getTransitionAnchorPos(coord_t lower_bead_count) const override { return parent->getTransitionAnchorPos(lower_bead_count); }
+
+        virtual std::string toString() const override { return std::string("RedistributeBeadingStrategy+") + parent->toString(); }
```
jellespijker, line 52: `virtual` is redundant since allready declared `override`

src/BeadingStrategy/RedistributeBeadingStrategy.h
```diff
@@ -0,0 +1,61 @@
...
+
+namespace cura
+{
+    /*!
+     * Made a meta-beading-strat. that takes outer and inner wall widths into account. `
+     *
+     * The idea is as follows: Initially, for the parent (base) beading strategy, set the 'optimum' width as a weighted average defined by
+     * 'what the total thickness of the fill would be if max-bead width is always reached' (so take the situation where the actually
+     * requested total wall width is reached as the ground truth we base all the rest on). The base beading-strategy will thus produce the
+     * 'right' number of walls but the wrong bead-widths (and locations) -- becasue it assumes the walls are some average. This is then
```
jellespijker, line 17: typo: `because`

src/BeadingStrategy/RedistributeBeadingStrategy.cpp
```diff
@@ -0,0 +1,72 @@
+//Copyright (c) 2020 Ultimaker B.V.
+//CuraEngine is released under the terms of the AGPLv3 or higher.
+
+#include "RedistributeBeadingStrategy.h"
+
+#include <algorithm>
+
+namespace cura
+{
+    BeadingStrategy::Beading RedistributeBeadingStrategy::compute(coord_t thickness, coord_t bead_count) const
```
jellespijker, line 10: shouldn't `bead_count` be of `size_t`

src/BeadingStrategy/RedistributeBeadingStrategy.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+        // Early out when the only walls are outer, the parent can have been trusted to handle it.
+        if (bead_count < 3)
+        {
+            return ret;
+        }
+
+        // Calculate the factors with which to multiply the outer and inner walls:
+        // (There may be a way to do this with all-integer math, but the speedup won't probably be that much and this will be easier to maintain).
+        const float current_total_outer_walls_width = ret.bead_widths.front() + ret.bead_widths.back();
```
jellespijker, line 26: all of these float types introduces a lot of implicit and narrowing conversions, maybe reconsider the types

jellespijker:
I think it is a good approach I will do the component testing later today. A few remarks, nothing to serious

src/BeadingStrategy/BeadingStrategy.h
```diff
@@ -51,9 +50,8 @@ class BeadingStrategy
      */
     AngleRadians transitioning_angle;
 
-    BeadingStrategy(coord_t optimal_width_outer, coord_t optimal_width_inner, coord_t default_transition_length, float transitioning_angle = M_PI / 3)
-    : optimal_width_outer(optimal_width_outer)
-    , optimal_width_inner(optimal_width_inner)
+    BeadingStrategy(coord_t optimal_width, coord_t default_transition_length, float transitioning_angle = M_PI / 3)
```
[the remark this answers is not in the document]
rburema: One the one hand, the compiler will likely make it a constant anyway. On the other hand, I like the constexpr function idea, that is, if our build-system accepts it. I'll try it out.

src/BeadingStrategy/BeadingStrategyFactory.cpp
```diff
@@ -35,12 +42,17 @@ std::string to_string(StrategyType type)
 
 BeadingStrategy* BeadingStrategyFactory::makeStrategy(StrategyType type, coord_t preferred_bead_width_outer, coord_t preferred_bead_width_inner, coord_t preferred_transition_length, float transitioning_angle, const std::unique_ptr<coord_t>& min_bead_width, const std::unique_ptr<coord_t>& min_feature_size,  const coord_t max_bead_count)
 {
+    const coord_t preferred_weighted_average_wall_width =
```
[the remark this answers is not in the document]
rburema: Yeah, not sure what to call this.

Not sure about bar either, as that requires a comment to those that didn't run into physics much (still bitter about the CS paper by physicists that didn't explain that the dot above a variable meant 'derivative over time over this variable' ... I was stuck on that for _two days_ as I thought it meant 'this is a vector').

I think I'll rename it to your suggestion. It requires a comment anyway as well, I'll add that too.

src/BeadingStrategy/BeadingStrategyFactory.cpp
```diff
@@ -35,12 +42,17 @@ std::string to_string(StrategyType type)
...
+        max_bead_count > 2 ?
```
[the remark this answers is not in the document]
rburema: Lambda is a bit heavy-weight where I just need 2 implicit ifs, but make_strategy doesn't get called that much anyway. Will have another look.

src/BeadingStrategy/CenterDeviationBeadingStrategy.cpp
jellespijker: It makes me wonder a bit if the logic is sound, eeeing how these equations simplify.

src/BeadingStrategy/RedistributeBeadingStrategy.h
rburema: ... because I copied it from another beading strategy /shameful/ :-p

src/BeadingStrategy/RedistributeBeadingStrategy.h
rburema: I'll remove it, but I don't think override is completely redundant, as it makes it clear that this is not the base class (which also has to be virtual for these functions, but isn't 'override' there).

src/BeadingStrategy/RedistributeBeadingStrategy.cpp
rburema: First, `size_t` is a pointer size, it is often used for counts of things, since, as a pointer size, it's also the max of what fits into sequences like vector or list and the like. However, while there is a small number of beads in a vector in the end, the majority of bead count as used in the code is to do math with things that very much _should_ be a `coord_t`, so I'd like to keep it that way.

jellespijker: Yes but see commit below ;-)

Ghostkeeper: `coord_t` has the same bit size. We specialise `coord_t` to always be `int64_t`. `size_t` is usually the same as `intptr_t` which is 64 bits on all of our builds (since we only ship 64-bit Cura versions now).

rburema: True, but `coord_t` is signed, and `size_t` is unsigned. In calculations, it can be useful to go below 0, without having to cast, even if the input is a count.

Ghostkeeper: In those cases it would be better to use a plain integer then rather than `coord_t`.

PR #1358, opened 2020-11-14
Cura 8148 extruders switch retraction initialization

sisu70:
As discussed here: https://github.com/Ultimaker/Cura/issues/8148

Still working only locally to my machine (and outside any git) for what concerns the "fdmprinter.def.json" file, where I added two new properties that are referenced by the code of this pull request: attaching that version of the file here, for reference.

[fdmprinter.def.zip](https://github.com/Ultimaker/CuraEngine/files/5541120/fdmprinter.def.zip)


sisu70:
An automated test is failing, I guess because of the official and not coherent version of the "fdmprinter.def.json" file used for the test, but not part of this repository.
I just opened a parallel pull request for that change in Cura: https://github.com/Ultimaker/Cura/pull/8772.
I have no clue on how to link the two pull requests in order to fix the failing test.

Ghostkeeper:
It's not actually using fdmprinter from a different repository. The failing test is using a static set of settings defined here: https://github.com/Ultimaker/CuraEngine/blob/0852f3cd5e909c9ebb459befb499292460da97d4/tests/LayerPlanTest.cpp#L94-L179

To fix the test, you should add the setting there if it is necessary for constructing a layer plan or for adding a travel move to it. Unfortunately the layer plan touches quite a lot of settings, but it's best to keep the settings list fixed and linked to this test so that the test doesn't suddenly start failing due to changes in a different repository or something like that. So just add a line there that sets your new setting to `false` or something, the basic use case.

I haven't done a review yet (there are still quite a few others before you) but I've seen that you change the start extruder to the extruder with the greatest adhesion tendency. That will not be accepted, because existing start g-codes depend on the start extruder being the one with the lowest extruder number.

Thanks for working on this! It's not something that we at Ultimaker will otherwise be able to make time for, since it's not affecting our printers, but it's something that quite a few people have asked for so I hope it'll be useful.

Ghostkeeper:
Hmm, that didn't seem to work then.

sisu70:
> Hmm, that didn't seem to work then.

Please note that I had pushed also the commit with the merging with the master branch along with the original pull request and that commit is still marked as failing: could that explain ?

src/PrimeTower.cpp
```diff
@@ -26,6 +26,13 @@ PrimeTower::PrimeTower()
 : wipe_from_middle(false)
 {
     const Scene& scene = Application::getInstance().current_slice->scene;
+
+    {
+        EPlatformAdhesion adhesion_type = scene.current_mesh_group->settings.get<EPlatformAdhesion>("adhesion_type");
+
+        multiple_extruders_on_first_layer = scene.settings.get<bool>("machine_extruders_share_nozzle") && ((adhesion_type == EPlatformAdhesion::RAFT) || (adhesion_type == EPlatformAdhesion::NONE));
```
Ghostkeeper, line 33: This line is the reason why the tests are failing. You're requesting this setting from the scene, i.e. global settings. The LayerPlan test only sets the mesh group settings (settings for a group of meshes that get sliced together from bottom to top, i.e. the entire scene unless using one-at-a-time mode).

```suggestion
        multiple_extruders_on_first_layer = scene.current_mesh_group->settings.get<bool>("machine_extruders_share_nozzle") && ((adhesion_type == EPlatformAdhesion::RAFT) || (adhesion_type == EPlatformAdhesion::NONE));
```

sisu70:
It looks fine now, thank you !

> I haven't done a review yet (there are still quite a few others before you) but I've seen that you change the start extruder to the extruder with the greatest adhesion tendency. That will not be accepted, because existing start g-codes depend on the start extruder being the one with the lowest extruder number.

This is a bit puzzling me: the start extruder was already driven by some logic linked to parameters (e.g. adhesion_type + adhesion_extruder_nr), so I doubt that start g-codes could and can make any 'blind' assumption. Please note that the start g-code that I put in place for my shared-nozzle multiple-extruder printer basically ignores the extruder that it founds selected at the beginning and absolutely needs to potentially change the selected extruder in order to initialize/prime all the extruders: for this reason I also needed to add a re-selection of the start extruder at the end of machine start gcode execution.
Anyway I consider the current implementation of extruders initialization in charge of machine start gcode to be a suboptimal solution, as for example it does not enable the initialization/priming with extruder/material-specific temperatures.

> Thanks for working on this! It's not something that we at Ultimaker will otherwise be able to make time for, since it's not affecting our printers, but it's something that quite a few people have asked for so I hope it'll be useful.

Welcome, I appreciate you openess and help for the integration.

Ghostkeeper:
This has been sitting on the shelf for too long :( I'm trying to get it into the 4.9 release though.

src/PrimeTower.cpp
```diff
@@ -26,6 +26,13 @@ PrimeTower::PrimeTower()
 : wipe_from_middle(false)
 {
     const Scene& scene = Application::getInstance().current_slice->scene;
+
+    {
+        EPlatformAdhesion adhesion_type = scene.current_mesh_group->settings.get<EPlatformAdhesion>("adhesion_type");
+
+        multiple_extruders_on_first_layer = scene.current_mesh_group->settings.get<bool>("machine_extruders_share_nozzle") && ((adhesion_type == EPlatformAdhesion::RAFT) || (adhesion_type == EPlatformAdhesion::NONE));
```
Ghostkeeper, line 33: So basically if I understood correctly, this change is necessary because:

We normally only print the first layer prime tower with a single extruder, the adhesion extruder. This is done to improve stability of the prime tower. However that requires that all extruders are primed by the start g-code or the firmware at the start of the print. That assumption is dropped when extruders share a nozzle. Thus it is necessary for the prime tower to prime all extruders that are used on every layer, including the first one.

It would be good to document this, because it's pretty hard to figure this out reading the code later, especially for someone not involved with the [original issue](https://github.com/Ultimaker/Cura/issues/8148).

I think the fact that multiple extruders on the first layer are not allowed is a property of the skirt and brim, which both have extra skirt lines around the adhesion with the other extruders. So it'd be best to flip the check for the adhesion types around (with the same final result, until we add adhesion types).

src/PrimeTower.cpp
```diff
@@ -40,9 +47,32 @@ PrimeTower::PrimeTower()
...
         const Ratio adhesion_a = scene_pointer->extruders[extruder_nr_a].settings.get<Ratio>("material_adhesion_tendency");
         const Ratio adhesion_b = scene_pointer->extruders[extruder_nr_b].settings.get<Ratio>("material_adhesion_tendency");
-        return adhesion_a < adhesion_b;
+
+        EPlatformAdhesion adhesion_type = scene_pointer->current_mesh_group->settings.get<EPlatformAdhesion>("adhesion_type");
+        const unsigned int adhesion_extr = scene_pointer->current_mesh_group->settings.get<ExtruderTrain&>("adhesion_extruder_nr").extruder_nr;
+
+        if (adhesion_type == EPlatformAdhesion::RAFT)
+        {
+            retval = (extruder_nr_a == adhesion_extr);
```
Ghostkeeper, line 60: Instead of this `retval`, it's more semantic to just use `return`.
```suggestion
            return extruder_nr_a == adhesion_extr;
```

Ghostkeeper, line 60: So basically the idea of this is that if using the raft, the raft extruder is used as outer shell (would also be good to document that; it's pretty hard to read that from this!)
This presents a problem when using a raft on >2 extruder prints. The rest of the extruders are then no longer sorted by adhesion tendency.

I'd guess that you would need a check like this instead:

```
if(adhesion_type == EPlatformAdhesion::RAFT && extruder_nr_a == adhesion_extr)
{
    return true;
}
```

That way other extruders are still properly ordered among each other in the `else` cases.

I also think this reordering is somewhat out of scope. Is there a reason this needs to change to support multiple-extruders-single-nozzle printers?

src/PrimeTower.cpp
```diff
@@ -40,9 +47,32 @@ PrimeTower::PrimeTower()
...
+        {
+            retval = (adhesion_a > adhesion_b);
+        }
+        else if (adhesion_type != EPlatformAdhesion::NONE)
+        {
+            retval = (extruder_nr_a == adhesion_extr);
+        }
+        else
+        {
+            retval = (extruder_nr_a < extruder_nr_b);
```
Ghostkeeper, line 72: This would be unnecessary since we're doing a stable sort.

Ghostkeeper (changes requested):

src/PrimeTower.cpp
sisu70: Yes, in fact I did not enable multiple extruders on the base layer of the prime tower in case the adhesion method is skirt or brim. I wanted to address only the problematic cases and not to change anything in the others.
I guess that your suggestion for 'documenting it' targets adding some comments in the code: I tried to use very informative variable names and kept the code readable, but I agree that adding some rationale could help in the future.
Unfortunately I'm still not used to this code review system, so ... I will do my best.

src/PrimeTower.cpp
sisu70: Well ... this is a very, very long story ! E.g. https://softwareengineering.stackexchange.com/questions/118703/where-did-the-notion-of-one-return-only-come-from.
I started coding (when I was 10 or so) in the era of 'structured programming' and in my code I'm still following those principles (unless I'm using C++'s exceptions, but doesn't seem to be the case in CuraEngine).
Anyway, I'm not fighting for a line of code out of thousands, so ... ok :-)

sisu70: My idea was to still target printing prime tower's external wall with the same extruder as it's single-extruder base, in case that single-extruder base is still there. After I needed to drop the single-extruder base when the base is at the same level as the first layer of the model (but only in case of shared heater/nozzle), the raft is the only remaining opportunity.
But you are most probably right that my code is messing the print order up, apart the first extruder, when more than 2 extruders are used. This for sure needs to be fixed.
About the need for reordering, as said it is simply an opportunity for ensuring the best adhesion of the prime tower external (solid) wall to its base and should be specific
- to the shared heater/nozzle case
- to the first extruder in the sequence, leaving the order of the others unchanged.

src/PrimeTower.cpp
sisu70: Yes, agreed.

sisu70:
So, what's the best way forward ? Shall I implement the requested/suggested changes in my fork/branch and the push them there ?

src/PrimeTower.cpp
Ghostkeeper: My preference would go towards not changing the order at all. If this actually changes the order, it would involve a bunch of print reliability tests for us. In general, it is good style to keep the scope of pull requests small. If we need to improve the order of the prime tower shells, let's do that separately.

Ghostkeeper:
Yeah, if you implement the suggested changes (at least as far as you agree) then they'll pop up in this pull request and we'll review them again. This PR is currently in our sprint so it'll keep in our attention.

Don't worry too much about the merge conflict. Those are typically pretty easy to fix for us anyway.

sisu70:
I just pushed two additional commits in my fork, addressing the two discussions above, but I'm not sure whether that's enough or  not for updating the pull request as well.
The first commit has no functional impact and is expected to address your remarks about the clarity of the code.
The second commit is a fixed version of the extruders/adhesion ordering: let's say it's my updated proposal, but I agree that it could be left out from the merge in the main fork/branch in case it is still considered unsafe.

src/PrimeTower.cpp
sisu70: Please have a look at the updated version.
Take special care of the "adhesion_a > adhesion_b" (that I did not modify compared to my earlier push), that is opposite compared to the "adhesion_a < adhesion_b" that I originally started from and that I still see in the main fork/branch.

Ghostkeeper:
Sorry, this currently introduces a bug that the `{initial_extruder_nr}` as predicted in the front-end is no longer correct if the adhesion type is "None". The `ExtruderManager.getInitialExtruderNr` function would need to be updated accordingly.

While I agree that it would be better in case of ties of adhesion tendency to use the adhesion extruder, it would still be better to remove that change from this PR and put it in a separate PR. Especially if the ordering with adhesion tendency is reversed (due to that `<` reversal), I don't think I can merge it like this without getting the process engineers of Ultimaker involved to test the stability of the prime tower.

It's also useful to know that a researcher at Ultimaker is currently trying to get the bed adhesion to work per extruder, so you could get the brim/raft to use a different extruder for different parts of the print. That would also remove the usefulness of this change.

src/PrimeTower.cpp
Ghostkeeper: Yeah I saw that. I was wondering if it was now stored in reverse order and interpreted the other way around somehow, and was planning to test the order thoroughly.

sisu70: Well, I do see a conflict between the comment and the code in the original version:

//Sort from high adhesion to low adhesion.
...
return (adhesion_a < adhesion_b);

Maybe the comment is reversed, but to me it makes sense to build the tower from the outer shell towards the inner and with decreasing adhesion tendency, as the inner parts are anyway confined by the outer ones and therefore more robust aginst poor adhesion (also because apart the outer shell, some or many layers could be fully missing for inner shells and in that case the adhesion of the finally printer inner upper layer is 0 by construction, as there is nothing below to adhere to).
To me it looks that with the code currently in the main branch, the outer shall of the prime tower is printed with the extruder/material with the lowest (worst, I guess) adhesion, which doesn't sound a great idea. Please note that in my setup I found most of the material configured with the same adhesion tendency, so this re-ordering, correct or wrong, was never really performed.

sisu70:
Ok, as said I agree not to pull this part, in order to have more time for a closer look. Shall I do anything in my trunk/branch or you can cherry-pick while pulling ?

Ghostkeeper:
I've cherry-picked the ones not related to the order of the prime tower and the start extruder. And also omitted a fix you made to the initialisation of two variables for MSVC that we also made in the meanwhile (so that resolved the merge conflict). With that, the feature seems to be working correctly. Several key points of my attention:

* With the default settings, there is no significant change in g-code output.
* When the shared nozzle is enabled, each extruder gets properly primed (so in the prime tower both are part of the first layer).
* Testing with raft or without, since this is a bit of a special case.
* The prime tower is still ordered correctly in all cases. Indeed, PLA goes on the outside due to better adhesion tendency than ABS, so the current `<` is correct then. This is unchanged from 4.8.
* The initial retraction setting assumes the nozzle starts from a retracted position.

I've made slight changes to the front-end setting for Shared Nozzle Initial Retraction: Give it capitalisation, a minimum value of 0mm and disable the setting if Shared Nozzle is disabled.

I think for 4.9 we'll just leave this as-is and give some time to test it in the field. If we don't get any complains I'd like to have this and the Shared Heater back in the Machine Settings dialogue.

Ghostkeeper:
This is merged via a cherry-pick, so Github won't pick up on it. See the merge commit here: https://github.com/Ultimaker/CuraEngine/commit/e508c6044845213489230af68505f161092567ed

sisu70:
I just realized on Cura 4.9.1 that my https://github.com/Ultimaker/CuraEngine/pull/1358/commits/bb53e8b5ec511bbe95ac416dc07892c16194b3a3 commit, for selecting once again the start extruder just after the execution of the machine start gcode, had not been pulled, so a model that I was expecting to be printed with my second extruder only (as shown in Cura's preview pane) has been printed with the first one instead.
The reasons why a second selection is in my opinion needed are:
- my machine start gcode needs to work on multiple extruders, so it needs to select new ones during its execution
- in my understanding there is no gcode's code for pushing/popping the current extruder, so the script itself cannot restore the extruder happened to be the selected one at its start time
- the extruder selection commanded earlier than machine start gcode execution is still needed for compatibility with other types of machine start gcodes

Ghostkeeper:
> in my understanding there is no gcode's code for pushing/popping the current extruder, so the script itself cannot restore the extruder happened to be the selected one at its start time

If the printer requires it, you can write `T{initial_extruder_nr}` in the start g-code to make it go to the start extruder. The reason why I don't expect that to happen at the end of the start g-code automatically is because some printers need things to happen after switching to the start extruder (e.g. a prime line) so that would take away power from them.

sisu70:
> If the printer requires it, you can write `T{initial_extruder_nr}` in the start g-code to make it go to the start extruder.

I was not aware and that actually prevents/solves the residual issue that I had. Thank you !

PR #1360, opened 2020-11-18
CURA-7751: Simplify arachne toolpaths

konskarm:
Simplify the toolpaths created by libArachne to reduce small segments.

Things to notice:

1. The simplification was split from the polygon simplification to avoid having a "one-size-fits-all" simplify
2. This code borrows the simplify implementation of polygons, adapted to work with `ExtrusionLines`
3. `ExtrusionLines` can either be open polylines (first and last junctions have different coordinates) or closed polylines (first and last junctions have the same -or almost the same- coordinates)
4. The simplification of `ExtrusionLines` differs from the simplification of polygons in the following:
   **i.** When simplifying them we should always bypass the first and last junctions, contrary to the polygon simplification (because of 3.)
   **ii.** When a set of `current` and `previous` junctions are intended to be replaced by an `intersection_point`, the `intersection point` will take the width of the `current` junction
   **iii.** When removing a junction that is in the middle of a long linear segment, the entire line will take as a width the weighted average of the two segments. In this case, an error is introduced in the extrusion area covered by these segments, as one of them will lose some area while the other one will gain some area. If the sum of the errors introduced by removing the intermediate junction is smaller than the maximum allowed, then that junction can be safely removed and the entire line will change its width. 

image

Merge along with https://github.com/Ultimaker/Cura/pull/8804.

CURA-7751


src/utils/ExtrusionLine.cpp
```diff
@@ -27,4 +28,143 @@ void ExtrusionLine::appendJunctionsTo(LineJunctions& result) const
...
+        {
+            // We're allowed to always delete segments of less than 5 micron. The width in this case doesn't matter that
+            // much.
+            continue;
+        }
+
+        const coord_t area_removed_so_far = accumulated_area_removed + negative_area_closing; // close the shortcut area polygon
+        const coord_t base_length_2 = vSize2(next - previous);
+
+        if (base_length_2 == 0) //Two line segments form a line back and forth with no area.
```
konskarm, line 102: I am not sure if this case can occur in libArachne toolpaths at all, since libArachne should generally produce separate ExtrusionLines for cases like this. Nevertheless, I kept it in just to be sure.

rburema: Keep in mind that eventually, your method might be called by other things than just libArachne, or Arachne might change. So keeping these things is more than sensible.

rburema (changes requested):
Good explanation! I like most of what's in here, but I do have some serious doubts about how the width is handled right now, so I have some change-requests.

src/utils/ExtrusionLine.cpp
```diff
@@ -27,4 +28,143 @@ void ExtrusionLine::appendJunctionsTo(LineJunctions& result) const
...
+        {
+            const coord_t next_length2 = vSize2(current - next);
+            if (next_length2 > smallest_line_segment_squared)
+            {
+                // Special case; The next line is long. If we were to remove this, it could happen that we get quite noticeable artifacts.
+                // We should instead move this point to a location where both edges are kept and then remove the previous point that we wanted to keep.
+                // By taking the intersection of these two lines, we get a point that preserves the direction (so it makes the corner a bit more pointy).
+                // We just need to be sure that the intersection point does not introduce an artifact itself.
+                Point intersection_point;
+                bool has_intersection = LinearAlg2D::lineLineIntersection(previous_previous.p, previous.p, current.p, next.p, intersection_point);
```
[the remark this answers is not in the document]
konskarm: Yup, that was also my thought on that. The point at index 1 will never actually be removed. But I thought it would be more confusing if I would start at index 2. Nevertheless, it will save one iteration for every toolpath, so I will start it at index 2 instead and I will explain why this happens in the function.

src/utils/ExtrusionLine.cpp
```diff
@@ -27,4 +28,143 @@ void ExtrusionLine::appendJunctionsTo(LineJunctions& result) const
...
+            else
+            {
+                continue; //Remove the vertex.
+            }
+        }
+        //Don't remove the vertex.
+        accumulated_area_removed = removed_area_next; // so that in the next iteration it's the area between the origin, [previous] and [current]
+        previous_previous = previous;
+        previous = current; //Note that "previous" is only updated if we don't remove the vertex.
+        new_junctions.push_back(current);
```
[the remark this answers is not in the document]
konskarm: Good observation. I was a bit hesitant of going down the "average" route but maybe a _weighted_ average will do the trick. I'll try it out.

src/utils/ExtrusionLine.cpp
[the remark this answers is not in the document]
konskarm: Wait, come to think of it, point at index 1 can still be removed completely from the path if its distance from the previous is less than 5 microns. So we should check it.

src/WallToolPaths.h
```diff
@@ -76,6 +76,13 @@ class WallToolPaths
      */
     static bool removeEmptyToolPaths(VariableWidthPaths& toolpaths);
 
+    /*!
+     * Simplifies The generated toolpaths
```
nallath, line 80: A little bit more information about this would be nice. I understand that calling this function will simplify the generated toolpaths because that's the name of the function ;) 

Something like: "This will simplify the toolpath by calling the simplify on every line in the toolpath using the provideded settings" Would make more sense to me.

src/utils/ExtrusionLine.cpp
```diff
@@ -27,4 +28,180 @@ void ExtrusionLine::appendJunctionsTo(LineJunctions& result) const
...
+     * ---------------                                         **************
+     * |             |                                         ------------------------------------------
+     * |             |--------------------------|  B removed   |            |***************************|
+     * |             |                          |  --------->  |            |                           |
+     * |             |--------------------------|              |            |***************************|
+     * |             |                                         ------------------------------------------
+     * ---------------             ^                           **************
+     *       ^                    C.w                                             ^
+     *      B.w                                                     new_width = weighted_average_width
+     * */
```
nallath, line 188: Perhaps also mention what the ****** in the "drawing" mean?

src/utils/ExtrusionLine.cpp
```diff
@@ -27,4 +28,180 @@ void ExtrusionLine::appendJunctionsTo(LineJunctions& result) const
...
+     From this area we compute the height of the representative triangle using
+     the standard formula for a triangle area: A = .5*b*h
+     */
+    coord_t accumulated_area_removed = previous.p.X * current.p.Y - previous.p.Y * current.p.X; // Twice the Shoelace formula for area of polygon per line segment.
+
+    for (size_t point_idx = 1; point_idx < junctions.size() - 1; point_idx++)
+    {
+        current = junctions.at(point_idx);
+
+        // don't spill over if the [next] vertex will then be equal to [previous]
```
nallath, line 77:
```suggestion
        // Don't spill over if the [next] vertex will then be equal to [previous]
```

src/utils/ExtrusionLine.cpp
```diff
@@ -27,4 +28,180 @@ void ExtrusionLine::appendJunctionsTo(LineJunctions& result) const
...
+
+        const coord_t length2 = vSize2(current - previous);
+        if (length2 < 25)
+        {
+            // We're allowed to always delete segments of less than 5 micron. The width in this case doesn't matter that
+            // much.
+            continue;
+        }
+
+        const coord_t area_removed_so_far = accumulated_area_removed + negative_area_closing; // close the shortcut area polygon
```
nallath, line 98:
```suggestion
        const coord_t area_removed_so_far = accumulated_area_removed + negative_area_closing; // Close the shortcut area polygon
```

src/utils/ExtrusionLine.cpp
```diff
@@ -27,4 +28,180 @@ void ExtrusionLine::appendJunctionsTo(LineJunctions& result) const
...
+        if (point_idx + 1 == junctions.size() && new_junctions.size() > 1)
+        {
+            next = new_junctions[0]; //Spill over to new polygon for checking removed area.
+        }
+        else
+        {
+            next = junctions.at(point_idx + 1);
+        }
+        const coord_t removed_area_next = current.p.X * next.p.Y - current.p.Y * next.p.X; // Twice the Shoelace formula for area of polygon per line segment.
+        const coord_t negative_area_closing = next.p.X * previous.p.Y - next.p.Y * previous.p.X; // area between the origin and the short-cutting segment
```
nallath, line 87:
```suggestion
        const coord_t negative_area_closing = next.p.X * previous.p.Y - next.p.Y * previous.p.X; // Area between the origin and the short-cutting segment
```

src/utils/ExtrusionLine.cpp
```diff
@@ -27,4 +28,180 @@ void ExtrusionLine::appendJunctionsTo(LineJunctions& result) const
...
+        //A = 1/2 * b * h     [triangle area formula]
+        //L = b * h           [apply above two and take out the 1/2]
+        //h = L / b           [divide by b]
+        //h^2 = (L / b)^2     [square it]
+        //h^2 = L^2 / b^2     [factor the divisor]
+        const coord_t height_2 = area_removed_so_far * area_removed_so_far / base_length_2;
+        coord_t weighted_average_width;
+        const coord_t extrusion_area_error = calculateExtrusionAreaDeviationError(previous, current, next, weighted_average_width);
+        if ((height_2 <= 1 //Almost exactly colinear (barring rounding errors).
+             && LinearAlg2D::getDistFromLine(current.p, previous.p, next.p) <= 1) // make sure that height_2 is not small because of cancellation of positive and negative areas
```
nallath, line 116:
```suggestion
             && LinearAlg2D::getDistFromLine(current.p, previous.p, next.p) <= 1) // Make sure that height_2 is not small because of cancellation of positive and negative areas
```

src/utils/ExtrusionLine.cpp
```diff
@@ -27,4 +28,180 @@ void ExtrusionLine::appendJunctionsTo(LineJunctions& result) const
...
+                        previous = previous_previous;
+                    }
+                }
+            }
+            else
+            {
+                continue; //Remove the vertex.
+            }
+        }
+        //Don't remove the vertex.
```
nallath, line 163: Isn't this a "Vertex isn't removed" (Because it's a state)

PR #1361, opened 2020-11-23
Fix simplify application

BagelOrb:
The simplify function has been fixed in recent versions, but it's application was still shoddy.
Sometimes it was applied where it shouldn't have been (in a getter function)
Sometimes it wasn't applied where it should have been,
Sometimes it was applied with user-settings even though it wasn;t a printed polygon
Sometimes it  was applied without user settings even though it was a printed polygon
Sometimes the user settings weren't obtained from the correct settings container (mesh settings)

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -651,6 +651,9 @@ void FffGcodeWriter::processRaft(const SliceDataStorage& storage)
     constexpr double fill_overlap = 0; // raft line shouldn't be expanded - there is no boundary polygon printed
     constexpr int infill_multiplier = 1; // rafts use single lines
     constexpr int extra_infill_shift = 0;
+    coord_t max_resolution = mesh_group_settings.get<ExtruderTrain&>("adhesion_extruder_nr").settings.get<coord_t>("meshfix_maximum_resolution");
+    coord_t max_deviation = mesh_group_settings.get<ExtruderTrain&>("adhesion_extruder_nr").settings.get<coord_t>("meshfix_maximum_deviation");
+
     Polygons raft_polygons; // should remain empty, since we only have the lines pattern for the raft...
     std::optional<Point> last_planned_position = std::optional<Point>();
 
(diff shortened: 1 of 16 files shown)


Ghostkeeper:
A few general remarks:
* Instead of passing around individual settings, try to pass around `Settings` objects. It scales better if there are more settings necessary in a function.
* Instead of using some sort of resolve function on per-object settings, we normally grab the setting value from the per-extruder settings in a context that's not specific to a mesh.

BagelOrb:
Good idea about the settings objects.

About the second remark: I thought there were some inherent problems when you don't get a setting from the stack it should be defined on. You wouldn't get travel speed from the global container if no extruder is available for some reason, because the global container doesn't have a value for travel speed, right?

I'm not sure, though. Can you verify whether per-object settings also always have a value on the extruder stack?

Ghostkeeper:
There are inherent problems with it indeed, but it's a bit different than what you expect. The global settings object does have a value for the travel speed, but it's the global travel speed which is not shown in the interface. Likewise for per-extruder settings and per-object settings.
As a result, grabbing it from the most specific settings object available gives the result that when drawing a specific mesh, the MaxResolution of the per-object settings will be used, but when drawing something common like support, the per-extruder settings will be used for that. You'll get a mix. But I think that is expected behaviour in this case.

Just to verify, I ran this code in the `MeshGroup::finalize` function (where `settings` refers to the mesh group settings):

```
std::cout << "Travel speed: " << settings.get<double>("speed_travel") << std::endl;
std::cout << "Infill speed: " << settings.get<double>("speed_infill") << std::endl;
```

That gives the following output:

```
2020-11-24 11:11:27,914 - DEBUG - [EngineOutputThread] UM.Backend.Backend._backendLog [110]: [Backend] Travel speed: 150.000000
2020-11-24 11:11:27,914 - DEBUG - [EngineOutputThread] UM.Backend.Backend._backendLog [110]: [Backend] Infill speed: 35.000000
```

I had overridden the travel speed to be 120 in the first extruder, but that didn't show up here. The infill speed was overridden by the profile to be 70, but that also didn't show up here. But they do always have a value.

BagelOrb:
Ok, but how is for getting object settings from the extruder? Does the extruder always have the value that's displayed in settings bar in Cura? Yeah right?

Ghostkeeper:
Yes, it gets the one from the list of settings shown in the interface.

Internally, the extruder instance itself only stores the per-extruder settings and then refers to its parent `Settings` instance for any setting you request that it doesn't have. This parent is set to the current mesh group being processed. The mesh group will always be empty at the moment but itself refers to the global `Settings` instance which always has all settings.

BagelOrb:
Ok I've fixed some settings in a better way now.

I haven't really changed the parameters to settings objects, though. I didn't feel that was a hard requirement, but I can also change that if you'd like.

rburema (commented):
Looks good.

I don't really have a problem with the parameters, since Infill already passes parameters instead of using settings. I understand that fixing this is a bit out of scope for the task at hand.

I'll component-test it and then will merge if that works.

BagelOrb:
That went quicker than I expected! :D

PR #1362, opened 2020-11-23
Fix polyline handling

BagelOrb:
This PR combines the past few PRs of mine which ar all related to processing polylines.
This concerns polylines in surface mode, but also for all infill patterns.

- the simplify() function is edited so that it can also process polylines
- the pathOrderOptimizer is extended so that it can handle polylines of size > 2
- in various places in the engine we make use of this by actually generating polylines instead of immediately cutting them up into 2 vert segments
- a PolylineStitcher class is introduced which further stitches together polylines into chains
- the chaining functionality which used to be inside pathOrderOptimizer is removed
- simplification is now applied in between chaining and order optimization
- the infill unit test has various bugs fixed
- a new test case showing a bug in clipper is added (which currently **always fails**)
- the modifier mesh functionality has also been applied to surface mode meshes - making it *possible to design and test an infill pattern outside of Cura*

As such this PR has the following advantages:
- surface mode for high poly models can now be printed without constant stuttering
- surface mode doesn't start somewhere in the middle of a polyline anymore, which would leave blobs in the middle of curved surfaces
- connected infill also doesn't start somewhere in the middle of a polyline anymore. I can recall a bug which caused the connections to be printed in the wrong directions sometimes (see below), but I'm not sure that failure mode is still present.
- surface mode meshes can now be used as infill, meaning that people (both internal and external) can more easily prototype a novel infill pattern - without actually implementing it.
- for some models the computation time is considerably reduced (9%)
- for some models the print time is also reduced. (`simplifyPolylines()` removes verts, which removes the need to slow down for a corner there. )
- fixed an issue where surface mode meshes would be printed twice

PS: the zigzag order bug I recall: Left infill line is printed first, right one last.
image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -217,6 +217,7 @@ set(engine_SRCS # Except main.cpp.
 
 # List of tests. For each test there must be a file tests/${NAME}.cpp.
 set(engine_TEST
+    ClipperTest
     GCodeExportTest
     InfillTest
     LayerPlanTest
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -209,6 +209,7 @@ set(engine_SRCS # Except main.cpp.
     src/utils/PolygonProximityLinker.cpp
     src/utils/polygonUtils.cpp
     src/utils/polygon.cpp
+    src/utils/PolylineStitcher.cpp
     src/utils/ProximityPointLink.cpp
     src/utils/SVG.cpp
     src/utils/socket.cpp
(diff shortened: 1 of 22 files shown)


BagelOrb:
[UM3_gyroid.3mf.rename.zip](https://github.com/Ultimaker/CuraEngine/files/5584189/UM3_gyroid.3mf.rename.zip)
on 4.8: gcode of 31MB, 2h1min print time

BagelOrb:
See above three linked PRs for more example results.

TheRooster13:
Has this been abandoned? These features would be so helpful for me...

Ghostkeeper:
Yeah, this was sort of abandoned in favour of Arachne (#1210). We are unsure if there are still pieces of this that we now need to merge in. There's a lot of changes here though so it's hard to say.

TheRooster13:
> Yeah, this was sort of abandoned in favour of Arachne (#1210). We are unsure if there are still pieces of this that we now need to merge in. There's a lot of changes here though so it's hard to say.

The part that I'm most interested in is correctly handling polylines for surface mode. Currently, surface mode doesn't respect some settings like resolution, so if the mesh is highly detailed, the prints take a long time because the polyline created from slicing has tons of tiny line segments. Sometimes the print order of the polylines is odd too. Like starting in the middle of what should be one long line segment.

I'm not sure if Arachne handles those cases or not. If so, great!

PR #1363, opened 2020-11-25
CURA-7874:  remove insets in spiralize and surface mode

konskarm:
Previously, the vase (spiralize) mode always used the insets[0] when creating the outer spiral for a vase. This meant that the spiral inset was always dependend on the generation of insets. Since in libArachne there are no insets any more, spiralize mode can't work. We had two options for that:

- Generate the bottom layers (base) of the vase using libArachne toolpaths and the rest of the vase (spiral part) using the old-fashioned offets (to produce one spiral inset)
- Generate both the base and the spiral part of the vase using libArachne toolpaths

Considering that the second option would require a significant change, we decided to go for the first approach. This PR implements exactly that. It separates the spiral part into a new variable called spiral_inset and generates that only when spiralize mode is enabled. 

**Known issue:** Currently, due to another bug, libArachne moves all the generated toolpaths a bit to the bottom left of the model, leading to a misalignment with the part outline. Although this isn't that apparent in general, it becomes visible now, since it can be directly compared to the offsetted wall generated for the spiral.

image

Once this (separate) bug is solved, the vase will look as it is intended.

Contributes to CURA-7874 subtask of CURA-7678


src/FffGcodeWriter.cpp
```diff
@@ -1781,20 +1781,20 @@ bool FffGcodeWriter::partitionInfillBySkinAbove(Polygons& infill_below_skin, Pol
...
+    const coord_t infill_skin_overlap = mesh.settings.get<coord_t>((part.wall_toolpaths.size() > 1) ? "wall_line_width_x" : "wall_line_width_0") / 2;
     const Polygons infill_below_skin_overlap = infill_below_skin.offset(-(infill_skin_overlap + tiny_infill_offset));
 
     return !infill_below_skin_overlap.empty() && !infill_not_below_skin.empty();
 }
 
 void FffGcodeWriter::processSpiralizedWall(const SliceDataStorage& storage, LayerPlan& gcode_layer, const PathConfigStorage::MeshPathConfigs& mesh_config, const SliceLayerPart& part, const SliceMeshStorage& mesh) const
 {
-    if (part.insets.size() == 0 || part.insets[0].size() == 0)
+    if (part.spiral_insets.size() == 0 || part.spiral_insets[0].size() == 0)
```
jellespijker, line 1792: Check against empty()

src/sliceDataStorage.cpp
```diff
@@ -284,7 +284,7 @@ bool SliceMeshStorage::getExtruderIsUsed(const size_t extruder_nr, const LayerIn
     {
         for (const SliceLayerPart& part : layer.parts)
         {
-            if (part.insets.size() > 1 && part.insets[1].size() > 0)
+            if (part.wall_toolpaths.size() > 1 && part.wall_toolpaths[1].size() > 0)
```
jellespijker, line 287: Check against empty() for > 0

jellespijker (approved):
Missed a couple of small check against empty

jellespijker (approved):

PR #1364, opened 2020-11-25
Cura 7678 remove inset usage

jellespijker:
Removed insets from CuraEngine.

Also see Cura-PR Ultimaker/Cura#8834

libArachne takes care of the generation of walls. A lot of legacy code still used insets, which where generated in the "old" way. This pull request fixes the legacy-code to use the the walls and correlated areas generated by libArachne.
Since insets were intertwined and used in a lot functions, this PR needs to be tested extensively for various scenarios. Below is a list of processes and methods that are changed, describing the change and possible (side) effects on various Cura settings:

* **Combing**
  Previously the combing algorithm used multiple offset from the inset to generate the allowed area's. These are now replaced by the outline, and precomputed infill and inner_areas. (Note to devs see PR #1371) 
* **Path ordering for parts**
  Previously either the outline, or if available, the outer inset of a part was returned to determine the optimal order of the parts to be printed. Now it will just return the outline, since computing the outer wall polygon would be to expensive to calculate and the gain in precision negligible. (Note to devs see PR #1370) 
* **Infill cubic sub division**
  Previously Cubic Subdivision infill used the 2nd wall (or the outer wall if there was no second) to check if it needed to be divided further. However, that check was not entirely correct, since the model could have more than just two walls. It now checks against the already generated infill area. (Note to devs see PR #1373 for more detail)
* **Spiralize**
  Spiralizing is still done with the usage of an inset, since libArachne will not generate a continues spiraling bead. When *surface mode: both* or *surface mode: normal* is checked in Cura the bottom part will be generated by libArachne and the spiraled surface is generated the old way. Since libArachne currently shifts the whole lines a couple of micrometers there could be a visible shift seen in layers generated between libArachne and the spiral inset. This should be addressed in CURA-7853 (Note to devs see #1363 )
  In addition, there is a know bug where in some very specific models the top-most bottom layer may be missing. There is a separate ticket in the backlog (CURA-7920) to fix that.
* **Fuzzy skin**
 Fuzzy skin is disabled for now and will be made to work at a later stage, see CURA-7887. The settings are hidden in Cura in PR Ultimaker/Cura#8834 . (Note to devs see #1363 )
* **Support**
  - Changes are made in the combine support infill layers, were the gradual infill areas are combined per density. Here a check was removed that skipped the combing if there were no insets present.
  - Changes are made in the detection of overhang points
* **Skin edge support**
  The generation of the area which will be the skin edge support area (with supporting wall) used to check for insets is empty now it checks if wall_toolpaths are empty
* **Documentation**
  Not yet done and taken out of the scope of this ticket, see CURA-7887

Fixes CURA-7678 with sub-tickets: CURA-7560,  CURA-7874, CURA-7875, CURA-7876, CURA-7877, CURA-7880, CURA-7881, CURA-7882, CURA-7883 and CURA-7884


src/LayerPlan.cpp
```diff
@@ -1059,7 +1029,7 @@ void LayerPlan::addWalls(const PathJunctions& walls, const SliceMeshStorage& mes
 void LayerPlan::addLinesByOptimizer(const Polygons& polygons, const GCodePathConfig& config, SpaceFillType space_fill_type, bool enable_travel_optimization, int wipe_dist, float flow_ratio, std::optional<Point> near_start_location, double fan_speed)
 {
     Polygons boundary;
-    if (enable_travel_optimization && comb_boundary_inside2.size() > 0)
+    if (enable_travel_optimization && comb_boundary_preferred.size() > 0)
```
konskarm, line 1037: Check against empty()

src/LayerPlan.cpp
```diff
@@ -135,121 +135,92 @@ ExtruderTrain* LayerPlan::getLastPlannedExtruderTrain()
...
-    if (combing_mode == CombingMode::OFF)
+    if (combing_mode == CombingMode::OFF || (layer_nr < 0 && combing_mode == CombingMode::NO_SKIN))
     {
-        return Polygons();
+        return false;
     }
-    if (layer_nr < 0)
-    { // when a raft is present
-        if (combing_mode == CombingMode::NO_SKIN)
+    return true;
```
Ghostkeeper, line 145: Perhaps more expressive to:

```suggestion
    return combing_mode != CombingMode::OFF && (layer_nr >=0 || combing_mode != CombingMode::NO_SKIN)
```
```

File: src/LayerPlan.h [hunk 3]
```diff
@@ -334,17 +334,40 @@ class LayerPlan : public NoCopy
...
+     * Determines if the combing boundary need to be computed
+     * \return true if there should be a combing designated surface on this layer, otherwise false.
      */
-    Polygons computeCombBoundaryInside(const size_t max_inset);
+    bool CombBoundaryRequired();
+
+    /*!
+     * Compute the minimum combing boundary.
+     *
+     * This is in most cases within half a wall line width smaller then the part outlines. Except for raft layers.
```
Ghostkeeper, line 350:
```suggestion
     * This is in most cases within half a wall line width smaller than the part outlines. Except for raft layers.
```

src/LayerPlan.h
```diff
@@ -334,17 +334,40 @@ class LayerPlan : public NoCopy
...
+     * It will then return a boundary 0.1 mm bigger then the raft outline. When combing mode is OFF it
```
Ghostkeeper, line 351:
```suggestion
     * It will then return a boundary 0.1 mm bigger than the raft outline. When combing mode is OFF it
```

src/LayerPlan.h
```diff
@@ -334,17 +334,40 @@ class LayerPlan : public NoCopy
...
+     */
+    Polygons computeMinimumCombBoundary();
+
+    /*!
+     * Compute the preferred combing boundary
+     *
+     * It will different boundaries depending on the requested combing mode:
+     *  - OFF: an empty Polygon
+     *  - ALL: a boundary half a wall line width smaller then the part outline
+     *  - NOT_SKIN: a boundary half a wall line width smaller then the part outline and the inside of the inner wall,
```
Ghostkeeper, line 364:
```suggestion
     *  - NO_SKIN: a boundary half a wall line width smaller than the part outline and the inside of the inner wall,
```

src/LayerPlan.h
Ghostkeeper, line 350: Currently the minimum combing boundary is a full Outer Wall Line Width smaller than the outlines. Not half.

This might change if you implement my earlier concern that this area is wrong though.

src/WallsComputation.cpp
```diff
@@ -125,10 +127,17 @@ void WallsComputation::generateInsets(SliceLayerPart* part)
...
+    if (spiralize)
+    {
+        generateSpiralInsets(part, line_width_0, wall_0_inset, recompute_outline_based_on_outer_wall);
+    }
     // Call on libArachne:
     WallToolPaths wall_tool_paths(part->outline, line_width_0, line_width_x, inset_count, settings);
     part->wall_toolpaths = wall_tool_paths.getToolPaths();
     part->inner_area = wall_tool_paths.getInnerContour();
+    part->print_outline = part->outline;
+    // TODO Recompute the print_outline of the part according to the generated wall_toolpaths
```
Ghostkeeper, line 140: Oh, but this could actually be a pretty big problem, right?

src/WallsComputation.cpp
```diff
@@ -161,4 +170,45 @@ void WallsComputation::generateInsets(SliceLayer* layer)
...
+        // - There are fewer parts to the polygon (fits better in slim areas).
+        // - The polygon area is largely unaffected.
+        Polygons alternative_inset;
+        alternative_inset = part->outline.offset(-(line_width_0 - try_smaller) / 2 - wall_0_inset);
+
+        if (alternative_inset.size() < inset_part_count - minimum_part_saving) //Significantly fewer parts (saves more than 3 parts).
+        {
+            part->spiral_insets[0] = alternative_inset;
+        }
+    }
```
Ghostkeeper, line 192: This part here is meant to reduce the effect of the predecessor of the "wobbly lines" issue (currently in the sprint), where a slightly alternating polygon width could cause a wall to split up into many tiny little contours, because it alternates between slightly over 1 wall width and slightly under 1 wall width. This code tries a few line widths to see which leads to the smallest number of paths and chooses that then.

I think this is not necessary for spiralize mode. There's only 1 wall, so it would only apply for parts of around the thickness of 1 outer wall width. Spiralize doesn't deal well with thin walls anyway, especially now that wall overlap compensation is removed. So I think to make this code simpler, it's best to remove this.

src/support.cpp
```diff
@@ -1287,8 +1287,9 @@ std::pair<Polygons, Polygons> AreaSupport::computeBasicAndFullOverhang(const Sli
 void AreaSupport::detectOverhangPoints(const SliceDataStorage& storage, SliceMeshStorage& mesh)
 {
     const ExtruderTrain& infill_extruder = mesh.settings.get<ExtruderTrain&>("support_infill_extruder_nr");
-    const coord_t support_line_width = infill_extruder.settings.get<coord_t>("support_line_width");
-    const coord_t max_tower_supported_diameter = mesh.settings.get<coord_t>("support_tower_maximum_supported_diameter");
+    const int offset = - static_cast<int>(infill_extruder.settings.get<coord_t>("support_line_width")) / 2;
```
Ghostkeeper, line 1290: `coord_t` is a signed integer and can be negative. No cast is necessary.

```suggestion
    const coord_t offset = -infill_extruder.settings.get<coord_t>("support_line_width") / 2;
```

src/support.cpp
```diff
@@ -1287,8 +1287,9 @@ std::pair<Polygons, Polygons> AreaSupport::computeBasicAndFullOverhang(const Sli
...
+    const double max_tower_supported_diameter = static_cast<double>(mesh.settings.get<coord_t>("support_tower_maximum_supported_diameter"));
```
Ghostkeeper, line 1291: Why would you make this `double`? It is a size, so it should be `coord_t`:

```suggestion
    const coord_t max_tower_supported_diameter = mesh.settings.get<coord_t>("support_tower_maximum_supported_diameter");
```

Ghostkeeper (changes requested):

src/WallsComputation.cpp
konskarm: Generally, yes, it won't be easy to implement. Nevertheless, we can just keep it as the part->outline offsetted by the outer wall inset, as the deviation between the outer side of the outer wall generated by libArachne and the print outline should be insignificant.

src/WallsComputation.cpp
konskarm: But spiralize reduces the entire shape to only its outer polygon. So technically, if the outer polygon becomes smaller by an outer wall line width, it may still still result in a multi-part polygon, which is not ideal for spiralization, considering that only the first part will always be spiralized. So it _may_ then still be a problem then, right?

Although I do understand that the possibility of such a case occuring is tremendously low.

src/LayerPlan.cpp
```diff
@@ -135,121 +135,92 @@ ExtruderTrain* LayerPlan::getLastPlannedExtruderTrain()
...
+            {
+                const SliceLayer& layer = mesh.layers[static_cast<size_t>(layer_nr)];
+                if (mesh.settings.get<bool>("infill_mesh")) // don't process infill_mesh
+                {
+                    continue;
+                }
+                const coord_t offset = -10 - mesh.settings.get<coord_t>("wall_line_width_0");
+                for (const SliceLayerPart& part : layer.parts)
+                {
+                    comb_boundary.add(part.outline.offset(offset));
```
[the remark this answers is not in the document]
jellespijker: Yes that was the old "correct" behaviour. @konskarm raised the same question but we both thought this implementation might be in improvement. That being said, this will be technically be a change in the behaviour so I will change the code per your suggestions.

src/LayerPlan.cpp
```diff
@@ -135,121 +135,92 @@ ExtruderTrain* LayerPlan::getLastPlannedExtruderTrain()
...
-
-                        outer = outer.offset(outer_to_outline_dist/2+10).unionPolygons(outline_where_outer_is_missing.offset(outer_to_outline_dist/2+10)).offset(-(outer_to_outline_dist/2+10));
+                        comb_boundary.add(part.inner_area);
                     }
-
-                    if (num_insets == 0)
+                }
+                else if (combing_mode == CombingMode::NO_SKIN)
+                {
+                    const coord_t inner_walls_offset = (mesh.settings.get<coord_t>("wall_line_count") - 1) / 2 * mesh.settings.get<coord_t>("wall_line_width_x") / 2;
```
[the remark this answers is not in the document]
jellespijker: Thanks for the clarification.
But it seems that we both missed an additional bug: `mesh.settings.get<coord_t>("wall_line_count")` should be `mesh.settings.get<size_t>("wall_line_count")`. The `get<coord_t>` does a factor 1000 multiplication.

src/WallsComputation.cpp
Ghostkeeper: Yeah that's true, it would only apply for parts of around the thickness of 1 outer wall width. Those could get broken up in this case. Spiralize doesn't deal well with thin walls anyway due to its nature of gradually increasing the Z coordinate, especially now that wall overlap compensation is removed.

Also consider that this only tries 10 micrometres smaller line widths. This is intended to guard against rounding errors in the insets. The chance that it'll help with a very thin part is going to be very rare.

src/WallsComputation.cpp
```diff
@@ -21,30 +21,30 @@ WallsComputation::WallsComputation(const Settings& settings, const LayerIndex la
...
     {
-        inset_count += ((layer_nr % 2) + 2) % 2;
+        wall_count += ((layer_nr % 2) + 2) % 2;
     }
 
-    if (inset_count == 0)
+    if (wall_count == 0)
     {
-        part->insets.push_back(part->outline);
+        part->spiral_insets.push_back(part->outline);
```
Ghostkeeper, line 41: I'd say that the spiralize path should not get 1/2 line width offset if the Wall Line Count is set to 0, right? So this also needs to be inset by half a line width again here.

Ghostkeeper, line 41: You only need to generate this if Spiralize is enabled. And wasn't that done in `FffPolygonGenerator` too?

I'd expect that the `generateSpiralInsets` function is responsible for this.

src/sliceDataStorage.cpp
```diff
@@ -50,6 +50,24 @@ const Polygons& SliceLayerPart::getOwnInfillArea() const
     }
 }
 
+bool SliceLayerPart::hasWallAtInsetIndex(size_t inset_idx) const
+{
+    if(!wall_toolpaths.empty())
```
Ghostkeeper, line 55: This if-statement is superfluous. The `for` loop below already does that check.

src/sliceDataStorage.cpp
```diff
@@ -240,7 +175,7 @@ bool SliceMeshStorage::getExtruderIsUsed(const size_t extruder_nr, const LayerIn
     {
         for (const SliceLayerPart& part : layer.parts)
         {
-            if (part.insets.size() > 0 && part.insets[0].size() > 0)
+            if ((!part.wall_toolpaths.empty() && part.hasWallAtInsetIndex(0)) || (!part.spiral_insets.empty() && !part.spiral_insets[0].empty()))
```
Ghostkeeper, line 178: `hasWallAtInsetIndex(0)` already includes checking if `wall_toolpaths` is empty.
```suggestion
            if(part.hasWallAtInsetIndex(0) || (!part.spiral_insets.empty() && !part.spiral_insets[0].empty()))
```

src/sliceDataStorage.cpp
```diff
@@ -268,23 +203,11 @@ bool SliceMeshStorage::getExtruderIsUsed(const size_t extruder_nr, const LayerIn
...
-                return true;
-            }
-        }
-    }
     if ((settings.get<size_t>("wall_line_count") > 1 || settings.get<bool>("alternate_extra_perimeter")) && settings.get<ExtruderTrain&>("wall_x_extruder_nr").extruder_nr == extruder_nr)
     {
         for (const SliceLayerPart& part : layer.parts)
         {
-            if (part.insets.size() > 1 && part.insets[1].size() > 0)
+            if (!part.wall_toolpaths.empty() && part.hasWallAtInsetIndex(1))
```
Ghostkeeper, line 210: `hasWallAtInsetIndex(1)` already includes checking if `wall_toolpaths` is empty.
```suggestion
            if(part.hasWallAtInsetIndex(1))
```

src/utils/ExtrusionLine.h
```diff
@@ -60,6 +60,11 @@ struct ExtrusionLine
      */
     coord_t getLength() const;
 
+    /*!
+     * Get the minimal width of this path
+     */
+    coord_t getMinimalWidth() const;
```
Ghostkeeper, line 66: This new function seems to be unused.

Ghostkeeper (changes requested):

src/LayerPlan.cpp
```diff
@@ -135,121 +135,52 @@ ExtruderTrain* LayerPlan::getLastPlannedExtruderTrain()
...
-            }
-            const CombingMode combing_mode = mesh.settings.get<CombingMode>("retraction_combing");
-            if (combing_mode == CombingMode::NO_SKIN)
+            for (const SliceMeshStorage& mesh : storage.meshes)
             {
-                // we need to include the walls in the comb boundary otherwise it's not possible to tell if a travel move crosses a skin region
-
-                const coord_t line_width_0 = mesh.settings.get<coord_t>("wall_line_width_0");
+                const SliceLayer& layer = mesh.layers[static_cast<size_t>(layer_nr)];
+                if (mesh.settings.get<bool>("infill_mesh")) // don't process infill_mesh
```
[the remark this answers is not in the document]
jellespijker: Good point. But this could also decrease the print quality of the frist layer if a travel is oozing material on the plate where at a later time a mesh is printed.

src/utils/ExtrusionLine.h
jellespijker: remnant of an old solution, kept in because it seemed like a logical interface which can be used in the future.

src/LayerPlan.cpp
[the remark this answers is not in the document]
jellespijker: The function `mesh.isPrinted()` return `true` if the mesh is not infill, cutting or overhang. So it isn't a status if the mesh is already printed and imo a misnomer.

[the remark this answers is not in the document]
jellespijker: That also means that it is allowed to go through the skin of a cutting mesh.

image
Above picture is with the suggested change. Here a travel can occur through an outside wall when cut with a cutting mesh.

image
But changing it to `mesh.settings.get<bool>("infill_mesh") && mesh.settings.get<bool>("anti_overhang_mesh")` gives beter results.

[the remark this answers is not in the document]
Ghostkeeper: Hmm, maybe the cutting mesh is "cropped" to the shape that it cuts? I don't remember exactly how this worked any more.

As long as it doesn't go avoiding the full shape of the cutting mesh (which could be way bigger than the actual normal mesh) then it's fine like this.

src/utils/ExtrusionLine.h
Ghostkeeper: I would still remove it in the interest of reducing maintenance burden.

However my opinion about this isn't very strong, because I see your point too. You can decide for yourself if you think it's worth it to keep then.

src/LayerPlan.cpp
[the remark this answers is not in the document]
Ghostkeeper: Well the problem is that it then doesn't effectively work any more as `NO_SKIN` will still go through skin at the top and bottom of the model and such.

Ghostkeeper (approved):

Ghostkeeper (approved):

PR #1366, opened 2020-11-26
CURA-7876: Disable fuzzy skin in libArachne

konskarm:
Fuzzy skin will be re-enabled when it is made to work with the new wall_toolpaths.
Until then it will be disabled.

Contributes to CURA-7678 via CURA-7876


jellespijker:
Is there also a PR for libarachne_rebased in Cura where the setting is disabled?

PR #1371, opened 2020-11-30
Cura 7877 combing without insets

jellespijker:
Reverted to old "correct" combing area generation

### DIFF
--- a/src/LayerPlan.cpp
+++ b/src/LayerPlan.cpp
@@ -95,8 +95,8 @@ LayerPlan::LayerPlan(const SliceDataStorage& storage, LayerIndex layer_nr, coord
 , last_extruder_previous_layer(start_extruder)
 , last_planned_extruder(&Application::getInstance().current_slice->scene.extruders[start_extruder])
 , first_travel_destination_is_inside(false) // set properly when addTravel is called for the first time (otherwise not set properly)
-, comb_boundary_minimum(computeMinimumCombBoundary())
-, comb_boundary_preferred(computePreferredCombBoundary())
+, comb_boundary_minimum(computeCombBoundary(true))
+, comb_boundary_preferred(computeCombBoundary(false))
 , comb_move_inside_distance(comb_move_inside_distance)
 , fan_speed_layer_time_settings_per_extruder(fan_speed_layer_time_settings_per_extruder)
 {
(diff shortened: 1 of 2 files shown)


src/LayerPlan.cpp
```diff
@@ -95,8 +95,8 @@ LayerPlan::LayerPlan(const SliceDataStorage& storage, LayerIndex layer_nr, coord
 , last_extruder_previous_layer(start_extruder)
 , last_planned_extruder(&Application::getInstance().current_slice->scene.extruders[start_extruder])
 , first_travel_destination_is_inside(false) // set properly when addTravel is called for the first time (otherwise not set properly)
-, comb_boundary_minimum(computeMinimumCombBoundary())
-, comb_boundary_preferred(computePreferredCombBoundary())
+, comb_boundary_minimum(computeCombBoundary(true))
+, comb_boundary_preferred(computeCombBoundary(false))
```
konskarm, line 99: Rename the function computeCombBoundary to computeCombBoundaryWithMinimumBoundary, or else the boolean input can be confusing.

```suggestion
, comb_boundary_minimum(computeCombBoundaryWithMinimumBoundary(true))
, comb_boundary_preferred(computeCombBoundaryWithMinimumBoundary(false))
```

src/LayerPlan.cpp
```diff
@@ -135,49 +135,11 @@ ExtruderTrain* LayerPlan::getLastPlannedExtruderTrain()
...
-    const CombingMode combing_mode = Application::getInstance().current_slice->scene.current_mesh_group->settings.get<CombingMode>("retraction_combing");
-    if (combing_mode == CombingMode::OFF || (layer_nr < 0 && combing_mode == CombingMode::NO_SKIN))
-    {
-        return false;
-    }
-    return true;
-}
-
-Polygons LayerPlan::computeMinimumCombBoundary()
+Polygons LayerPlan::computeCombBoundary(const bool minimumBoundary)
```
konskarm, line 138: Rename the variable to not use camel case.

```suggestion
Polygons LayerPlan::computeCombBoundary(const bool minimum_boundary)
```

src/LayerPlan.h
```diff
@@ -338,36 +339,23 @@ class LayerPlan : public NoCopy
...
+     * Preferred combing boundary:
+     *  - If CombingMode::ALL: Add the increased outline offset (skin, infill and part of the inner walls).
+     *  - If CombingMode::NO_SKIN: Add the increased outline offset, subtract skin (infill and part of the inner walls).
+     *  - If CombingMode::INFILL: Add the infill (infill only).
      *
-     * \return the preferred boundary or an empty Polygons if no combing is required
+     * \return the combing boundary or an empty Polygons if no combing is required
      */
-    Polygons computePreferredCombBoundary();
+    Polygons computeCombBoundary(const bool minimumBoundary);
```
konskarm, line 358:
```suggestion
    Polygons computeCombBoundaryWithMinimumBoundary(const bool minimum_boundary);
```

konskarm (changes requested):
Minor code style changes

src/LayerPlan.cpp
konskarm: Come to think of it, we can keep the function name the same and just introduce an enum CombingBoundary with values MINIMUM and PREFERRED. It will be more readable.

jellespijker: Like the idea of an enum. I will use that

PR #1376, opened 2020-12-01
Use MM2INT for hard-coded distances

Ghostkeeper:
This should not bring any real change to the behaviour, but will make it easier in the future to change the scale of our units.


Ghostkeeper:
Seems that Github Actions had some internal errors when these tests were ran. Please verify by running the tests locally beforehand.

rburema (approved):

rburema:
`100% tests passed, 0 tests failed out of 21`

Runs fine. I don't think we need a ticket for this one do we? What will the testers even do?

rburema:
I did notice that squared values weren't replaced (like areas), but I'm not sure if we want that?

Ghostkeeper:
We should probably replace those squared values too. There are alternative macros called `MM2_2INT` and `MM3_2INT` for higher dimensionalities. Where are they?

To the testers there are provably no differences unless the unit scale is actually changed. I had before tested the behaviour when changing the unit scale to 0.1um and found no more issues, but it could be that I missed a few details.

PR #1377, opened 2020-12-02
Outer wall inset

nallath:
Fixes the outer wall inset functionality to work with LibArachne style path generation. 

CURA-7830

### DIFF
--- a/src/BeadingStrategy/OuterWallInsetBeadingStrategy.cpp
+++ b/src/BeadingStrategy/OuterWallInsetBeadingStrategy.cpp
@@ -15,14 +15,13 @@ namespace cura
         bead_count = std::count_if(ret.bead_widths.begin(), ret.bead_widths.end(), [](const coord_t width) { return width > 0; });
 
         // Early out when the only walls are outer. 
-        if (bead_count < 3)
+        if (bead_count < 2)
         {
             return ret;
         }
         
-        // Actually move the outer wall inside. 
-        ret.toolpath_locations[0] = ret.toolpath_locations[0] + outer_wall_offset;
-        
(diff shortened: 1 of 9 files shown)


src/BeadingStrategy/OuterWallInsetBeadingStrategy.cpp
```diff
@@ -0,0 +1,29 @@
...
+        bead_count = std::count_if(ret.bead_widths.begin(), ret.bead_widths.end(), [](const coord_t width) { return width > 0; });
+
+        // Early out when the only walls are outer. 
+        if (bead_count < 3)
+        {
+            return ret;
+        }
+        
+        // Actually move the outer wall inside. 
+        ret.toolpath_locations[0] = ret.toolpath_locations[0] + outer_wall_offset;
```
Ghostkeeper, line 24: I wonder what would happen if this moves the bead beyond the range of the edge in question. Worth paying attention to in testing I guess.

Ghostkeeper: This seems to then get clipped to the skeletal centroid, which is fine.

src/BeadingStrategy/OuterWallInsetBeadingStrategy.cpp
```diff
@@ -15,14 +15,13 @@ namespace cura
         bead_count = std::count_if(ret.bead_widths.begin(), ret.bead_widths.end(), [](const coord_t width) { return width > 0; });
 
         // Early out when the only walls are outer.
```
Ghostkeeper, line 17: This comment is outdated.
I'll update it myself to reduce the ping-pong.

Ghostkeeper (approved):

rburema (approved):

PR #1378, opened 2020-12-04
CURA-7830: Apply the Outer Wall Inset by regenerating outer toolpath

konskarm:
The process used is the following:

1.   Generate the wall toolpaths as if `Outer Wall Inset = 0`
2.   If the `Outer Wall Inset` is not 0, regenerate a second list of toolpaths, taking the inset into account (by insetting the outline before generating the toolpaths)
3.   Replace the wall toolpath at index 0 with the insetted wall toolpaths at index 0

This technique, although more resource intensive (as it generates the Skeletal Trapezoidation twice), it is more reliable as it does not suffer from the problem of how to inset outer walls at thin parts that already have only two beads. In such places, insetting the walls would lead in overlapping and overextrusion. Using the regenerated toolpaths, it is certain that libArachne will properly deal with these thin parts by generating a single ExtrusionLine

image

**Note:** I was considering generating only one toolpath the second time the skeletal trapezoidation is applied (using inset_count = 2) but that would then always generate a toolpath that has a width of exactly one outer wall line width. By generating all the toolpaths the second time, it is ensured that the beading strategy will do its job and reduce/increase the width in some places according to the total width, thus being closer to the original outer wall toolpath.

CURA-7830

### DIFF
--- a/src/BeadingStrategy/BeadingStrategyFactory.cpp
+++ b/src/BeadingStrategy/BeadingStrategyFactory.cpp
@@ -41,7 +41,7 @@ std::string to_string(StrategyType type)
 }
 
 
-coord_t get_weighted_average(const coord_t preferred_bead_width_outer, const coord_t preferred_bead_width_inner, const coord_t max_bead_count)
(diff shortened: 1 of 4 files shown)


nallath:
Closing this in favor of https://github.com/Ultimaker/CuraEngine/pull/1377

PR #1379, opened 2020-12-06
Cura issue #8844, fix for recesses being removed at the bottom of models when using "Make Overhangs Printable"

OpusMcN:
Using "Make Overhangs Printable" was removing any recessed areas at the bottom of the model due to the way this was calculated.  The top of the recessed area was solid and therefore completely filled the layer below it removing the hole.  This propagated down and filled the entire hole.

This fix checks each hole in a layer to see if it's completely occluded by the layer above.  If so its the top surface of a hole and we don't want to remove it. The hole polygon gets subtracted from the layer above before it gets offset and added to the layer below.   This only needs to happen on the top layer and it stops the propagation of the fill down through the hole.   Any time the hole polygon isn't completely occluded by the layer above no changes are made and the process carries on as per the initial implementation.   

Two test STL files attached, one showing the correct behavior for various arrangements of holes (i.e. not filling them) and one showing the correct behaviour for overhangs inside holes (where the overhang gets corrected but the hole isn't filled).

Also added two initial values into Comb.cpp, the MSVC debugger would throw an exception in debug mode when these values were passed to later functions (even though the values were never used)

### DIFF
--- a/src/ConicalOverhang.cpp
+++ b/src/ConicalOverhang.cpp
@@ -9,33 +9,66 @@
 
 namespace cura {
 
-void ConicalOverhang::apply(Slicer* slicer, const Mesh& mesh)
-{
-    const AngleRadians angle = mesh.settings.get<AngleRadians>("conical_overhang_angle");
-    const double tan_angle = tan(angle);  // the XY-component of the angle
-    const coord_t layer_thickness = mesh.settings.get<coord_t>("layer_height");
-    coord_t max_dist_from_lower_layer = tan_angle * layer_thickness; // max dist which can be bridged
+	void ConicalOverhang::apply(Slicer* slicer, const Mesh& mesh)
+	{
+		const AngleRadians angle = mesh.settings.get<AngleRadians>("conical_overhang_angle");
+		const double tan_angle = tan(angle);  // the XY-component of the angle
+		const coord_t layer_thickness = mesh.settings.get<coord_t>("layer_height");
+		coord_t max_dist_from_lower_layer = tan_angle * layer_thickness; // max dist which can be bridged
 
-    for (unsigned int layer_nr = slicer->layers.size() - 2; static_cast<int>(layer_nr) >= 0; layer_nr--)
-    {
-        SlicerLayer& layer = slicer->layers[layer_nr];
-        SlicerLayer& layer_above = slicer->layers[layer_nr + 1];
-        if (std::abs(max_dist_from_lower_layer) < 5)
-        { // magically nothing happens when max_dist_from_lower_layer == 0
-            // below magic code solves that
-            int safe_dist = 20;
-            Polygons diff = layer_above.polygons.difference(layer.polygons.offset(-safe_dist));
-            layer.polygons = layer.polygons.unionPolygons(diff);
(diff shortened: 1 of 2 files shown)


Ghostkeeper:
Fixes https://github.com/Ultimaker/Cura/issues/8844

(Linking here for Github. It doesn't seem to recognise that in the title.)

rburema:
Sorry for the wait.

This PR will remove _all_ holes from consideration of 'Make Overhang Printable'.

However, consider the case of a flipped over box (like in the screenshots... lid hovering above the build-plate on the left, flipped over box on the right).

box_top_view
box_mid_view

In this case, the overhang (that would be the bottom of the box if it weren't flipped over), is not 'Made Printable' at all, even though it would most definitely sag.

OpusMcN:
If I understand you correctly, it seems that's what I think should happen?   With the existing code the flipped over empty box would become a solid filled box which is doing something much more destructive that just fixing overhangs.   Leaving it as a hole with a (potentially) bridgeable roof is the behavior I'd like to see in this case.    The ideal solution would be to create a conical shaped roof with a printable slope, but that's beyond the current algorithm, and not meant to be fixed by this.   I can see the case for making an option for this fix or the original behavior depending on the situation though.

If I've misunderstood you please let me know!

OpusMcN:
If I'm understanding you correctly I feel that's describing the "correct" behavior I'd like to see.   With the existing code the flipped over box would be printed as a solid filled box which is a much more destructive change to the model than just conforming an overhang to an printable angle.    Ideally we'd be adding a sloping roof the the hole that matches the required angle, but that's beyond the current algorithm and not meant to be fixed by this.     Leaving the hole with a (potentially bridgeable) roof makes this feature useable for a much larger range of models.   (for example, the current models I'm printing have mounting holes on the bottom that require this fix, and lots of fine detail on walls that print beautifully without support by fixing the overhang angles)

It feels like filling in holes like this and adding slopes to overhangs are two different use cases, so I can see that an option for choosing the behavior make sense.

rburema:
If the box had only a slight hole in the side it would fill right up again though. That's a very large difference in the print for something that can be a very small change.

OpusMcN:
> If the box had only a slight hole in the side it would fill right up again though. That's a very large difference in the print for something that can be a very small change.

I don't see a hole in the side making a difference in to either the old code or the fix.  The old code will fill the entire box regardless of a hole (making the top of the hole "printable" if appropriate), the new code will keep the empty box, and make the top of the hole "printable" if it's needed.
image
slices to
image
with this fix applied.

I've added an option for limiting the area of filled recesses, a default value of zero leaves the original behaviour, increasing this can selectively bring back empty areas of any size required.  I'll update the code

OpusMcN:
Link to Ultimaker/Cura#9162 for the new option in fdmprinter.def.json

rburema:
Hi! Thanks for your contribution.

After a quick discussion with my colleagues, we found that since the extra setting keeps the current default (as we prefer, so an upgrade from last version isn't necessary ... for opening old projects and the like), we've decided to accept it basically as is (now).

P.S. With a hole I really meant one extending through to the bottom (of the BP, top when viewed from the upside down box' perspective). That's a bit of a moot point now though :-)

PR #1380, opened 2020-12-08
[CURA-7148] re-enable infill gap-filling w.r.t. 'variable line width'

rburema:
In case of zig-zagify (connect infill lines, on by default for newest UM printers) the infill is offset half a line width to make place for the potential extra zigs and zags. Because this happens on both sides of the polygons, this can result in areas of up to approximately a line-width not being filled. Previously, this was handled by our precarious and buggy gap filling techniques. The aim of this is to replace that with a call to Arachne (variable line width). So, in case zig-zagify or similar happens, make walltoolpaths with arachne, for a single inset, _but only keep the ones that are 'middle' (is_odd). Because the inset is max one, the only 'middle' walls will be what we'd call 'thin walls' if they where on the outside of the model. This corresponds 1:1 with what we want the infill gap-filling to be (in my opinion anyway). Besides that, it's 'just' a task to make sure the gap-infill and the 'normal' zig-zagified infill don't overlap, and it's done. Note that some code has been moved from the helper _generate function to the 'main' generate function so that wall-toolpaths could be called without too much hassle (and also it may actually fit better here: potentially reducing the outline was originally placed _after_ checking if the outline was empty or not).

Note that there is possibly a potential to get this a little bit more efficient if, instead of adding the lines to a new vector, we use remove_if with lambda .... However, we can't use erase after that, since we're inside a loop of vectors over vectors, and I couldn't get it done with just the remove_if (maybe it doesn't register as 'empty' any more?): The effect was an extra wall _always_ not just where the gaps needed to be filled.

*update*: See the conversation below, otherwise the accompanying frontend PR ( https://github.com/Ultimaker/Cura/pull/8918 ) will make little sense ;-)


src/infill.cpp
```diff
@@ -63,6 +63,58 @@ void Infill::generate(VariableWidthPaths& toolpaths, Polygons& result_polygons,
...
+                    }
+                    gap_filled_areas.add(path);
+                }
+            }
+            if (! thin_walls_only.empty())
+            {
+                toolpaths.push_back(thin_walls_only);
+            }
+        }
+        gap_filled_areas = gap_filled_areas.offsetPolyLine(infill_line_width / 2).unionPolygons();
```
Ghostkeeper, line 112: Previously this feature was dependent on the `fill_perimeter_gaps` setting. Maybe it should only be computed if that setting is active?

We can also consider to remove the setting entirely and just always fill these gaps. If we don't delete the setting though I think it should be renamed because its original function is no longer functioning.

Ghostkeeper (changes requested):
This approach is good! Quite elegant.

I think three things are missing from this PR. All of them involve deleting code:
* The MergeInfillLines class is still being called upon from `LayerPlan`. As far as I can tell, this could still have an effect on skin and infill lines at >=100% density but that may not be desirable any more, and in practice I never see this happen. There are many preconditions for it so my theory is probably wrong on this. What's more, I think this feature is intended to remove MergeInfillLines entirely.
* The `fill_perimeter_gaps` setting is still active and used in the code, but not your code. This starts from `FffGcodeWriter::processTopBottom`, which creates a "perimeter gaps output" polygon that ultimately gets passed to the `Infill` class who fills it in `generateConcentricInfill` only. This polygon then gets filled with a lines pattern in `FffGcodeWriter::processPerimeterGaps` (and later merged with the `MergeInfillLines` class which you should have deleted then). The purpose of the setting is caught by your current changes with the toolpaths generated in the half line width offset. So in effect, this would remove several hundreds of lines of code, defeat a dragon and remove a bunch of parameters from functions that have too many of those. It should be done!
* Similarly, we have the `fill_outline_gaps` setting (i.e. "Print Thin Walls") which was repurposed. However the outline gaps are still being computed in `FffPolygonGenerator::processOutlineGaps`. This is somewhat out of scope if you look at the title of this PR but not if you look at the title and description of the Jira ticket. This function can be removed as well, which will save some processing time. It fills the `SliceLayerPart::outline_gaps` field which is then not used at all except in a broken way to determine if an extruder is used.

src/infill.cpp
konskarm: Note: The processOutlineGaps has already been removed by https://github.com/Ultimaker/CuraEngine/pull/1364 (which is currently in review).

rburema: Conclusion: Setting removed.

rburema: (See also frontend PR: https://github.com/Ultimaker/Cura/pull/8918 )

PR #1381, opened 2020-12-10
[CURA-7148] "Alternate Ending"

rburema:
Alternate for https://github.com/Ultimaker/CuraEngine/pull/1380

If the last fix on that one doesn't work, we could decide to go this way. Not really anyone's preference, but at least we can continue with the plans tomorrow.


konskarm:
Closing this in favor of https://github.com/Ultimaker/CuraEngine/pull/1380.

PR #1394, opened 2021-01-20
Fix subtracting Durations

Ghostkeeper:
This was a small typo. Should be a minus sign here.

Gave me quite a bit of confusion today!


rburema (approved):
Usually I'd expect _very_ clear one liners as this directly on master, but I did my due dilligence and  asked nallath as well :-)

PR #1397, opened 2021-01-25
Limits hole offsets to holes larger than configured only instead of all.

alexjx:
The original hole horizon expansion will expand all internal outlines. This will create problems in certain cases.

For example, with 1mm expansion (exaggerated), with following model, the compensation create different outlines for layers with and without holes in the vertical walls.

Model:
image

Expanding (since the wall forms a enclosed circle, which considered as a "hole"):
image

Not expanding (since the hole in the vertical wall breaks the wall outline):
image

This behavior limits the applicable scenarios that this feature is usable. 
I think by adding an option to limit when the hole will receive the compensation might help.

Cura Options: https://github.com/Ultimaker/Cura/pull/9182

### DIFF
--- a/src/layerPart.cpp
+++ b/src/layerPart.cpp
@@ -55,6 +55,7 @@ void createLayerWithParts(const Settings& settings, SliceLayer& storageLayer, Sl
         result = layer->polygons.splitIntoParts(union_layers || union_all_remove_holes);
     }
     const coord_t hole_offset = settings.get<coord_t>("hole_xy_offset");
+    const coord_t max_hole_length = settings.get<coord_t>("hole_xy_offset_max_length");
     for(unsigned int i=0; i<result.size(); i++)
     {
         storageLayer.parts.emplace_back();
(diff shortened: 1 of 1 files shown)


rburema:
Just discussed this with colleagues, and while I can see the case for separate settings, we think it would be best to merge them at the current juncture.

The other behaviour that this is influencing will not be likely to make the print any worse. In worst case, if different parts of the model require different considerations, then modifier meshes can be used. Lastly, since all the settings are in experimental at the moment anyway, we have much more leeway to change it later if it later turns out that it _is_ often requested that they should be different.

(Another solution that could be considered is using the existing child-setting 'small_feature_max_length' for the slow-down, as it is now, and then using the parent-setting 'small_hole_max_size' for the expansion. Though I'm not sure how robust and maintainable that would be going forward.)

I was on the verge of doing this myself, but it seemed more polite to first discuss this.

alexjx:
Ah, sorry for replying so late. Please go ahead to implement they way you see proper. I'm just posting something that I think it's useful. It's just a simple patch. I will close this PR. Thanks for taking it. :)

PR #1398, opened 2021-01-25
upgrade to cpp 17

BagelOrb:
Use newer version of C++

Now we don't need to use our botched version of std::optional anymore and we can make use of so much more *standard* functionality!

"Don't let yourself be held back by needless barriers. Free yourself from the shackles of legacy. Advance!"


rburema:
Duplicate of #1273 

.... which is on hold because of the build-servers.

BagelOrb:
Is that also the reason for this build error? https://github.com/Ultimaker/CuraEngine/actions/runs/514847238

rburema:
> Is that also the reason for this build error? https://github.com/Ultimaker/CuraEngine/actions/runs/514847238

Hm, no I think that might be something else (the OS used here is a newer version of Ubuntu, not the ageing CentOS distro / old mac / old Win we use on the build-server).

P.S. I'll reply to the mini-review when I have more time (I want to check the docs to see if the move version of that operator would already take care of the complaint).

PR #1400, opened 2021-01-28
4.9 Add new Printer

Rupiert:
[Ender-6 3D Printer](https://www.creality.com/goods-detail/ender-6-3d-printer)


rburema:
@Rupiert Making a random pull request (which is a tool we use for coding the application), is _not_ the way to ask for a new printer. Especially not in the engine, since printer configurations are managed in the frontend. Furthermore, we rely on the community to provide definitions for 3rd party printers.

Information on how to add new machine definitions yourself can be found here: https://github.com/Ultimaker/Cura/wiki/Adding-new-machine-profiles-to-Cura

Other issues (bugs and feature requests), should be reported here: https://github.com/Ultimaker/Cura/issues

PR #1403, opened 2021-02-02
Fix Tree Support Floor layers missing every second layer.

### LINKED ISSUE #1402: Tree Support: Support Floor missing every second layer
**Application Version**
4.7.1

**Platform**
Windows 10

**Steps to Reproduce**
Slice a model with Tree Support and Support Floor enabled.

**Actual Results**
Every second layer does not contain a Support Floor

**Expected results**
Every layer up to Support Floor Thickness worth of layers should have a support floor.

**Additional Information**
grafik

PS. I found the issue and will link a pull request shortly.

ThomasRahm:
As unionPolygons does not work in place the result is not saved, which seems to cause issues down the line.
Should fix #1402

While fixing issues with my new tree support implementation, I found this issue, which is also present in the live version. So while I know that this fixes the issue I had, and that the issue is also present in the live version, I did not test that this fixes it in the live version. But the code is similar enough that I am pretty sure that it will behave identically here.

### DIFF
--- a/src/TreeSupport.cpp
+++ b/src/TreeSupport.cpp
@@ -176,7 +176,7 @@ void TreeSupport::drawCircles(SliceDataStorage& storage, const std::vector<std::
                 constexpr bool no_prime_tower = false;
                 floor_layer.add(support_layer.intersection(storage.getLayerOutlines(sample_layer, no_support, no_prime_tower)));
             }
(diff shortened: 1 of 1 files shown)


Ghostkeeper (approved):
Indeed. This was properly done for the support roof and support infill, but apparently not for the support floor. Good catch!

PR #1405, opened 2021-02-05
[CURA-7979] Lock outer wall in place and width as much as possible

rburema:
Outer walls react to what happens inside. While technically correct behaviour given the basic premise of Arachne, this can cause noticeable artefacts in what would otherwise be a straight wall. This would already be worthy of fixing, but it seems these artefacts can be exacerbated by the printing process, making it a at least a critical and probably even a blocker for release.

~~This commit sets up the basic premise. Add a parameter to adjust how close the outer walls are to their 'optimal' width (and, because they are outer walls therefore also placement, which is what is needed), with 1.0 meaning completely optimal (mostly ignore any width changes to the outer walls coming in, let the inner walls 'solve it as much as possible') and 0.0 being the old 'correct' behaviour (no compensation is applied to the previous strategy other than the 'relative widths for inner and outer' redistribution which was already happening).~~

~~(See also front-end PR: https://github.com/Ultimaker/Cura/pull/9245)~~

__Jelle says__:
I have been thinking really hard how a beading strategy could be applied to two different wall regions (inner/outer) and still be true to the strategy and the difference in wall width specified, and this is basically the only way I see how.

I think of having the "Lock Wall 0" feature always on (no check-box in the front-end) just normal behavior.

Rational behind this:

> If two different wall thicknesses are specified for outer and inner walls, the user wants that outer wall to have that dimension. Since the outer wall always consist of 1 wall there isn't anything to distribute. It either is the specified dimension or less if the thickness of the outline is less then (2 x the optimal outer wall width + 1 x minimal inner wall width).
Now it is entirely possible that the inner wall has more then 1 walls (even when 1 inner wall is specified, for the cases of converging walls) In this case the strategies define the distributed width.

lock

The gif above shows how the outer wall will no longer react to changes on the inside, until it arrives at thickness which is less then twice the optimal outer width and once the minimum inner width (see image below) It will then keep the third minimum wall constant till there is only space for two beads.

smaller_then

Note 1:
> I the divergence of the wall position in certain places is not part of this ticket (although I original thought it was, Sorry @rburema )  
This is more likely related to CURA-7853 and CURA-7854
location

Note 2:
> The center deviation  will have some left over space in case of an even number of walls, this is a separate bug imo and I will create a separate ticket for this. Then over/under extrusion which I found during my own review for the distributed strategies were a part of this ticket tho.
center_even2


jellespijker (dismissed):
The code style and the intention are clear.
But the placement of the center line of inner walls seems to suffer from an accumulating error, which is more evident in certain cases when the wall count is increased to a high enough number to fill the shape completely.

location

I also see a problem when handling the width of the inner central wall in those cases.
centerwall

tests/WallsComputationTest.cpp
```diff
@@ -62,6 +62,7 @@ class WallsComputationTest : public testing::Test
         settings.add("wall_transition_length", "1");
         settings.add("wall_transition_threshold", "50");
         settings.add("wall_x_extruder_nr", "0");
+        settings.add("wall_0_lock_factor", "0.5");
```
rburema, line 65: Remove unused setting completely.

src/BeadingStrategy/RedistributeBeadingStrategy.cpp
```diff
@@ -4,65 +4,126 @@
...
     {
-        Beading ret = parent->compute(thickness, bead_count);
+        const coord_t inner_transition_width = optimal_width_inner * minimum_variable_line_width;
+        const coord_t outer_bead_width =
+            getOptimalOuterBeadWidth(thickness, optimal_width_outer, inner_transition_width);
 
-        // Actual count and thickness as represented by extant walls. Don't count any potential zero-width 'signalling' walls.
-        bead_count = std::count_if(ret.bead_widths.begin(), ret.bead_widths.end(), [](const coord_t width) { return width > 0; });
-        thickness -= ret.left_over;
+        // Outer wall is locked in size en position for wall regions of 3 and higher which have at least a
```
rburema, line 20: 'size en position' -> 'size _and_ position'

rburema (commented):
Other than the cosmetic changes I propose in my review, I think the code looks good.

If my component test(s) pass(es) as well, I'll make the cosmetic changes myself (otherwise I will of course put it back into Todo).

PR #1408, opened 2021-02-09
CURA-7927: Fix extruders alternating each layer for inner and outer walls

konskarm:
When multiple extruders are used for a print, the `optimize()` function of the `InsetOrderOptimizer` is called once for every extruder. This function would then add all the wall insets for all extruders in the final layer data. As a result, the outer and inner walls were printed multiple times, once for each extruder, but Cura would render only the color of the first extruder used for each wall, leading to the following effect:

image

Why is this happening?

To optimize travelling (i.e. to make sure that the extruder that ends in the previous layer is the extruder that starts in the next layer), the list of extruders is iterated with a different direction on each consecutive layer in multi-extruder prints (e.g. layer 0: 0->1->2->3, layer 1: 3->2->1->0, layer 2: 0->1->2->3 etc.). This was leading to the print seeming like it alternates extruders in consecutive layers. In truth, it was just instructed to print all the walls with all the extruders.

This PR fixes that by making sure that each wall type (outer or inner) is printed only once with the extruder indicated by the user (the outer wall should be printed with the `wall_0_extruder_nr` and the inner wall with the `wall_x_extruder_nr`).

The PR also ensures that infill walls and skin walls will not take into account the setting "Outer Before Inner Walls" when the wall_0 and wall_x extruders are different (considering that the "Outer Before Inner Walls" setting is disabled in the front-end in that case). 

CURA-7927

### DIFF
--- a/src/InsetOrderOptimizer.cpp
+++ b/src/InsetOrderOptimizer.cpp
@@ -53,6 +53,7 @@ bool InsetOrderOptimizer::optimize(const WallType& wall_type)
     size_t start_inset;
     size_t end_inset;
     int direction;
+    //If the entire wall is printed with the current extruder, print all of it.
     if((wall_type == WallType::OUTER_WALL && wall_0_extruder_nr == wall_x_extruder_nr && wall_x_extruder_nr == extruder_nr) ||
             (wall_type == WallType::EXTRA_SKIN && extruder_nr == top_bottom_extruder_nr) ||
             (wall_type == WallType::EXTRA_INFILL && extruder_nr == infill_extruder_nr))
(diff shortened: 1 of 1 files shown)


src/InsetOrderOptimizer.cpp
```diff
@@ -36,31 +36,69 @@ bool InsetOrderOptimizer::optimize(const WallType& wall_type)
...
+        {
+            return added_something;
+        }
     }
     else
     {
-        start_inset = insets.size() - 1;
-        end_inset = -1;
-        direction = -1;
+        return added_something;
```
Ghostkeeper, line 98: Took quite some thought as to what each of these cases means now, so I'd like to add some documentation, in particular to this case.

Ghostkeeper (changes requested):

PR #1414, opened 2021-02-15
Properly round Voronoi coordinates to IntPoint coordinates

Ghostkeeper:
Instead of just casting them, actually round them. The Voronoi calculation has some minor inaccuracies in due to floating point rounding. If the coordinates are rounded down this can cause a vertex from the source polygon to change position, causing it to no longer be recognised as a vertex from the source polygon.

Contributes to issue CURA-7970. It is not a complete solution though.


Ghostkeeper:
Tests fail with the same error as what this is attempting to solve. Interesting. There are also still models left that were not fixed by this.

jellespijker (approved):

PR #1416, opened 2021-02-19
CURA-7952: Fix running into edges without a twin

konskarm:
The process of cleaning polygons to prepare them for SkeletalTrapezoidation sometimes would leave almost-colinear segments, which would lead to generating graph edges without a twin.

This PR fixes that by making sure that the `removeColinearSegments()` will properly cover all cases. It also increases the maximum_angle_deviation (i.e. the maximum angle allowed for two segments to be considered as colinear) when checking for colinearity.

Before:
stages_of_outline_cleaning_before

As it can be seen, the remove_colinear_edges actually leaves a point that is part of colinear segments, which cannot be handled by the generation of the graph and will lead to graph edge without a twin.

After:
stages_of_outline_cleaning_after

The spike mentioned above no longer exists in the output of `removeColinearEdges()`.

CURA-7952


Ghostkeeper:
Replaced by https://github.com/Ultimaker/CuraEngine/pull/1421.

PR #1419, opened 2021-02-23
Do not wipe for continuously linked paths

jellespijker:
The problem is currently that the walls are fed as small segments and
the PathOrderOptimizer ensures they are processed in a correct order and
it really doesn't matter that all segments are small and broken up.
Now all is well, you would think, But then the outer wipe distance comes
in to play. Add the end of a small wall segment it thinks: He, I need to
wipe. I will traverse all long the path a do a travel afterwards. This
occurs after each small wall segment. So when a wall is split in two
walls it does the outer wall wipe thingy.

By storing the end point of the last path and checking if this is
aligned with the start point of the new point I will know it this path
is part of a continuously linked path consisting of smaller segments.
If that is the case the outer wall wipe isn't performed.

Contributes to CURA-7949_outer_wall_wipe_distance

The current solution depends on the order that it provides from the 
PathOrderOptimizer that is why there might still be some cases where
a unnecessary wipe might occur. See gif below. I don't consider that part
of this ticket and I will raise at the eCCB.

Peek 2021-02-23 17-53

### DIFF
--- a/src/LayerPlan.cpp
+++ b/src/LayerPlan.cpp
@@ -740,8 +740,10 @@ void LayerPlan::addWall(ConstPolygonRef wall, int start_idx, const Settings& set
         ewall.emplace_back(p, nominal_line_width, dummy_perimeter_id);
     });
     ewall.emplace_back(*wall.begin(), nominal_line_width, dummy_perimeter_id);
-
-    addWall(ewall, start_idx, settings, non_bridge_config, bridge_config, wall_0_wipe_dist, flow_ratio, always_retract, true, false, false);
+    constexpr bool is_closed = true;
+    constexpr bool is_reversed = false;
+    constexpr bool is_linked_path = false;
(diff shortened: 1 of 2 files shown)


src/LayerPlan.cpp
```diff
@@ -741,10 +741,10 @@ void LayerPlan::addWall(ConstPolygonRef wall, int start_idx, const Settings& set
     });
     ewall.emplace_back(*wall.begin(), nominal_line_width, dummy_perimeter_id);
 
-    addWall(ewall, start_idx, settings, non_bridge_config, bridge_config, wall_0_wipe_dist, flow_ratio, always_retract, true, false);
+    addWall(ewall, start_idx, settings, non_bridge_config, bridge_config, wall_0_wipe_dist, flow_ratio, always_retract, true, false, false);
```
nallath, line 744: That's nasty booltrap. I don't know why it wasn't caught in the previous reviews, but 3 bools without context in a row is getting a bit out of hand.

I suggest swapping em out for constexpr bool statements.

PR #1420, opened 2021-02-23
[CURA-7946] fix concentric support interface

rburema:
Concentric support interface didn't show up in Arachne. Turns out if paths result from the output of the infill-comp (rather than just polygons or lines), we where ignoring it, instead of calling add walls on it.

This nescesitated a small refactor in AddWall(s) and the like, but since it's simpler now, it's all for the better.

**FUN FACT** _Did you know?_ Spiralize is the only reason we keep the 'old' AddWalls around. Could be deleted if we got rid of that!

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -2622,7 +2622,7 @@ bool FffGcodeWriter::addSupportRoofsToGCode(const SliceDataStorage& storage, Lay
     VariableWidthPaths roof_paths;
     Polygons roof_lines;
     roof_computation.generate(roof_paths, roof_polygons, roof_lines, roof_extruder.settings);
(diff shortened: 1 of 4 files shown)


jellespijker:
I was thinking about the spiralize just the other day. It really bugs me that we have to keep those insets just for this edge case.
To more I get to know the code, the more I think it should be fairly simple to replace spiralize with arachne generated wall.
We didn't do that at the time due to bugs in the path ordering, travel optimizer etc.
Those are all fixed now. Also the lock outer wall should keep the line fairly consistent.
It should be fairly simp to generate a limited bead of 1 for each layer and at a later stage steadily increase the z to the next layer height.

jellespijker (approved):

rburema:
> Those are all fixed now.

Go for it as one of those 'Technical Improvements'. Put it on the backlog :-)

PR #1422, opened 2021-02-24
Fix for bad bug that duplicates infill in skin support regions

smartavionics:
As discussed in https://github.com/smartavionics/Cura/issues/98.

I'm amazed this bug hasn't been spotted before now.

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -1618,7 +1618,7 @@ bool FffGcodeWriter::processSingleLayerInfill(const SliceDataStorage& storage, L
                                infill_angle, gcode_layer.z, infill_shift, max_resolution, max_deviation, skin_below_wall_count, infill_origin,
                                perimeter_gaps, connected_zigzags, use_endpieces, skip_some_zags, zag_skip_count,
                                pocket_size);
-            infill_comp.generate(infill_polygons, infill_lines_here, mesh.cross_fill_provider, &mesh);
+            infill_comp.generate(infill_polygons_here, infill_lines_here, mesh.cross_fill_provider, &mesh);
             if (density_idx < last_idx)
             {
                 const coord_t cut_offset =
@@ -1627,7 +1627,9 @@ bool FffGcodeWriter::processSingleLayerInfill(const SliceDataStorage& storage, L
                 infill_lines_here.cut(tool);
             }
             infill_lines.add(infill_lines_here);
+            infill_polygons.add(infill_polygons_here);
             infill_lines_here.clear();
+            infill_polygons_here.clear();
             // normal processing for the infill that isn't below skin
             in_outline = infill_not_below_skin;
             if (density_idx == last_idx)
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -1627,6 +1627,7 @@ bool FffGcodeWriter::processSingleLayerInfill(const SliceDataStorage& storage, L
                 infill_lines_here.cut(tool);
             }
             infill_lines.add(infill_lines_here);
+            infill_lines_here.clear();
             // normal processing for the infill that isn't below skin
             in_outline = infill_not_below_skin;
             if (density_idx == last_idx)


rburema:
Hi again! Wanted to have a look at this one as well today (still). While I'm still going to argue for it's inclusion regardless (after I do some testing -- since the code is more correct), I can't seem to reproduce the original problem reported to your repo either for our 4.8 or our current main. Perhaps I am looking at the wrong thing? I don't see any duplicated infill lines in that project in layer 111 is what I'm saying.

smartavionics:
It was definitely occurring when using my builds and I checked the UM master branch visually to see that the buggy code was there also so it should be affecting your builds as well. Maybe I goofed there. I will check again tomorrow.

rburema:
The _code_ is definitely still there, yes. I'm just not seeing the behaviour. That could be on me though.

smartavionics:
There's an easy way to check if the bug is occurring. See if the total filament distance changes when the same project is sliced with or without this PR in place. If the bug is there, the distance will increase.

rburema:
> If the bug is there, the distance will increase.

I did just that. No difference.

@jellespijker you refactored a large part of this code recently. Do you have any opinion on how to proceed? I think it can do little harm and might be more correct, despite not solving a bug in our current code base. O.t.o.h. I don't want to mess with things too much if there's no real problem.

jellespijker:
Yes I recently refactored a lot in this function. 
At a first glance (on my phone) I'm unsure about the clear() calls, since I seem to recall that multiply infill needs to store previous generated lines and polygons.
But that is just on top of my hat and without look at the code in the greater context.

I will take a look on Monday.

jellespijker:
I can also not observe this behavior, certainly not with the original project file in smartavionics/Cura#98

The proposed fix changes the infill computations for infill which is positioned below the skin and needs an edge support layer. It is after all in the scope of the `if (hasSkinEdgeSupport)`  and not in the general Infill computation done on line 1650.  The original project file sets the skin edge support layers to 0. meaning that that specific infill computation isn't called.

I think that the original buggy behavior has been taken care of when I worked on CURA-7115 **Gradual Infill + Connect Infill = Overextrusion**

@rburema I agree with your reluctance to change this piece of code, certainly because these polygons and multiple infill instances are used by gradual infill, infill support, skin edge support layer, connected lines, connected polygons, and multiple features that rely on previously generated lines and polygons. Certainly, if the buggy behavior can't be replicated.

smartavionics:
Just a thought, try not calling gcode_layer.optimizePaths() in FffGcodeWriter::processLayer() and see if the duplicate infill appears. It may be the optimization removes the duplicates? I don't use that function in my builds so that could be why I am seeing the duplicates and you are not.

smartavionics:
I tried enabling the optimizePaths() function and it made no difference so ignore that idea.

jellespijker:
I.m.o. it basically boils down to the question: _Is it a problem?_  
Since both @rburema and I can't reproduce the buggy behaviour, I'm inclined to say that we're trying to solve something here that doesn't need to be solved at this time.

smartavionics:
Well, I guess time will tell...

smartavionics:
I am still convinced there's a real problem here. I just tried using the current master branch of the CuraEngine (that's your master branch not my variant) and the issue is present without this PR and gone away with the PR. I can't explain why you are not seeing that.

jellespijker:
@smartavionics could you share the project file, that shows this behavior? 

Because I'm really not seeing it and the fact that you do makes me a bit nervous and if it really is a bug that it should most certainly be fixed.

That being said I still find it strange that this fix targets the Infill instance used in creating the skin edge support area's. I would expect to be targeted at the Infill instance used for the "regular" infill.

smartavionics:
Ah, in Infill::_generate(), if zig_zagify is true, for certain infill patterns, the result polygon is cleared first so I guess that's why you aren't seeing the problem.

smartavionics:
Nope, it looks like that isn't it either, still looking...

jellespijker:
@smartavionics you're persistent, I will certainly give you that ;-) 

I will give it a deeper look tomorrow if my agenda allows it. Otherwise, it will be the day after that.

Ghostkeeper:
The bug does reproduce in Cura 4.8, but not just with the project file. The code changes made here are in a piece of code that only applies if Skin Edge Support is used, but it's set to 0 layers in that project file. Setting Skin Edge Support Layers to 1 reproduces the issue in layer 111.

The same doesn't apply for the Arachne branch because there the `infill_lines_here` variable doesn't get filled at all. Instead, the cut works by the segmentation of the infill area into `infill_below_skin` and `infill_not_below_skin`. The `infill_below_skin` is filled directly into `infill_lines` (not `infill_lines_here` any more) and the rest is printed when it generates the normal infill.

jellespijker:
Thank you @Ghostkeeper for picking this up.

PR #1426, opened 2021-03-02
Cura 7756 infill layer thickness

jellespijker:
The problem was two fold:

1. ~~The wall toolpaths were generated in the function `processMultiLayerInfill` but never written to the GCode~~
2.  Wall toopaths should always be one layer in height with the optimal line width. But the lines should be a the intersect of the multiple layers. Since the area needed by the lines is depended on the inner contour of the wall toolpaths and these toolpaths should follow the normal infill area per layer. The generation of the infill walls for the sparsest area should be performed before the combining of the layers. The generation of the walls them self should then be removed from the processSingleLayerInfill and processMultiLayerInfill functions.

The function generateWalltoolpaths is introduced and is also used by the normal generate function. This ensures that other `Infill` class usage doesn't break with the regular interface.

> Note!
> setting the infill layer height to a multitude of the regular height and setting the skin edge support layers to a number higher then 1 does not generate that number of skin edge support layers. This also happens in master and I consider it a separate bug and not part of this one. I will raise it at the eCCB for tomorrow.

### DIFF
--- a/src/infill.cpp
+++ b/src/infill.cpp
@@ -43,19 +43,14 @@ static inline int computeScanSegmentIdx(int x, int line_width)
 
 namespace cura {
 
-void Infill::generate(VariableWidthPaths& toolpaths, Polygons& result_polygons, Polygons& result_lines, const Settings& settings, const SierpinskiFillProvider* cross_fill_provider, const SliceMeshStorage* mesh)
+Polygons& Infill::generateWalltoolpaths(VariableWidthPaths& toolpaths, const size_t number_of_walls, const coord_t wall_line_width, const Settings& settings)
 {
-    // generate walls
-    if (outer_contour.empty())
-    {
-        return;
-    }
     outer_contour = outer_contour.offset(infill_overlap);
 
-    if (wall_line_count > 0)
+    if (number_of_walls > 0)
     {
         constexpr coord_t wall_0_inset = 0; //Don't apply any outer wall inset for these. That's just for the outer wall.
-        WallToolPaths wall_toolpaths(outer_contour, infill_line_width, wall_line_count, wall_0_inset, settings);
+        WallToolPaths wall_toolpaths(outer_contour, wall_line_width, number_of_walls, wall_0_inset, settings);
         wall_toolpaths.pushToolPaths(toolpaths);
         inner_contour = wall_toolpaths.getInnerContour();
     }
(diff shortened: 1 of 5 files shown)


src/infill.cpp
```diff
@@ -43,26 +43,33 @@ static inline int computeScanSegmentIdx(int x, int line_width)
 
 namespace cura {
 
-void Infill::generate(VariableWidthPaths& toolpaths, Polygons& result_polygons, Polygons& result_lines, const Settings& settings, const SierpinskiFillProvider* cross_fill_provider, const SliceMeshStorage* mesh)
+Polygons Infill::generateWalltoolpaths(VariableWidthPaths& toolpaths, Polygons& outer_contour, const size_t wall_line_count, const coord_t line_width, const coord_t infill_overlap, const Settings& settings)
```
konskarm, line 46:
```suggestion
Polygons Infill::generateWallToolPaths(VariableWidthPaths& toolpaths, Polygons& outer_contour, const size_t wall_line_count, const coord_t line_width, const coord_t infill_overlap, const Settings& settings)
```

src/infill.cpp
```diff
@@ -43,26 +43,33 @@ static inline int computeScanSegmentIdx(int x, int line_width)
...
+}
+
+void Infill::generate(VariableWidthPaths& toolpaths, Polygons& result_polygons, Polygons& result_lines, const Settings& settings, const SierpinskiFillProvider* cross_fill_provider, const SliceMeshStorage* mesh)
+{
+    if (outer_contour.empty())
+    {
+        return;
+    }
+
+    inner_contour = generateWalltoolpaths(toolpaths, outer_contour, wall_line_count, infill_line_width, infill_overlap, settings);
```
konskarm, line 72:
```suggestion
    inner_contour = generateWallToolPaths(toolpaths, outer_contour, wall_line_count, infill_line_width, infill_overlap, settings);
```

src/infill.h
```diff
@@ -105,6 +105,19 @@ class Infill
...
+     *
+     * \param toolpaths [out] The generated toolpaths
+     * \param outer_contour [in,out] the outer contour, this is offsetted with the infill overlap
+     * \param wall_line_count [in] The number of walls that needs to be generated
+     * \param line_width [in] The optimum wall line width of the walls
+     * \param infill_overlap [in] The overlap of the infill
+     * \param settings [in] A settings storage to use for generating variable-width walls.
+     * \return The inner contour of the wall toolpaths
+     */
+    static Polygons generateWalltoolpaths(VariableWidthPaths& toolpaths, Polygons& outer_contour, const size_t wall_line_count, const coord_t line_width, const coord_t infill_overlap, const Settings& settings);
```
konskarm, line 120:
```suggestion
    static Polygons generateWallToolPaths(VariableWidthPaths& toolpaths, Polygons& outer_contour, const size_t wall_line_count, const coord_t line_width, const coord_t infill_overlap, const Settings& settings);
```

src/skin.cpp
```diff
@@ -461,7 +465,7 @@ void SkinInfillAreaComputation::generateGradualInfill(SliceMeshStorage& mesh)
         {
             assert((part.infill_area_per_combine_per_density.empty() && "infill_area_per_combine_per_density is supposed to be uninitialized"));
 
-            const Polygons& infill_area = part.getOwnInfillArea();
+            const Polygons& infill_area = Infill::generateWalltoolpaths(part.infill_wall_toolpaths, part.getOwnInfillArea(), infill_wall_count, infill_wall_width, infill_overlap,  mesh.settings);
```
konskarm, line 468:
```suggestion
            const Polygons& infill_area = Infill::generateWallToolPaths(part.infill_wall_toolpaths, part.getOwnInfillArea(), infill_wall_count, infill_wall_width, infill_overlap,  mesh.settings);
```

konskarm (changes requested):
Minor naming issue, change the `generateWalltoolpaths` to `generateWallToolPaths` so that it is consistent with the rest of the code.

PR #1427, opened 2021-03-02
Don't reuse unused extruder plans as they will have the wrong fan and retraction settings.

smartavionics:
Here's the fix for https://github.com/Ultimaker/Cura/issues/9301.

### DIFF
--- a/src/LayerPlan.cpp
+++ b/src/LayerPlan.cpp
@@ -288,15 +288,8 @@ bool LayerPlan::setExtruder(const size_t extruder_nr)
             addTravel(end_pos); //  + extruder_offset cause it
         }
     }
-    if (extruder_plans.back().paths.empty() && extruder_plans.back().inserts.empty())
-    { // first extruder plan in a layer might be empty, cause it is made with the last extruder planned in the previous layer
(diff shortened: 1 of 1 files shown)


rburema:
As always, thanks for submitting (this) PR(s).

While working on this, with your solution, I noticed that T1 only gets an M104 S0 at the very end of the file, instead of at the position in the screenshot. I don't think there are any other temperature commands the second extruder receives when T0 is printing the mesh, so that would mean T1 stays at printing temperature while idle.

fanspeed_pr_situation

When I put the if-statement back, but don't put the rest in the else, the third variation happens. A bit verbose and unnecessary, but I don't want to experiment with what happens if I just outright delete the empty layer right now.

It's getting late right now, so I thought I might as well throw it out here before I continue some time hopefully tomorrow.

rburema:
Alright, did some tests and merged. It's up to QA now if they want to send it back or not.

Thanks!

smartavionics:
Putting that if statement back in and updating the extruder_nr looks bogus to me because then two elements in the extruder_plans vector have the same extruder_nr. Even if there's no bad side effect of that, it looks dodgy.

PR #1429, opened 2021-03-03
[CURA-8060] Fix combine layers (vertically) for support as well.

rburema:
Layers for support wouldn't combine vertically where this was needed, like for PVA printing. This relies on the work done in jira ticket CURA-7756 where almost, but not quite, the same thing happens for infill. In fact, it relies on that PR for how the infill generation class now has promoted the wall generating part to a publicly callable function, as one of the problems of both this ticket and its precursor was that (extra) walls _shouldn't_ combine. (And the 'normal' lines of course _should_ that's the whole point.)

See also:  #1426  --- please merge that one first!

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -2497,7 +2497,7 @@ bool FffGcodeWriter::processSupportInfill(const SliceDataStorage& storage, Layer
             const coord_t support_line_width = default_support_line_width * (combine_idx + 1);
 
             Polygons support_polygons;
-            VariableWidthPaths dummy_wall_toolpaths;
+            VariableWidthPaths wall_toolpaths = part.wall_toolpaths;
             Polygons support_lines;
             const size_t max_density_idx = part.infill_area_per_combine_per_density.size() - 1;
             for (size_t density_idx = max_density_idx; (density_idx + 1) > 0; --density_idx)
@@ -2519,21 +2519,22 @@ bool FffGcodeWriter::processSupportInfill(const SliceDataStorage& storage, Layer
                                                         gcode_layer.getLayerNr(), support_pattern, support_line_width,
                                                         support_brim_line_count);
 
+                constexpr size_t wall_count = 0; // Walls are generated somewhere else, so their layers aren't vertically combined.
                 Infill infill_comp(support_pattern, zig_zaggify_infill, connect_polygons, area,
                                    support_line_width, support_line_distance_here, current_support_infill_overlap - (density_idx == max_density_idx ? 0 : wall_line_count * support_line_width),
                                    infill_multiplier, support_infill_angle, gcode_layer.z, support_shift,
                                    max_resolution, max_deviation,
-                                   0 /*walls are generated somewhere else, so their layers aren't vertically combined*/, infill_origin, support_connect_zigzags,
+                                   wall_count, infill_origin, support_connect_zigzags,
                                    use_endpieces, skip_some_zags, zag_skip_count, pocket_size);
-                infill_comp.generate(dummy_wall_toolpaths, support_polygons, support_lines, infill_extruder.settings, storage.support.cross_fill_provider);
+                infill_comp.generate(wall_toolpaths, support_polygons, support_lines, infill_extruder.settings, storage.support.cross_fill_provider);
             }
 
             setExtruder_addPrime(storage, gcode_layer, extruder_nr); // only switch extruder if we're sure we're going to switch
             gcode_layer.setIsInside(false); // going to print stuff outside print object, i.e. support
 
-            if (! part.wall_toolpaths.empty())
+            if (! wall_toolpaths.empty())
             {
-                const BinJunctions bins = InsetOrderOptimizer::variableWidthPathToBinJunctions(part.wall_toolpaths);
+                const BinJunctions bins = InsetOrderOptimizer::variableWidthPathToBinJunctions(wall_toolpaths);
                 for (const PathJunctions& paths : bins)
                 {
                     for (const LineJunctions& line : paths)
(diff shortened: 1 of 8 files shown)


src/FffGcodeWriter.cpp
```diff
@@ -2498,7 +2497,7 @@ bool FffGcodeWriter::processSupportInfill(const SliceDataStorage& storage, Layer
             const coord_t support_line_width = default_support_line_width * (combine_idx + 1);
 
             Polygons support_polygons;
-            VariableWidthPaths wall_toolpaths;
+            VariableWidthPaths dummy_wall_toolpaths;
```
jellespijker, line 2500: I would assign the pre-generated wall_toolpaths to this variable. Since the `generate()` still makes some use of them.
so in effect `VariableWidthPaths wall_toolpaths = part.wall_toolpaths`

src/FffGcodeWriter.cpp
```diff
@@ -2524,22 +2523,22 @@ bool FffGcodeWriter::processSupportInfill(const SliceDataStorage& storage, Layer
                                    support_line_width, support_line_distance_here, current_support_infill_overlap - (density_idx == max_density_idx ? 0 : wall_line_count * support_line_width),
                                    infill_multiplier, support_infill_angle, gcode_layer.z, support_shift,
                                    max_resolution, max_deviation,
-                                   (density_idx == max_density_idx ? wall_line_count : 0), infill_origin, support_connect_zigzags,
+                                   0 /*walls are generated somewhere else, so their layers aren't vertically combined*/, infill_origin, support_connect_zigzags,
```
jellespijker, line 2526: `constexpr` the wall line count of 0

src/FffGcodeWriter.cpp
```diff
@@ -2524,22 +2523,22 @@ bool FffGcodeWriter::processSupportInfill(const SliceDataStorage& storage, Layer
                                    support_line_width, support_line_distance_here, current_support_infill_overlap - (density_idx == max_density_idx ? 0 : wall_line_count * support_line_width),
```
jellespijker, line 2523: Dubbel check if the support infill overlap is correct. This infill overlap is also used when generating the wall toolpaths and the contour for the lines (used here) is generated from that "offset" inner_contour.

src/FffGcodeWriter.cpp
```diff
@@ -2524,22 +2523,22 @@ bool FffGcodeWriter::processSupportInfill(const SliceDataStorage& storage, Layer
...
                                    use_endpieces, skip_some_zags, zag_skip_count, pocket_size);
-                infill_comp.generate(wall_toolpaths, support_polygons, support_lines, infill_extruder.settings, storage.support.cross_fill_provider);
+                infill_comp.generate(dummy_wall_toolpaths, support_polygons, support_lines, infill_extruder.settings, storage.support.cross_fill_provider);
             }
 
             setExtruder_addPrime(storage, gcode_layer, extruder_nr); // only switch extruder if we're sure we're going to switch
             gcode_layer.setIsInside(false); // going to print stuff outside print object, i.e. support
 
-            if (!wall_toolpaths.empty())
+            if (! part.wall_toolpaths.empty())
```
jellespijker, line 2534: If you assign the part.wall_toolpaths to the wall_toolpaths as suggested above. Don't forget to rollback these ;-)

jellespijker (changes requested):

src/FffGcodeWriter.cpp
rburema: Yeah I wanted to ask about that ... why a copy and not a reference?

src/FffGcodeWriter.cpp
rburema: How is that different?

src/FffGcodeWriter.cpp
rburema: If I do that, I won't forget ;-)
... I hope.

src/FffGcodeWriter.cpp
jellespijker: I think it I did that because this function was const, and the generate could modify the wall tool paths.
Might be worth looking into if we can use reference

src/FffGcodeWriter.cpp
jellespijker: Readability there are a lot of arguments in the function generate. Most IDE display the argument names but not all.
Can't hurt

src/FffGcodeWriter.cpp
rburema: What function is const? part.(infill_)wall_toolpaths isn't? combineSupportInfillLayers isn't?

src/FffGcodeWriter.cpp
jellespijker: I thought so I have high regards for your mental capabilities

src/FffGcodeWriter.cpp
jellespijker: No but the FFGodewriter processSingleLayerInfill and processMultiLayerInfill is.

jellespijker: quick call?

src/FffGcodeWriter.cpp
rburema: Haha, already decided to fix it as 'readability' after all before reading this comment :-)

src/FffGcodeWriter.cpp
rburema: Turns out I can't do it either :-p

src/FffGcodeWriter.cpp
rburema: Turns out I didn't forget.

jellespijker (approved):
Yet again you managed to get my stamp of approval. Once @nallath gives his approval to the changes made in PR #1426 I will merge this!

PR #1444, opened 2021-04-26
Truncated octahedron infill

gringer:
This is the backend code for 3D Honeycomb / Truncated Octahedron infill, as seen in Slic3r (but not [yet] PrusaSlicer). The infill pattern only bridges once per "module", at the smallest bridge point on the square horizontal faces (with a bridge span of approximately [line_distance / 2]). This produces a strong infill pattern which should work at very low infill percentages.

Note that this is only the backend code. It will need to be bolted onto the frontend interface. To make that easier, I've tried to make sure that the public function syntax is exactly the same as the Gyroid infill, so hopefully it should just be a matter of copying the Gyroid definitions, and replacing "Gyroid" with "Troct".

See the following image, taken from my fork of PrusaSlicer, demonstrating a 5% infill with no walls, where I've changed the colour at each direction change. I've been having trouble testing this code (see below); this image is based onthe PrusaSlicer equivalent of the code in this branch:

density_5pct

In other words, this code, in its current form, probably won't work. I'm hoping that someone with Cura building capabilities can test it, or help me build a working test setup on my Debian system.

I have also attempted to tweak the density calculation to make it more accurate. Assuming I've got the right density calculation (i.e. [infill_line_width / line_distance]), it should have a similar accuracy to the gyroid and rectilinear infills.

Note: this is essentially the same change I've opened a pull request for in PrusaSlicer. There are some additional minor bug fixes due to pattern start/end points, but probably more introduced bugs due to the porting process.

p.s. I'm having a really hard time getting the Lulzbot Cura GUI to compile on my computer, and the test scripts for CuraEngine aren't working for me. Here's an example that I tried to run, and its output:

```
grinja@hera:~/install/CuraEngine$ ./build/CuraEngine slice -v -s machine_extruder_count="1" -j ../Cura/resources/definitions/fdmextruder.def.json -j ../Cura/resources/extruders/prusa_i3_mk3_extruder_0.def.json -j ../Cura/resources/definitions/prusa_i3_mk3.def.json -o "output/test.gcode" -l "/model_1.stl"

Cura_SteamEngine version master
Copyright (C) 2020 Ultimaker

This program is free software: you can redistribute it and/or modify
it under the terms of the GNU Affero General Public License as published by
the Free Software Foundation, either version 3 of the License, or
(at your option) any later version.

This program is distributed in the hope that it will be useful,
but WITHOUT ANY WARRANTY; without even the implied warranty of
MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
GNU Affero General Public License for more details.

You should have received a copy of the GNU Affero General Public License
along with this program.  If not, see <http://www.gnu.org/licenses/>.
[ERROR] Couldn't find definition file with ID: fdmextruder
[ERROR] Inherited JSON file "fdmextruder" not found.
[ERROR] Failed to load JSON file: ../Cura/resources/extruders/prusa_i3_mk3_extruder_0.def.json

```
I don't need a GUI; I just need some way to test out this pattern and produce GCode.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -182,6 +182,7 @@ set(engine_SRCS # Except main.cpp.
     src/infill/SierpinskiFillProvider.cpp
     src/infill/SubDivCube.cpp
     src/infill/GyroidInfill.cpp
+    src/infill/TroctInfill.cpp
 
     src/pathPlanning/Comb.cpp
     src/pathPlanning/GCodePath.cpp
--- a/src/infill/TroctInfill.cpp
+++ b/src/infill/TroctInfill.cpp
@@ -1,0 +1,237 @@
+//Copyright (c) 2021 David Eccles (gringer)
+//CuraEngine is released under the terms of the AGPLv3 or higher.
+
+/*
+Creates a contiguous sequence of points at a specified height that make
+up a horizontal slice of the edges of a space filling truncated
+octahedron tesselation. The octahedrons are oriented so that the
+square faces are in the horizontal plane with edges parallel to the X
+and Y axes.
+
+Note: this code is derived from David Eccles' truncated octahedron
+infill (3D honeycomb) from Slic3r.
+*/
+
+#include "TroctInfill.h"
+#include "../utils/AABB.h"
+#include "../utils/linearAlg2D.h"
+#include "../utils/polygon.h"
+
+namespace cura {
+
+TroctInfill::TroctInfill() {
+}
+
+TroctInfill::~TroctInfill() {
+}
+
+// sign function
+template <typename T> int sgn(T val) {
+  return (T(0) < val) - (val < T(0));
+}
+  
+// triangular wave function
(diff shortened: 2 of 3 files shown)


rburema:
While we're not completely convinced by the presumed advantages you list (more rigid and printable at lower density are doubtful for instance), the actual big sticking point is that this isn't fully integrated (not just with the front-end, also with the engine parts that handle infill) or tested in any way at all (either just the unit-tests we've made for infill or printed in real life).

To put it bluntly: Basically it just compiles, and that's it. We're not going to put in the extra work ourselves (given all other tickets/issues/etc. that require attention).

In other words: We can help you by answering questions and pointing you to documentation, but unless this becomes a complete package that we can just (code) review, test and merge (if all is well), we're not likely to pick this up.

Joeydelarago:
This PR appears to be abandoned, closing.

PR #1457, opened 2021-06-18
Add option to print centre wall last

Ghostkeeper:
This implementation adds a new option for the order in which walls are printed. Instead of just printing from inside to outside, or from outside to inside, it is now also possible to print from inside to outside, except to print the centre lines last.

If selected, the printer will first complete all the doubled-up / looped walls:
image

... and then print the odd-numbered centre walls afterwards in a separate pass:
image

The benefit in print quality of this has not yet been tested, only theorised. There are two theoretical advantages to this option:
* The odd-numbered centre walls are thought to have a problem with overextrusion, moreso than other walls (although why they overextrude is not yet known). This overextrusion propagates towards the outside, causing the surface to become more irregular.
* Printing all loops first ensures that loops are closed all at once, reducing the number of short travel moves during the printing of those loops in some cases, in favour of longer travel moves towards the end.

The new option is a drop-down called Wall Ordering. It replaces the previous setting "Outer Before Inner Walls" which was always a bit of a difficult name for a setting, and then adds this as a third option to what was formerly a boolean setting.
image

This implementation depends on a change to the list of settings, which is made here: https://github.com/Ultimaker/Cura/pull/9997

Implements CURA-8109.

### DIFF
--- a/src/InsetOrderOptimizer.cpp
+++ b/src/InsetOrderOptimizer.cpp
@@ -165,7 +165,7 @@ size_t InsetOrderOptimizer::getOuterRegionId(const VariableWidthPaths& toolpaths
     return outer_region_id;
 }
 
-BinJunctions InsetOrderOptimizer::variableWidthPathToBinJunctions(const VariableWidthPaths& toolpaths, const bool& ignore_inner_inset_order, const bool& pack_regions_by_inset, std::set<size_t>* p_bins_with_index_zero_insets)
+BinJunctions InsetOrderOptimizer::variableWidthPathToBinJunctions(const VariableWidthPaths& toolpaths, const bool ignore_inner_inset_order, const bool pack_regions_by_inset, std::set<size_t>* p_bins_with_index_zero_insets)
 {
     // Find the largest inset-index:
     size_t max_inset_index = 0;
--- a/src/InsetOrderOptimizer.cpp
+++ b/src/InsetOrderOptimizer.cpp
@@ -1,4 +1,4 @@
-//Copyright (c) 2020 Ultimaker B.V.
+//Copyright (c) 2021 Ultimaker B.V.
 //CuraEngine is released under the terms of the AGPLv3 or higher.
 
 #include "ExtruderTrain.h"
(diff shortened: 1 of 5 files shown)


src/InsetOrderOptimizer.h
```diff
@@ -77,19 +77,15 @@ class InsetOrderOptimizer
...
      * \return A bin of walls, consisting of a vector of paths consisting of vector of lines
      */
-    static BinJunctions variableWidthPathToBinJunctions
-    (
-        const VariableWidthPaths& toolpaths,
-        const bool& ignore_inner_inset_order = false,
-        const bool& pack_regions_by_inset = true,
-        std::set<size_t>* p_bins_with_index_zero_insets = nullptr
-    );
+    static BinJunctions variableWidthPathToBinJunctions(
```
rburema, line 84: not in a proper review yet, but brackets on the same line, when the statement is multiline is a Python-ism.

rburema: ... but the rest looks good so I'll change this one cosmetic thing myself.

src/InsetOrderOptimizer.cpp
```diff
@@ -197,20 +197,17 @@ BinJunctions InsetOrderOptimizer::variableWidthPathToBinJunctions(const Variable
...
             const bool in_hole_region = line.region_id != outer_region_id && line.region_id != 0;
-            if(ignore_inner_inset_order)
+            if(center_last && line.is_odd)
+            {
+                bin_index = 0;
+            }
+            else if(pack_regions_by_inset)
             {
-                bin_index = std::min(inset_index, static_cast<size_t>(1)) + 2 * (in_hole_region ? line.region_id : 0);
+                bin_index = std::min(inset_index, static_cast<size_t>(1)) + 2 * (in_hole_region ? line.region_id : 0) + center_last;
```
rburema, line 206: Side Note: Though I'm not sure where we stand (as a team or company) re:code-style on implicit bool conversion like you did with `... + center_last`, I personally think it's fine. After all, it's both readable, and _explicitly_ in the C++ standard that `false = 0` and `true = 1` :-)

Ghostkeeper: I'm fine with changing this as well. I agree it's a bit hard to read.

PR #1460, opened 2021-06-30
[CURA-8354] Change the actual width instead of a copy.

rburema:
Also optimize as well.

Want to have another dev. have a look at this as well and maybe make a unit test, so I'll make this a draft for now.


rburema:
Connected to internal ticket: CURA-8354

nallath (approved):

Issue #1458, opened 2021-06-25
Feature request: unretract inside region before printing outer wall


richfelker:
When starting an outer wall - particularly, with `outer_inset_first`, and particularly on an overhang - the unretract can cause warts on the outer surface. Provided the region is more than the nozzle width thick at the starting point, a point inset just slightly inside the outer wall (preferably by a configurable amount, default to half nozzle width or so) could be used to perform the unretract, followed by travel to the real starting point to start extrusion. This is sort of an "inverse wipe".

Ghostkeeper:
It's expected to already do this. That's what this little hook is for when travelling to the outer wall.
![image](https://user-images.githubusercontent.com/2448634/124446959-371b7c80-dd81-11eb-8b30-d505b0554e47.png)

However when I test this, it appears like it's unretracting too late. It's supposed to be unretracting while away from the outer wall, but it seems like that unretract happens one move later when it actually starts to print.

richfelker:
So the intended behavior is what I described, but there seems to be a bug preventing it from actually doing what was intended here?

Ghostkeeper:
Indeed. When moving to print the outer wall, it first finds a spot next to the outer wall where it can unretract. It should move there instead, unretract, and then move to the start of the outer wall. From what I can see it currently does move there, then move to the outer wall, and then unretract, which is wrong. Probably the code to unretract is missing and then automatically filled in when it actually starts printing the outer wall.

The spot near the outer wall does have a limit for distance though, I think it's like 2mm. So it won't function for thin walls (where everything is outer wall) and very sharp tips.

Ghostkeeper:
Developers, see issue CURA-8365.

richfelker:
> The spot near the outer wall does have a limit for distance though, I think it's like 2mm

Do you mean there needs to be space to move back 2mm into the shape? That's large enough that it wouldn't work for most models I print. 1-1.5 nozzle widths should be plenty, and even half a nozzle width should be able to achieve the desired goal most of the time. 2mm sounds like way too much; it would allow oozing between the unretract and the wall start, yielding a weak wall.

I think the ideal behavior should be tunable, by default trying to go in 1 nozzle width, but going less (e.g. just half that if the entire region is 2 outer walls butted against each other) if the default would cross an opposing outer wall. This is probably computed by insets, so it would be something like "find a point within distance (default: nozzle width) in the 1 nozzle width inset, repeat with reduced inset if none exist".

[...]

PR #1474, opened 2021-07-28
CURA-8365: Unretract prior to the last travel move before an outer wall

konskarm:
When moving to print an outer wall, the printer should not unretract while the nozzle is already on top of the outer wall. Unretracting there can cause dimensional accuracy issues and visual artefacts since the nozzle is sitting on top of the outer wall while unretracting and melting the material, which can leave a blip.

Instead, the printer should unretract before the last travel move when going to print an outer wall. After the unretraction is complete, it can apply the last travel move and start printing the outer wall.

The unretraction can happen when combing decides that it should add extra travel moves when moving from the object A (left cube) to the outer wall of object B (right cube).

**Before:** Unretraction happens while the nozzle is already on top of the outer wall
```
;MESH:cube.stl
G0 F9000 X160.587 Y124.43
G0 X158.115 Y124.43
G0 X158.46 Y124.825
M204 S500
M205 X5 Y5
;TYPE:WALL-OUTER
G1 F1500 E30.72717 <-- Unretraction
G1 F1200 X148.81 Y124.825 E30.78011
G1 X148.81 Y115.175 E30.83306
```
image

**After:** The unretraction happens before the last travel move that would lead to the outer wall starting point
```
;MESH:cube.stl
G0 F9000 X160.587 Y124.43
G0 X158.115 Y124.43
G1 F1500 E30.72717 <-- Unretraction
G0 F9000 X158.46 Y124.825
M204 S500
M205 X5 Y5
;TYPE:WALL-OUTER
G1 F1200 X148.81 Y124.825 E30.78011
G1 X148.81 Y115.175 E30.83306
```
image

Fixes https://github.com/Ultimaker/CuraEngine/issues/1458

CURA-8365


src/LayerPlan.h
```diff
@@ -477,7 +477,7 @@ class LayerPlan : public NoCopy
      * \param p The point to travel to.
      * \param force_comb_retract Whether to force a retraction to occur.
      */
-    GCodePath& addTravel(const Point p, const bool force_retract = false);
+    GCodePath& addTravel(const Point p, const bool force_retract = false, const bool unretract_before_last_travel_move = false);
```
nallath, line 480: Documentation should also be updated (it's already outdated, since it mentiones force_comb_retract)

nallath:
Would it also be possible to add a test for the new behaviour? It seems like something that can be unit tested to a degree.

src/FffGcodeWriter.cpp
```diff
@@ -2064,12 +2064,12 @@ bool FffGcodeWriter::processInsets(const SliceDataStorage& storage, LayerPlan& g
...
+                            gcode_layer.addWalls(outer_wall, mesh, mesh_config.inset0_config, mesh_config.bridge_inset0_config, wall_overlap_computation, z_seam_config, mesh.settings.get<coord_t>("wall_0_wipe_dist"), flow, retract_before_outer_wall, /*unretract_before_last_travel_move = */true);
                         }
                         else
                         {
                             WallOverlapComputation wall_overlap_computation(outer_wall, mesh_config.inset0_config.getLineWidth());
-                            gcode_layer.addWalls(outer_wall, mesh, mesh_config.inset0_config, mesh_config.bridge_inset0_config, &wall_overlap_computation, z_seam_config, mesh.settings.get<coord_t>("wall_0_wipe_dist"), flow, retract_before_outer_wall);
+                            gcode_layer.addWalls(outer_wall, mesh, mesh_config.inset0_config, mesh_config.bridge_inset0_config, &wall_overlap_computation, z_seam_config, mesh.settings.get<coord_t>("wall_0_wipe_dist"), flow, retract_before_outer_wall, /*unretract_before_last_travel_move = */true);
```
jellespijker, line 2072: whynot make it a `constexpr bool unretract_before_last_travel_move = True` and add that variable. Would technically be the same but more readable and not such a long line. If you add if above the `if` statement you can use it for both calls

src/InsetOrderOptimizer.cpp
```diff
@@ -283,7 +283,7 @@ void InsetOrderOptimizer::processHoleInsets()
             {
                 if (extruder_nr == mesh.settings.get<ExtruderTrain&>("wall_0_extruder_nr").extruder_nr)
                 {
-                    gcode_layer.addWalls(hole_outer_wall, mesh, mesh_config.inset0_config, mesh_config.bridge_inset0_config, wall_overlapper_0, z_seam_config, wall_0_wipe_dist, flow, retract_before_outer_wall);
+                    gcode_layer.addWalls(hole_outer_wall, mesh, mesh_config.inset0_config, mesh_config.bridge_inset0_config, wall_overlapper_0, z_seam_config, wall_0_wipe_dist, flow, retract_before_outer_wall, /*unretract_before_last_travel_move = */true);
```
jellespijker, line 286: See previous comment about `constexpr`

src/InsetOrderOptimizer.cpp
```diff
@@ -311,7 +311,7 @@ void InsetOrderOptimizer::processHoleInsets()
                 gcode_layer.addWalls(hole_inner_walls, mesh, mesh_config.insetX_config, mesh_config.bridge_insetX_config, wall_overlapper_x);
                 if (extruder_nr == mesh.settings.get<ExtruderTrain&>("wall_0_extruder_nr").extruder_nr)
                 {
-                    gcode_layer.addWall(hole_outer_wall[0], outer_poly_start_idx, mesh, mesh_config.inset0_config, mesh_config.bridge_inset0_config, wall_overlapper_0, wall_0_wipe_dist, flow, retract_before_outer_wall);
+                    gcode_layer.addWall(hole_outer_wall[0], outer_poly_start_idx, mesh, mesh_config.inset0_config, mesh_config.bridge_inset0_config, wall_overlapper_0, wall_0_wipe_dist, flow, retract_before_outer_wall, /*unretract_before_last_travel_move = */true);
```
jellespijker, line 314: See previous comment about `constexpr`

src/InsetOrderOptimizer.cpp
```diff
@@ -330,11 +330,11 @@ void InsetOrderOptimizer::processHoleInsets()
             {
                 // align z-seam of hole with z-seam of outer wall - makes a nicer job when printing tubes
                 const unsigned point_idx = PolygonUtils::findNearestVert(start_point, hole_outer_wall.back());
-                gcode_layer.addWall(hole_outer_wall.back(), point_idx, mesh, mesh_config.inset0_config, mesh_config.bridge_inset0_config, wall_overlapper_0, wall_0_wipe_dist, flow, retract_before_outer_wall);
+                gcode_layer.addWall(hole_outer_wall.back(), point_idx, mesh, mesh_config.inset0_config, mesh_config.bridge_inset0_config, wall_overlapper_0, wall_0_wipe_dist, flow, retract_before_outer_wall, /*unretract_before_last_travel_move = */true);
```
jellespijker, line 333: See previous comment about `constexpr`

src/InsetOrderOptimizer.cpp
```diff
@@ -330,11 +330,11 @@ void InsetOrderOptimizer::processHoleInsets()
...
             }
             else
             {
-                gcode_layer.addWalls(hole_outer_wall, mesh, mesh_config.inset0_config, mesh_config.bridge_inset0_config, wall_overlapper_0, z_seam_config, wall_0_wipe_dist, flow, retract_before_outer_wall);
+                gcode_layer.addWalls(hole_outer_wall, mesh, mesh_config.inset0_config, mesh_config.bridge_inset0_config, wall_overlapper_0, z_seam_config, wall_0_wipe_dist, flow, retract_before_outer_wall, /*unretract_before_last_travel_move = */true);
```
jellespijker, line 337: See previous comment about `constexpr`

src/InsetOrderOptimizer.cpp
```diff
@@ -428,15 +428,15 @@ void InsetOrderOptimizer::processOuterWallInsets(const bool include_outer, const
             constexpr bool always_retract = false;
             for (unsigned int wall_idx : orderOptimizer.polyOrder)
             {
-                gcode_layer.addWall(part_inner_walls[wall_idx], orderOptimizer.polyStart[wall_idx], mesh, mesh_config.insetX_config, mesh_config.bridge_insetX_config, wall_overlapper_x, wall_0_wipe_dist, flow_ratio, always_retract);
+                gcode_layer.addWall(part_inner_walls[wall_idx], orderOptimizer.polyStart[wall_idx], mesh, mesh_config.insetX_config, mesh_config.bridge_insetX_config, wall_overlapper_x, wall_0_wipe_dist, flow_ratio, always_retract, /*unretract_before_last_travel_move = */false);
```
jellespijker, line 431: See previous comment about `constexpr` well you get the drift

src/LayerPlan.cpp
```diff
@@ -1644,10 +1646,16 @@ void LayerPlan::writeGCode(GCodeExport& gcode)
                     // Prevent the final travel(s) from resetting to the 'previous' layer height.
                     gcode.setZ(final_travel_z);
                 }
-                for(unsigned int point_idx = 0; point_idx < path.points.size(); point_idx++)
+                for(unsigned int point_idx = 0; point_idx < path.points.size() - 1; point_idx++)
```
jellespijker, line 1649: why not use:
```c++
for(const auto& point : path.points)
{
   gcode.writeTravel(point, speed)
}
```
untested and maybe use the actual type instead of auto but we should no longer use old style loops imo

src/LayerPlan.cpp
```diff
@@ -1644,10 +1646,16 @@ void LayerPlan::writeGCode(GCodeExport& gcode)
...
                 {
                     gcode.writeTravel(path.points[point_idx], speed);
                 }
+                if (path.unretract_before_last_travel_move)
+                {
+                    // We need to unretract before the last travel move of the path if the next path is an outer wall.
+                    gcode.writeUnretractionAndPrime();
+                }
+                gcode.writeTravel(path.points[path.points.size() - 1], speed);
```
jellespijker, line 1658: use `vector::back`

src/LayerPlan.h
```diff
@@ -578,7 +579,7 @@ class LayerPlan : public NoCopy
      * \param flow_ratio The ratio with which to multiply the extrusion amount
      * \param always_retract Whether to force a retraction when moving to the start of the wall (used for outer walls)
      */
-    void addWall(ConstPolygonRef polygon, int start_idx, const SliceMeshStorage& mesh, const GCodePathConfig& non_bridge_config, const GCodePathConfig& bridge_config, WallOverlapComputation* wall_overlap_computation, coord_t wall_0_wipe_dist, float flow_ratio, bool always_retract);
+    void addWall(ConstPolygonRef polygon, int start_idx, const SliceMeshStorage& mesh, const GCodePathConfig& non_bridge_config, const GCodePathConfig& bridge_config, WallOverlapComputation* wall_overlap_computation, coord_t wall_0_wipe_dist, float flow_ratio, bool always_retract, bool unretract_before_last_travel_move);
```
jellespijker, line 582: Update documentation

src/LayerPlan.h
```diff
@@ -592,7 +593,7 @@ class LayerPlan : public NoCopy
      * \param flow_ratio The ratio with which to multiply the extrusion amount
      * \param always_retract Whether to force a retraction when moving to the start of a wall (used for outer walls)
      */
-    void addWalls(const Polygons& walls, const SliceMeshStorage& mesh, const GCodePathConfig& non_bridge_config, const GCodePathConfig& bridge_config, WallOverlapComputation* wall_overlap_computation, const ZSeamConfig& z_seam_config = ZSeamConfig(), coord_t wall_0_wipe_dist = 0, float flow_ratio = 1.0, bool always_retract = false);
+    void addWalls(const Polygons& walls, const SliceMeshStorage& mesh, const GCodePathConfig& non_bridge_config, const GCodePathConfig& bridge_config, WallOverlapComputation* wall_overlap_computation, const ZSeamConfig& z_seam_config = ZSeamConfig(), coord_t wall_0_wipe_dist = 0, float flow_ratio = 1.0, bool always_retract = false, bool unretract_before_last_travel_move = false);
```
jellespijker, line 596: Update documentation

src/LayerPlanBuffer.cpp
```diff
@@ -94,10 +94,11 @@ void LayerPlanBuffer::addConnectingTravelMove(LayerPlan* prev_layer, const Layer
...
+        const Settings& prev_extruder_settings = Application::getInstance().current_slice->scene.extruders[prev_layer->extruder_plans.back().extruder_nr].settings;
+        const Settings& next_extruder_settings = Application::getInstance().current_slice->scene.extruders[newest_layer->extruder_plans.back().extruder_nr].settings;
         prev_layer->setIsInside(new_layer_destination_state->second);
-        const bool force_retract = extruder_settings.get<bool>("retract_at_layer_change") ||
-          (mesh_group_settings.get<bool>("travel_retract_before_outer_wall") && (mesh_group_settings.get<bool>("outer_inset_first") || mesh_group_settings.get<size_t>("wall_line_count") == 1)); //Moving towards an outer wall.
+        const bool force_retract = prev_extruder_settings.get<bool>("retract_at_layer_change") ||
+                                   (mesh_group_settings.get<bool>("travel_retract_before_outer_wall") && (mesh_group_settings.get<bool>("outer_inset_first") || mesh_group_settings.get<size_t>("wall_line_count") == 1)); //Moving towards an outer wall.
```
jellespijker, line 101: Add linebreak after `&&`

src/LayerPlanBuffer.cpp
```diff
@@ -107,6 +108,16 @@ void LayerPlanBuffer::addConnectingTravelMove(LayerPlan* prev_layer, const Layer
             // we force the path to use retraction
             path.retract = true;
         }
+        if (path.retract && !path.unretract_before_last_travel_move &&
+            next_extruder_settings.get<bool>("outer_inset_first") &&
+            (next_extruder_settings.get<coord_t>("infill_line_distance") == 0 || !next_extruder_settings.get<bool>("infill_before_walls")))
```
jellespijker, line 113: Add line break after `||`

tests/LayerPlanTest.cpp
```diff
@@ -398,6 +398,24 @@ class AddTravelTest : public LayerPlanTest, public testing::WithParamInterface<s
...
     }
+
+    /*!
+     * Runs the test, adding a travel move to the layer plan with the specified input.
+     * \param unretract_before_last_travel_move Whether to unretract before the last travel move of the travel path
+     * \return The resulting g-code path.
+     */
+    GCodePath run(bool unretract_before_last_travel_move)
+    {
+        const Point destination(500000, 500000);
```
jellespijker, line 409: What was the reasoning for the values? Also use `constexpr`

jellespijker (dismissed):
I think the approach you took is good only got some remarks about  code-style. We have a modern C++17 style now use it! ;-) 
- Use `constexpr` instead of `\* var_name *\
- Use `for(const auto& item : collection)`
- Use `vector::back`

And you missed some documentation of functions

src/LayerPlan.cpp
konskarm: Because it reaches up to the `path.points.size() - 1` point.

src/FffGcodeWriter.cpp
konskarm: Good suggestion!

src/LayerPlan.cpp
konskarm: I always forget that. Good suggestion

tests/LayerPlanTest.cpp
konskarm: Basically adding a long travel move.

Ghostkeeper: We can't make this variable `constexpr` without also adding a `constexpr` constructor. That can be done, of course, but is a modification to ClipperLib.

src/LayerPlanBuffer.cpp
```diff
@@ -107,6 +110,17 @@ void LayerPlanBuffer::addConnectingTravelMove(LayerPlan* prev_layer, const Layer
...
         }
+        if (path.retract && !path.unretract_before_last_travel_move &&
+            next_extruder_settings.get<bool>("outer_inset_first") &&
+            (next_extruder_settings.get<coord_t>("infill_line_distance") == 0 ||
+                    !next_extruder_settings.get<bool>("infill_before_walls")))
+        {
+            // If there is a retraction and the next layer's first path is an outer wall, then we need to make sure that
+            // we unretract before the last travel move of the previous layer. This way we make sure that the unretract
+            // does not happen while the nozzle is on top of the new layer's outer wall, which could cause artefacts and
+            // decrease dimensional accuracy.
```
Ghostkeeper, line 121: There's a lot more to detecting if the next layer's first path is an outer wall... For instance, if there is any support, it will be printed before the outer wall. In fact, I've written a 160-line function once [here](https://github.com/Ultimaker/CuraEngine/blob/flow_advance/src/FffGcodeWriter.cpp#L3044) to determine which structure it should start with on the next layer. Things that could occur before an outer wall:
* Raft (not applicable)
* Prime tower
* Brim
* Skirt
* Ooze shield
* Draft shield
* Support infill
* Support roof
* Support floor
* Different meshes with per-mesh settings saying they have no walls, outer inset not first, etc.
* Surface mode

I'll do some testing to see if it actually goes wrong, but we might have a big simplification here of a complex problem.

src/LayerPlanBuffer.cpp
```diff
@@ -107,6 +110,17 @@ void LayerPlanBuffer::addConnectingTravelMove(LayerPlan* prev_layer, const Layer
             // we force the path to use retraction
             path.retract = true;
         }
+        if (path.retract && !path.unretract_before_last_travel_move &&
+            next_extruder_settings.get<bool>("outer_inset_first") &&
+            (next_extruder_settings.get<coord_t>("infill_line_distance") == 0 ||
+                    !next_extruder_settings.get<bool>("infill_before_walls")))
+        {
+            // If there is a retraction and the next layer's first path is an outer wall, then we need to make sure that
```
Ghostkeeper, line 118: Next layer? Or next extruder?

Ghostkeeper, line 118: What happens if the next object has per-object settings that change these setting values? I think it's getting them from the wrong settings object.

Getting the first line is a complex business. Other things that could be printed in between are:
* Support
* Draft shield
* Ooze shield
* Prime tower
* Meshes with per-object settings causing the outer wall to not be the first thing printed.

For the Flow Advance algorithm I wrote a 160-line function to get the flow rate of the first extrusion on the next layer. Complex business.

src/LayerPlanBuffer.cpp
```diff
@@ -94,10 +94,13 @@ void LayerPlanBuffer::addConnectingTravelMove(LayerPlan* prev_layer, const Layer
     if (!prev_layer->last_planned_position || *prev_layer->last_planned_position != first_location_new_layer)
     {
         const Settings& mesh_group_settings = Application::getInstance().current_slice->scene.current_mesh_group->settings;
-        const Settings& extruder_settings = Application::getInstance().current_slice->scene.extruders[prev_layer->extruder_plans.back().extruder_nr].settings;
+        const Settings& prev_extruder_settings = Application::getInstance().current_slice->scene.extruders[prev_layer->extruder_plans.back().extruder_nr].settings;
+        const Settings& next_extruder_settings = Application::getInstance().current_slice->scene.extruders[newest_layer->extruder_plans.back().extruder_nr].settings;
```
Ghostkeeper, line 98: If there are multiple extruders on `newest_layer`, then this will get the last of those extruders, not the first. Is this what you meant?

```suggestion
        const Settings& next_extruder_settings = Application::getInstance().current_slice->scene.extruders[newest_layer->extruder_plans.front().extruder_nr].settings;
```

src/LayerPlan.h
```diff
@@ -577,8 +581,12 @@ class LayerPlan : public NoCopy
      * \param wall_0_wipe_dist The distance to travel along the wall after it has been laid down, in order to wipe the start and end of the wall together
      * \param flow_ratio The ratio with which to multiply the extrusion amount
      * \param always_retract Whether to force a retraction when moving to the start of the wall (used for outer walls)
+     * \param unretract_before_last_travel_move Whether to unretract before the last travel move of the travel path,
```
Ghostkeeper, line 584: Architecturally it seems a bit weird for the `addWall` function to get this parameter.

Should the `unretract_before_last_travel_move` parameter of paths not just be set by whatever code adds the travel move that the unretract is necessary for?

Ghostkeeper (changes requested):

src/LayerPlan.cpp
jellespijker: @konskarm Good catch and stupid suggestion of me. But it so makes me wish that we add the range-v3 library (which we now can). So that we can do this save, modern and lazy. https://ericniebler.github.io/range-v3/ 
I wished we used the header only range-v3 library that would make this a breach:
```c++
for(const auto& point : path.points| ranges::views::take(path.points.size() - 1))
{
   gcode.writeTravel(point, speed)
}
```

tests/LayerPlanTest.cpp
jellespijker: Most certainly out of scope

src/LayerPlan.h
konskarm: Indeed, but the codes that add travel moves do not know if they are adding outer or inner wall to make a decision on the unretraction. The knowledge of whether we are adding an outer wall lies in the code that actually adds that wall.

src/LayerPlanBuffer.cpp
konskarm: Yes, indeed. Good catch

src/LayerPlan.h
Ghostkeeper: But that code does know whether to make that small travel move towards the position away from the outer wall, right? How does it decide on that then, if it doesn't know whether it's printing an outer or an inner wall?

konskarm: As discussed, this is now fixed, since the decision on whether we should unretract is taken by the combing (considering that this is the place where the extra travel move to the inside of an outer wall is added).

konskarm:
The failing tests are unrelated to this ticket.

src/pathPlanning/Comb.cpp
```diff
@@ -106,7 +106,11 @@ bool Comb::calc(const ExtruderTrain& train, Point startPoint, Point endPoint, Co
     {
         PolygonsPart part = partsView_inside_optimal.assemblePart(start_part_idx);
         combPaths.emplace_back();
-        return LinePolygonsCrossings::comb(part, *inside_loc_to_line_optimal, startPoint, endPoint, combPaths.back(), -offset_dist_to_get_from_on_the_polygon_to_outside, max_comb_distance_ignored, fail_on_unavoidable_obstacles);
+        const bool combing_succeeded = LinePolygonsCrossings::comb(part, *inside_loc_to_line_optimal, startPoint, endPoint, combPaths.back(), -offset_dist_to_get_from_on_the_polygon_to_outside, max_comb_distance_ignored, fail_on_unavoidable_obstacles);
+        // If the endpoint of the travel path changes with combing, then it means that we are moving to an outer wall
+        // and we should unretract before the last travel move when travelling to that outer wall
+        unretract_before_last_travel_move = combing_succeeded && endPoint != travel_end_point_before_combing;
```
rburema, line 112: This isn't very likely to occur, if it indeed _can_ occur, but is there a possibility of the end-point just _happening to be_ on that specific spot on an outer wall?

rburema (approved):
Looks good enough to go to component test at least!

src/LayerPlan.cpp
```diff
@@ -405,7 +405,8 @@ GCodePath& LayerPlan::addTravel(const Point p, const bool force_retract)
         // Multiply by 2 because if two lines start and end points places very close then will be applied combing with retractions. (Ex: for brim)
         const coord_t max_distance_ignored = extruder->settings.get<coord_t>("machine_nozzle_tip_outer_diameter") / 2 * 2;
 
-        combed = comb->calc(*extruder, *last_planned_position, p, combPaths, was_inside, is_inside, max_distance_ignored);
+        bool unretract_before_last_travel_move = false; // Decided when calculating the combing
+        combed = comb->calc(*extruder, *last_planned_position, p, combPaths, was_inside, is_inside, max_distance_ignored, unretract_before_last_travel_move); //Here is the magic happening
```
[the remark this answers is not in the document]
konskarm: Whoops, I put the comment to remember where things should be changed. Will remove and push immediately

src/pathPlanning/Comb.cpp
konskarm: Initially the end-point is indeed on the outer wall, but combing decides to move it on the inside to avoid blobs and inaccuracies on the outer wall. If combing decides to leave it on the outer wall, then it means that combing did not succeed to find a proper combing path. In such cases, it is reasonable to not unretract before the last travel move, as there is no combing either.

PR #1475, opened 2021-07-29
[CURA-8442] Wall min flow doesn't make sense for Arachne.

rburema:
Any walls we get from Arachne should be printed, as such thin wall considerations are already taken into account when that is run.

(Diff is easier to see with -w as the changes are largely whitespace because of removed if statements.)

See frontend PR: https://github.com/Ultimaker/Cura/pull/10204


jellespijker (approved):
Looks good

PR #1476, opened 2021-08-02
Monotonic ordering of skin, top surface and ironing

Ghostkeeper:
This is a feature that implements printing lines in a monotonic order. That is, all adjacent lines should be printed in the same major direction side-by-side. The purpose of this printing order is to cause the micro-Fresnel reflection of light on the surface to appear the same. For instance, when a series of parallel line segments is printed from East to West, then ordering them from North to South causes the reflection of the printed surface to be different from when they are ordered from South to North. This technique will ensure that the lines are always ordered in the same direction if they are adjacent. For instance, all from North to South.

Here is an example of the monotonic ordering in action:
https://user-images.githubusercontent.com/2448634/127877046-761ddc12-6957-4c08-a986-285540a060ca.mp4

Current features include:
* Supporting lines as well as polylines.
* Available for skin, top surface and ironing. It could be used for others as well, but these are the relevant ones for the beauty of the top surface skin.
* Polylines get connected to ensure they are printed in one piece.
* Polylines still get broken up into pieces if multiple polylines are adjacent to one another in a joint.
* Integration tests enforce the monotonic constraint and a few others.
* Class layout ensures that it integrates more easily with libArachne, at the cost of being consistent with the current PathOrderOptimizer and LineOrderOptimizer classes.

Explicitly taken out of scope:
* Within the constraints of monotonic order, you would prefer to still keep an optimal ordering to reduce printing time. This is currently not implemented, but the algorithm was made to allow implementing this in the future. The algorithm currently creates segments of adjacent/connected lines, and prints those segments in a monotonic order which guarantees that everything is monotonic. Which of the monotonic segments gets sorted first depends on where they start in the monotonic direction, or arbitrary/rounding error if they are exactly equal.
* Printing infill or support in monotonic order. It is considered irrelevant.

Contributes to issue CURA-7928. Implements https://github.com/Ultimaker/Cura/issues/8909.


Ghostkeeper:
As of this writing this PR is still a draft because of two issues:
* In rare cases where many segments join together (like teeth exactly aligned in the monotonic direction) it seems that the joint line can still appear out of monotonic order. This is what the integration tests are indicating.
* When the endcap of a zigzag pattern is exactly aligned to the perpendicular direction, the algorithm ends up in an infinite loop for some reason.

src/TopSurface.cpp
```diff
@@ -122,7 +122,14 @@ bool TopSurface::ironing(const SliceMeshStorage& mesh, const GCodePathConfig& li
...
         }
 
-        layer.addLinesByOptimizer(ironing_lines, line_config, SpaceFillType::PolyLines);
+        if(!mesh.settings.get<bool>("ironing_monotonic"))
+        {
+            layer.addLinesByOptimizer(ironing_lines, line_config, SpaceFillType::PolyLines);
+        }
+        else
+        {
+            layer.addLinesMonotonic(ironing_lines, line_config, SpaceFillType::PolyLines, (direction + 90) / 180.0 * M_PI, line_spacing * 1.1);
```
rburema, line 131: Another raw `1.1` (and less important, another `X + 90 / 180.0 * M_PI`).

tests/PathOrderMonotonicTest.cpp
```diff
@@ -0,0 +1,286 @@
...
+#include <clipper.hpp>
+
+#include "../src/infill.h"
+#include "../src/utils/linearAlg2D.h"
+#include "../src/utils/math.h"
+#include "../src/PathOrderMonotonic.h"
+#include "../src/utils/polygon.h"
+#include "ReadTestPolygons.h"
+
+#define TEST_PATHS_SVG_OUTPUT
```
rburema, line 16: Do we want to leave this on during CI?

tests/PathOrderMonotonicTest.cpp
```diff
@@ -0,0 +1,286 @@
...
+            size_t section_b_id = 0;
+            for (const auto& section_b : sections)
+            {
+                ++section_b_id;
+                if (section_a_id >= section_b_id)
+                {
+                    continue; // <-- So section B will always be 'later' than section A.
+                }
+
+                // Already tested for A start < B start in the monotonic direction,
```
rburema, line 221: We should remove this comment (and the next line), as it makes little sense anymore.

src/PathOrderMonotonic.h
```diff
@@ -0,0 +1,449 @@
...
+                path->start_vertex = (path->start_vertex == 0) ? path->converted->size() - 1 : 0;
+                path->backwards = !path->backwards;
+            }
+        }
+
+        if(result.size() == 1)
+        {
+            result[0]->start_vertex = result[0]->converted->size(); //Reset start vertex as "unknown" again if it's not a string of polylines.
+        }
+        return result;
```
rburema, line 359: For consistencies sake: could use a `// Uses RVO!` comment like you did below (or remove that one).

src/PathOrderMonotonic.h
```diff
@@ -0,0 +1,449 @@
...
+                       && vSize2(found_path.point - first_endpoint) < SQUARED_COINCIDENT_POINT_DISTANCE; //And only find close lines.
+            });
+        }
+        while(close_line_after != lines_after.end())
+        {
+            Path* last = close_line_after->val;
+            result.push_back(last);
+            size_t farthest_vertex = getFarthestEndpoint(last, close_line_after->point); //Get to the opposite side.
+            last->start_vertex = (farthest_vertex == 0) ? last->converted->size() - 1 : 0;
+            last->backwards = (farthest_vertex != 0);
```
rburema, line 332: For consistencies sake: Either use brackets `(`, like you do here, or don't, like in the block above.

src/PathOrderMonotonic.h
```diff
@@ -0,0 +1,449 @@
...
+        while(close_line_before != lines_before.end())
+        {
+            Path* first = close_line_before->val;
+            result.push_front(first); //Store this one in the sequence. It's a good one.
+            size_t farthest_vertex = getFarthestEndpoint(first, close_line_before->point); //Get to the opposite side.
+            first->start_vertex = farthest_vertex;
+            first->backwards = farthest_vertex != 0;
+            first_endpoint = (*first->converted)[farthest_vertex];
+            lines_before = line_bucket_grid.getNearby(first_endpoint, COINCIDENT_POINT_DISTANCE);
+            close_line_before = std::find_if(lines_before.begin(), lines_before.end(), [first_endpoint, result](SparsePointGridInclusiveImpl::SparsePointGridInclusiveElem<Path*> found_path) {
```
rburema, line 321: So a couple of things about these specific inline closures.
- I don't think the capture of 'result' is necessary here.
- Propose: If you capture the other one (`..._endpoint`) as a reference or pointer instead, you could make one (named) closure, set the reference, and then feed the same one to the `.find(...)` each time. (Or make an actual lambda or function, and feed the reference via a `std::bind`.)

(Related: Perhaps the two earlier blocks could also be folded into the while loops, but let's not go overboard here.)

src/PathOrderMonotonic.h
```diff
@@ -0,0 +1,449 @@
...
+        std::vector<Path*> overlapping_lines;
+        for(auto overlapping_line = polyline_it + 1; overlapping_line != polylines.end(); overlapping_line++)
+        {
+            //Don't go beyond the maximum adjacent distance.
+            const coord_t start_their_projection = dot((*overlapping_line)->converted->front(), monotonic_vector);
+            const coord_t end_their_projection = dot((*overlapping_line)->converted->back(), monotonic_vector);
+            const coord_t their_farthest_projection = std::max(start_their_projection, end_their_projection);
+            const coord_t their_closest_projection = std::min(start_their_projection, end_their_projection);
+            if(their_closest_projection - my_farthest_projection > max_adjacent_distance * 1000
+                    || my_closest_projection - their_farthest_projection > max_adjacent_distance * 1000)  //Distances are multiplied by 1000 since monotonic_vector is not a unit vector, but a vector of length 1000.
```
rburema, line 417: Either make (global) `constexpr`, or use `MM2INT(...)`.

rburema (changes requested):
While I do request changes (or have made enough proposals and remarks that I want you to have a look at them at least), none of these should change the behaviour of the code. As such, I'll continue with the component test tomorrow. We could even merge, then have the changes as a separate task, to be completed later.

P.S. Also, I'm really impressed what you did, especially considering the little time you had! The comments where also very welcome and mostly really excellent (excepting there's one instance of under-explanation, see the review, and '5 cases for overlapping range' is a bit excessive in that I had an easier time reading the actual code than the comment, but you can leave that, as not everyone reads these the same way).

src/FffGcodeWriter.cpp
```diff
@@ -2509,26 +2511,44 @@ void FffGcodeWriter::processSkinPrintFeature(const SliceDataStorage& storage, La
...
+            constexpr Ratio flow = 1.0_r;
+            if(pattern == EFillMethod::GRID || pattern == EFillMethod::LINES || pattern == EFillMethod::TRIANGLES || pattern == EFillMethod::CUBIC || pattern == EFillMethod::TETRAHEDRAL || pattern == EFillMethod::QUARTER_CUBIC || pattern == EFillMethod::CUBICSUBDIV)
+            {
+                gcode_layer.addLinesMonotonic(skin_lines, config, SpaceFillType::Lines, monotonic_direction, config.getLineWidth() * 1.1, mesh.settings.get<coord_t>("infill_wipe_dist"), flow, fan_speed);
+            }
+            else
+            {
+                const SpaceFillType space_fill_type = (pattern == EFillMethod::ZIG_ZAG) ? SpaceFillType::PolyLines : SpaceFillType::Lines;
+                constexpr coord_t wipe_dist = 0;
+                gcode_layer.addLinesMonotonic(skin_lines, config, space_fill_type, monotonic_direction, config.getLineWidth() * 1.1, wipe_dist, flow, fan_speed);
```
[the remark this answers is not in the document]
Ghostkeeper: Done: c641818be85da0f1c727ee12da58c4a7606d94c8

src/FffGcodeWriter.cpp
```diff
@@ -2509,26 +2511,44 @@ void FffGcodeWriter::processSkinPrintFeature(const SliceDataStorage& storage, La
...
-        constexpr float flow = 1.0;
-        if (pattern == EFillMethod::GRID || pattern == EFillMethod::LINES || pattern == EFillMethod::TRIANGLES || pattern == EFillMethod::CUBIC || pattern == EFillMethod::TETRAHEDRAL || pattern == EFillMethod::QUARTER_CUBIC || pattern == EFillMethod::CUBICSUBDIV)
+        if(monotonic)
         {
-            gcode_layer.addLinesByOptimizer(skin_lines, config, SpaceFillType::Lines, enable_travel_optimization, mesh.settings.get<coord_t>("infill_wipe_dist"), flow, near_start_location, fan_speed);
+            const AngleRadians monotonic_direction = (skin_angle + 90) / 180.0 * M_PI;
+            constexpr Ratio flow = 1.0_r;
+            if(pattern == EFillMethod::GRID || pattern == EFillMethod::LINES || pattern == EFillMethod::TRIANGLES || pattern == EFillMethod::CUBIC || pattern == EFillMethod::TETRAHEDRAL || pattern == EFillMethod::QUARTER_CUBIC || pattern == EFillMethod::CUBICSUBDIV)
+            {
+                gcode_layer.addLinesMonotonic(skin_lines, config, SpaceFillType::Lines, monotonic_direction, config.getLineWidth() * 1.1, mesh.settings.get<coord_t>("infill_wipe_dist"), flow, fan_speed);
```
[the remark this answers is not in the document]
Ghostkeeper: Done: c641818be85da0f1c727ee12da58c4a7606d94c8

src/TopSurface.cpp
Ghostkeeper: Done: c641818be85da0f1c727ee12da58c4a7606d94c8 and ca28df1077e733d3aea67407385f017518117da4.

tests/PathOrderMonotonicTest.cpp
Ghostkeeper: Definitely not! Must've accidentally left this in while testing.
Done here: cf051f0d8e32f2b1ff34d5a74aa32ecbab0a80bf

tests/PathOrderMonotonicTest.cpp
Ghostkeeper: Indeed. Done here: 57ad843b67160b7055544f98b36cf81416301934

src/PathOrderMonotonic.h
Ghostkeeper: I've opted to remove the other one. It's something I'm paying attention to during development, and thus think I need to write down. But for the most part a developer doesn't really need to be reminded of this. Removed here: 57ad843b67160b7055544f98b36cf81416301934

src/LayerPlan.cpp
```diff
@@ -1124,6 +1125,65 @@ void LayerPlan::addLinesByOptimizer(const Polygons& polygons, const GCodePathCon
...
 
+void LayerPlan::addLinesMonotonic(const Polygons& polygons, const GCodePathConfig& config, const SpaceFillType space_fill_type, const AngleRadians monotonic_direction, const coord_t max_adjacent_distance, const coord_t wipe_dist, const Ratio flow_ratio, const double fan_speed)
+{
+    const Point last_position = getLastPlannedPositionOrStartingPosition();
+    PathOrderMonotonic<ConstPolygonRef> order(monotonic_direction, max_adjacent_distance, last_position);
+    for(size_t line_idx = 0; line_idx < polygons.size(); ++line_idx)
+    {
+        order.addPolyline(polygons[line_idx]);
+    }
+    order.optimize();
```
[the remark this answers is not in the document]
Ghostkeeper: Indeed this is the way it'll need to be called in Arachne, so we can unify them there.

The difficulty of unifying them now lies in that the result of the optimisation needs to be obtained differently. This makes the loop condition different (because the size is in the `paths` member, not in `polyOrder`) and also the method to obtain the actual paths. The old optimizer returned a list of indices which then need to be looked up in the original array. The new method contains the actual data in its optimised order. So in the new method the path is obtained with `order.paths[order_idx].vertices` while in the old method it is obtained with `polygons[orderOptimizer.polyOrder[order_idx]]`. This is done in two places, one nested very deeply, so it's hard to pull out.

One could write an adapter to convert one API to be the same as the other. But I don't think that's worth the effort considering it needs to be removed for Arachne anyway. The adapter is likely to give us more maintenance than this repetition, if we need to make a change here, too.

Opinions welcome of course. If you disagree I could still work on an adapter.

src/LayerPlan.h
```diff
@@ -607,6 +607,25 @@ class LayerPlan : public NoCopy
      */
     void addLinesByOptimizer(const Polygons& polygons, const GCodePathConfig& config, SpaceFillType space_fill_type, bool enable_travel_optimization = false, int wipe_dist = 0, float flow_ratio = 1.0, std::optional<Point> near_start_location = std::optional<Point>(), double fan_speed = GCodePathConfig::FAN_SPEED_DEFAULT);
 
+    /*!
+     * Add polygons to the g-code with monotonic order.
+     * \param polygons The lines to add.
+     * \param config The settings to print those lines with.
+     * \param space_fill_type The type of space filling used to generate the
+     * line segments (should be either Lines or PolyLines!)
```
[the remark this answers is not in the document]
Ghostkeeper: Thing is, this `space_fill_type` variable is not used at all at the moment. It's necessary to give it a `space_fill_type` because the `GCodePath` class stores it (and it's a constructor argument). However if you look at the usages of that `space_fill_type` member you'll find that it is only ever written by the constructor and never actually read.

This `space_fill_type` should be removed, in my opinion. It's only slowing the slicing down (albeit not very much, really). This documentation has been copied from the `addPathsByOptimizer` function. There is no real reason why it should be Lines or PolyLines since it's not actually used, and the optimizer itself supports polygons just as well.

src/LayerPlan.cpp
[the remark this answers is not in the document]
rburema: I'm fine with postponing it until Arachne is in. My worry was just about it being permanent.

src/LayerPlan.h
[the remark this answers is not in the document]
rburema: Ok, let's put that out of scope for the time being then. (Unless it's really easy to remove.)

src/PathOrderMonotonic.h
rburema: Done, so can be resolved :-)

src/PathOrder.h
```diff
@@ -0,0 +1,197 @@
...
+    /*!
+     * In the current set of paths, detect all loops and mark them as such.
+     *
+     * This will go through all polylines in the paths. If the endpoints of
+     * these polylines are close enough together, it considers them a polygon.
+     * It will then mark these polylines as being polygons for the future.
+     */
+    void detectLoops()
+    {
+        constexpr coord_t coincident_point_distance = 10; //If the endpoints are closer together than 10 units, it is considered pretty much a closed loop.
```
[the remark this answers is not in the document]
Ghostkeeper: I'd argue that it is "considered closed if the endpoints are coincident". So "coincident to within a margin of error" is also okay.

But you make a good point that we have two constants here effectively defining the same thing. We can add a concept of "coincident to within a margin of error" to the entire `PathOrder` class and re-use that.

I've done that here: f5033e612a93ec2c6ad86e5be6556acf2290c5b2

src/PathOrder.h
```diff
@@ -0,0 +1,197 @@
...
+         * Whether the path should be traversed in backwards direction.
+         *
+         * For a polyline it may be more efficient to print the path in
+         * backwards direction, if the last vertex is closer than the first.
+         */
+        bool backwards;
+    };
+
+    /*!
+     * After reordering, this contains the path that need to be printed in the
```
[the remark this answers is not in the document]
Ghostkeeper: Done here, secretly: f5033e612a93ec2c6ad86e5be6556acf2290c5b2

src/PathOrder.h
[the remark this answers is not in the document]
rburema: That reasoning also makes sense. It was one or the other (rename or make value equal), really, so fine by me!

src/PathOrderMonotonic.h
```diff
@@ -0,0 +1,449 @@
...
+ * closest to the source direction of the monotonicity vector.
+ */
+template<typename PathType>
+class PathOrderMonotonic : public PathOrder<PathType>
+{
+public:
+    using typename PathOrder<PathType>::Path;
+
+    PathOrderMonotonic(const AngleRadians monotonic_direction, const coord_t max_adjacent_distance, const Point start_point)
+    : monotonic_vector(std::cos(monotonic_direction) * 1000, std::sin(monotonic_direction) * 1000)
```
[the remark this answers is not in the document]
Ghostkeeper: Indeed, but the aim is not for the vector to be a certain physical distance long, so `MM2INT` is not used correctly here. The aim is just to get enough spatial resolution so that rounding errors from the `coord_t` type don't significantly alter the direction. If the vector had a length of 1, it could really only point in the 8 cardinal directions due to rounding to integers (and in fact it rounds down here, so probably you just get 0,0).

Ideally you'd make the other uses of the `1000` constant multiply by the length of the vector instead. But since that length is already constant and known it's really quite unnecessary.

I'll define a constant here to remove the duplication. Done here: a0656e670f3470197d3def1612186ec759478ebb

src/PathOrderMonotonic.h
```diff
@@ -0,0 +1,449 @@
...
+            //If we're part of a string of polylines, connect up the whole string and mark all of them as being connected.
+            if(polystring.size() > 1)
+            {
+                starting_lines.insert(polystring[0]);
+                for(size_t i = 0; i < polystring.size() - 1; ++i) //Iterate over every pair of adjacent polylines in the string (so skip the last one)!
+                {
+                    connections[polystring[i]] = polystring[i + 1];
+                    connected_lines.insert(polystring[i + 1]);
+                    const std::vector<Path*> overlapping_lines = getOverlappingLines(std::find(polylines.begin(), polylines.end(), polystring[i]), perpendicular, polylines);
+                    for(Path* overlapping_line : overlapping_lines) //Last one is done in the else case below.
```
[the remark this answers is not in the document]
Ghostkeeper: Indeed, that's important to document well.

Also, you didn't remark this, but the remaining comment on this line was actually outdated too.

Added new documentation and removed the outdated documentation here: 430edc166df19083824c3714319a16f820104cfc

src/PathOrderMonotonic.h
Ghostkeeper: Made it a static constexpr here: a0656e670f3470197d3def1612186ec759478ebb

src/PathOrderMonotonic.h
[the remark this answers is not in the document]
rburema: I see. A constant is fine in that case, yes.

src/PathOrderMonotonic.h
Ghostkeeper: Yeah I considered factoring this repeated bit of code out. I think I'll do that. It's even caused bugs for me before because I forgot to update the two calls above the loop.

The `result` capture is indeed a leftover from a previous implementation. A bugfix on Tuesday to fix those infinite loops made this bit a lot simpler and removed the need to capture `result`.

I'm opting for the second option, feeding the point as an argument to an actual function. Storing the two points in the same variable is a bad idea because they are both used below (which needs the old value of the front still). That could be resolved too by interweaving that code, but I think this is clearer. And if it's a named function you'll have to give it that argument anyway.

Since `std::bind_front` is not available until C++20, I'll just have to use a lambda.

Done here: 47fe5b9e5a520de0682edad044ab79a70a281861

src/PathOrderMonotonic.h
Ghostkeeper: Done secretly inside this unrelated commit: 47fe5b9e5a520de0682edad044ab79a70a281861

src/PathOrderMonotonic.h
```diff
@@ -0,0 +1,449 @@
...
+     * For a given path, make sure that it is configured correctly to start
+     * printing from the best endpoint.
+     *
+     * This changes the path's ``start_vertex`` and ``backwards`` fields, and
+     * also adjusts the \ref current_pos in-place.
+     *
+     * If the path already had a ``start_vertex`` set, this will not be
+     * adjusted. Only the ``current_pos`` will be set then.
+     *
+     * Will cause a crash if given a path with 0 vertices!
```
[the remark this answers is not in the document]
Ghostkeeper: We do that check here: https://github.com/Ultimaker/CuraEngine/pull/1476/files#diff-4441e292f911dd779107cb41ee83900436da2df15d3830ab0c9dcdfcdbd48b2eR101

Since this is an internal function, we don't need to re-check all edge cases within every nested function call. This documentation is there to hopefully ensure that whoever maintains this later will make sure that empty paths are filtered out before it gets to this internal function.

src/PathOrderMonotonic.h
rburema: Yes, this is what I wanted. (I have a hard time keeping track of what's in C++17 versus what's in C++20 for some reason.)

src/PathOrderMonotonic.h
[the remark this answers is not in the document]
rburema: Fair enough.

rburema (approved):

nallath (approved):

PR #1478, opened 2021-08-06
[CURA-7279] Back-Pressure Compensation

rburema:
Since the total (filament) pressure in a feeder-system is not only dependant on the pressure that exists between the nozzle and the feed-mechanism (which should be near-constant on our setup), but _also_ between the nozzle and the last-printed layer. This last type is called 'back-pressure'. In this PR, properties of the outflow are adjusted so that the back-pressure is compensated for. This is probably especially important if the printer has a Bowden-tube style setup.

You'll need the front-end PR https://github.com/Ultimaker/Cura/pull/10245


Piezoid:
This changes look really interesting. Few questions, after trying this branch:

- Now that the `speed_equalize_flow_enabled` is gone for good, how do you enable flow equalization? Is it enabled by default? The setting description,
> A factor above zero means that thin lines are adjusted to be even faster than simple flow equalization would speed them up to

make it seems like there is a base linear flow equalization. When `speed_equalize_flow_width_factor`=0 (or really close to zero), I don't see any flow change wrt. to line width.

- How does it interact with linear / pressure advance? If I understand the PR correctly, this is a different correction, complementary to linear advance (which itself correspond to the correction of the "near-constant on our setup" part).

- Looks like it is currently possible to get negative speeds/flows. Here is the last top layer connected to the base of the benchy's chimney (lw x lh = .2 x .08, `speed_equalize_flow_width_factor`=1):
image

Thanks

rburema:
Hi @Piezoid!

First, note that this PR is still in draft mode. We're still working on it. Also note that it will be merged to (the) Arachne (branch), not master.

That said, thanks for trying it out! It can definitely help when the community takes an interest at this early stage as well :-) Definitely a good point about the negative speeds, we'll really need to fix that before merge.

So, let's try to answer these questions:

- The original flow equalization setting applied only to gap filling. Since gap filling isn't really a thing anymore in Arachne, (because the walls can just vary widths from the start) this could be removed. The setting introduced here indeed does similar things.
- For this setting, 'zero' means off, no back-pressure compensation will take place. It is initially not enabled, but we're hoping to get the materials team to have a look at it (well, if it works at all), so we can set a sensible default instead of 'off'.
- Yes, linear adavance / pressure advance, etc. is separate... sort of 'the other side of the coin' of what this is trying to achieve.
- As I said, thanks, negative speeds should definitely be avoided (unless perhaps someone invents time-travel I suppose). Can't imagine what it would do to a printer with less than great firmware, for example.

Piezoid:
> First, note that this PR is still in draft mode. We're still working on it. Also note that it will be merged to (the) Arachne (branch), not master.
I'm well aware of that. I'm following the development of the Arachne branch and I've been doing flow equalization with a custom gcode post-processor that yield constant flow, so I though I would give a try to this PR.

> That said, thanks for trying it out! It can definitely help when the community takes an interest at this early stage as well :-) Definitely a good point about the negative speeds, we'll really need to fix that before merge.

On that note, I've ran CuraEngine/Arachne with sanitizers (memory, UB and leak) and wrote a bunch of patches to improve stability. It is in a very draft state, not well informed, probably doesn't honor the coding style, and some commits might be outdated (as not needed anymore, after months of rebasing). I haven't got the time to clean this and make a proper detailed PRs. 
 However you can [find the patches here](https://github.com/Piezoid/CuraEngine/compare/Arachne_merged...Piezoid:Arachne_fixes?expand=1), maybe it solves some of the bugs you are currently chasing. If it helps, you have my permission to cherry-pick without attribution.

>     * For this setting, 'zero' means off, no back-pressure compensation will take place. It is initially not enabled, but we're hoping to get the materials team to have a look at it (well, if it works at all), so we can set a sensible default instead of 'off'.

Ok, I thought this setting was complementary to a "simple flow equalization" that I couldn't find in the settings. Maybe the description should be more explicit that this is the (only) flow equalization.

Thanks

rburema:
@Piezoid Good that you had a look! I skimmed through the code, and while a lot of things do look like they might be neat improvements, I'd _strongly_ recommend splitting this in (a lot of?) tiny/smaller PR's, and then actually making a pull requests.

The way we merge PR's internally here is that all the (coder) devs take a look at the pulls here on github, then say in our slack something like 'This is the PR I'm starting on today, can I have another dev to confirm that the _idea_ behind it is something we want.'

If I then have to go, when they ask 'well, what does it do?', it's a lot harder to sell 'well, I dunno, a lot? they're not sure about it and it's still in progress, also, not really submitted as a proper PR yet' than 'it fixes these related memory leaks' or 'it improves code style', etc. Smaller PR's also speed up the process in other, similar ways when we hand it over to QA Furthermore, as someone who works on merging these; it's a lot easier to have to concentrate on a (relatively) small piece of code, as opposed to a massive set of changes that may or may not be related.

Lastly, we can then also determine per subject if we want something. (For example, memory leak fixes might be in, code style changes might be out.)

I hope you understand!

Piezoid:
> I hope you understand!

Totally! 

This is exactly why I didn't make a batch PR. As you said this should split in a lot of independent PRs, therefore requiring a lot of work. My point is that I won't have time to do a proper cleaning and splitting of these patches for some time. I haven't touched the code for a while and don't even remember the exact bugs and how to reproduce them. (A lack of organization on my side :sweat_smile:)

I was only mentioning it for internal use in testing and bug chasing, if it can be of any help.

But I'll certainly try to do the work, next time I put my hand in the code.

rburema:
@Piezoid  No problem! Take your time. We _are_ working on Arachne this sprint, and I _do_ now know about your branch, so who knows.

Back on the original topic: We've rewritten the formula we use for this, so now the flow equalization should already be happening at the default factor or '0.0'. (Also the tests should now pass.) If you still find negative speeds, please let us know!

nallath (approved):

Piezoid:
Hi @rburema !

d17b3d3 has indeed fixed the negative flows and speeds issues with overboard settings. :+1: 

About the factor value: it seems to me that the value '1.0' yields a constant flow per feature, with a speed scaled by `speed_back_pressure_factor = nominal_width_for_path / line_width_for_path`, while '0.0' leaves the speeds constant. 
I'm only nitpicking about the description: "A factor above zero means that thin lines are adjusted to be even faster than simple flow equalization would speed them up to." Although I may misunderstand something.
Also, the factor appears to be dimensionless?

Sorry for being such a pain :sweat_smile:

rburema:
Hi @Piezoid ,

> Sorry for being such a pain 😅

When a colleague does that, we call it 'code review', and a regular part of my work is to receive (and give) them :-)

> About the factor value: it seems ...

You're right! I think I know how this confusion got started. The formula as written in the ticket says to add one to the factor before multiplication, but we kept that out here in order to keep the behaviour of 0.0 the same as 'do nothing'. Glad you caught that.

PR #1483, opened 2021-08-13
Actually give the simplify the squared values

nallath:
Whoops... Well, at least that makes sense why the simplify wasn't doing what we expected it to do...

CURA-8132


rburema (approved):
Good find! :-)

PR #1485, opened 2021-08-16
Fix combing 

nallath:
CURA-8117


rburema (approved):
While I had some concerns about what happens when there is an entire scene of objects encircled within a polygon (like when you have a vase with something(s) in it), it seems it now performs way closer to what the main branch is doing, which is a good thing. Approved.

PR #1486, opened 2021-08-17
Cura 8069 centerline

jellespijker:
While working on this I discovered that the `getOptimalThickness` of the Redistribution class (also used by the other strategies) only used the width of the inner bead. This would result in an under-/overestimation of the thickness of the whole region. Resulting in variances of the innermost inner beads compared with outermost inner beads. This occurs when the shape wasn't completely filled and the beads could have their optimal width. The first commit fixes this, by using the optimal width of the two outer walls and the optimal width of the other innerwalls.

The center deviation strategy now generates a centerline again. When this strategy is used it will either deviate the width
of both inner lines, in the case of an even number of beads, or deviate the inner line, when there is an uneven number.

It now also handles the splitting and adding of the centerline correctly.

Fixes CURA-8069

### DIFF
--- a/src/BeadingStrategy/CenterDeviationBeadingStrategy.cpp
+++ b/src/BeadingStrategy/CenterDeviationBeadingStrategy.cpp
@@ -1,69 +1,86 @@
-//Copyright (c) 2021 Ultimaker B.V.
-//CuraEngine is released under the terms of the AGPLv3 or higher.
+// Copyright (c) 2021 Ultimaker B.V.
+// CuraEngine is released under the terms of the AGPLv3 or higher.
+#include <algorithm>
 
 #include "CenterDeviationBeadingStrategy.h"
 
 namespace cura
 {
-    CenterDeviationBeadingStrategy::CenterDeviationBeadingStrategy(const coord_t pref_bead_width, const AngleRadians transitioning_angle, const Ratio wall_split_middle_threshold, const Ratio wall_add_middle_threshold)
-    : BeadingStrategy(pref_bead_width, pref_bead_width / 2, transitioning_angle)
-    , minimum_line_width_split(pref_bead_width * wall_split_middle_threshold)
-    , minimum_line_width_add(pref_bead_width* wall_add_middle_threshold)
-    {
-        name = "CenterDeviationBeadingStrategy";
-    }
+CenterDeviationBeadingStrategy::CenterDeviationBeadingStrategy(const coord_t pref_bead_width,
+                                                               const AngleRadians transitioning_angle,
+                                                               const Ratio wall_split_middle_threshold,
+                                                               const Ratio wall_add_middle_threshold)
+  : BeadingStrategy(pref_bead_width, pref_bead_width / 2, transitioning_angle),
+    minimum_line_width_split(pref_bead_width * wall_split_middle_threshold),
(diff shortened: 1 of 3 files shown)


nallath (approved):

src/BeadingStrategy/RedistributeBeadingStrategy.cpp
```diff
@@ -24,7 +24,8 @@ RedistributeBeadingStrategy::RedistributeBeadingStrategy(   const coord_t optima
 
 coord_t RedistributeBeadingStrategy::getOptimalThickness(coord_t bead_count) const
 {
-    return parent->getOptimalThickness(bead_count);
+    const coord_t inner_bead_count = bead_count > 2 ? bead_count - 2 : 0;
```
Piezoid, line 27: @jellespijker 
This is problematic when `bead_count == 1` or `0`:

we then get `getOptimalThickness(1) == getOptimalThickness(2)` because in both cases `inner_bead_count == 0`.

This causes `getTransitionAnchorPos` to return `inf` which produces an `end_pos=-9223372036854775761` in `generateTransitionEnds`

Note there is potentially other cases for `getTransitionAnchorPos` to return `inf`, for example, `LimitedBeadingStrategygetTransitionThickness(bead_count) == LimitedBeadingStrategygetTransitionThickness(bead_count+1)` will return when `bead_count > max_bead_count`

Piezoid: Ah sorry, I didn't realized that the next line from last commit fixed it.

jellespijker: @Piezoid no worries. I missed it indeed but lucky for me @nallath caught it during component testing and fixed it.

nallath: Ha, good to see that both of us came to the same conclusion on this :)

src/BeadingStrategy/CenterDeviationBeadingStrategy.cpp
```diff
@@ -1,69 +1,86 @@
...
 namespace cura
 {
-    CenterDeviationBeadingStrategy::CenterDeviationBeadingStrategy(const coord_t pref_bead_width, const AngleRadians transitioning_angle, const Ratio wall_split_middle_threshold, const Ratio wall_add_middle_threshold)
-    : BeadingStrategy(pref_bead_width, pref_bead_width / 2, transitioning_angle)
-    , minimum_line_width_split(pref_bead_width * wall_split_middle_threshold)
-    , minimum_line_width_add(pref_bead_width* wall_add_middle_threshold)
-    {
-        name = "CenterDeviationBeadingStrategy";
-    }
+CenterDeviationBeadingStrategy::CenterDeviationBeadingStrategy(const coord_t pref_bead_width,
```
rburema, line 9: If you're doing this, then please put all arguments & brackets on their own line. This isn't python!

jellespijker:
The crash that occurred in one of @rburema his models, happens when processing the layer plan. You slowly see your memory growing before the crash occurs. I'm not using Valgrind to identify the problem, but I'm fairly confident that it is unrelated to the changes introduced in this PR. I will therefore merge this PR in libArachne_rebased as agreed with @rburema

PR #1487, opened 2021-08-17
Inset Order Optimizer: fix out of bound error when region_ids were removed 

Piezoid:
[`variableWidthPathToBinJunctions`](https://github.com/Ultimaker/CuraEngine/blob/a0fc3a2bb934cbc244d73a7b32e2e12aa04fc276/src/InsetOrderOptimizer.cpp#L169) allocates a number of bins according to the number of regions computed by [`getOuterRegionId`](https://github.com/Ultimaker/CuraEngine/blob/a0fc3a2bb934cbc244d73a7b32e2e12aa04fc276/src/InsetOrderOptimizer.cpp#L130). `bin_id`, computed from `region_id`, could be out of bound when some lower `region_id` has been removed.

This is a fix attempt, that allocates a number of bins based on the max `region_id rather than the number of regions.
I'm still not settle on whether this is a symptom of another issue, since hitting this bug requires a specific combination of model and settings.

Here's a 3DBenchy 3mf that fail to slice on the current arachne branch:
[CFFFP_3DBenchy.zip](https://github.com/Ultimaker/CuraEngine/files/7001105/CFFFP_3DBenchy.zip)

Also, I included a small optimization that remove the unneeded manual insertion of a default-initialized `AARB` when a new `region_id` is encountered. I would prefer using the more explicit `try_emplace` but it is C++17 only.


jellespijker:
@Piezoid I didn't check any code (so no comments on that) but I just wanted to inform you that we should be C++17 compatible now. So if you want to use C++17 feature go ahead.

PR #1488, opened 2021-08-17
Fix error when usign both bridge and top surfaces

Piezoid:
[This 3mf file](https://github.com/Ultimaker/CuraEngine/files/7002231/CFFFP_VZCube.zip) fails to slice with `Coordinate outside allowed range`. This happens when using both bridges settigns and top surface layers.

This is caused by processing an empty polygon with `bridgeAngle` causing an invalid AARB computation.

This patch skip `processTopBottom` on `skin_part` when `skin_part.skin_fill.empty()`.


src/bridge.cpp
```diff
@@ -12,6 +12,7 @@ namespace cura
 
 int bridgeAngle(const Settings& settings, const Polygons& skin_outline, const SliceDataStorage& storage, const unsigned layer_nr, const unsigned bridge_layer, const SupportLayer* support_layer, Polygons& supported_regions)
 {
+    assert(skin_outline.size() > 0);
```
jellespijker, line 15: Why not use '! skin_outline.empty()'  it is not guaranteed that size() is a constant-time function, and it is generally more efficient and also shows clearer intent to use empty().

jellespijker (changes requested):
The suggested fix looks simple and straight forward. I will test it tomorrow morning and if I don't encounter any weird behavior merge it.
One small remarks is the assert. Don't check against size but check against empty.

Love to see a PR from the community against libArachne btw.

jellespijker:
Hi @Piezoid Finally got around to test your change. Your suggested fix will skip the whole processing of the TopBottom structure. That would mean it will also skip adjusting the fan speed and monotonic ordering of those area's  Although it is not immediately clear to me which situations can occur if the skin_fill is empty. I personally think it might be best if we put your check for empty closer to where it actually breaks.

I think the lambda function handle_bridge_skin should return a `false` if the `skin_fill` is empty. That would ensure all other processing steps in the layer are executed except for bridging the skin. Which is imo a safer fix.

Do you agree?

Piezoid:
I may indeed have bypassed to much generation. I'll investigate this.

I think this requires a test with a staircase model that features both normal top infill and top surface infill on the same layer.

Piezoid:
@jellespijker 
After taking a closer look at `processTopBottom`, it appears that this method should produce no side effects when given an empty polygon vector. While this method and `processSkinPrintFeature` do a lot of things, it is only in preparation for emitting gcode features generated from polygons. 

After much planing, `processSkinPrintFeature` is called. Inside this method, the `Infill` generator doesn't produce any features when working on an empty vector of polygons. With no features to write, the following large if statement cuts off any side effect. This is the only place where `gcode_layer&` is mutated. The computed `fan_speed` is only consumed by the the various `gcode_layer.addLines`* that never get called for the same reason.

I agree that it's is really hard to prove that a C++ method has no side effects under some runtime conditions. The best I could do is exercise the code with [this patch](https://github.com/Piezoid/CuraEngine/commit/assert-empty_skinfill_no_side_effects), checking at runtime that:
  * `added_something` is not set to true by `processTopBottom` if there is no polygons,
  * `Infill` doesn't generates any path, polygons or lines when given an empty vector (there is no other geometry input),
  * the gcode produced is identical if calls to `processTopBottom` are skipped when `skin_fill` is empty (compiling with `bool do_empty_skin_fill = false;`, ran with `-m 1` for reproducibility)

This is tested by slicing this [test piece](https://github.com/Ultimaker/CuraEngine/files/7034442/test_piece.zip) and other more organic models.
  
What do you think?
The change you suggested to harden the `handle_bridge_skin` lambda is already in the test patch linked above. I can easily replace this PR with this 3 lines change.

jellespijker:
I think that your analysis is correct, but above all I/we really appreciate your effort here. Just wanted to let you know that.
I would suggest that you create a commit from the patch. I will do a second test. Probably tomorrow morning (today is swamped with other work for me) just to be sure.
But I don't foresee any complications or unwanted side effects at first glance.

PR #1492, opened 2021-08-25
Fix memory leaks

Piezoid:
This fixes leaks of three different type of objects: `BeadingStrategies`, `LocToLineGrid` and `InfillLineSegment`. This first two are handled by generalizing the use of `unique_ptr` at the factory level.

The first is specific to the Arachne branch. If wanted, I can backport the others. Last commit is optional and unimportant.

On a big model, the memory consumption went down by 153MB (from 282MB to 129MB, max RSS, [detailed leak sanitizer report](https://github.com/Ultimaker/CuraEngine/files/7049026/LSan_report.txt). 

---

`InfillLineSegment`s are allocated by `generateLinearBasedInfill` and their pointers is stored in the vector `crossings_on_line`. They are leaked when zigzaggify is disabled and `connectLines` is not called to consume them. The solution is to not generate them in the first place. A new boolean member tells whether `connectLines` will be called.

It is not easy to use `unique_ptr` without more logic changes, because a pointer is stored in two `crossings_on_line` entries, which share the ownership. I didn't change the memory management in `connectLines`, which seems correct, if a bit hard to follow.

Also `generateLinearBasedInfill` doesn't need to call `addLineInfill` when the lines will be generated by `connectLines`. I hoisted the if statement from `addLineInfill` that disable line generation into `generateLinearBasedInfill`.

I'm hesitant to include the last commit, which is mostly cosmetics and some not so useful fine tuning. I might instead do another PR: I'm a bit bothered that `generateLinearBasedInfill` fills three different outputs (`cut_list`, `crossings_per_scanline`/`crossings_on_line`, and `zigzag_connector_processor`) that are not always consumed:
* Non zigzagfified linear infills use only `cut_list`,
* Zigzagified linear infills use only `crossings_on_line`,
* Zigzag uses `zigzag_connector_processor` and `cut_list`.
Some static polymorphism might be helpfull here. 

On the subject of unifying, there also the TODO about using the `connectLines` algorithm for more kinds of infills. I probably don't get all the details to tackle that. For example, there are some differences between ZigZag and connected Lines which I don't quite understand yet.


rburema:
Hi @Piezoid,

Thanks so much for the fixes & refactors here. Great work! :-)

> I'm hesitant to include the last commit, which is mostly cosmetics and some not so useful fine tuning.

Cosmetics usually help with readability.

> For example, there are some differences between ZigZag and connected Lines which I don't quite understand yet.

Zig-zag is a specific pattern, for 'line based infill' (by which we mean parallel lines): extrude to one side, go 'sideways' via the border, extrude back to the other side, go sideways again, etc. With connect lines, you 'just' want the extrusions, wether they are line based or not, to be connected (which could result in a zig-zag pattern of course) via the sides.

(for our internal reference: ticket CURA-8532)

PR #1499, opened 2021-09-09
Fix centerline being twice as wide as needs to be for uneven beads

nallath:
CURA-8069

### DIFF
--- a/src/BeadingStrategy/CenterDeviationBeadingStrategy.cpp
+++ b/src/BeadingStrategy/CenterDeviationBeadingStrategy.cpp
@@ -45,7 +45,7 @@ CenterDeviationBeadingStrategy::Beading CenterDeviationBeadingStrategy::compute(
             {
                 return compute(thickness, bead_count - 1);
             }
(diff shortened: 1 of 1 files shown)


Ghostkeeper (approved):

src/BeadingStrategy/CenterDeviationBeadingStrategy.cpp
```diff
@@ -34,7 +34,7 @@ CenterDeviationBeadingStrategy::Beading CenterDeviationBeadingStrategy::compute(
         {
             if (inner_bead_widths < minimum_line_width_add)
             {
-                return compute(thickness, bead_count - 1);
+                return compute(thickness, bead_count + 1);
```
Ghostkeeper, line 37: I'm not sure how this would fix the issue we're debugging now, since it's a case where we have an odd number of walls, not even?

Also, this seems to be saying that if the line width is too small, we're going to try with a greater number of lines, which would result in an even smaller line width, right?

jellespijker (approved):

PR #1505, opened 2021-09-28
CURA-7976: Combing strategy to avoid top and bottom most surface skins

konskarm:
New combing mode "Not in Outer Surfaces" to avoid going through the skin, only on the bottom-most and too-most outer surfaces. Compared to the "Not in Skin" option, which avoids combing through all the skin surfaces regardless if they are visible, the "Not in Outer Surfaces" avoids only the skin surfaces that are actually exposed and will be visible in the printed model. This has the added benefit that it saves time while preserving visual quality.

The "Not in Outer Surfaces" option will be enabled by default in the following cases:
1. Monotonic Top/Bottom Order is enabled (in any extruder)
2. Ironing is enabled (in any extruder)
3. Top Surface Skin Layers > 0 and Monotonic Top Surface Order is enabled (in any extruder)

Backend of https://github.com/Ultimaker/Cura/pull/10536.

CURA-7976

### DIFF
--- a/src/skin.cpp
+++ b/src/skin.cpp
@@ -147,6 +147,8 @@ void SkinInfillAreaComputation::generateSkinsAndInfill()
         generateSkinInsetsAndInnerSkinInfill(&part);
 
         generateRoofing(part);
+
+        generateTopAndBottomMostSkinSurfaces(part);
     }
 }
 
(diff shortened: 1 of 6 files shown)


src/skin.cpp
```diff
@@ -545,6 +546,31 @@ Polygons SkinInfillAreaComputation::generateNoAirAbove(SliceLayerPart& part)
...
 
+/*
+ * This function is executed in a parallel region based on layer_nr.
+ * When modifying make sure any changes does not introduce data races.
+ *
+ * this function may only read the skin and infill from the *current* layer.
+ */
+    Polygons SkinInfillAreaComputation::generateNoAirBelow(SliceLayerPart& part, size_t flooring_layer_count)
+    {
+        const size_t wall_idx = std::min(size_t(2), mesh.settings.get<size_t>("wall_line_count"));
```
jellespijker, line 557: Use constexpr and add early out
```suggestion
        if (layer_nr <= flooring_layer_count)
        {
            return {};
        }
        constexpr size_t min_wall_line_count = 2;
        const size_t wall_idx = std::min(min_wall_line_count, mesh.settings.get<size_t>("wall_line_count"));
```

src/skin.cpp
```diff
@@ -545,6 +546,31 @@ Polygons SkinInfillAreaComputation::generateNoAirAbove(SliceLayerPart& part)
...
+            {
+                for (int layer_nr_below = layer_nr - flooring_layer_count + 1; layer_nr_below < layer_nr; layer_nr_below++)
+                {
+                    Polygons outlines_below = getWalls(part, layer_nr_below, wall_idx);
+                    no_air_below = no_air_below.intersection(outlines_below);
+                }
+            }
+            return no_air_below;
+        }
+        return {};
```
jellespijker, line 571: Store `layer_nr - flooring_layer_count` in variable for reuse.
Don't do calculations as `for (int layer_nr_below = layer_nr - flooring_layer_count + 1;` in a for loop, most modern compilers should be able to make that distinction but in complex files/functions they can make mistakes and do the calculation for each pass.

The previous early out ensures that the subtractions never result in `< 0` so we can keep on using `size_t` removing implicit casts.

> **NOTE**  
> Untested code below

```suggestion
       const size_t non_flooring_layer = layer_nr - flooring_layer_count;
       Polygons no_air_below = getWalls(part, non_flooring_layer, wall_idx);
       if (!no_small_gaps_heuristic)
       {
           const size_t next_non_flooring_layer = non_flooring_layer + 1;
           for (size_t layer_nr_below = next_non_flooring_layer; layer_nr_below < layer_nr; layer_nr_below++)
           {
               Polygons outlines_below = getWalls(part, layer_nr_below, wall_idx);
               no_air_below = no_air_below.intersection(outlines_below);
           }
       }
       return no_air_below;
```

jellespijker (changes requested):
Looks good only some minor issues with the function `generateNoAirBelow``

PR #1517, opened 2021-10-28
Correctly fix the infill going through skin when infill overlap is set

nallath:
CURA-8644


Ghostkeeper:
This is not the approach that we're looking for. This indeed removes the infill overlapping with skin, but it also adds a gap between the infill and the walls if there is an overlap set, which is opposite of what the user would want. It effectively makes the setting negative.

PR #1519, opened 2021-11-01
Fix fuzzy skin overshoots

BagelOrb:
Fuzzy skin has a problem where it sometimes produces weird long overshoots:
https://github.com/Ultimaker/Cura/issues/10585

image

The problem can be made extra clear by setting the point distance high while the thickness should be low. In usch a case you would still expect a very low deviation from the surface, but it is actually rather high.
image

I know you will have to reimplement the feature for Arachne, but then you'll probably look at the old implementation first, so it might be nice to merge this fix or at least keep it in the back of your minds when working on CURA-7887

Looking just at the comment and doing the math you can see how the new line does adhere to the math in the comment on line 1179:
```
+    int64_t dist_last_point = ???; // so that p0p1_size - dist_last_point evaulates to dist_left_over - p0p1_size
```
the comment refers to line 1189 where we perform `dist_left_over = p0p1_size - dist_last_point;`.
The code in line 1179 is executed for the case where the for loop in between those two lines doesn't get executed because it's a short segment.

If you use the math in the comment to come up with what should be in the place of `???` then you'll also come to the conclusion that the old code was wrong and the new code is correct. Also given how the bug presents itself it was exactly this line of code which I was expecting to be wrong. So it makes a lot of sense to check this fix.

### DIFF
--- a/src/FffPolygonGenerator.cpp
+++ b/src/FffPolygonGenerator.cpp
@@ -1176,7 +1176,7 @@ void FffPolygonGenerator::processFuzzyWalls(SliceMeshStorage& mesh)
                 { // 'a' is the (next) new point between p0 and p1
                     Point p0p1 = p1 - *p0;
                     int64_t p0p1_size = vSize(p0p1);
-                    int64_t dist_last_point = dist_left_over + p0p1_size * 2; // so that p0p1_size - dist_last_point evaulates to dist_left_over - p0p1_size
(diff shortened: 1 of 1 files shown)


BagelOrb:
image

Never mind. The problem is in lines 1189 and 1187

Ghostkeeper:
Fixed through https://github.com/Ultimaker/CuraEngine/pull/1523.

PR #1523, opened 2021-11-02
Fuzzy skin fix

BagelOrb:
Fuzzy skin has a problem where it sometimes produces weird long overshoots:
https://github.com/Ultimaker/Cura/issues/10585

image

The problem can be made extra clear by setting the point distance high while the thickness should be low. In usch a case you would still expect a very low deviation from the surface, but it is actually rather high.
image

I know you will have to reimplement the feature for Arachne, but then you'll probably look at the old implementation first, so it might be nice to merge this fix or at least keep it in the back of your minds when working on CURA-7887

Totoro is happy with this PR:
image

At a way too high sample point distance you get this error which was always a problem in fuzzy skin, but at least no overshoots any more.
image

Dual fuzzy squirtle is ambivalent about these changes:
image

Surface image logo is very flashy with these changes:
image


rburema:
Devs: Was reviewed/is processed as internal ticket CURA-8675.

rburema:
@BagelOrb Nice pics! :-)

PR #1524, opened 2021-11-03
[CURA-8573] Replace OpenMP with native async for Support Generation

rburema:
Needed for Mac support of multi-threading.

### DIFF
--- a/src/TreeSupport.cpp
+++ b/src/TreeSupport.cpp
@@ -10,12 +10,15 @@
 #include "settings/EnumSettings.h"
 #include "settings/types/Angle.h" //Creating the correct branch angles.
 #include "settings/types/Ratio.h"
+#include "utils/algorithm.h"
 #include "utils/IntPoint.h" //To normalize vectors.
 #include "utils/logoutput.h"
 #include "utils/math.h" //For round_up_divide and PI.
 #include "utils/MinimumSpanningTree.h" //For connecting the correct nodes together to form an efficient tree.
 #include "utils/polygon.h" //For splitting polygons into parts.
 #include "utils/polygonUtils.h" //For moveInside.
+
+#include <mutex>
 
 #define SQRT_2 1.4142135623730950488 //Square root of 2.
 #define CIRCLE_RESOLUTION 10 //The number of vertices in each circle.
@@ -107,10 +110,10 @@ void TreeSupport::drawCircles(SliceDataStorage& storage, const std::vector<std::
     const double diameter_angle_scale_factor = sin(mesh_group_settings.get<AngleRadians>("support_tree_branch_diameter_angle")) * layer_height / branch_radius; //Scale factor per layer to produce the desired angle.
     const coord_t line_width = mesh_group_settings.get<coord_t>("support_line_width");
     const coord_t resolution = mesh_group_settings.get<coord_t>("support_tree_collision_resolution");
-    size_t completed = 0; //To track progress in a multi-threaded environment.
-#pragma omp parallel for shared(storage, contact_nodes)
-    // Use a signed type for the loop counter so MSVC compiles (because it uses OpenMP 2.0, an old version).
-    for (int layer_nr = 0; layer_nr < static_cast<int>(contact_nodes.size()); layer_nr++)
+
+    std::atomic<size_t> completed = 0; //To track progress in a multi-threaded environment.
+    std::mutex critical_sections;
+    cura::parallel_for<size_t>(0, contact_nodes.size(), 1, [&](const size_t layer_nr)
     {
         Polygons support_layer;
         Polygons& roof_layer = storage.support.supportLayers[layer_nr].support_roof;
@@ -186,23 +189,26 @@ void TreeSupport::drawCircles(SliceDataStorage& storage, const std::vector<std::
             storage.support.supportLayers[layer_nr].support_infill_parts.emplace_back(part, line_width, wall_count);
         }
 
-#pragma omp critical (support_max_layer_nr)
         {
+            std::lock_guard<std::mutex> critical_section_support_max_layer_nr(critical_sections);
+
             if (!storage.support.supportLayers[layer_nr].support_infill_parts.empty() || !storage.support.supportLayers[layer_nr].support_roof.empty())
             {
                 storage.support.layer_nr_max_filled_layer = std::max(storage.support.layer_nr_max_filled_layer, static_cast<int>(layer_nr));
             }
         }
-#pragma omp atomic
-        completed++;
-#pragma omp critical (progress)
+
+        ++completed;
+
         {
+            std::lock_guard<std::mutex> critical_section_progress(critical_sections);
+
             const double progress_contact_nodes = contact_nodes.size() * PROGRESS_WEIGHT_DROPDOWN;
             const double progress_current = completed * PROGRESS_WEIGHT_AREAS;
             const double progress_total = completed * PROGRESS_WEIGHT_AREAS;
             Progress::messageProgress(Progress::Stage::SUPPORT, progress_contact_nodes + progress_current, progress_contact_nodes + progress_total);
         }
-    }
+    });
 }
 
 void TreeSupport::dropNodes(std::vector<std::vector<Node*>>& contact_nodes)
--- a/src/support.cpp
+++ b/src/support.cpp
@@ -23,6 +23,7 @@
 #include "settings/EnumSettings.h" //For EFillMethod.
 #include "settings/types/Angle.h" //To compute overhang distance from the angle.
 #include "settings/types/Ratio.h"
+#include "utils/algorithm.h"
 #include "utils/logoutput.h"
 #include "utils/math.h"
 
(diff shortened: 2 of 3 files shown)


Piezoid:
Similarly, I've re-implemented `GcodeLayerThreader` as I wasn't satisfied with the 500ms wait for work. I used `std::condition_variable` and a different logic to the produce out of order and consume in order. I got better CPU utilization and wall clock time. 
I kept the omp parallel used for spawning the workers, allowing nested omp parallelism to behave correctly. On the other hand, it is not recommended in principle to use std/thread synchronization primitives with the omp runtime (but there is no condition wait in omp, without tasks). In practice, I found that it's not problematic on gcc/linux, but YMMV.
If you are interested in this, what thread runtime would you recommend using at this level?

rburema:
> Similarly, I've re-implemented `GcodeLayerThreader` as I wasn't satisfied with the 500ms wait for work. I used `std::condition_variable` and a different logic to the produce out of order and consume in order. I got better CPU utilization and wall clock time. I kept the omp parallel used for spawning the workers, allowing nested omp parallelism to behave correctly. On the other hand, it is not recommended in principle to use std/thread synchronization primitives with the omp runtime (but there is no condition wait in omp, without tasks). In practice, I found that it's not problematic on gcc/linux, but YMMV. If you are interested in this, what thread runtime would you recommend using at this level?

Hi @Piezoid ,

We're going to replace all OpenMP instances with native threading, as it's a world of hassle to get that working on Mac (a sizeable chunk of our user-base). As a consequence, none of our Mac builds where using threading in the engine.

In order to not be completely overwhelmed & split chunks of programming into more predictable work, (and keep it reviewable), we decided to replace a small chunk first, to test the waters.

I suppose that does mean we're interested though! :-)

So I think that answers your question?: We're eventually just going to use native over OpenMP.

Piezoid:
Hi @rburema,

I have some reservations about using `std::async`. With many of the current implementations of stdc++, `parallel_for` immediately [spawns](https://ddanilov.me/std-async-implementations/#how-does-stdlaunchasync-work-in-different-implementations) one thread per layer. (One exception being MSCV's, where async calls `_Task_async_state` that start a task in what appears to be a thread pool). Beside the overhead of thread creation, it will create more CPU tasks than there is available CPU core, potentially resulting in scheduling overhead. This could cause scaling issues with bigger sliced objects.

I made a [branch](https://github.com/Piezoid/CuraEngine/blob/remove-openmp/src/utils/threadpool.h) that uses a minimalistic thread pool owned by the `Application` singleton. I have re-implemented both `parallel_for` and `GcodeLayerThreader` using it. OpenMP is entirely [stripped](https://github.com/Piezoid/CuraEngine/commit/9c6cdf91) from the code base. 

If you're ok with this approach, I will make a PR and a bit more cleanup.

rburema:
Hi @Piezoid !

Sorry for the slow response.

> With many of the current implementations of stdc++

Thanks for the link to the analysis!

That's on me for just assuming these sort-of higher-level calls have all that sort of stuff sorted out under the hood (like MSVC apparently already does -- though reading up on that, this _may_ make them technically not standard-compliant).

The only part that used `async` however, is hopefully  temporary. That said, like we always say in the team _"there is nothing more permanent than a temporary solution"_. So that's not a real argument.

Since there are no thread-pools used (internally), I think the general idea of limiting the number of threads is a decent/good one.

The general problem I have with this is:

This (as in, my original PR) was tested against the code the way it was before, and it already seems to be a little improvement over what the (ancient version of) OpenMP was doing. At the very least not worse. Possibly because each of the many tasks that are launched is quite large in an of itself (even though there may be many of them). I suspect the overhead of calling a new thread compared to what the tasks are actually doing is, in most cases, negligible.

My point here is: Is the increase in complexity worth it, especially since C++ threading is a subject where lots of things are still changing? (Though part of what makes the linked file larger is a function that doesn't seem to be used directly by the linked code.)

A more specific question is about this part of the code:

~~~
        else
        {   // Too many tasks queued: do work on the main thread
            func(index, lock);
        }
~~~

This is placed in a for-loop, that may run to full completion before any thread has had a chance to get going. That being the case; if one thread finishes before the entire list of parallel-for-each tasks is done, how will that thread ever pick a new one? Won't this just result in the majority of the tasks (all except the first `4n`, where `n` is the number of threads active in the pool) being executed on the main thread of typical scenario's? (Number of layers, or tree-support contact-points, etc.)

Given our use-cases (since the tasks are large, but expected to have roughly the same run-time if averaged over a large number of tasks), I think a better approach might be to divide the tasks into chunks of `n`, then just link all of those together, so that if task `i * n` finishes, it starts `(i + 1) * n` on the same thread. Then wait until those `n` threads have finished. I think that might be simpler to implement as well.

Piezoid:
> Hi @Piezoid !
> 
> Sorry for the slow response.
> 
> > With many of the current implementations of stdc++
> 
> Thanks for the link to the analysis!
> 
> That's on me for just assuming these sort-of higher-level calls have all that sort of stuff sorted out under the hood (like MSVC apparently already does -- though reading up on that, this _may_ make them technically not standard-compliant).
> 
> The only part that used `async` however, is hopefully temporary. That said, like we always say in the team _"there is nothing more permanent than a temporary solution"_. So that's not a real argument.
> 
> Since there are no thread-pools used (internally), I think the general idea of limiting the number of threads is a decent/good one.
> 
> The general problem I have with this is:
> 
> This (as in, my original PR) was tested against the code the way it was before, and it already seems to be a little improvement over what the (ancient version of) OpenMP was doing. At the very least not worse. Possibly because each of the many tasks that are launched is quite large in an of itself (even though there may be many of them). I suspect the overhead of calling a new thread compared to what the tasks are actually doing is, in most cases, negligible.

I agree that a least for the support code that was converted to async, the size of the works completed was enough to mitigate the 

> 
> My point here is: Is the increase in complexity worth it, especially since C++ threading is a subject where lots of things are still changing? (Though part of what makes the linked file larger is a function that doesn't seem to be used directly by the linked code.)
> 
> A more specific question is about this part of the code:
> 
> ```
>         else
>         {   // Too many tasks queued: do work on the main thread
>             func(index, lock);
>         }
> ```
> 
> This is placed in a for-loop, that may run to full completion before any thread has had a chance to get going. That being the case; if one thread finishes before the entire list of parallel-for-each tasks is done, how will that thread ever pick a new one? Won't this just result in the majority of the tasks (all except the first `4n`, where `n` is the number of threads active in the pool) being executed on the main thread of typical scenario's? (Number of layers, or tree-support contact-points, etc.)

No because the loop is synchronized with a mutex to the queue. It is held while 

> 
> Given our use-cases (since the tasks are large, but expected to have roughly the same run-time if averaged over a large number of tasks), I think a better approach might be to divide the tasks into chunks of `n`, then just link all of those together, so that if task `i * n` finishes, it starts `(i + 1) * n` on the same thread. Then wait until those `n` threads have finished. I think that might be simpler to implement as well.

rburema:
> I agree that a least for the support code that was converted to async, the size of the works completed was enough to mitigate

Looking at the rest of the code, I think we don't have that much small works.

> No because the loop is synchronized with a mutex to the queue. It is held

Ah yes, you're right. (If we go through with this) It's probably good to document that, because the connection to the max-in-queue isn't explicit since the extra optimization of running on the main thread when the thread-pool queue is full 'takes care' of that part as well.

Ok, I think the added complexity will be worth it if the speedup is significant (at least a few percent or so) to my original approach. Ideally, I'd like to have that tested because of the (hopefully negligible) overhead you added.

rburema:
Hi @Piezoid 

There are three main things here I think:

- I was originally misreading the code slightly, but your misfired email already set me straight, so the parts where you explain the code can be considered done now, I think.
- Your PR is written with a certain future in mind where we'd use this generally all throughout the code, for completely different purposes. The current proposed solution is written with the assumption that, when it comes to threading, any future use cases within the code base will be similar or small enough. (Though there's a chance your implementation will work better even under the latter assumptions.) -- I'm was a bit on the fence here. You seem to have done solid work and it would have been a shame to throw away potential optimizations, on the other hand, this feels close to the 'premature optimization' that we're always warned of, in a way.
- With 'complexity' I didn't mean so much big-Oh or other asymptotic notation related things -- I meant the complexity of the code, w.r.t. maintainability / readability. Thought that shouldn't be _that_ much of an issue with some added comments.

Also, keep in mind: Ideally, I'd just like to use the 'parallel for' and other such constructs that confirming C++ 17 implementations _should_ provide. That was in fact the original plan for this ticket. We'd probably have very little reason to use any self-rolled thread-pool if that was the case. I bring this up because we might switch 'back' to that once provided.

All of that said however (I had a quick think about how to do it differently, but you always end up with at least some kind of thread-safe queue or vector, and that point, why not go the full thread pool approach?) -- I think the potential benefits of your solution outweigh any negatives I can bring to bear. -- Especially since you said you already parallelized the rest? That would take care of a ticket for us 😁 

So, please make the PR, so we can see if it works for all platforms 🙂 

P.S.
> Maybe I don't get it, but as I understand this doesn't allows work redistribution between threads.

No you do get it. That is a downside of that approach, I just thought it might be an acceptable one. (Again, given our current use cases...)

P.P.S. I know its really small and a bit silly to point out, but it's been bothering me every time I reply to this: 'until' is spelled with only one 'l'.

PR #1530, opened 2021-11-16
[CURA-8702] Don't fill the entire outline when processing each part.

rburema:
~~Draft PR so we don't accidentally merge to 4.12 if we decide not to, otherwise good to be reviewed.~~ We're going to make the patch!

Lightning infill was first generated, then applied to the part. However, the entire fill would be applied to each part. If a model consisted of more than one part, the lightning paths would be retraced for each part of the model in that layer. Now it limits the output to the outlines that are currently processed. It's relatively efficient, since it only has to check one node of each tree against the outlines of a single part.

fixes Cura/#10793

### DIFF
--- a/src/infill/LightningLayer.cpp
+++ b/src/infill/LightningLayer.cpp
@@ -250,7 +250,7 @@ void LightningLayer::reconnectRoots
 }
 
 // Returns 'added someting'.
-Polygons LightningLayer::convertToLines(const coord_t line_width) const
+Polygons LightningLayer::convertToLines(const Polygons& limit_to_outline, const coord_t line_width) const
 {
     Polygons result_lines;
     if (tree_roots.empty())
@@ -260,7 +260,12 @@ Polygons LightningLayer::convertToLines(const coord_t line_width) const
 
     for (const auto& tree : tree_roots)
     {
-        tree->convertToPolylines(result_lines, line_width);
+        // If even the furthest location in the tree is inside the polygon, the entire tree must be inside of the polygon.
+        // (Don't take the root as that may be on the edge and cause rounding errors to register as 'outside'.)
+        if (limit_to_outline.inside(tree->getFurthestLocation()))
+        {
+            tree->convertToPolylines(result_lines, line_width);
+        }
     }
 
     // TODO: allow for polylines!
--- a/src/infill/LightningLayer.cpp
+++ b/src/infill/LightningLayer.cpp
@@ -262,7 +262,10 @@ Polygons LightningLayer::convertToLines(const Polygons& limit_to_outline, const 
     {
         // If even the furthest location in the tree is inside the polygon, the entire tree must be inside of the polygon.
         // (Don't take the root as that may be on the edge and cause rounding errors to register as 'outside'.)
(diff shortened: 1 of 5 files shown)


rburema:
Link back to internal tickets: This is CURA-8702, but will likely also fix CURA-8633.

src/infill/LightningTreeNode.cpp
```diff
@@ -37,6 +37,23 @@ const Point& LightningTreeNode::getLocation() const
...
+    for (const auto& child : children)
+    {
+        const Point cp{ child->getFurthestLocation() };
+        const coord_t dist2 = vSize2(cp - result);
+        if (dist2 >= furthest_dist2)
+        {
+            furthest_dist2 = dist2;
+            result = cp;
+        }
+    }
```
Ghostkeeper, line 53: This seems like the sort of stuff that `std::max_element` should be used for?

src/infill/LightningTreeNode.cpp
```diff
@@ -37,6 +37,23 @@ const Point& LightningTreeNode::getLocation() const
...
 }
 
+Point LightningTreeNode::getFurthestLocation() const
+{
+    Point result{ p };
+    coord_t furthest_dist2 = 0;
+    for (const auto& child : children)
+    {
+        const Point cp{ child->getFurthestLocation() };
+        const coord_t dist2 = vSize2(cp - result);
```
Ghostkeeper, line 47: This will measure the distance between the current child's farthest leaf and the previously farthest leaf, rather than between the current child and the node itself. I think you meant this:
```suggestion
        const coord_t dist2 = vSize2(cp - p);
```

src/infill/LightningTreeNode.cpp
rburema: Ah, yes, so you'd leave the `getFurthestLocation` function, but replace the loop and do the recursion in the comparator function argument.

src/infill/LightningLayer.cpp
```diff
@@ -260,7 +260,15 @@ Polygons LightningLayer::convertToLines(const coord_t line_width) const
 
     for (const auto& tree : tree_roots)
     {
-        tree->convertToPolylines(result_lines, line_width);
+        // If even the furthest location in the tree is inside the polygon, the entire tree must be inside of the polygon.
+        // (Don't take the root as that may be on the edge and cause rounding errors to register as 'outside'.)
+        constexpr coord_t epsilon = 5;
+        Point should_be_inside = tree->getLocation();
+        PolygonUtils::moveInside(limit_to_outline, should_be_inside, epsilon, epsilon * epsilon);
+        if (limit_to_outline.inside(should_be_inside))
```
Ghostkeeper, line 268: Effectively, `moveInside` already checks if it's inside. This extra check (`O(N)`) is unnecessary if we look at how far the `moveInside` needs to move the point instead of relying on the maximum move distance:

```suggestion
        PolygonUtils::moveInside(limit_to_outline, should_be_inside);
        if (vSize2(tree->getLocation() - should_be_inside) < epsilon * epsilon)
```

We decided however that since this is only a performance issue, we move that out of scope for now and make that improvement to Master, and merge this for 4.12.

PR #1533, opened 2021-11-23
Add SET_RPATH option to CMake

### LINKED ISSUE #1522: Redundant /usr/lib in CMAKE_INSTALL_RPATH
When compiling CuraEngine with the default `CMakeLists.txt`, `CMAKE_INSTALL_RPATH` is set to `CMAKE_INSTALL_PREFIX/lib`, which is redundant in most cases. Changing this, or adding an option to adjust this configuration setting would be helpful in efforts for better packaging of CuraEngine for linux distros, particularly [alpine](https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/26934). Before opening up a PR however, I'd like to make sure that I'm understanding what purpose these configuration options serve and maybe contribute to better documentation as there is nothing on `LIB_SUFFIX` anywhere from what I can find.

boomanaiden154:
Fixes #1522.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,10 +55,14 @@ if(APPLE AND CMAKE_CXX_COMPILER_ID MATCHES "Clang")
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libc++")
 endif()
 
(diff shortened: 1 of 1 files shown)


rburema (approved):

rburema:
Devs: Refer back to internal ticket CURA-8727

PR #1534, opened 2021-11-23
[CURA-8713] Revert "Prevent accumulation of error?"

rburema:
This reverts commit 55d0e58de3d8d22e4470431fc9b9f8992779dc93.

That commit was done as an unsure 'well if it can't harm' kind of way in order to _maybe_ halfway fix a bug that wasn't possibly wasn't affecting our prints and now appears to have been fixed in other ways, as I can't even reproduce it anymore. However, the now reverted fix may have casued the issue in CURA-8713 -- retracts in gcode preventing the UM2+ from finishing prints.

### DIFF
--- a/src/LayerPlan.cpp
+++ b/src/LayerPlan.cpp
@@ -1558,9 +1558,8 @@ void LayerPlan::writeGCode(GCodeExport& gcode)
     Communication* communication = Application::getInstance().communication;
     communication->setLayerForSend(layer_nr);
     communication->sendCurrentPosition(gcode.getPositionXY());
-
-    gcode.resetExtrusionValue();
     gcode.setLayerNr(layer_nr);
+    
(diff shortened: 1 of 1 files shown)


Ghostkeeper (commented):
This is indeed likely a fix, or a workaround around a firmware bug. `resetExtrusionValue` should then not be called when the material is retracted.

I looked for other usage of `resetExtrusionValue`. One is at the start of the print, which is safe. Another after an unretract, which is also not during a retraction. There were also two instances though where it's called during an extruder switch while the material is retracted. This is also potentially a problem when using firmware retraction, though not for the UM2+ where we encountered the issue (since it had only 1 extruder). Do you think we should fix that too?

Ghostkeeper:
After discussion, we've decided to defer that potential issue and keep it at this fix. Proceeding with testing.

PR #1536, opened 2021-11-28
Fixed extra z-hops in Surface slicing mode

IkonOne:
closes #1532 

Open poly lines were never cleaned up the same way that polygons are.

PR #1537, opened 2021-12-01
Prevent super small travel moves

nallath:
Instead of printing super small travel moves (which cause all kinds of issues with monotonic & different settings), just add an extrude move with zero extrusion.

CURA-8692

### DIFF
--- a/src/LayerPlan.cpp
+++ b/src/LayerPlan.cpp
@@ -1068,12 +1068,12 @@ void LayerPlan::addLinesByOptimizer
 (
     const Polygons& polygons,
     const GCodePathConfig& config,
-    SpaceFillType space_fill_type,
-    bool enable_travel_optimization,
-    int wipe_dist,
-    float flow_ratio,
-    std::optional<Point> near_start_location,
-    double fan_speed
+    const SpaceFillType space_fill_type,
+    const bool enable_travel_optimization,
+    const coord_t wipe_dist,
+    const Ratio flow_ratio,
+    const std::optional<Point> near_start_location,
+    const double fan_speed
 )
 {
     Polygons boundary;
@@ -1125,7 +1125,8 @@ void LayerPlan::addLinesByOptimizer
             // Instead of doing a small travel that is shorter than the line width (which is generally done at pretty high jerk & move) do a
             // "fake" extrusion move
             addExtrusionMove(p0, config, space_fill_type, 0, false, 1.0, fan_speed);
-        } else
+        }
+        else
         {
             addTravel(p0);
         }
@@ -1240,7 +1241,8 @@ void LayerPlan::addLinesMonotonic
             // Instead of doing a small travel that is shorter than the line width (which is generally done at pretty high jerk & move) do a
             // "fake" extrusion move
             addExtrusionMove(p0, config, space_fill_type, 0, false, 1.0, fan_speed);
-        } else
+        }
+        else
         {
             addTravel(p0);
         }
--- a/src/LayerPlan.cpp
+++ b/src/LayerPlan.cpp
@@ -1064,7 +1064,17 @@ unsigned LayerPlan::locateFirstSupportedVertex(ConstPolygonRef wall, const unsig
     }
 }
 
-void LayerPlan::addLinesByOptimizer(const Polygons& polygons, const GCodePathConfig& config, SpaceFillType space_fill_type, bool enable_travel_optimization, int wipe_dist, float flow_ratio, std::optional<Point> near_start_location, double fan_speed)
+void LayerPlan::addLinesByOptimizer
+(
+    const Polygons& polygons,
+    const GCodePathConfig& config,
+    SpaceFillType space_fill_type,
+    bool enable_travel_optimization,
+    int wipe_dist,
+    float flow_ratio,
+    std::optional<Point> near_start_location,
+    double fan_speed
+)
 {
     Polygons boundary;
     if (enable_travel_optimization && comb_boundary_inside2.size() > 0)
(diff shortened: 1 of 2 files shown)


IkonOne:
It appears that our PR's may be related: https://github.com/Ultimaker/CuraEngine/pull/1536

Would this be a potential fix for the issue I addressed.  And if so, would yours be the preferred approach?

nallath:
It could be, yeah. I'm going to have a look (although some other developers should also have a look, since one of the solutions is mine)

nallath:
They aren't quite fixing the same thing. The issue I was trying to solve is still present with your changes:

image

src/LayerPlan.cpp
```diff
@@ -1064,7 +1064,17 @@ unsigned LayerPlan::locateFirstSupportedVertex(ConstPolygonRef wall, const unsig
...
 }
 
-void LayerPlan::addLinesByOptimizer(const Polygons& polygons, const GCodePathConfig& config, SpaceFillType space_fill_type, bool enable_travel_optimization, int wipe_dist, float flow_ratio, std::optional<Point> near_start_location, double fan_speed)
+void LayerPlan::addLinesByOptimizer
+(
+    const Polygons& polygons,
+    const GCodePathConfig& config,
+    SpaceFillType space_fill_type,
+    bool enable_travel_optimization,
+    int wipe_dist,
```
Ghostkeeper, line 1073: Since `wipe_dist` is a distance, it should be a `coord_t`.

src/LayerPlan.cpp
```diff
@@ -1108,7 +1118,17 @@ void LayerPlan::addLinesByOptimizer(const Polygons& polygons, const GCodePathCon
...
             continue;
         }
-        addTravel(p0);
+
+        if(vSize2(getLastPlannedPositionOrStartingPosition() - p0) < line_width_2)
+        {
+            // Instead of doing a small travel that is shorter than the line width (which is generally done at pretty high jerk & move) do a
+            // "fake" extrusion move
+            addExtrusionMove(p0, config, space_fill_type, 0, false, 1.0, fan_speed);
+        } else
```
Ghostkeeper, line 1127: Code style: `else` on a new line.
```suggestion
        }
        else
```

src/LayerPlan.cpp
```diff
@@ -1212,7 +1233,15 @@ void LayerPlan::addLinesMonotonic
         {
             continue;
         }
-        addTravel(p0);
+        if(vSize2(current_last_position - p0) < line_width_2)
+        {
+            // Instead of doing a small travel that is shorter than the line width (which is generally done at pretty high jerk & move) do a
+            // "fake" extrusion move
+            addExtrusionMove(p0, config, space_fill_type, 0, false, 1.0, fan_speed);
+        } else
```
Ghostkeeper, line 1241: Code style: `else` on a new line.
```suggestion
        }
        else
```

src/LayerPlan.cpp
```diff
@@ -1064,7 +1064,17 @@ unsigned LayerPlan::locateFirstSupportedVertex(ConstPolygonRef wall, const unsig
...
 
-void LayerPlan::addLinesByOptimizer(const Polygons& polygons, const GCodePathConfig& config, SpaceFillType space_fill_type, bool enable_travel_optimization, int wipe_dist, float flow_ratio, std::optional<Point> near_start_location, double fan_speed)
+void LayerPlan::addLinesByOptimizer
+(
+    const Polygons& polygons,
+    const GCodePathConfig& config,
+    SpaceFillType space_fill_type,
+    bool enable_travel_optimization,
+    int wipe_dist,
+    float flow_ratio,
```
Ghostkeeper, line 1074: Should be a `Ratio`, right?

Ghostkeeper (commented):
A few minor things about the code itself. I can easily fix those.

I'm not sure if this change is entirely without risk. I'm going to need to test this well. One particular area of concern is how this influences filling small gaps, which uses travel moves to determine if the gap filling lines need to get merged together. And of course we'll need to test if there is no real-world case where a travel move could just be smaller than a line width where an actual travel is desired.

src/LayerPlan.cpp
nallath: I only changed the formatting here, but you're right!

src/LayerPlan.cpp
nallath: I only changed the formatting here. But I think you're right.

Ghostkeeper:
One of the unintended consequences of this change is that the lines pattern starts to look weird and inconsistent based on rounding errors. They are spaced exactly 1 line width apart, so the threshold being at 1 line width makes them merge together if the rounding errors place the distance slightly lower.

image

It's not necessarily a big problem because the lines are printed with 0 flow. But it will be somewhat inconsistent with acceleration and jerk, a pattern that might become visible in the print. I did a test print of a tiny cube and see something that could be the result of this (a vertical stripe of stronger ringing), but it's edging on noise because of the other ringing involved. We discussed in a chat that this could be solved by slightly lowering the threshold to be below 1 line width. It'll work for most cases as normal.

A more serious issue is that this indeed interferes with the small gap filling. Not always, but in some cases. This is a print before this change:
image

And this is how it looks after this change:
image

This is done with the default settings of the UMS5. Here is a project file: [UMS5_web.zip](https://github.com/Ultimaker/CuraEngine/files/7643112/UMS5_web.zip)

For this last one we'll need to really look into this issue again for a fix or a better solution.

nallath:
Fortunately, this seems to be one of the cases where a single solution fixes both problems. By reducing the the distance where the travel is removed (from a full line width to half a line width) both issues are fixed. 

It should now never interfere with the gap filling, as that can never have travel moves that are longer than half the line width in the first place, so the merging should work as intended again.

Ghostkeeper (approved):
Plausible solution! Will continue to testing.

PR #1538, opened 2021-12-04
Improve mesh loading

Piezoid:
This is a work in progress of a re-implementation of vertices merging during mesh loading, made to be more precise and more efficient. A 3DBenchy no longer triggers the warning `Mesh has overlapping faces!`.

**Background:** `Mesh::addFace` lockups vertices in a spatial hash map. This recovers the connectivity while accommodating for inaccuracies in STLs. Vertices closer than a 0.03 mm are merged together.

Previously, a vertex was inserted into a 0.03 mm wide cubic bucket. During lookup, a single bucket is searched for a neighboring vertex. Vertices close the the side of one cube will miss merge candidates from neighboring cubes. This could lead to unexpected behavior, and exacerbate the order sensitivity of the greedy merge algorithm.

This implementation uses half-open 0.064mm wide cubic buckets for merging vertices closer than 0.032 mm. It is optimized for queries: on insertion, a vertex is inserted in all of the 8 buckets overlapping with the 0.032mm radius sphere centered on the vertex. A query gets all candidates vertices from a single bucket.

For this purpose I wrote a specialized hash map. Iterating over a bucket is accomplished by iteratively probing the hash map. This gets some collisions during lookup, but they are filtered by the distance checks anyway. Using `std::unordered_map`, as in the current implementation, requires an indirection to dynamically sized buckets and can't combine collision checks with distance checks.

The hash map implementation is minimalist but supports rehashing when the number of vertices can't be estimated beforehand. Some cache locality is recovered from the inherent spatial locality of meshes: 
 * Point coordinates are hashed to their [Morton codes](https://en.wikipedia.org/wiki/Z-order_curve),
 * The probing strategy is copied from [Python's dict](https://github.com/python/cpython/blob/master/Objects/dictnotes.txt#L120),
 * Load factor is 2/3, growth factor is 4. Map sizes are powers of two, starting from 2^10.

This was a fun exercise, but I'd like to know if there is an interest in this before polishing it.
I'm in the process of benchmarking, and experimenting with different strategies and constants. (I will update with the results). But overall, the one above seems to perform the best for it's simplicity.

### DIFF
--- a/include/mesh.h
+++ b/include/mesh.h
@@ -77,7 +77,7 @@ public:
 
     /*!
      * Insert a new 3D point into the map and returns its identifier.
-     * If a previously inserted point is found in a sphere of vertex_meld_distance radius, returns its identifier instead.
+     * If a previously inserted point is found, returns its identifier instead.
      * \param vertices Vector mapping identifiers to already inserted points. New identifier are generated from it's size. Read when rehashing.
      */
     Item insert(const Point3& p, const std::vector<MeshVertex>& vertices);
@@ -90,13 +90,6 @@ private:
     uint8_t bits = min_bits; //!< log2 size of map
 
     void init(); //!< Allocate the map. bits must be set beforehand.
-
-    /*!
-     * Inserts a point and its associated value in each of the 8 cubes overlapping
-     * the sphere of radius=vertex_meld_distance centered on the point.
-     * \param value Vertex identifier plus one (0 is the marker for empty slot)
-     */
-    void insert8(const Point3& p, Item value);
 
     /*!
      * Probing primitive.
--- a/include/mesh.h
+++ b/include/mesh.h
@@ -6,8 +6,9 @@
 
 #include "settings/Settings.h"
 #include "utils/AABB3D.h"
-#include "utils/floatpoint.h"
 #include "utils/FMatrix4x3.h"
+#include "utils/NoCopy.h"
+#include "utils/floatpoint.h"
 
 namespace cura
 {
@@ -59,7 +60,7 @@ A Mesh is the most basic representation of a 3D model. It contains all the faces
 
 See MeshFace for the specifics of how/when faces are connected.
 */
-class Mesh
+class Mesh : public NoCopy
 {
     //! The vertex_hash_map stores a index reference of each vertex for the hash of that location. Allows for quick retrieval of points with the same location.
     std::unordered_map<uint32_t, std::vector<uint32_t> > vertex_hash_map;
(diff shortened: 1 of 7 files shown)


Ghostkeeper:
> Thus, maybe the implementation should only perform strict merging, and then rely on the stitching of 2D polygons to close malformed paths. This would give another ~8x improvement in memory usage and CPU time to the `HasMap3D` implementation.

I agree with this. Since we're using it for debugging, it is nice if the mesh loading in CuraEngine works the same way as the mesh loading in Cura's front-end. This uses numpy-stl which does no deduplication. Improved performance and simplicity are bonus then.

Piezoid:
I've benchmarked the **mesh loading code** with a large STL (383M binary STL, 5.4M vertices and 8M faces). Mesh is loaded through libArcus. To skip the rest of the slicing process, a `exit(3)` is added at the end of `ArcusCommunication::Private::readMeshGroupMessage`. 
For each test `/usr/bin/time -v ./CuraEngine connect 127.0.0.1:49675` is ran three times, taking the run with the best wall-clock time.

`master` ran in **8.35s** wall-clock (7s user, 1.1s system) with a maximum RSS of **1.21Go** and 755K page faults.

The present code in this PR, with vertices merging, took 5.84s (4.53s user, 0.91s system) with a maximum RSS of 1.09Go and 606K page faults. This is a 30% reduction of run time compared to master.

Removing the merging of close vertices, the run time drops to 4.24s (3.29s user, 0.77s system) with a maximum RSS of 1.1Go and 512K page faults. This adds a reduction of 27%, (49% compared to `master`)

I noticed that the maximum RSS was not affected by skipping deduplication. This is because the peak is reached during the `shrink_to_fit()` call on vertices and faces vectors, which double the memory occupation during the copy.
I added guards that only performs shrinking when more than 1/3 of the space is wasted by over-allocations. No shrinking happends when using libArcus which provide the number of faces beforehand.
This lowers the runtime to **3.96s** (3.20s user, 0.47s system) with a peak RSS of **824Mo** and 388K page faults. The size of allocations inside `Mesh` are comparable to those happening inside libArcus/Protobuf.

This was best improvement I could get by tuning this change set: **53%** reduction in mesh loading time compared to `master`.

The tests above were performed using `boost::container::small_vector`
to store the connected faces ids in instances of `MeshVertex`.
Using `std::vector` instead increases the runtime to 5.02s (+27%, 4.20s user, 0.55s system) with a peak RSS of 886Mo and 542K page faults. Memory usage and page faults increase a bit from fragmentation.


Running the **full slicing** process on the test STL with the fdmprinter's Normal - 0.15 mm profile, took 14.9s (137.6s user, 2.17s system) with a peak RSS of 1.4Go and 844K page faults, resulting in 292 layers. This is with the optimized mesh loading and `small_vector`.
Doing the same with `master` took 20.8s (143.2s user, 2.35s system) with a peak RSS of 1.93Go and 1.53M page faults.
Again, I took the best of 3 runs, but this time the results were much more volatile because of the multithreaded scheduling (on a old NUMA machine, 2x 6C12T). Still, I wasn't expecting a 27% improvement, something could be off...

Here's a screenshot of what it went through :sweat_smile:  
image

PR #1541, opened 2021-12-10
[CURA-8079] Alternate Wall Directions

rburema:
Adds a setting to alternate the direction a wall is printed every other layer (and inset, if applicable).

See also front-end pull: https://github.com/Ultimaker/Cura/pull/11049

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -1363,7 +1363,7 @@ void FffGcodeWriter::addMeshLayerToGCode(const SliceDataStorage& storage, const 
     {
         part_order_optimizer.addPolygon(&part);
     }
-    part_order_optimizer.optimize();
+    part_order_optimizer.optimize(gcode_layer.getLayerNr() % 2 == 0);
     for(const PathOrderOptimizer<const SliceLayerPart*>::Path& path : part_order_optimizer.paths)
     {
         addMeshPartToGCode(storage, mesh, extruder_nr, mesh_config, *path.vertices, gcode_layer);
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -1363,7 +1363,8 @@ void FffGcodeWriter::addMeshLayerToGCode(const SliceDataStorage& storage, const 
     {
         part_order_optimizer.addPolygon(&part);
     }
-    part_order_optimizer.optimize(gcode_layer.getLayerNr() % 2 == 0);
+    const bool should_alternate_walls = mesh.settings.get<bool>("material_alternate_walls") && (gcode_layer.getLayerNr() % 2 == 0);
+    part_order_optimizer.optimize(should_alternate_walls);
     for(const PathOrderOptimizer<const SliceLayerPart*>::Path& path : part_order_optimizer.paths)
     {
         addMeshPartToGCode(storage, mesh, extruder_nr, mesh_config, *path.vertices, gcode_layer);
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -1363,8 +1363,7 @@ void FffGcodeWriter::addMeshLayerToGCode(const SliceDataStorage& storage, const 
     {
         part_order_optimizer.addPolygon(&part);
     }
-    const bool should_alternate_walls = mesh.settings.get<bool>("material_alternate_walls") && (gcode_layer.getLayerNr() % 2 == 0);
-    part_order_optimizer.optimize(should_alternate_walls);
+    part_order_optimizer.optimize();
     for(const PathOrderOptimizer<const SliceLayerPart*>::Path& path : part_order_optimizer.paths)
     {
         addMeshPartToGCode(storage, mesh, extruder_nr, mesh_config, *path.vertices, gcode_layer);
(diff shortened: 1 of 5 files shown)


casperlamboo:
Suggestion from @nallath; might be a good time to boyscout `src/PathOrderOptimizer.h` and `src/LayerPlan.h` by moving all logic to the `cpp` files rather then the header files.

src/PathOrderOptimizer.h
```diff
@@ -179,18 +179,25 @@ class PathOrderOptimizer
...
         if(paths.empty())
         {
             return;
         }
 
+        const std::function<Point(const Point&)> maybe_mirror =
+            [&mirrored](const Point& p)
+            {
+                return Point{ p.X, p.Y * (mirrored ? -1 : 1) };
+            };
```
casperlamboo, line 193: I'm not entirely sure what this function aims to do, if I'm not correct let me know. The aim of this mirror function is to not only reverse the order of the polygon itself, but also reverse the order of the polygons itself. I'm guessing this is specifically needed for infill as the print direction of these extrusions should also alternate.

I'm not sure if this solves the issue though. Conciser the image attached. The tool paths are traversed as if they were in the mirrored position, however using the "travel to closest endpoint" path optimizing strategy does not result in an reversed order of tool paths for this case; in fact, the order is not changed at all.
Untitled Diagram

May I suggest two alternative approach
1. Instead of mirroring move the point to the _opposite location of the build plate_. So `point(X, Y) -> point(build_plate_width - X, build_plate_height - Y)`. Problem with this solution is that this wouldn't work as well if the starting location is close to the center of the build plate. (also as the distances between mirrored points is the same as before mirroring it is semantically the same to _only_ mirror the starting position for alternating layers)
2. Calculate the optimize path order normally (without mirror), and if the tool paths should be reversed reverse the order of the polygons in the optimized order, and for all open polygons reverse the order of the points. The initial travel moves will be a bit illogical as a result. but that is a compromise.

casperlamboo (commented):
@nallath suggest to apply some boyscouting as well regarding the `src/LayerPlan.h` and `src/PathOrderOptimizer.h` files, and move logic from the header to the cpp files

Ghostkeeper:
> @nallath suggest to apply some boyscouting as well regarding the src/LayerPlan.h and src/PathOrderOptimizer.h files, and move logic from the header to the cpp files

While this is possible for LayerPlan, it is not possible for PathOrderOptimizer because it is a template class. For LayerPlan it really depends on how far you'd take the boyscouting. Typically I'd only boyscout briefly around the code that was touched, not, like, the entire file which could lead to much confusion later if we look through the blames.

src/LayerPlan.cpp
```diff
@@ -1029,7 +1051,8 @@ void LayerPlan::addWalls(const PathJunctions& walls, const Settings& settings, c
     }
 
     cura::Point p_end {0, 0};
-    order_optimizer.optimize();
+    const bool alternate_walls = settings.get<bool>("material_alternate_walls") && (layer_nr % 2 == (alternate_direction_modifier ? 1 : 0));
```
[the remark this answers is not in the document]
Ghostkeeper: To me the tertiary there is more clear than an xor. When I read `layer_nr % 2 == 0` it makes me read "on even layers" and `layer_nr % 2 == 1` it makes me read "on odd layers". The tertiary to switch between even and odd is pretty clear.

But I do agree with the rename of the variable, and it could also use some documentation regardless.

Implemented in 93eb68413031a84741a259cc29a809f7ec19ceda.

src/FffGcodeWriter.cpp
```diff
@@ -1363,7 +1363,8 @@ void FffGcodeWriter::addMeshLayerToGCode(const SliceDataStorage& storage, const
     {
         part_order_optimizer.addPolygon(&part);
     }
-    part_order_optimizer.optimize();
+    const bool should_alternate_walls = mesh.settings.get<bool>("material_alternate_walls") && (gcode_layer.getLayerNr() % 2 == 0);
```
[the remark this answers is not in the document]
Ghostkeeper: In particular, another reason to reverse the path is for polylines to have the wrong endpoint be the closer one for travel optimisation.

src/PathOrderOptimizer.h
Ghostkeeper: I'm not really sure I see a reason to reverse the order of open polylines at all. The infill is not expected to change direction, is it?

Regardless of in which direction the polygons are printed, you'd start and end on the same spot, so for the travel length optimisation it shouldn't matter.

So I'm inclined to go with your second option but also simplify it to not adjust open polylines any more.

Ghostkeeper:
The current solution also causes the open side of open polylines to get closed sometimes, because it shifts the starting vertex. This will need adjusting in where the optimised paths are consumed.
image

src/PathOrderOptimizer.h
rburema: Great analysis.

I think this can be removed if it gives us cause to think this deeply about it: It was just something extra that isn't in the acceptance criteria; propose to remove.

Ghostkeeper: This is changed in 8368fc28a90fb457202be19483c2f3799bd40b6c. The new solution is hopefully simpler, although it does involve another copy of the `Path` instance.

This turns out to have been the cause of the misplaced lines as well. With polylines, the direction of the line is calculated based on the optimal starting vertex. This was reversed as well, causing the direction of polylines to be the wrong way around again. In the new implementation everything is reversed exactly once so the extra lines are gone.

nallath:
> Also, typically I'd only boyscout briefly around the code that was touched, not, like, the entire file which could lead to much confusion later if we look through the blames. 

I prefer having a cleaner code over having an extra click for the blame.

nallath (approved):

PR #1545, opened 2021-12-16
Cura 8080 alternate direction of support walls and support pattern

casperlamboo:
WIP

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -2638,6 +2638,8 @@ bool FffGcodeWriter::processSupportInfill(const SliceDataStorage& storage, Layer
             setExtruder_addPrime(storage, gcode_layer, extruder_nr); // only switch extruder if we're sure we're going to switch
             gcode_layer.setIsInside(false); // going to print stuff outside print object, i.e. support
 
+            const bool alternate_layer_print_direction = gcode_layer.getLayerNr() % 2 == 1;
+
             if (! wall_toolpaths.empty())
             {
                 const BinJunctions bins = InsetOrderOptimizer::variableWidthPathToBinJunctions(wall_toolpaths);
@@ -2661,8 +2663,17 @@ bool FffGcodeWriter::processSupportInfill(const SliceDataStorage& storage, Layer
 
             if (!support_lines.empty())
             {
+                const bool enable_travel_optimization = false;
+                const coord_t wipe_dist = 0;
+                const Ratio flow_ratio = 1.0;
+                const std::optional<Point> near_start_location = std::optional<Point>();
+                const double fan_speed = GCodePathConfig::FAN_SPEED_DEFAULT;
+                const bool reverse_print_direction = infill_extruder.settings.get<bool>("material_alternate_walls") && alternate_layer_print_direction;
+
                 gcode_layer.addLinesByOptimizer(support_lines, gcode_layer.configs_storage.support_infill_config[combine_idx],
-                                                (support_pattern == EFillMethod::ZIG_ZAG) ? SpaceFillType::PolyLines : SpaceFillType::Lines);
+                                                (support_pattern == EFillMethod::ZIG_ZAG) ? SpaceFillType::PolyLines : SpaceFillType::Lines,
+                                                enable_travel_optimization, wipe_dist, flow_ratio, near_start_location, fan_speed, reverse_print_direction);
+
                 added_something = true;
             }
         }
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -2654,8 +2654,8 @@ bool FffGcodeWriter::processSupportInfill(const SliceDataStorage& storage, Layer
                     constexpr bool always_retract = false;
                     const bool alternate_inset_direction_modifier = alternate_inset_direction && (alternate_layer_print_direction ^ alternate_bin_direction);
                     alternate_bin_direction = !alternate_bin_direction;
-                    gcode_layer.addWalls(paths, infill_extruder.settings, gcode_layer.configs_storage.support_bottom_config,
-                                         gcode_layer.configs_storage.support_bottom_config, z_seam_config, wall_0_wipe_dist,
+                    gcode_layer.addWalls(paths, infill_extruder.settings, gcode_layer.configs_storage.support_infill_config[0],
+                                         gcode_layer.configs_storage.support_infill_config[0], z_seam_config, wall_0_wipe_dist,
                                          flow_ratio, always_retract, alternate_inset_direction_modifier);
                 }
                 added_something = true;
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -2652,11 +2652,10 @@ bool FffGcodeWriter::processSupportInfill(const SliceDataStorage& storage, Layer
                     constexpr coord_t wall_0_wipe_dist = 0;
                     constexpr float flow_ratio = 1.0;
                     constexpr bool always_retract = false;
-                    const bool alternate_inset_direction_modifier = alternate_inset_direction && (alternate_layer_print_direction ^ alternate_bin_direction);
-                    alternate_bin_direction = !alternate_bin_direction;
                     gcode_layer.addWalls(paths, infill_extruder.settings, gcode_layer.configs_storage.support_infill_config[0],
                                          gcode_layer.configs_storage.support_infill_config[0], z_seam_config, wall_0_wipe_dist,
-                                         flow_ratio, always_retract, alternate_inset_direction_modifier);
+                                         flow_ratio, always_retract, alternate_bin_direction);
+                    alternate_bin_direction = !alternate_bin_direction;
                 }
                 added_something = true;
             }
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -2642,12 +2642,18 @@ bool FffGcodeWriter::processSupportInfill(const SliceDataStorage& storage, Layer
 
             if (! wall_toolpaths.empty())
             {
+                bool alternate_inset_print_direction = false;
                 const BinJunctions bins = InsetOrderOptimizer::variableWidthPathToBinJunctions(wall_toolpaths);
                 for (const PathJunctions& paths : bins)
                 {
                     for (const LineJunctions& line : paths)
                     {
-                        gcode_layer.addInfillWall(line, gcode_layer.configs_storage.support_infill_config[0], false);
+                        const bool reverse_order = infill_extruder.settings.get<bool>("material_alternate_walls") &&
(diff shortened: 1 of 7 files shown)


src/InsetOrderOptimizer.cpp
```diff
@@ -216,6 +218,25 @@ BinJunctions InsetOrderOptimizer::variableWidthPathToBinJunctions(const Variable
             }
         }
     }
+
+    if (alternate_inset_direction) {
```
jellespijker, line 222: code-style bracket on newline
```suggestion
    if (alternate_inset_direction)
    {
```

src/InsetOrderOptimizer.cpp
```diff
@@ -216,6 +218,25 @@ BinJunctions InsetOrderOptimizer::variableWidthPathToBinJunctions(const Variable
...
+        bool alternate_inset_print_direction = reverse_initial_inset;
+        for (PathJunctions &paths: insets) {
+            for (LineJunctions &line: paths) {
+                if (alternate_inset_print_direction) {
```
jellespijker, line 226: code-style brackets on newline
```suggestion
        for (PathJunctions &paths: insets)
        {
            for (LineJunctions &line: paths)
            {
                if (alternate_inset_print_direction)
                {
```

src/InsetOrderOptimizer.cpp
```diff
@@ -216,6 +218,25 @@ BinJunctions InsetOrderOptimizer::variableWidthPathToBinJunctions(const Variable
...
+                    int left = 0;
+                    int right = line.size() - 1;
+                    while (left < right) {
```
jellespijker, line 229: code-style bracket on newline
```suggestion
                    while (left < right)
                    {
```

src/utils/ExtrusionJunction.cpp
```diff
@@ -20,4 +20,17 @@ ExtrusionJunction::ExtrusionJunction(const Point p, const coord_t w, const coord
       region_id(region_id)
 {}
 
+void ExtrusionJunction::Swap(ExtrusionJunction* other) {
```
jellespijker, line 23: code-style bracket on newline and naming, we use lowerCamelCase for functions:
```suggestion
void ExtrusionJunction::swap(ExtrusionJunction* other)
{
```

src/utils/ExtrusionJunction.cpp
```diff
@@ -20,4 +20,17 @@ ExtrusionJunction::ExtrusionJunction(const Point p, const coord_t w, const coord
...
+  if (other == this) return;
+  InternalSwap(other);
+}
+void ExtrusionJunction::InternalSwap(ExtrusionJunction* other) {
```
jellespijker, line 27: code-style newline between functions, bracket on newline and naming, we use lowerCamelCase for functions
```suggestion
}

void ExtrusionJunction::internalSwap(ExtrusionJunction* other)
{
```

src/utils/ExtrusionJunction.h
```diff
@@ -47,6 +47,14 @@ struct ExtrusionJunction
     ExtrusionJunction(const Point p, const coord_t w, const coord_t perimeter_index, const size_t region_id = 0);
 
     bool operator==(const ExtrusionJunction& other) const;
+
+    void InternalSwap(ExtrusionJunction* other);
+
+    void Swap(ExtrusionJunction* other);
+    friend void swap(ExtrusionJunction& a, ExtrusionJunction& b)
```
jellespijker, line 54: newline between functions
```suggestion

    friend void swap(ExtrusionJunction& a, ExtrusionJunction& b)
```

src/utils/ExtrusionJunction.h
```diff
@@ -47,6 +47,14 @@ struct ExtrusionJunction
     ExtrusionJunction(const Point p, const coord_t w, const coord_t perimeter_index, const size_t region_id = 0);
 
     bool operator==(const ExtrusionJunction& other) const;
+
+    void InternalSwap(ExtrusionJunction* other);
```
jellespijker, line 51: code-style naming we use lowerCamelCase for functions:
```suggestion
    void internalSwap(ExtrusionJunction* other);
```

src/utils/ExtrusionJunction.h
```diff
@@ -47,6 +47,14 @@ struct ExtrusionJunction
...
+
+    void Swap(ExtrusionJunction* other);
```
jellespijker, line 53: code-style naming we use lowerCamelCase for functions:
```suggestion
    void swap(ExtrusionJunction* other);
```

src/utils/ExtrusionJunction.cpp
```diff
@@ -20,4 +20,17 @@ ExtrusionJunction::ExtrusionJunction(const Point p, const coord_t w, const coord
...
+  using std::swap;
+
+  swap(p, other->p);
+  swap(w, other->w);
+  swap(perimeter_index, other->perimeter_index);
+  swap(region_id, other->region_id);
```
jellespijker, line 33: I would make it absolutely clear this is the `std::swap` which is used for all four function call, certainly because due to our naming conventions (function: lowerCamelCase) we now have an actual class function called `swap(...)`
```suggestion
  std::swap(p, other->p);
  std::swap(w, other->w);
  std::swap(perimeter_index, other->perimeter_index);
  std::swap(region_id, other->region_id);
```

src/utils/ExtrusionJunction.cpp
```diff
@@ -20,4 +20,17 @@ ExtrusionJunction::ExtrusionJunction(const Point p, const coord_t w, const coord
       region_id(region_id)
 {}
 
+void ExtrusionJunction::Swap(ExtrusionJunction* other) {
+  if (other == this) return;
+  InternalSwap(other);
```
jellespijker, line 25: code-style naming we use lowerCamelCase for functions:
```suggestion
  internalSwap(other);
```

src/FffGcodeWriter.cpp
```diff
@@ -2655,14 +2665,33 @@ bool FffGcodeWriter::processSupportInfill(const SliceDataStorage& storage, Layer
...
+                gcode_layer.addPolygonsByOptimizer(support_polygons, gcode_layer.configs_storage.support_infill_config[combine_idx],
+                       z_seam_config, wall_0_wipe_dist, spiralize, flow_ratio, always_retract, reverse_order, start_near_location);
                 added_something = true;
             }
 
             if (!support_lines.empty())
             {
+                const bool enable_travel_optimization = false;
+                const coord_t wipe_dist = 0;
+                const Ratio flow_ratio = 1.0;
```
jellespijker, line 2686: use `constexpr`
```suggestion
                constexpr bool enable_travel_optimization = false;
                constexpr coord_t wipe_dist = 0;
                constexpr Ratio flow_ratio = 1.0;
```

src/FffGcodeWriter.cpp
```diff
@@ -2655,14 +2665,33 @@ bool FffGcodeWriter::processSupportInfill(const SliceDataStorage& storage, Layer
...
+                const std::optional<Point> near_start_location = std::optional<Point>();
+                const double fan_speed = GCodePathConfig::FAN_SPEED_DEFAULT;
```
jellespijker, line 2688: use `constexpr`
```suggestion
                constexpr double fan_speed = GCodePathConfig::FAN_SPEED_DEFAULT;
```

src/InsetOrderOptimizer.cpp
```diff
@@ -216,6 +218,25 @@ BinJunctions InsetOrderOptimizer::variableWidthPathToBinJunctions(const Variable
...
+                        swap(line[left], line[right]);
+                        left++;
+                        right--;
```
jellespijker, line 232: can be shortened to:
```suggestion
                        swap(line[left++], line[right--]);
```
Since the `++` and `--` values are in-/decremented after they are used. If you use the `++left` they are incremented before the value is used.
But I'm okay if you leave it like this.

jellespijker (changes requested):
A lot of code-style issues. Are you familiar with our Meta repository? https://github.com/Ultimaker/Meta/ 

I create suggestions for each change. Did you know that if you go into the `Files changed` tab on Github of this Pull Request, you can add each suggestion to the batch and create one big commit instead of multiple separate commits for each. 

I did notice some strange behavior in the ZigZag patterns. This is often used in the support interface layers. We seem to break up the lines and print them in an unlogical order on alternating layers. See the animated gif.

I also noticed that alternative patterns in the Support Interface don't follow the alternating patterning and are all printed in the same direction.

Peek 2021-12-20 13-40

src/InsetOrderOptimizer.cpp
Ghostkeeper: I shortened it even further by using `std::reverse` instead of this custom implementation.

src/utils/ExtrusionJunction.h
Ghostkeeper: Agreed, but I removed this code. Swap on POD-members like the ones in ExtrusionJunction will end up moving the data anyway. No need to specify that to swap this class, you need to swap each of its members.

src/utils/ExtrusionJunction.h
Ghostkeeper: Code was removed.

src/utils/ExtrusionJunction.h
Ghostkeeper: Code was removed.

src/utils/ExtrusionJunction.cpp
Ghostkeeper: While not in our code style standards, I agree that this is better practice.
Code was removed entirely, so this change is no longer necessary.

src/utils/ExtrusionJunction.cpp
Ghostkeeper: Code was removed, so this is no longer necessary.

src/InsetOrderOptimizer.cpp
```diff
@@ -164,7 +164,8 @@ size_t InsetOrderOptimizer::getOuterRegionId(const VariableWidthPaths& toolpaths
     return outer_region_id;
 }
 
-BinJunctions InsetOrderOptimizer::variableWidthPathToBinJunctions(const VariableWidthPaths& toolpaths, const bool pack_regions_by_inset, const bool center_last, std::set<size_t>* p_bins_with_index_zero_insets)
+BinJunctions InsetOrderOptimizer::variableWidthPathToBinJunctions(const VariableWidthPaths& toolpaths, const bool pack_regions_by_inset,
```
rburema, line 167: I'm not going to change it (since frankly it isn't _that_ important and we're inconsistant with it as well), but in C++ I think our code style is that if it's too long for a single line, then:

~~~
return-type className::functionName
(
    type parameter0,
    type parameter2,
    ....
)
~~~

casperlamboo: https://github.com/Ultimaker/CuraEngine/pull/1545/commits/f90088a316b15e675230ced08344dd2426262b09

rburema (approved):

PR #1547, opened 2021-12-21
Add GCode Flavor Used by some Flashforge Printers

STKFLT:
This PR adds a GCode flavor to support some Flashforge printers that use a modified Makerbot flavor (Flashforge/Monoprice Inventor I/II at least, not sure of others).

Most of the changes are pretty minor and are related to how multi-nozzle printing and temperature changes are handled. 

Everything was tested on a Flashforge Inventor I. The flavor might not be specific to flashforge and doesn't apply to all flashforge printers so it may make sense to change the name of the flavor.

I think I have followed all of the code styling suggestions. Please let me know if there's anything I need to add or clarify.

After this I will submit a PR that adds a profile using the Flavor to the Cura repo


Ghostkeeper:
Sorry, we have long since stepped away from having separate g-code flavours for all the slightly different firmwares out in the field. It would become unmaintainable. There are literally thousands of flavours we could add.

Instead, we're now going for having individual settings for the adjustments in the output g-code necessary. So e.g. in this case you would need to make a setting to switch between `T#` and `M108` or something.

> Currently CuraEngine still emits a "T<0|1>" command before the start GCode of my printer is inserted. I have to manually delete this line or the printer will reject the file because it needs "M108 T<0|1> instead. I wasn't able to identify where in the code that's added.

This is done here:
https://github.com/Ultimaker/CuraEngine/blob/92fd0abb7867ca3c86fdac67fc950497562c60f1/src/FffGcodeWriter.cpp#L626

Or here (for Griffin flavour):
https://github.com/Ultimaker/CuraEngine/blob/92fd0abb7867ca3c86fdac67fc950497562c60f1/src/FffGcodeWriter.cpp#L598

STKFLT:
Thanks for your reply - it definitely felt weird adding a new GCode flavor for whats probably only a couple printers so that policy makes total sense. The work was still really useful for figuring out what was broken. 

When I get a chance I will refactor everything into discrete settings. Do you have any examples of good PRs that add GCode settings? I just want to make sure I'm adding settings in a way that keeps things clear and maintainable.

Ghostkeeper:
I think this is a pretty clear one, and quite easy to understand: https://github.com/Ultimaker/CuraEngine/pull/1197

That PR adds a new machine setting, which can be separately enabled and disabled in the printer's definition file. And if it's enabled, it changes one of the g-code commands to also include the extruder it applies to.

Looking a bit more closely at it, I think it would be good to also read through this thread for the solutions we'd prefer to this problem: https://github.com/Ultimaker/Cura/issues/7395

PR #1548, opened 2021-12-22
Break long line segments in pieces if deviation is exceeded

Ghostkeeper:
This breaks a long line into multiple pieces if the Maximum Extrusion Area Deviation is exceeded. Long line segments weren't broken up by the Skeletal Trapezoidation if there was no additional rib in the skeletal graph. There would only be additional ribs if the number of lines changes, or if there is a vertex in the contour. This fixes that issue: It will now test if a line needs to be broken up into multiple pieces, by evaluating how much a line deviates in its area from the maximum deviation.

If it is exceeded, it will calculate how many pieces it needs to use to stay beneath the maximum extrusion area deviation. If this causes the pieces to be shorter than the maximum resolution, the maximum resolution is held. Thus this implements the following constraints:
* Lines should not get broken up into pieces smaller than the Maximum Resolution.
* If possible with the first constraint, lines should get broken up into the fewest amount of pieces needed to hold the Maximum Extrusion Area Deviation.

Each piece gets its line width calculated from the average width along its length.

Implements CURA-8170.

### DIFF
--- a/src/LayerPlan.cpp
+++ b/src/LayerPlan.cpp
@@ -821,11 +821,6 @@ void LayerPlan::addWall(const LineJunctions& wall, int start_idx, const Settings
     coord_t distance_to_bridge_start = 0; // will be updated before each line is processed
 
     const coord_t min_bridge_line_len = settings.get<coord_t>("bridge_wall_min_length");
-    const coord_t small_feature_max_length = settings.get<coord_t>("small_feature_max_length");
-    const bool is_small_feature = (small_feature_max_length > 0) && cura::shorterThan(wall, small_feature_max_length);
-    Ratio small_feature_speed_factor = settings.get<Ratio>((layer_nr == 0) ? "small_feature_speed_factor_0" : "small_feature_speed_factor");
-    const Velocity min_speed = fan_speed_layer_time_settings_per_extruder[getLastPlannedExtruderTrain()->extruder_nr].cool_min_speed;
-    small_feature_speed_factor = std::max((double)small_feature_speed_factor, (double)(min_speed / non_bridge_config.getSpeed()));
 
     const Ratio nominal_line_width_multiplier = 1.0 / Ratio(non_bridge_config.getLineWidth()); // we multiply the flow with the actual wanted line width (for that junction), and then multiply with this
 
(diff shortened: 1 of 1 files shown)


src/LayerPlan.cpp
```diff
@@ -921,26 +923,56 @@ void LayerPlan::addWall(const LineJunctions& wall, int start_idx, const Settings
...
+          Area Deviation setting, unless required by the first constraint.
+        Since breaking up a line segment into N pieces (each with averaged
+        width) divides the area deviation by N, we can simply check how many
+        pieces we'd want to get low enough deviation, then check if each piece
+        is not too short at the end.
+        */
+        const coord_t delta_line_width = p1.w - p0.w;
+        const Point line_vector = p1.p - p0.p;
+        const coord_t line_length = vSize(line_vector);
+        const coord_t line_area_deviation = std::abs(delta_line_width) * line_length / 8; //How much the line would deviate from the trapezoidal shape if printed at average width.
```
casperlamboo, line 953: When I calculate the area of the deviation I end up with 
```cpp
const coord_t line_area_deviation = std::abs(delta_line_width) * line_length / 4;


```
My reasoning:
Untitled Diagram drawio

src/LayerPlan.cpp
```diff
@@ -921,26 +923,56 @@ void LayerPlan::addWall(const LineJunctions& wall, int start_idx, const Settings
...
+        size_t pieces = std::max(size_t(1), round_up_divide(line_area_deviation, max_area_deviation)); //How many pieces we'd need to stay beneath the max area deviation.
```
casperlamboo, line 954: I have a slight preference for something like
```cpp
const size_t max_pieces_area_deviation = round_up_divide(line_area_deviation, max_area_deviation);
const size_t max_pieces_resolution = size_t(line_length / max_resolution);
const size_t pieces = std::max(size_t(1), std::min(max_pieces_area_deviation, max_pieces_resolution));
```
Results in equal or less computations, and is a bit clearer imo.

casperlamboo (commented):
Really nice clear code, was very readable and clean solution to something that looked complex at first. Had two small remarks. I came to a different result when calculating the area deviation, please see my comment. And I have a slightly different notation preference, please conciser, up to you if you think it is an improvement.

src/LayerPlan.cpp
casperlamboo: Actually, I believe we don't even need to calculate the deviation area. We can directly calculate how much the area deviates if the line is partitioned in `n` parts.

If a line is partitioned in `n` parts then `4n` of these deviation triangles are created. Each of this triangles have corresponding shapes with a width that is equal to `line_width / n` and a height that is equal to `abs(delta_width) / 4n` and a width that is equal to `line_length / n`. The area such a triangle is equal to 
```
triangle_area_deviation = (w * h) / 2 
                        = ((abs(delta_width) / 4n) * (line_length / n)) / 2 
                        = (abs(delta_width) * line_length) / (8n^2)
```
Multiply this by the number of triangles, `4n`, we get 
```
total_area_deviation = (abs(delta_width) * line_length) / (8n^2) * 4n 
                     = (abs(delta_width) * line_length) / (4n)
```
So if we partition a line into `n` parts the total area deviation will be `(abs(delta_width) * line_length) / (4n)`, Solve for `n`
```
total_area_deviation      = (abs(delta_width) * line_length) / (4n)
total_area_deviation * 4n = abs(delta_width) * line_length
                        n = (abs(delta_width) * line_length) / (total_area_deviation * 4)
```
Now we can compute how many pieces we need if we want to bound the total area deviation by `max_area_deviation` by simply substituting the variable.
```cpp
size_t pieces = round_up_divide(abs(delta_width) * line_length, max_area_deviation * 4)
```

Ghostkeeper: Yes, indeed I did something similar on a piece of paper when I got that formula. Maybe I mistakenly divided the length by 2 as well back then or something. I'm now arriving at `/ 4` as well.

src/LayerPlan.cpp
Ghostkeeper: I don't think this will make a noticeable difference in performance. I also don't feel strongly about which solution is easier to read so I've changed it here: ea3f1e00b53dbeb1e7ba05285e7bb0a7725c926d

PR #1354, opened 2020-11-06
Simplify polylines for Surface mode

BagelOrb:
Applies the Max Resolution settings to Surface mode as well.

Tested extremely zoomed in with an extremely small line width, so you can easily see the Max resolution setting at work.

Before:
image

After:
image

Project:
image
[tiny_open_mesh_surface_mode.3mf.rename.zip](https://github.com/Ultimaker/CuraEngine/files/5502373/tiny_open_mesh_surface_mode.3mf.rename.zip)

PR #1551, opened 2022-01-07
Simplify polylines for Surface mode

BagelOrb:
Rebase of https://github.com/Ultimaker/CuraEngine/pull/1354

I've verified that it still works and does the same

---- Copy of that PR's text

Applies the Max Resolution settings to Surface mode as well.

Tested extremely zoomed in with an extremely small line width, so you can easily see the Max resolution setting at work.

Before:
image

After:
image

Project:
image
[tiny_open_mesh_surface_mode.3mf.rename.zip](https://github.com/Ultimaker/CuraEngine/files/5502373/tiny_open_mesh_surface_mode.3mf.rename.zip)

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -1312,6 +1312,10 @@ void FffGcodeWriter::addMeshLayerToGCode_meshSurfaceMode(const SliceDataStorage&
         polygons.add(part.outline);
     }
 
+    coord_t max_resolution = mesh.settings.get<coord_t>("meshfix_maximum_resolution");
+    coord_t max_deviation = mesh.settings.get<coord_t>("meshfix_maximum_deviation");
+    polygons.simplify(max_resolution, max_deviation);
+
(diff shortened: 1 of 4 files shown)


Ghostkeeper (approved):
Looks good. Quite logical changes.

Ghostkeeper:
It seems to work, but I found an issue of which I'm unsure if it should be within the scope of this fix or not.
The polylines are indeed simplified correctly. However closed polygons are not simplified when using surface mode.

Here is a screenshot showcasing the problem:
image
Or perhaps this one, where it's in the same model but closed on different layers:
image
[Project file](https://github.com/Ultimaker/CuraEngine/files/7882088/surface_mode_simplification_in_loops.zip)

The model on the left is a cylinder with two slits on the side (making it not watertight). The model on the right is the same cylinder, but the original without the two slits. The third model is a cone section with 1 triangle missing (making it closed due to snapping in stitching at the top). The simplification is properly applied to the polylines , but as soon as the model creates a loop, it stops simplifying. The brim is properly simplified too however.

BagelOrb:
That issue has now been fixed.

Ghostkeeper (approved):

BagelOrb:
Just did a rebase, because the newer PR is based on a newer rebase of these same commits. This reduces future merge conflicts.

nallath (approved):

PR #1560, opened 2022-01-21
[CURA-8852] Revert "Don't re-apply Support Z Offset for support roof"

rburema:
Reverts the commit that's causing Tree Support to go through the model. Seems like this commit was meant as an extra clean-up after CURA-8635, so reverting it shouldn't damage the existing solution as far as I can see.

This reverts commit 62a61544f4067c3824dcb16c07effb2effd74c3f.


rburema:
Contains an other error. Also not based off of master maybe. Better to close this.

PR #1563, opened 2022-01-24
Cura 8852 tree support going through model

jellespijker:
Reverts the commit that's causing Tree Support to go through the model. Seems like this commit was meant as an extra clean-up after CURA-8635, so reverting it shouldn't damage the existing solution. But I had to take the support XY distance into account when  dropping nodes. Otherwise the difference operation could result in polygon parts which weren't connected to the branch and hanging in mid-air.

This reverts commit 62a6154.

Fixes ultimaker/cura#11258
Fixes ultimaker/cura#11306


jellespijker:
CI/CT is failing due to issues on the build server

nallath (approved):

PR #1565, opened 2022-01-28
Changes needed to support SIP 6.x

nallath:
We were still using SIP 4.x, but it's on end of life for quite some time now. 
https://github.com/Ultimaker/libArcus/pull/129
https://github.com/Ultimaker/Uranium/pull/776


jellespijker:
Close in favor of #1639

PR #1566, opened 2022-01-29
Replace OpenMP with custom thread pool

Piezoid:
Hi!

Following discussion in #1524, this PR:
* replaces uses of OpenMP with `cura::parallel_for`. After the first commit `GcodeLayerThreader` is the only piece of code still depending on OpenMP.
* Introduce a custom thread pool, with client functions like `parallel_for` (replacing rburema's original implementation) and `run_multiple_producers_ordered_consumer` (replacing `GcodeLayerThreader`).
* Removes OpenMP compiler flags.
* Fixes some race conditions found with `-fsanitize=thread`. For example, `sendLayerComplete` was [called in parallelized code](https://github.com/Ultimaker/CuraEngine/commit/43753e4b), causing some layers to show up at z=0 in the preview. Another race condition is mutation of state variables in Arcus' sockets (that's for another PR).

#### Thread pool vs. `std::async`
The original implementation of `parallel_for` use `std::async`, spawning a task for each item processed. Most stdc++ implementation don't reuse the spawned threads. On Linux and maxOS platforms [`pthread_create` is called for every items](https://ddanilov.me/std-async-implementations/#how-does-stdlaunchasync-work-in-different-implementations). This also over-provision the number of threads / cpus, with potential overhead from the OS scheduler trying to be fair. Memory consumption is also more difficult to manage with an unbounded number of tasks.

I tried to keep `ThreadPool` implementation minimalist. It is composed of a FIFO, a fleet of threads and some C++11 synchronization primitives. This class 

The `ThreadPool` lifetime is tied to `Application` singleton's. The communication classes sets the number of thread from `-m` command line argument, like they did with OpenMP.

#### Parallelization primitives
`ThreadPool` is not meant for direct consumption and should be used through functions like `parallel_for`. It's fairly easy to add other parallelization primitives, like an async runner producing `std::futures`, but there is currently no uses for it in the code base.

`parallel_for` split the range into chunks. By default it uses 8 chunk per worker. It's also possible to make the chunk size a multiple of a number. This allows code generating support stairs to handle layers with the same stair step in the same thread.

`run_multiple_producers_ordered_consumer` does exactly what did `GcodeLayerThreader` but utilizing the thread pool. In my opinion, the explicit synchronization between thread is better as it doesn't require spinning with `sleep_for`.

There is probably other alternatives and also concerns about using a custom implementation to resolve shortcomings in the standard library's implementation.
In the end, the first commit is probably what interests you the most.


src/utils/ThreadPool.h
```diff
@@ -0,0 +1,384 @@
...
+/// `std::make_signed_t` fails for non integral types in a way that doesn't allows SFINAE fallbacks. This alias solves that.
+template<typename T> using make_signed_if_integral_t = typename std::enable_if_t<std::is_integral_v<T>, std::make_signed<T>>::type;
+
+using std::distance;
+/// Overloads `std::distance()` to work on integral types
+template<typename Int, typename Signed=make_signed_if_integral_t<Int>>
+inline Signed distance(const Int& first, const Int& last)
+{
+    return static_cast<Signed>(last) - static_cast<Signed>(first);
+}
```
Piezoid, line 96: I'm not sure where to put these declarations.
This enables the same implementation of `parallel_for` to work on integers and random access iterators.
Also, this could be replaced with `<` and `-`.

src/utils/ThreadPool.h
```diff
@@ -0,0 +1,384 @@
...
+{
+    ThreadPool* thread_pool = Application::getInstance().thread_pool;
+    assert(thread_pool);
+    assert(max_pending_per_worker > 0);
+    const size_t max_pending = max_pending_per_worker * (thread_pool->thread_count() + 1);
+    MultipleProducersOrderedConsumer(first, last, std::forward<P>(producer), std::forward<C>(consumer), max_pending).run(*thread_pool);
+}
+
+template<typename Producer, typename Consumer>
+class MultipleProducersOrderedConsumer
```
Piezoid, line 236: This class stores user closures and state variables together such that they can referenced by spawned tasks through a single reference. `parallel_for` does the same but with a scoped struct having no methods.

State for `run_multiple_producers_ordered_consumer` is a bit more involved and I found that it's easier to use a full fledged class. On the other hand, this class should not be used directly as I don't want to pretend that there is other lifecycles than constructor->`run()`->destructor.

src/utils/logoutput.cpp
```diff
@@ -3,16 +3,15 @@
...
-#ifdef _OPENMP
-    #include <omp.h>
-#endif // _OPENMP
 #include "logoutput.h"
 
 namespace cura {
 
 static int verbose_level;
 static bool progressLogging;
+static std::mutex log_mutex;
```
Piezoid, line 14: Added a global mutex for standard output synchronization. Mainly to make `-fsanitize=thread` happy and less verbose.
This is not critical and can be omitted.

src/utils/macros.h
```diff
@@ -4,17 +4,20 @@
...
 #define RUN_ONCE(runcode) \
 { \
-    static bool code_ran = 0; \
+    static std::atomic<bool> code_ran = false; \
     if(!code_ran){ \
-        code_ran = 1; \
-        runcode; \
+        bool expected = false; \
+        if(code_ran.compare_exchange_strong(expected, true)) \
+        { runcode; }\
```
Piezoid, line 20: Same as above (`logoutput.cpp`). Not critical.

smartavionics:
> The original implementation of parallel_for use std::async, spawning a task for each item processed. Most stdc++ implementation don't reuse the spawned threads. On Linux and maxOS platforms [pthread_create is called for every items](https://ddanilov.me/std-async-implementations/#how-does-stdlaunchasync-work-in-different-implementations). This also over-provision the number of threads / cpus, with potential overhead from the OS scheduler trying to be fair. Memory consumption is also more difficult to manage with an unbounded number of tasks.

FYI, It has been noticed on a 32 bit ARM build of Cura 4.13 that the use of tree supports can cause aborts in the parallel_for code. As a workaround, I added the following:

```
diff --git a/src/utils/algorithm.h b/src/utils/algorithm.h
index f9f7e52f7..54ebcd68a 100644
--- a/src/utils/algorithm.h
+++ b/src/utils/algorithm.h
@@ -86,6 +86,17 @@ void parallel_for(T from, T to, T increment, const std::function<void(const T)>&
     for (size_t index = from; index != to; index += increment)
     {
         scope_guard.push_back(std::async(std::launch::async, func, index));
+#if __arm__
+        // hack to avoid crash on 32 bit ARM
+        if (scope_guard.size() == 25)
+        {
+            while (!scope_guard.empty())
+            {
+                // removing item will invoke destructor and cause task to be executed
+                scope_guard.pop_back();
+            }
+        }
+#endif
     }

```

Without that code, certain models+settings would always crash the slicer.

I note that, my current builds for x86_64, aarch64 and armhf all still abort occasionally but not in a repeatable fashion and I'm sure that's fairly new behaviour (arrived in the last couple of UM releases).

So I don't believe the current parallel_for implementation is reliable on Linux.

Hope this helps!

Piezoid:
> FYI, It has been noticed on a 32 bit ARM build of Cura 4.13 that the use of tree supports can cause aborts in the parallel_for code. 

Thank for this information. Do you get an error message?

By default on Linux, each new thread reserves 8 MiB of address space. On 32 bit systems, there is only 3 GiB available for userspace. It is expected to reach the limit quickly with enough slice layers.
Limiting the number of concurrent threads will indeed solve this problem, but in my opinion, it is best to avoid the overhead of thread creation by reusing them.

> I note that, my current builds for x86_64, aarch64 and armhf all still abort occasionally but not in a repeatable fashion and I'm sure that's fairly new behaviour (arrived in the last couple of UM releases).

This a bit more unexpected. Do you have an error message or stack trace of the failure?
Does this problem goes away with you tasks limiting patch, or with this PR?

smartavionics:
> Thank for this information. Do you get an error message?

So, without the above mentioned hack, certain models+settings will cause the 32 bit ARM engine to crash every time like this...

```
Thread 1 "CuraEngine" received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) where
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0xb6c4c230 in __GI_abort () at abort.c:79
#2  0x00424fcc in __gnu_cxx::__verbose_terminate_handler() ()
#3  0x003931b8 in __cxxabiv1::__terminate(void (*)()) ()
#4  0x0039322c in std::terminate() ()
#5  0x00393b84 in __cxa_rethrow ()
#6  0x0012f708 in std::future<std::result_of<std::decay<cura::parallel_for<unsigned int>(unsigned int, unsigned int, unsigned int, std::function<void (unsigned int)> const&)::{lambda(unsigned int)#1} const&>::type (std::result_of<unsigned int&>::type)>::type> std::async<cura::parallel_for<unsigned int>(unsigned int, unsigned int, unsigned int, std::function<void (unsigned int)> const&)::{lambda(unsigned int)#1} const&, unsigned int&>(std::launch, cura::parallel_for<unsigned int>(unsigned int, unsigned int, unsigned int, std::function<void (unsigned int)> const&)::{lambda(unsigned int)#1} const&, (std::decay<cura::parallel_for<unsigned int>(unsigned int, unsigned int, unsigned int, std::function<void (unsigned int)> const&)::{lambda(unsigned int)#1} const&>::type&&)...) ()
#7  0x00130b10 in void cura::parallel_for<unsigned int>(unsigned int, unsigned int, unsigned int, std::function<void (unsigned int)> const&) ()
#8  0x001260e0 in cura::AreaSupport::generateOverhangAreasForMesh(cura::SliceDataStorage&, cura::SliceMeshStorage&) ()
#9  0x0012627c in cura::AreaSupport::generateOverhangAreas(cura::SliceDataStorage&) ()
#10 0x00160ba0 in cura::FffPolygonGenerator::slices2polygons(cura::SliceDataStor--Type <RET> for more, q to quit, c to continue without paging--
age&, cura::TimeKeeper&) ()
#11 0x00161658 in cura::FffPolygonGenerator::generateAreas(cura::SliceDataStorage&, cura::MeshGroup*, cura::TimeKeeper&) ()
#12 0x0012060c in cura::Scene::processMeshGroup(cura::MeshGroup&) ()
#13 0x000c66d8 in cura::Slice::compute() ()
#14 0x00019d70 in cura::ArcusCommunication::sliceNext() ()
#15 0x00016294 in cura::Application::run(unsigned int, char**) ()
#16 0x00013af4 in main ()
(gdb) 

```


>This a bit more unexpected. Do you have an error message or stack trace of the failure?
Does this problem goes away with you tasks limiting patch, or with this PR?

With the task limiting patch, I am get crashes like the following on 32 bit ARM but not on every run...

```
Thread 1 "CuraEngine" received signal SIGTERM, Terminated.
__GI___nanosleep (remaining=0xbe8c4728, requested_time=0xbe8c4728)
    at ../sysdeps/unix/sysv/linux/nanosleep.c:28
28	../sysdeps/unix/sysv/linux/nanosleep.c: No such file or directory.
(gdb) where
#0  __GI___nanosleep (remaining=0xbe8c4728, requested_time=0xbe8c4728)
    at ../sysdeps/unix/sysv/linux/nanosleep.c:28
#1  __GI___nanosleep (requested_time=0xbe8c4728, remaining=0xbe8c4728)
    at ../sysdeps/unix/sysv/linux/nanosleep.c:25
#2  0x00072c68 in cura::GcodeLayerThreader<cura::LayerPlan>::act() ()
#3  0x0005f24c in cura::GcodeLayerThreader<cura::LayerPlan>::run() [clone ._omp_fn.0] ()
#4  0xb6e03b60 in GOMP_parallel ()
   from /tmp/.mount_Cura-mAGVydU/usr/bin/libgomp.so.1
#5  0x00064be8 in cura::FffGcodeWriter::writeGCode(cura::SliceDataStorage&, cura::TimeKeeper&) ()
#6  0x00120eb0 in cura::Scene::processMeshGroup(cura::MeshGroup&) ()
#7  0x000c6eb8 in cura::Slice::compute() ()
#8  0x00019d70 in cura::ArcusCommunication::sliceNext() ()
#9  0x00016294 in cura::Application::run(unsigned int, char**) ()
#10 0x00013af4 in main ()
(gdb) 
```

And I do see occasional crashes on 64 bit ARM and x86_64 also and I'm sure these have only started occurring quite recently.

Ghostkeeper:
We had something similar planned also, but didn't start with that yet other than a try-out in support generation. Developers, see CURA-8369. I hope this helps get started, though it would be good to have the maintainers from Ultimaker involved too so that they can also maintain it better in the future and have a say in the architecture.

rburema:
Hi @Piezoid ... Sorry this took so very long. PR's have unfortunately taken a backseat during 5.0, but we should be picking up the pace again soon.

As you can see I made a separate PR, because there where some conflicts with main and it didn't out-of-the-box compile for Windows (`ssize_t` is posix, not 'standard') and Mac (compilation order was different, which laid bare that the a header was missing `#include vector`).

The good news is that, since it was in _just_ before the 5.1 feature-freeze, unless this causes issues, it will be in 5.1 :-)

See: https://github.com/Ultimaker/CuraEngine/pull/1681

PR #1576, opened 2022-02-10
print support walls, even if there are no support infill areas

sebastiankuzminsky:
Hi there, we sometimes want to print supports with several walls, and this is currently broken in master.  If a support layer consists of only walls, no infill, then it does not get printed:

Screenshot from 2022-02-10 12-31-49

This PR fixes the issue:

Screenshot from 2022-02-10 12-33-12

I'm not sure this fix is the best way to resolve the issue and I welcome feedback.

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -2644,6 +2644,18 @@ bool FffGcodeWriter::processSupportInfill(const SliceDataStorage& storage, Layer
         // process sub-areas in this support infill area with different densities
         if ((default_support_line_distance <= 0 && support_structure != ESupportStructure::TREE) || part.infill_area_per_combine_per_density.empty())
         {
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Ah, thanks for taking the initiative.

We hadn't seen this PR yet, but we also made a fix last week for this issue: https://github.com/Ultimaker/CuraEngine/pull/1579
The issue was a bit more widespread than just support: It was also happening for infill. This was a critical issue for the Arachne release.

The fix I made for support is similar in its idea but has a bit of a different structure, which is why Github doesn't indicate this as a merge conflict. But we don't need this any more then, as the issue is already fixed.

Regardless, thanks for the contribution, even though we can't merge it any more now. It's nice to see people working along to solve the issues they are seeing.

sebastiankuzminsky:
All good, thanks for finding a better fix!  :-)

PR #1578, opened 2022-02-14
swap M104 and M109

rijkvanmanen:
The idle nozzle can start cooling down, while printing nozzle is heating up. This saves a few seconds of oozing, which is significant for materials like TPU. PP-122

### DIFF
--- a/src/LayerPlan.cpp
+++ b/src/LayerPlan.cpp
@@ -1706,11 +1706,6 @@ void LayerPlan::writeGCode(GCodeExport& gcode)
 
             const ExtruderTrain& extruder = Application::getInstance().current_slice->scene.extruders[extruder_nr];
 
-            { // require printing temperature to be met
-                constexpr bool wait = true;
-                gcode.writeTemperatureCommand(extruder_nr, extruder_plan.required_start_temperature, wait);
-            }
-
             if (extruder_plan.prev_extruder_standby_temp)
             { // turn off previous extruder
                 constexpr bool wait = false;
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
This is a dangerous one!
We had it this way around at first. However if the estimated heating/cooling rates are a bit inaccurate, this results in a runaway effect where the nozzle is cooler than CuraEngine thinks, and it doesn't pre-heat in time. It then needs to wait a long time on the `M109`. During this time, the other nozzle cools down further, causing that nozzle to be even cooler than CuraEngine thinks. This is a feedback loop that causes the printing time to increase drastically.

That's why we now first pre-heat with `M109` and only then let it cool down the other nozzle. It effectively syncs up the nozzle temperatures with what CuraEngine thinks they should be.

rijkvanmanen:
@Ghostkeeper and I talked about the change and came to conclusion we had some misunderstanding about the change. I descriped the change in more depth in PP-122. Also within the print profile CCB we approved the change.

Within dual extrusion prints there is quite complex heating and cooling down behaviour, which is shown for one nozzle in the attached picture. The idle nozzle is preheated with the M104 marked with a C, so it ready to start printing directly after the nozzle switch. Directly after the nozzle switch there is a double check with a M109 marked with an A, that the minimum printing temperature is indeed reached. If not, it will wait on that temperature. Directly after A, there is a M104 for the idle nozzle to allow it to cool down.
However, A can also be placed directly before C. This allows the idle nozzle to cooldown earlier in case the active nozzle needs to wait on reaching the minimum printing temperature.

image

Ghostkeeper (approved):

PR #1581, opened 2022-02-17
Fix some more polyline issues

BagelOrb:
LayerPlan::addLinesByOptimizer wasn't able to deal with polylines which turned out to be closed polygons.

Modifier meshes would result in empty layers being processed.

### DIFF
--- a/src/FffPolygonGenerator.cpp
+++ b/src/FffPolygonGenerator.cpp
@@ -135,7 +135,10 @@ bool FffPolygonGenerator::sliceModel(MeshGroup* meshgroup, TimeKeeper& timeKeepe
         //Find highest layer count according to each mesh's settings.
         for(const Mesh& mesh : meshgroup->meshes)
         {
-            if ( ! mesh.isPrinted()) continue;
+            if ( ! mesh.isPrinted())
+            {
+                continue;
+            }
             const coord_t mesh_height = mesh.max().z;
             switch(mesh.settings.get<SlicingTolerance>("slicing_tolerance"))
             {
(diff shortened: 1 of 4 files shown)


src/FffPolygonGenerator.cpp
```diff
@@ -135,6 +135,7 @@ bool FffPolygonGenerator::sliceModel(MeshGroup* meshgroup, TimeKeeper& timeKeepe
         //Find highest layer count according to each mesh's settings.
         for(const Mesh& mesh : meshgroup->meshes)
         {
+            if ( ! mesh.isPrinted()) continue;
```
casperlamboo, line 138: lil codestyle
```suggestion
            if ( ! mesh.isPrinted())
            {
                continue;
            }
```

casperlamboo (approved):
Had one very minor code style remark, otherwise the code looks good! straightforward changes.

src/FffPolygonGenerator.cpp
BagelOrb: Is there some code conventions document? I forgot where to find it.

BagelOrb: Personally I choose between
```
if (condition) short_statement;
```
and
```
if (condition)
{
  multi
  line;
}
```
based on which would be better readable. Readability is most important.

I avoid the other option in order to prevent this mistake:
```
if (condition)
  part of condition;
  oops;
```

casperlamboo: code style conventions https://github.com/Ultimaker/Meta/. specifically the codeblock conventions
 https://github.com/Ultimaker/Meta/blob/master/general/generic_code_conventions.md#code-blocks-not-codeblocks-

PR #1582, opened 2022-02-17
Fix warnings

Piezoid:
Tested with Clang 13 and GCC 11.2.

Updating to the latest master of rapidjson solves the warnings from that code. This PR doesn't update it as there is no release since v1.1. But I'm not sure if there ever will be another release.

### DIFF
--- a/include/skin.h
+++ b/include/skin.h
@@ -157,9 +157,8 @@ protected:
      * Helper function to calculate and return the areas which are 'directly' above air.
      *
      * \param part Where to get the SkinParts to get the outline info from
-     * \param flooring_layer_count The number of layers below the layer which we are looking into
      */
-    Polygons generateNoAirBelow(SliceLayerPart& part, int flooring_layer_count);
+    Polygons generateNoAirBelow(SliceLayerPart& part);
 
     /*!
      * Helper function to recalculate the roofing fill and inner infill in roofing layers where the 
--- a/include/skin.h
+++ b/include/skin.h
@@ -159,7 +159,7 @@ protected:
      * \param part Where to get the SkinParts to get the outline info from
      * \param flooring_layer_count The number of layers below the layer which we are looking into
      */
-    Polygons generateNoAirBelow(SliceLayerPart& part, size_t flooring_layer_count);
+    Polygons generateNoAirBelow(SliceLayerPart& part, int flooring_layer_count);
 
     /*!
      * Helper function to recalculate the roofing fill and inner infill in roofing layers where the 
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -362,7 +362,7 @@ size_t FffGcodeWriter::getStartExtruder(const SliceDataStorage& storage)
     const EPlatformAdhesion adhesion_type = mesh_group_settings.get<EPlatformAdhesion>("adhesion_type");
     const ExtruderTrain& skirt_brim_extruder = mesh_group_settings.get<ExtruderTrain&>("skirt_brim_extruder_nr");
 
-    size_t start_extruder_nr;
+    size_t start_extruder_nr = 0;
     if (adhesion_type == EPlatformAdhesion::SKIRT && (skirt_brim_extruder.settings.get<int>("skirt_line_count") > 0 || skirt_brim_extruder.settings.get<coord_t>("skirt_brim_minimal_length") > 0))
     {
         start_extruder_nr = skirt_brim_extruder.extruder_nr;
@@ -2077,7 +2077,7 @@ void FffGcodeWriter::processSpiralizedWall(const SliceDataStorage& storage, Laye
     const bool is_top_layer = ((size_t)layer_nr == (storage.spiralize_wall_outlines.size() - 1) || storage.spiralize_wall_outlines[layer_nr + 1] == nullptr);
     const int seam_vertex_idx = storage.spiralize_seam_vertex_indices[layer_nr]; // use pre-computed seam vertex index for current layer
     // output a wall slice that is interpolated between the last and current walls
-    for (const ConstPolygonRef& wall_outline : part.spiral_wall)
+    for (const ConstPolygonRef wall_outline : part.spiral_wall)
     {
         gcode_layer.spiralizeWallSlice(mesh_config.inset0_config, wall_outline, ConstPolygonRef(*last_wall_outline), seam_vertex_idx, last_seam_vertex_idx, is_top_layer, is_bottom_layer);
     }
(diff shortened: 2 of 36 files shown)


Piezoid:
About: `warning: 'void* memset(void*, int, size_t)' clearing an object of non-trivial type 'class cura::TimeEstimateCalculator::Block'; use assignment or value-initialization instead [-Wclass-memaccess]`
Should I add [this commit](https://github.com/Piezoid/CuraEngine/commit/a4dd88398fb1e147d7272e72313c571a8ac44cec)?

src/infill/LightningTreeNode.cpp
```diff
@@ -406,7 +406,7 @@ void LightningTreeNode::removeJunctionOverlap(Polygons& result_lines, const coor
         PolygonRef polyline = *poly_it;
         if (polyline.size() <= 1)
         {
-            polyline = std::move(result_lines.back());
+            polyline = static_cast<PolygonRef&&>(result_lines.back());
```
jellespijker, line 409: I have to give this line a little extra thought tbh. But that is mostly because our `PolygonRef` class makes so little sense to me.  As far as I remember it is a sort of wrapper with references to the polygon, casting this to a `refref`  seems strange to me. But that is also partly because I watched this YouTube video last week https://www.youtube.com/watch?v=2Y9XbltAfXs and around 38minutes  he talks about Template argument deduction, he shows a table what the deduced argument would be and passing a `T&` to a `T&&` will still result in `T&`.

src/infill/LightningTreeNode.cpp
```diff
@@ -433,7 +433,7 @@ void LightningTreeNode::removeJunctionOverlap(Polygons& result_lines, const coor
 
         if (polyline.size() <= 1)
         {
-            polyline = std::move(result_lines.back());
+            polyline = static_cast<PolygonRef&&>(result_lines.back());
```
jellespijker, line 436: See comment above

src/utils/polygonUtils.cpp
```diff
@@ -1251,7 +1251,7 @@ bool PolygonUtils::polygonCollidesWithLineSegment(ConstPolygonRef poly, const Po
 
 bool PolygonUtils::polygonCollidesWithLineSegment(const Polygons& polys, const Point& transformed_startPoint, const Point& transformed_endPoint, PointMatrix transformation_matrix)
 {
-    for (ConstPolygonRef poly : polys)
+    for (const ConstPolygonRef poly : polys)
```
jellespijker, line 1254: Not really a comment on this change specific;  Just a remark on the strangeness which is CuraEngine: "Why do we even have those strange specialization for the polygon-class?" 
@rburema we should have a propper look at this class, together with @nallath

src/utils/SquareGrid.cpp
```diff
@@ -127,7 +127,7 @@ bool SquareGrid::processAxisAlignedTriangle
 {
     Point a = from;
     Point b = to;
-    if ((a.X < b.X == a.Y < b.Y) != to_the_right)
+    if (((a.X < b.X) == (a.Y < b.Y)) != to_the_right)
```
jellespijker, line 130: Good catch, this was just asking for trouble

src/utils/SVG.cpp
```diff
@@ -1,3 +1,4 @@
+
```
jellespijker, line 1:
```suggestion
```

src/FffGcodeWriter.cpp
```diff
@@ -384,6 +384,7 @@ size_t FffGcodeWriter::getStartExtruder(const SliceDataStorage& storage)
         }
         else
         {
+            start_extruder_nr = 0; // avoid maybe-unitialized warning (compiler cannot assume !extruder_is_used.empty())
```
jellespijker, line 387: Why not initialize the value at line 365 then?

src/InsetOrderOptimizer.cpp
```diff
@@ -227,7 +226,7 @@ std::unordered_set<std::pair<const ExtrusionLine*, const ExtrusionLine*>> InsetO
             grid.insert(LineLoc{ junction, line });
         }
     }
-    for (const std::pair<SquareGrid::GridPoint, LineLoc>& pair : grid)
+    for (const std::pair<const SquareGrid::GridPoint, LineLoc>& pair : grid)
```
jellespijker, line 229:
```suggestion
    for (const auto& pair : grid)
```

jellespijker (changes requested):
Nice! Great innititative.

Some minor code changes, added them as suggestions. Still need to look a bit mor into the detail of the replacement of `std::move(` with `static_cast<PolygonRef&&` but I'm a bit pressed for time. Maybe leave this changes out for now?

jellespijker:
I don't know which IDE you're using but we recently update our clang-tidy and clang-format, although they still don't fully specify our style, they're good enough to apply auto formatting. Don't be afraid to use the auto formatting on these files. If we see something that is really not according to our style, we can update those files accordingly. Step by step it will be a more uniform code base.

Piezoid:
My IDE didn't do the formatting, but I had to use `git filter branch` and `git-clang-format` to pre-format my edit for minimizing conflicts during rebase. And i forgot to apply it one more time after rebase.
So this is your formatting on my edits, but with the original context/indentation. I noticed my other PR have this issues to. I'll fix that.

src/infill/LightningTreeNode.cpp
Piezoid: If I recall correctly, both the above line are equivalent. I don't remember exactly why I did that, maybe the sole purpose was to silence a warning.
`PolygonRef` is a non owning pointer wrapper, so copying it ([like here](https://github.com/Ultimaker/CuraEngine/pull/1582/files#diff-0b8a0d9d420d4449bc50f348e965e7b418df3b843bc48572a58ceb2aee0eacedL1254-R1254)) shouldn't be an issue.

include/skin.h
```diff
@@ -159,7 +159,7 @@ class SkinInfillAreaComputation
      * \param part Where to get the SkinParts to get the outline info from
      * \param flooring_layer_count The number of layers below the layer which we are looking into
      */
-    Polygons generateNoAirBelow(SliceLayerPart& part, size_t flooring_layer_count);
+    Polygons generateNoAirBelow(SliceLayerPart& part, int flooring_layer_count);
```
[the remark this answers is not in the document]
Piezoid: Done

src/infill/LightningTreeNode.cpp
Piezoid: In the end I remove the moves.

Piezoid:
@jellespijker 
I've applied your suggestion and resolved few warnings recently added.
To be clear this only covers `-Wall -Wextra`. The `EXTENSIVE_WARNINGS` cmake flag is way too verbose. Maybe it shouldn't be enabled by default?

src/utils/polygonUtils.cpp
rburema: Eventually. Let's drill down until we got there. First we have to have that session about the architecture ;-)

Maybe there should be a place to collate these whichlist/fixme/todo's.

PR #1583, opened 2022-02-17
Replace abs() with std::abs()

Piezoid:
Some places in the code call the C function `abs`. Unlike the C++ function `std::abs()` which has overloads, C's `abs` cast its argument to `int`, risking overflows.

For example, [this block is dead code](https://github.com/Ultimaker/CuraEngine/blob/385e2ca6ce62af2b4b927d58352f378636418291/src/utils/LinearAlg2D.cpp#L211-L224) since `abs(x) > SQRT_LLONG_MAX_FLOOR` is always false.

The second commit, less important, changes `fabs` and `llabs` to `std::abs`. This is for consistency, but also avoid casting to floats when the argument is a double, like with `PolygonRef::area()`'s return value.

Other mathematical function like cos, sqrt, etc, operate on a double precision float argument, so it is less problematic to use the C variant.

### DIFF
--- a/src/ConicalOverhang.cpp
+++ b/src/ConicalOverhang.cpp
@@ -49,7 +49,7 @@ namespace cura {
 						{
 							Polygons holePoly;
 							holePoly.add(layerParts[part][hole_nr]);
-							if (maxHoleArea > 0.0 && INT2MM(INT2MM(fabs(holePoly.area()))) < maxHoleArea)
+							if (maxHoleArea > 0.0 && INT2MM2(std::abs(holePoly.area())) < maxHoleArea)
 							{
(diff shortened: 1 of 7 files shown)


Piezoid:
Is there any update on this? 
This is a simple fix and compiling with `-fsanitize=undefined,implicit-conversion` shows that this issue arise on most slice jobs. (see also #1585)

Ghostkeeper:
Sorry, we're not allowed to look at external pull requests due to a crunch happening to prepare the 5.0 release. I hope we'll get time to catch up in a bit.

Piezoid:
No problem, I'll rebase them when 5.0 is released.

rburema:
@Piezoid Thanks for the contribution, sorry for the wait!

PR #1585, opened 2022-02-17
Guard against signed overflows in IntPoint.h related code

Piezoid:
This work was done while investigating overflows when increasing the precision of fixed point `coord_t` coordinates. However some of the overflows happen with the normal micro-metric precision while slicing a Benchy on standard fdmprofile.

- First commit is already present in #1583 ,
- Second commit adds asserts, checking for overflows in `vSize2`. A `saturatingVSize2` is also added,
- Third commit reorders the short-circuiting disjunction in `PolygonRef::_simplify` such that, if the lines intersection is far away, `getDist2FromLine` is not called,
- Last one introduces the function `compareVSize2(a, b, c)` which returns `vSize2(a - c) - vSize2(b - c)` with less risks of overflows. One place where such overflow happens is [here in `findStartLocation`](https://github.com/Ultimaker/CuraEngine/blob/master/src/PathOrderOptimizer.h#L481).

I haven't investigated in depth potential changes of behavior. Since the overflows have gone unnoticed, it is likely that the code "do the right thing" in the presence of overflow. But it is not impossible that fixing them introduces other bugs...

There is other signed overflow cases not addressed by this PR. For example, the triangle [`height_2` computation in `PolygonRef::_simplify`](https://github.com/Ultimaker/CuraEngine/blob/385e2ca6ce62af2b4b927d58352f378636418291/src/utils/polygon.cpp#L532) sometimes overflows and produces a negative value (on most CPUs). A potential solution is to implement proper fixed point multiplication. However with a decimal factor, that would be slower than using floating point ([like so](https://github.com/Piezoid/CuraEngine/commit/d6e740115d4e01d63614addc26a0b921d68937bd#diff-80b19785c1dcfa41e7d311f9ecb5a3470cc66c58e16222613c73c730fd20210aR450-R542)). The overhead would likely be minimal, as the FPU ports were mostly unused in that code.

These issues were investigated using Clang's UndefinedBehaviorSanitizer (`-fsanitize=undefined,implicit-conversion,signed-integer-overflow`). In order to keep the reports minimal I've made [other changes](https://github.com/Piezoid/CuraEngine/commits/fix_ub) that I deemed too pedantic and of little importance to be included in this PR.

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -2205,7 +2205,7 @@ std::optional<Point> FffGcodeWriter::getSeamAvoidingLocation(const Polygons& fil
         return std::optional<Point>();
     }
     // now go to whichever of those vertices that is closest to where we are now
-    if (vSize2(pa.p() - last_position) < vSize2(pb.p() - last_position))
+    if (compareVSize2(pa.p(), pb.p(), last_position))
     {
         return std::optional<Point>(std::in_place, pa.p());
     }
--- a/src/LayerPlan.cpp
+++ b/src/LayerPlan.cpp
@@ -740,7 +740,7 @@ void LayerPlan::addWallLine(const Point& p0, const Point& p1, const Settings& se
                 Point b0 = bridge[0];
                 Point b1 = bridge[1];
 
-                if (vSize2f(cur_point - b1) < vSize2f(cur_point - b0))
+                if (compareVSize2(b0, b1, cur_point) > 0)
                 {
                     // swap vertex order
                     b0 = bridge[1];
@@ -882,7 +882,7 @@ void LayerPlan::addWall(const ExtrusionLine& wall, int start_idx, const Settings
                         Point b0 = bridge[0];
                         Point b1 = bridge[1];
 
-                        if (vSize2f(p0.p - b1) < vSize2f(p0.p - b0))
+                        if (compareVSize2(b0, b1, p0.p) > 0)
                         {
                             // swap vertex order
                             b0 = bridge[1];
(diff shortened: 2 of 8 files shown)


Piezoid:
I've rebased this branch. Changes:

- First commit was merged in #1583 
- Third commit (changes in `simplify`) was super-seeded by #1677 

However, please don't merge yet, First I need to verify that this change is still needed with the new decimation algorithm, and that new upstream code doesn't introduce new overflows when using higher `INT2MM` precision.

Piezoid:
Closing this (for now). The new decimator fixes some of the issues. Also, I found better solutions for hardening against overflows.

jellespijker:
Fyi I enabled all warnings on CuraEngine to give us (Cura devs) an incentive to do some proper defensive programming.

Part of my personal mission (in life) to push CURA engine to the modern standards. So I wanted to thank you for doing this on your own initiative 👍 

P.s. We have C++20 support now on our main branch. So that opens up `concepts`, `ranges` and `fmt`

Piezoid:
> Fyi I enabled all warnings on CuraEngine to give us (Cura devs) an incentive to do some proper defensive programming.

Nice ! :+1: 
#1582 could help improve the SNR of diagnostics :wink: 

To track overflow I use `-fsanitize=undefined` and sometimes `-fsanitize=implicit-conversion`. It might be useful to do some sort of integration tests with these flags enabled (also with address sanitizer).

Most segfaults I encountered are indirectly caused by overflows. For example, std::sort can segfault if the ordering it not consistent. Also the trapezoidal graph traversal fails if there is bad coordinates generated by overflows.

PR #1588, opened 2022-02-18
Add support for per-object retraction settings.

jeremysalwen:
This change adds support for per-object retraction settings to CuraEngine.
The main mechanism is through changing the code to track a pointer to the
SliceMeshStorage instead of just storing the mesh name as a string.

Additionally, the code is changed to group together all retraction and
wipe settings into a single object which allows for more easily switching
between the object-specified and the extruder-specified settings, depending
on the context.

Retraction settings will be applied based on the most recent mesh that was
extruded.  As @Ghostkeeper noted, this may cause imperfections on the next
mesh (albeit in a way that the user has control over).  However, in 
one-at-a-time print ordering, this will not be an issue.

For some retractions, occuring as part of setup or teardown, it does not
make sense to apply options from a specific mesh, so settings from the
mesh group or extruder are used instead.

This PR is related to https://github.com/Ultimaker/Cura/issues/3193

### DIFF
--- a/src/LayerPlan.cpp
+++ b/src/LayerPlan.cpp
@@ -99,7 +99,7 @@ void ExtruderPlan::applyBackPressureCompensation(const Ratio back_pressure_compe
 GCodePath* LayerPlan::getLatestPathWithConfig(const GCodePathConfig& config, SpaceFillType space_fill_type, const Ratio flow, bool spiralize, const Ratio speed_factor)
 {
     std::vector<GCodePath>& paths = extruder_plans.back().paths;
-    if (paths.size() > 0 && paths.back().config == &config && !paths.back().done && paths.back().flow == flow && paths.back().speed_factor == speed_factor && paths.back().mesh_id == current_mesh) // spiralize can only change when a travel path is in between
+    if (paths.size() > 0 && paths.back().config == &config && !paths.back().done && paths.back().flow == flow && paths.back().speed_factor == speed_factor && paths.back().mesh == current_mesh) // spiralize can only change when a travel path is in between
     {
         return &paths.back();
     }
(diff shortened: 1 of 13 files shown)


jeremysalwen:
Didn't realize there were some tests I wasn't building.  Fixed the tests to be compatible with this change.

jeremysalwen:
Ping @nallath can you run the CI workflow?

jeremysalwen:
Sweet! What is the review process like? If there's any feedback on changes needed to this PR I'm happy to make them.

nallath:
Sorry! We're in total crunch mode for the Cura 5.0 release. Reviewing / testing PR's is a bit on the backburner now. My apologies!

src/LayerPlan.cpp
```diff
@@ -1698,19 +1703,18 @@ void LayerPlan::writeGCode(GCodeExport& gcode)
...
             gcode.ResetLastEValueAfterWipe(prev_extruder);
 
-            const ExtruderTrain& prev_extruder_train = Application::getInstance().current_slice->scene.extruders[prev_extruder];
-            if (prev_extruder_train.settings.get<bool>("retraction_hop_after_extruder_switch"))
+            const RetractionAndWipeConfig& prev_retraction_config = storage.retraction_wipe_config_per_extruder[prev_extruder];
+            if (prev_retraction_config.retraction_hop_after_extruder_switch)
             {
-                z_hop_height = storage.extruder_switch_retraction_config_per_extruder[prev_extruder].zHop;
-                gcode.switchExtruder(extruder_nr, storage.extruder_switch_retraction_config_per_extruder[prev_extruder], z_hop_height);
+                //TODO: Is it intended that this modifies the Z-Hop down below?
```
rburema, line 1709: Yes, see the '// back to normal z hop' comment below, and commit 87dcc26e636aebb74e77388f6f60a2c370b8fe5e

(Hi, finally taking a look at this now. We're getting back on track with picking up PR's after a chaotic year. Sorry for the way too long wait!)

rburema:
Hi! We'd like to apologize to take this long to finally process this PR.

Thanks for your contribution! If this passes QA, it will be in 5.3

Since this PR was old, rather than asking you to update it, I merged the main branch in myself, and made a frontend PR to expose the fact that a lot of retraction settings can now also be accessed on a per-model basis.

(Devs, see jira ticket CURA-9876)

PR #1432, opened 2021-03-18
Fixes dangling references in Polygons::removeSmallAreas

Piezoid:
Since commit 377ee8f379a986fb81f5226ce9181b175a0492c3, #1156, `Polygons.removeSmallAreas(remove_holes=false)` also removes small holes contained within small outlines. 

Using the AdressSanitizer I found that it doesn't (no longer?) works with the `Polygons::remove(int)` and `ConstPolygonRef` implementation.  

`removeSmallAreas` creates `ConstPolygonRef`s  to small outlines for later reference at the small holes check. These outlines are overwritten by a call `remove(i)` which replace them with an element from the end. I think the heap-use-after-free reported by Asan was caused by the vector shrinking enough for it's storage to be reallocated : <details>

```
0x6080000169a8 is located 8 bytes inside of 96-byte region [0x6080000169a0,0x608000016a00)
freed by thread T20 here:
    #0 0x55ebd49c5c29 in operator delete(void*) (build-CuraEngine-Clang-Debug/CuraEngine+0x1ddc29)
    #1 0x55ebd4a7f1eb in __gnu_cxx::new_allocator<ClipperLib::IntPoint>::deallocate(ClipperLib::IntPoint*, unsigned long) /usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/10.2.0/../../../../include/c++/10.2.0/ext/new_allocator.h:133:2
    #2 0x55ebd4a7f1d8 in std::allocator_traits<std::allocator<ClipperLib::IntPoint> >::deallocate(std::allocator<ClipperLib::IntPoint>&, ClipperLib::IntPoint*, unsigned long) /usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/10.2.0/../../../../include/c++/10.2.0/bits/alloc_traits.h:492:13
    #3 0x55ebd4a7f1cd in std::_Vector_base<ClipperLib::IntPoint, std::allocator<ClipperLib::IntPoint> >::_M_deallocate(ClipperLib::IntPoint*, unsigned long) /usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/10.2.0/../../../../include/c++/10.2.0/bits/stl_vector.h:354:4
[...]

PR #1593, opened 2022-02-25
Fix polygon remove small areas

plaintoothpaste:
I was running cura engine in debug mode (compiling with visual studio 2019) and parts were crashing for me in polygons.cpp (polygons::removeSmallAreas). This was due to an iterator being decremented to before paths.begin(). Specifically, I was crashing at line 705 of the current version of polygons.cpp. In release mode this is not an issue as the iterator is not bounds checked.

To investigate this I set up some more unit tests with all the relevent geometric combinations of removeSmallAreas that I could think of. This led to a failure when a small area with a hole in it is removed but the hole is not.

To address these two issues I rewrote removeSmallAreas it no longer uses iterators and now pases the unit tests I set up. I have run 5 files and there is no performance difference for me, I am not sure if I can find changes in performance in github actions.

### DIFF
--- a/src/utils/polygon.cpp
+++ b/src/utils/polygon.cpp
@@ -678,63 +678,61 @@ void Polygons::removeEmptyHoles_processPolyTreeNode(const ClipperLib::PolyNode& 
 
 void Polygons::removeSmallAreas(const double min_area_size, const bool remove_holes)
 {
-    auto new_end = paths.end();
-    if(remove_holes)
-    {
-        for(auto it = paths.begin(); it < new_end; it++)
-        {
-            // All polygons smaller than target are removed by replacing them with a polygon from the back of the vector
-            if(fabs(INT2MM2(ClipperLib::Area(*it))) < min_area_size)
-            {
-                new_end--;
-                *it = std::move(*new_end);
-                it--; // wind back the iterator such that the polygon just swaped in is checked next
-            }
-        }
-    }
-    else
-    {
-        // For each polygon, computes the signed area, move small outlines at the end of the vector and keep references on small holes
-        std::vector<PolygonRef> small_holes;
-        for(auto it = paths.begin(); it < new_end; it++) {
-            double area = INT2MM2(ClipperLib::Area(*it));
-            if (fabs(area) < min_area_size)
-            {
-                if(area >= 0)
-                {
-                    new_end--;
-                    if(it < new_end) {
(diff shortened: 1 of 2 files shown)


Piezoid:
I'm the one to blame for that code (#1432) :sweat_smile: 

Someone else noticed it, so I published a patch in PR discussion, but it didn't make it into the merge.

plaintoothpaste:
Thanks for the feedback, sorry I didnt see the existing issue. I will close this and comment there.

Piezoid:
> I will close this and comment there.

Maybe don't, your tests add value. Even more so since you have isolated a failure case of my implementation.
I need a bit more time to have a deeper look in the next few days.

This PR also bring more readable and principled code, but I think it would be a good idea to quantify the added overhead of the range-insert, sort and reverse operations.

plaintoothpaste:
Okay thanks. Is there a standard way to quantify the overhead because I am happy to do it. Otherwise just let me know if you want any information.

Ghostkeeper:
Please beware of tabs vs. spaces for indentation. We use 4 spaces in this project.

There is no standard way specific to CuraEngine on how to measure performance overhead. Just the standard C++ techniques, like with the `steady_clock`. While working, we basically just apply good coding practice, attempting to prevent copies of polygon data where possible and such. Our testing engineers test slicing performance before every release, and if the slicing time increased significantly we tend to investigate where in the process it's now spending more time and to see if we can optimise something there. This approach is much more effective than finding micro-optimisations during development, in my experience.

One lazy-ish way of measuring performance of such functions is to simply add a loop somewhere in some of the unit tests that test this function, and get the before/after time around that loop. Depending on what you want to measure, you can also make the content of the test itself much bigger, e.g. using a high-resolution circle instead of a square. You have to add your own timestamps before/after though, because you want to exclude the setting up of the test from the measurement. You can't just look at the time it takes to run the test as a whole. Just remember to enable optimisations to get a more realistic measurement (we use `-O3` in GCC).

plaintoothpaste:
Thanks for the information @Ghostkeeper I will setup to compile in GCC and then begin some testing.

rburema:
Hi @plaintoothpaste Sorry for the wait. I understood (and can see for myself) from @Piezoid that they pulled in your changes with the PR that was mentioned above. (You're still credited as the author for the commits in the main branch, due to how cherry-picking works.) As such, I'm closing this PR.

PR #1595, opened 2022-03-02
Support gap filler fix

BagelOrb:
Some support lines had gone missing.

For example the line on the left here.
image
[zipped 3MF](https://github.com/Ultimaker/CuraEngine/files/8170140/SupportTest.zip)

When the Infill class generates any zigzag pattern then it performs an inset to the area so that the zigzag connectors don't overlap with the area to be filled, but lie half a line width inward. This inset can remove long thin areas which are then not filled anymore, which would be problematic for zizgag skin. The Infill class therefore generates gap filler lines for those areas removed.

However, it is not clear whether such gap filling areas also need to be filled when it comes to infill or support.

For support there could be a case made that if you don't use gap fillers then sometimes a thin overhang would not be supported. That's what this PR fixes.

However, for infill I see no reason to print gap fillers, since infill is generally allowed to be sparse. This is a TODO which is not fixed in this PR.

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -2689,7 +2689,7 @@ bool FffGcodeWriter::processSupportInfill(const SliceDataStorage& storage, Layer
             const coord_t support_line_width = default_support_line_width * (combine_idx + 1);
 
             Polygons support_polygons;
-            VariableWidthPaths wall_toolpaths_here;
+            VariableWidthPaths gap_filler_toolpaths; // In case of a zigzag infill pattern the Infill comp generates gap fillers for the initial offset from the outline to where the zigzag connectors will lie
             Polygons support_lines;
             const size_t max_density_idx = part.infill_area_per_combine_per_density.size() - 1;
             for(size_t density_idx = max_density_idx; (density_idx + 1) > 0; --density_idx)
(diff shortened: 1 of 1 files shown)


BagelOrb:
See https://github.com/Ultimaker/CuraEngine/commit/ae378b240bbad9edd6f213529b9036e04b39bb1b#commitcomment-67796239

MariskaMaas:
For Internal Use CURA-8995

Ghostkeeper:
For comparison, this seems to be the problem that the fix is solving:
image

The support there is hanging in mid-air because it would be too thin to print with the 0.5 line width inset being done for zigzag.

nallath:
Before the fix (project file, change profile to 0.15)
image

After the fix:

image

There is suddenly a floaty bit that wasn't there before.

nallath (changes requested):
The fix seems to introduce a bug that should be fixed.

BagelOrb:
This PR was to solve a hypothetical regression which was introduced in https://github.com/Ultimaker/CuraEngine/pull/1562, which built on https://github.com/Ultimaker/Cura/pull/11334.

When I checkout the master commit before that PR was merged (385e2ca6ce62af2b4b927d58352f378636418291 and checkout frontends with git checkout 'master@{Tue Feb 15 10:45:43 2022 +0100}') then we see the following behavior:
image

The gap filler for support was somehow not printed in the first place!

So this PR is trying to solve a bug which I thought I had introduced, but was already there in master before my Wall Ordering and polyline stitching PRs.

Do we want to print gap fillers for support in the first place??

BagelOrb:
This is 4.13 behavior:
image

No gap fillers either.

BagelOrb:
This PR has already been implemented in CURA-8067.

However, it is still unclear whether we want the behavior in the first place, given the bug noted by @nallath

BagelOrb:
Checking current master there are a lot of bugs in support:
image

image

BagelOrb:
Those disconnected segments in support might not print so well:
image

BagelOrb:
Closing because CURA-8067 implemented the same thing.

PR #1596, opened 2022-03-02
rename VariableWidthPaths to VariableWidthInsets

BagelOrb:
The naming was confusing and didn't reflect how it was used.

Conceptually 'path' and 'line' are not different enough, so the original name VariableWidthPaths was easily confused with VariableWidthLines.

Arachne produces a vector of insets, where each inset can consist of multiple lines.

I've made this into a PR, because it might clash with other PRs.


MariskaMaas:
For Internal Use CURA-8998

Ghostkeeper:
What if they're not used as insets, e.g. in the gap filling?

BagelOrb:
Gap filling lines are odd walls, right? So they are walls.

Ghostkeeper:
Yes, but not insets.

It's also being used for concentric patterns though, which are insets but not walls.

BagelOrb:
Then what do you propose?

"Paths" is definitely too vague, because it doesn't signify the order relation between the different elements of the vector.

BagelOrb:
Yeah sounds good.

Still it's quite easy to overlook the s in `VariableWidthLines`, so it's still easy to think that `std::vector<VariableWidthLines>` simply refers to a bag of extrusion lines. 

perhaps we should also rename VariableWidthLines to VariableWidthInset ?

Ghostkeeper:
It's still not always a variable-width inset though.

We have a similar problem with `Polygon` vs `Polygons`. I'd say we can leave it like this for now. This is already an improvement!

Ghostkeeper (approved):

PR #1597, opened 2022-03-03
Make Wall Ordering option Center Last into its own setting

BagelOrb:
Since the Wall Ordering PR was merged the ordering of center lines has changed such that the center is always printed after their enclosing wall.

The only thing that Center Last controls is whether to print all centers bunched up after all even walls have been printed, or interspersed. If we choose a strategy whereby the middle line has more width variation than the even walls then it could be nice to print all those center lines together (rather than interspersed) so that they don't influence the even walls. This choice is independent of which direction we print the walls in.

The difference becomes apparent when a layer has middle lines of different inset index. Even walls ABCDE, Middle Lines 12
image

Example orders:
not Optimize Wall Ordering & Inner to Outer & not Middle Last: E**2**DABC**1** (Note that 2 is after E because Middle lines are always printed after their enclosing wall)
not Optimize Wall Ordering & Inner to Outer & Middle Last: EDABC21
Optimize Wall Ordering & Inner to Outer & not Middle Last: E2ADCB1
Optimize Wall Ordering & Inner to Outer & Middle Last: EADCB21
not Optimize Wall Ordering & Outer to Inner & not Middle Last: ABC1DE2
not Optimize Wall Ordering & Outer to Inner & Middle Last: ABCDE12
Optimize Wall Ordering & Outer to Inner & not Middle Last: ACE2B1D
Optimize Wall Ordering & Outer to Inner & Middle Last: ACEBD12

Perhaps we might want each of these different behaviors in different situations, but the reason I introduce this setting change is not because we already have figured out that that's the case, but because we want separation of concerns; we want to be able to define the inset order direction now and look at the Center Last option later.

Frontend setting change: https://github.com/Ultimaker/Cura/pull/11592

### DIFF
--- a/src/InsetOrderOptimizer.cpp
+++ b/src/InsetOrderOptimizer.cpp
@@ -54,7 +54,7 @@ bool InsetOrderOptimizer::addToLayer()
     // Settings & configs:
     const bool pack_by_inset = ! settings.get<bool>("optimize_wall_printing_order");
     const InsetDirection inset_direction = settings.get<InsetDirection>("inset_direction");
-    const bool center_last = inset_direction == InsetDirection::CENTER_LAST;
+    const bool center_last = settings.get<bool>("wall_order_center_last");
     const bool alternate_walls = settings.get<bool>("material_alternate_walls");
 
     const bool outer_to_inner = inset_direction == InsetDirection::OUTSIDE_IN;
--- a/src/settings/Settings.cpp
+++ b/src/settings/Settings.cpp
@@ -569,11 +569,7 @@ template<> SlicingTolerance Settings::get<SlicingTolerance>(const std::string& k
 template<> InsetDirection Settings::get<InsetDirection>(const std::string& key) const
 {
     const std::string& value = get<std::string>(key);
-    if(value == "center_last")
-    {
-        return InsetDirection::CENTER_LAST;
-    }
-    else if(value == "outside_in")
+    if(value == "outside_in")
     {
         return InsetDirection::OUTSIDE_IN;
     }


MariskaMaas:
Thanks! For internal use: CURA-9028

Ghostkeeper:
The premise of this feature depends on a bug. When printing walls inside to outside, you'd expect the innermost wall to be printed first. I've reported this bug to the team to be discussed/prioritised tomorrow.

For this and for reducing the complexity to the user of determining the order in which walls are printed, we decided this is a feature that we don't want, for these reasons:
* It's nice for a user to be able to select an ordering in one go. If they are different settings, they might choose "inside to outside" and then break their heads on why this buggy software won't just print inside to outside, not knowing that there's a second setting that makes an exception on this.
* There's also the consideration that while all of the example orders given here might be possible, most of them probably don't have a significant niche in which it results in better print quality.
* We can still separate them by breaking up the enum into two booleans in the CuraEngine code, if separation becomes an issue for maintenance. And of course we can also still optimise settings by just looking at the first two options initially and then looking at the Center Last option later.

BagelOrb:
Currently, if you have a hollow cylinder with 3 walls it will always print the odd wall in the middle last - regardless of the wall ordering. This is *by design*, because the inner wall will have a more deviating width, so printing it afterwards will prevent the extrusion error of a more deviating width from being propagated outward. We decided not to make a user setting for this, because it would be yet another user setting. 

Getting users to understand is always secondary to having the correct behavior. We should never sacrifice printing quality for understandability.

PR #1608, opened 2022-03-15
Don't remove surface mode both cutting mesh areas.

BagelOrb:
The polylines are recomputed by the stitching, so they should be cleared.
However, the polygons should not be cleared for surface mode BOTH.
They are already cleared for surface mode SURFACE, 35 lines above at line 150.


BagelOrb:
I don't remember which bug this solves exactly, but judging from the description it should solve a bug where surface mode both would not generate any surface geometry:

image
[pyramid_mesh_surface_shield.zip](https://github.com/Ultimaker/CuraEngine/files/9283825/pyramid_mesh_surface_shield.zip)

BagelOrb:
Reproduced bug in main:
image

With the fix:
image

[cutting_surface_mode_both_bug.zip](https://github.com/Ultimaker/CuraEngine/files/9310594/cutting_surface_mode_both_bug.zip)

BagelOrb:
@devs see CURA-9534

PR #1623, opened 2022-03-25
CURA-8875 Add tie-breaker to seam score calculations

casperlamboo:
to create an even seem in the case where multiple candidate starting positions get the same score.

CURA-8875

### DIFF
--- a/src/PathOrderOptimizer.h
+++ b/src/PathOrderOptimizer.h
@@ -503,7 +503,7 @@ protected:
             }
 
             constexpr float EPSILON = 25.0;
-            if(best_score - EPSILON <= score && score <= best_score + EPSILON)
+            if (fabs(best_score - score) <= EPSILON)
             {
                 // add breaker for two candidate starting location with similar score
                 // if we don't do this then we (can) get an un-even seam
--- a/src/PathOrderOptimizer.h
+++ b/src/PathOrderOptimizer.h
@@ -208,7 +208,7 @@ public:
             assert(before_it != path_to_index.end());
             is_blocking[before_it->second].emplace_back(after_it->second);
         }
-        
+
 
         std::vector<bool> picked(paths.size(), false); //Fixed size boolean flag for whether each path is already in the optimized vector.
         Point current_position = start_point;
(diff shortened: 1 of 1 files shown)


src/PathOrderOptimizer.h
```diff
@@ -501,7 +501,19 @@ class PathOrderOptimizer
                     score += 1000; //1 meter penalty.
                 }
             }
-            if(score < best_score)
+
+            constexpr float EPSILON = 25.0;
+            if(best_score - EPSILON <= score && score <= best_score + EPSILON)
```
Joeydelarago, line 506:
```suggestion
            if(score >= best_score - EPSILON  && score <= best_score + EPSILON)
```
or something like 
`if (fabs(score - best_score) <= EPSILON)`
would be a bit clearer

src/PathOrderOptimizer.h
```diff
@@ -501,7 +501,19 @@ class PathOrderOptimizer
...
+            {
+                // add breaker for two candidate starting location with similar score
+                // if we don't do this then we (can) get an un-even seam
+                Point tie_breaker_point = target_pos != Point(0, 0)
+                    ? Point(0, 0)       // break ties by picking the point that is closest to the origin
+                    : Point(0, 100000); // unless the target position is the origin, then pick some other location
+                best_point = vSize2(best_point - tie_breaker_point) < vSize2(best_point - tie_breaker_point) ? best_point : here;
```
Joeydelarago, line 513: `vSize2(best_point - tie_breaker_point) < vSize2(best_point - tie_breaker_point)`

is always false I think? Did you mean to put something different on one side of this statement?

src/PathOrderOptimizer.h
```diff
@@ -501,7 +501,19 @@ class PathOrderOptimizer
...
+                best_score = std::min(best_score, score);
```
Joeydelarago, line 514: I think you could remove the above code and just leave this in as the tie breaker, that seems to be how it functions now.

Joeydelarago (changes requested):

src/PathOrderOptimizer.h
casperlamboo: Well there can still be a small difference (at most epsilon) between the two numbers, I'd prefer to keep it.

src/PathOrderOptimizer.h
casperlamboo: oops, should be
```suggestion
                 best_point = vSize2(best_point - tie_breaker_point) < vSize2(here - tie_breaker_point) ? best_point : here; 
```

casperlamboo: https://github.com/Ultimaker/CuraEngine/commit/4e137f5273716dedea5678f32183a601bc71fb6e

src/PathOrderOptimizer.h
Joeydelarago: I see where you are coming from, it is a lot easier to read when I think about it in that format 👌

casperlamboo: https://github.com/Ultimaker/CuraEngine/pull/1623/commits/fcf4eac5c4177f0d47b43383bf8e91db7b4cad70

Joeydelarago (approved):

Issue #377, opened 2016-08-04
MSVC build broken


sevaa:
At some point, CuraEngine was compatible with Microsoft Visual Studio. Now it's not. Here are some of the reasons:
- reliance on unistd.h
- "min"/"max" as variable names (they're #defines in MSVC, unless overridden)
- variable length arrays
- reliance on __WIN32 #define - it's not defined anymore, use WIN32 instead

Notably, libArcus and protobuf build under MSVC fine.

awhiemstra:
This is mostly due to MSVC's lacklustre support of C++11. Now that MSVC 2015 finally supports everything we need we can start looking at supporting MSVC again.

sevaa:
I'm building with MSVC 2015 Update 3; VLAs are still not there, at least not enabled by default. IIRC, it's a C99 feature, not a C++11 one. Anyways, those can be trivially replaced with alloca().

BagelOrb:
I don't think this is currently a problem for us, so I don't we will pick
it up any time soon.

We're always willing accept pull requests though.

Op 4 aug. 2016 16:06 schreef "Seva Alekseyev" <notifications@github.com>:

sevaa:
Addressed, see pull request #383. At least now it builds, I haven't tested properly, since I'm not sure how.

BagelOrb:
I'm sorry, we lost sight of this PR.
We will take a look soon; I've added an issue to our internal issue tracker.

jellespijker:
Fixed #1639 
Tested on Visual Studio 2019 and 2022

PR #1639, opened 2022-04-12
Cura 8640 PyQt6 upgrade

jellespijker:
In order to get this to work on our build-system and working for all three OSes we did a shit tons of boy scouting in our cmake. We removed old methods with variables and try to be consisted in a target-based approach. The idea is that we don't patch stuff down the line, but that the install should place everything in the correct path in a uniform way across all of Cura's dependencies.

Most of the changes had to do with how dependency targets were named and to make sure that external projects weren't downloaded automatically or that we prefer the usage over an external dependency instead of the shipped once. That means moving the third-party libs, such as clipper to another structure. I tried to be backward compatible with these changes but they are untested on my machine. I believe @Ghostkeeper ran into some issues recently.

Because of this target based approach and how the flags and settings are now set we can drop support for MinGW on Windows: CuraEngine (and its dependencies protobuf and Arcus) can now compile using the x64 MSVC toolchain on Windows. This is tested with the Visual Studio 2019 and 2022 version.

The question is if we are going put time into doing it properly for both scenario's: supporting shipped libs and external libs or are we saying don't waste the energy since these shipped libs will be removed anyhow in CURA-8830 . In fact people can install the third-party dependencies now anyhow using Conan. I myself am in favor of removing the third-party shipped dependencies t.b.h.

To setup Conan:
```bash
pip install conan
pip profile new default --detect
pip config install https://github.com/ultimaker/conan-config.git
```

To use Conan:
```
git clone https://github.com/ultimaker/curaengine.git
cd curaengine
conan install . if build -pr:b cura_build.jinja -pr:h cura_release.jinja --build=missing
cd build
cmake -G Ninja -DCMAKE_TOOLCHAIN_FILE=conan_toolchain.cmake -DArcus_ROOT=<root-path-arcus-install> .. 
ninja
```
This should download, (build if needed,) en install the dependencies, create a cmake toolchain containing the correct paths top dependencies and proper settings. You can also use make or nmake instead of Ninja but I advise Ninja since that can be used on all three systems.

There were also some changes needed in the namespace when using enums and the renaming of the Python modules Arcus and Savitar to their counterparts pyArcus and pySavitar. This was needed to ensure no naming conflicts downstream.

# Part of
- ultimaker/libarcus#134
- ultimaker/pynest2d#17
- ultimaker/libsavitar#41
- ultimaker/libnest2d#3
- ultimaker/cura-binary-data#17
- ultimaker/libnest2d#3
- ultimaker/uranium#804
- ultimaker/cura#11792
- ultimaker/cura-build-environment#126
- ultimaker/cura-build#280
- ultimaker/cura-rundeck-jobs#1

# Fixes
- #377
- 

# Todo
- [ ] Update Readme. But maybe best done in a separate ticket to prioritize with our other work for the 5.0 release
- [ ] Fix failing CI/CT due to out date build environment. I think it will be best if we fix this in a separate ticket.

> P.S. Developers try compiling with the `-DEXTENSIVE_WARNINGS=ON` we still have a bit of boy scouting in front of us.


CMakeLists.txt
```diff
@@ -1,355 +1,314 @@
...
+        $<$<BOOL:${ENABLE_ARCUS}>:ARCUS>
+        PRIVATE
+        VERSION=\"${CURA_ENGINE_VERSION}\"
+        $<$<BOOL:${WIN32}>:NOMINMAX>
+        $<$<CONFIG:Debug>:ASSERT_INSANE_OUTPUT>
+        $<$<CONFIG:Debug>:USE_CPU_TIME>
+        $<$<CONFIG:Debug>:DEBUG>
+        $<$<CONFIG:RelWithDebInfo>:ASSERT_INSANE_OUTPUT>
+        $<$<CONFIG:RelWithDebInfo>:USE_CPU_TIME>
+        $<$<CONFIG:RelWithDebInfo>:DEBUG>
```
Ghostkeeper, line 220: I think perhaps just `if(Debug OR RelWithDebInfo)` is easier syntax to read than this.

A bit of a matter of style though, so I'm fine either way.

Ghostkeeper (approved):
Only one minor remark that I'd point out in case you agree that it could be better, but the rest of the changes are good and I wouldn't block it.

In the libs folder I think all of the whitespace has changed to include a carriage return (Windows-style). This may be a weirdness in your Git config. I verified that no real changes were made with the `-w` option in `git diff`. The only actual change was the way `clipper.hpp` is imported, which is pointed out here on Github as well.

PR #1643, opened 2022-04-19
Allow tip of tree support to travel through support xy distance

casperlamboo:
# Overview

Some of the required tree-support failed to generate. For these instances the tree-support contact points were generated, but there were no supporting tree branches attached.

**Before**
Screenshot 2022-04-19 at 20 13 36
**After**
Screenshot 2022-04-19 at 20 12 39

# Analysis

As stated before the tree-support contact points were correctly being generated. As these points are within the overhang area they are almost always located within the xy support distance of the polygon. This can be seen in the image below (contact point tree-support: black dot, shape-outline: green, xy support distance: Red).
Screenshot 2022-04-19 at 20 20 21

If the support branch would fall within this xy support distance polygon
https://github.com/Ultimaker/CuraEngine/blob/1a7667832f918781ca7d2e2e008f02708b6be07c/src/TreeSupport.cpp#L368-L369 _and_ the distance to the boundary is less then the branches diameter https://github.com/Ultimaker/CuraEngine/blob/1a7667832f918781ca7d2e2e008f02708b6be07c/src/TreeSupport.cpp#L382-L390 _then_ the branch would be removed in it's entirely.

The rational behind this piece of code is optimization; if a branch polygon is completely within the xy distance offsetted polygon then it would be removed anyway in the finalization stage of the tree support generation. Removing the support branches preemptively prevents unnecessary branches from being maintained throughout the support generation.

This is the crux of the issue; as the support branch contact point is at the far end of the tip it's diameter is very small, only 100 microns for this example. This would in almost all case be less then the distance to the xy distance offsetted polygon, and thus the support branch is removed.

# Solution
The solution is fairly simple; in this PR we allow (the tip of the) support branches to travel through the xy distance.

## Asterisk

The original insensitive to remove the polygons still holds, the branches fall within the xy support distance and they should be removed here. 
https://github.com/Ultimaker/CuraEngine/blob/1a7667832f918781ca7d2e2e008f02708b6be07c/src/TreeSupport.cpp#L145-L146
I did not do anything to prevent this from happening, but the generated support seems correct... so not sure whats going on here.

Additionally, the behavior of the tree support seems to have changed between cura 4.13 and current master, while there has not been any changes to the relevant code. This is a bit suspicious and I started investigation what has changed between these versions. I compared some of the relevant setting values and compared the generated polygons, but I could not determine what has changed. I'm a bit puzzled how this has worked before.

CURA-9096


casperlamboo:
merged of the wrong branch, closing this in favour of #1644

PR #1644, opened 2022-04-19
[Cura] 9096 - Tree Support branches not being generated

casperlamboo:
# Overview

Some of the required tree-support failed to generate. For these instances the tree-support contact points were generated, but there were no supporting tree branches attached.

This PR aims to fix this issue.

**Before**Screenshot 2022-04-19 at 20 13 36

**After**Screenshot 2022-04-19 at 20 12 39

# Analysis

As stated previously the tree-support contact points were correctly generated. These points are created on overhanging areas, and are as a result almost always located within the xy support distance of the polygon. This can be seen in the image below (contact point tree-support: black dot, shape-outline: green, xy support distance: Red).
Screenshot 2022-04-19 at 20 20 21

This situation would trigger the following piece of code: If the support branch would fall within this xy support distance polygon
https://github.com/Ultimaker/CuraEngine/blob/1a7667832f918781ca7d2e2e008f02708b6be07c/src/TreeSupport.cpp#L368-L369 _and_ the distance to the boundary is less then the branch's diameter https://github.com/Ultimaker/CuraEngine/blob/1a7667832f918781ca7d2e2e008f02708b6be07c/src/TreeSupport.cpp#L382-L390 _then_ the branch would be removed in it's entirely.

The rational behind this piece of code is optimization; if a branch polygon is completely within the xy distance offsetted polygon then it would be removed anyway in the finalization stage of the tree support generation. Removing the support branches preemptively prevents unnecessary branches from being maintained throughout the support generation.

This is the crux of the issue; as the support branch contact point is at the far end of the tip it's diameter is very small, only 100 microns for this example. This would in almost all cases be less then the distance to the xy distance offsetted polygon, and thus the support branch is removed.

# Solution
The solution is fairly simple; in this PR I propose to allow (the tip of the) support branches to travel through the xy distance.

## Asterisk

The original insensitive to remove the polygons still holds, the branches fall within the xy support distance and they should be removed here. 
https://github.com/Ultimaker/CuraEngine/blob/1a7667832f918781ca7d2e2e008f02708b6be07c/src/TreeSupport.cpp#L145-L146
I did not do anything to prevent this from happening, but the generated support seems correct... so not sure whats going on here.

Additionally, the behavior of the tree support seems to have changed between cura 4.13 and current master, while there has not been any changes to the relevant code. This is a bit suspicious and I started investigation what has changed between these versions. I compared some of the relevant setting values and compared the generated polygons, but I could not determine what has changed. I'm a bit puzzled how this has worked before.

CURA-9096

### DIFF
--- a/src/TreeSupport.cpp
+++ b/src/TreeSupport.cpp
@@ -31,10 +31,9 @@
 namespace cura
 {
 
-TreeSupport::TreeSupport(const SliceDataStorage& storage)
+TreeSupport::TreeSupport(const SliceDataStorage& storage) :
+    volumes_(storage, Application::getInstance().current_slice->scene.current_mesh_group->settings)
 {
-    const Settings& mesh_group_settings = Application::getInstance().current_slice->scene.current_mesh_group->settings;
-    volumes_ = TreeModelVolumes(storage, mesh_group_settings);
 }
 
 void TreeSupport::generateSupportAreas(SliceDataStorage& storage)
--- a/src/TreeSupport.cpp
+++ b/src/TreeSupport.cpp
@@ -366,7 +366,9 @@ void TreeSupport::dropNodes(std::vector<std::vector<Node*>>& contact_nodes)
                     continue;
                 }
                 //If the branch falls completely inside a collision area (the entire branch would be removed by the X/Y offset), delete it.
-                if (group_index > 0 && volumes_.getCollision(0, layer_nr).inside(node.position))
+
+                Polygons collision = volumes_.getCollision(0, layer_nr);
+                if (group_index > 0 && collision.inside(node.position))
                 {
                     const coord_t branch_radius_node = [&]() -> coord_t
                     {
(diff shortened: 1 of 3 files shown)


src/TreeSupport.cpp
```diff
@@ -379,8 +379,15 @@ void TreeSupport::dropNodes(std::vector<std::vector<Node*>>& contact_nodes)
...
-                    if (vSize2(node.position - to_outside.location) >= branch_radius_node * branch_radius_node) //Too far inside.
+
+                    const ClosestPolygonPoint to_outside = PolygonUtils::findClosest(node.position, collision);
+                    bool node_is_tip = node.distance_to_top <= tip_layers;
+                    coord_t max_inside_dist = branch_radius_node;
+                    if (node_is_tip) {
+                        // if the node is part of the tip allow the branch to travel through the support xy distance
+                        max_inside_dist += (tip_layers - node.distance_to_top) * maximum_move_distance;
+                    }
+                    if (vSize2(node.position - to_outside.location) >= max_inside_dist * max_inside_dist) //Too far inside.
```
jellespijker, line 390:
```suggestion
                    if (vSize2(node.position - to_outside.location) >= max_inside_dist * max_inside_dist) // Too far inside.
```

src/TreeSupport.cpp
```diff
@@ -420,12 +427,11 @@ void TreeSupport::dropNodes(std::vector<std::vector<Node*>>& contact_nodes)
...
-                {
-                    //Avoid collisions.
-                    const coord_t maximum_move_between_samples = maximum_move_distance + radius_sample_resolution + 100; //100 micron extra for rounding errors.
-                    PolygonUtils::moveOutside(volumes_.getAvoidance(branch_radius_node, layer_nr - 1), next_layer_vertex, radius_sample_resolution + 100, maximum_move_between_samples * maximum_move_between_samples); //Some extra offset to prevent rounding errors with the sample resolution.
-                }
+
+                //Avoid collisions.
+                const coord_t maximum_move_between_samples = maximum_move_distance + radius_sample_resolution + 100; //100 micron extra for rounding errors.
+                Polygons avoidance = group_index == 0 ? volumes_.getAvoidance(branch_radius_node, layer_nr - 1) : volumes_.getCollision(branch_radius_node, layer_nr - 1);
+                PolygonUtils::moveOutside(avoidance, next_layer_vertex, radius_sample_resolution + 100, maximum_move_between_samples * maximum_move_between_samples);
```
jellespijker, line 434:
```suggestion
                constexpr size_t rounding_compensation = 100;
                const coord_t maximum_move_between_samples = maximum_move_distance + radius_sample_resolution + rounding_compensation;
                const Polygons avoidance = group_index == 0 ? volumes_.getAvoidance(branch_radius_node, layer_nr - 1) : volumes_.getCollision(branch_radius_node, layer_nr - 1);
                PolygonUtils::moveOutside(avoidance, next_layer_vertex, radius_sample_resolution + rounding_compensation, maximum_move_between_samples * maximum_move_between_samples);
```

jellespijker (changes requested):
Looks good, try to use `const` as much as possible. Also try to avoid magic numbers.

I'm still component testing atm, but if you check my comments and either fix or defer them, we can keep things moving forward

jellespijker (approved):

PR #1647, opened 2022-04-20
[CURA-9159] Alternate fix crash/deadlocks/threading Tree-support

rburema:
Reverted the old (as in: earlier today) fix, which would lead me to a more difficult route if I tried to re-fix the problems it still had.

Instead, introduced a new critical section in the parallel-for loop.

I also found out that there was a way to create deadlocks. Streamlined some code so that should now also not occur anymore. (Counter is no longer atomic, but handled by the 'progress' critical section, merging the 2 critical sections that had the same mutex.)

### DIFF
--- a/src/TreeModelVolumes.cpp
+++ b/src/TreeModelVolumes.cpp
@@ -104,8 +104,6 @@ const Polygons& TreeModelVolumes::calculateCollision(const RadiusLayerPair& key)
             collision_areas = collision_areas.unionPolygons(collision_model.offset(radius));
         }
     }
-
-    const std::lock_guard<std::mutex> lock(object_mutex_);
     const auto ret = collision_cache_.insert({key, std::move(collision_areas)});
     assert(ret.second);
     return ret.first->second;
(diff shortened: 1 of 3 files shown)


src/TreeModelVolumes.cpp
```diff
@@ -148,9 +144,8 @@ const Polygons& TreeModelVolumes::calculateInternalModel(const RadiusLayerPair&
 {
     const auto& radius = key.first;
     const auto& layer_idx = key.second;
-    const auto& internal_areas = getAvoidance(radius, layer_idx).difference(getCollision(radius, layer_idx));
 
-    const std::lock_guard<std::mutex> lock(object_mutex_);
+    const auto& internal_areas = getAvoidance(radius, layer_idx).difference(getCollision(radius, layer_idx));
```
jellespijker, line 148: auto.... vroem vroem! This makes things so much more readable imo. Keep using that

src/TreeSupport.cpp
```diff
@@ -31,9 +31,10 @@
 namespace cura
 {
 
-TreeSupport::TreeSupport(const SliceDataStorage& storage) :
-    volumes_(storage, Application::getInstance().current_slice->scene.current_mesh_group->settings)
+TreeSupport::TreeSupport(const SliceDataStorage& storage)
 {
+    const Settings& mesh_group_settings = Application::getInstance().current_slice->scene.current_mesh_group->settings;
+    volumes_ = TreeModelVolumes(storage, mesh_group_settings);
```
jellespijker, line 37: Again. Love the readability

jellespijker (approved):
Code looks clean and good.
I think there is nothing left to be done the component test. but if @evtrados agrees we could maybe move it along to QA and to the testing there.

PR #1648, opened 2022-04-20
Cura 9128 outer wall extruder crash

jellespijker:
There occurred a crash when using a different outer wall extruder.
This happened because the switching in iteration direction was done
"old skool"-style. Values like start en end inset where uninitialized
constructed. And the if-construct to set the values was overly complex
with cases which could never occur. Why are people so afraid of
`auto` and `for ()` and other standard algorithms?

Since we are still on C++17 I had to use a template helper function
to switch between forward begin()...end() and rbegin()...rend()
iterators. I can't wait till we support C++20 with ranges and views.

The crash is fixed atm but I'm still figuring out the offsets in the
iterators for when we use multiple extruders.

### DIFF
--- a/src/InsetOrderOptimizer.cpp
+++ b/src/InsetOrderOptimizer.cpp
@@ -91,13 +91,13 @@ bool InsetOrderOptimizer::addToLayer()
             {
                 return wallsToBeAdded(paths.rbegin(), paths.rend()); // Complete wall with one extruder
             }
-            return wallsToBeAdded(paths.rbegin(), std::prev(paths.rend())); // Ignore outer wall
+            return wallsToBeAdded(paths.rbegin(), std::prev(paths.rend())); // Ignore inner wall
         }
         if (use_one_extruder)
         {
             return wallsToBeAdded(paths.begin(), paths.end()); // Complete wall with one extruder
         }
-        return wallsToBeAdded(paths.begin(), std::next(paths.begin())); // Ignore inner wall
+        return wallsToBeAdded(paths.begin(), std::next(paths.begin())); // Ignore outer wall
     };
     const auto walls_to_be_added = get_walls_to_be_added(reverse, paths);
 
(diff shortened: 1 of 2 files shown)


src/InsetOrderOptimizer.cpp
```diff
@@ -52,81 +54,54 @@ InsetOrderOptimizer::InsetOrderOptimizer(const FffGcodeWriter& gcode_writer,
...
-        {
-            start_inset = 0;
-            end_inset = paths.size();
-            direction = 1;
-        }
-        else //INSIDE_OUT or CENTER_LAST.
+    const bool use_one_extruder = wall_0_extruder_nr == wall_x_extruder_nr;
+    const bool current_extruder_is_wall_x = wall_x_extruder_nr == extruder_nr;
+
+    auto should_reverse = [&](){
```
rburema, line 65: Functions are variables. Since the function _itself_ doesn't change (even tho' it can change other things) we can make it const :-)

I'll go over the code with a const hammer later.

src/InsetOrderOptimizer.cpp
```diff
@@ -52,81 +54,54 @@ InsetOrderOptimizer::InsetOrderOptimizer(const FffGcodeWriter& gcode_writer,
...
+        if (reverse)
         {
-            start_inset = paths.size() - 1;
-            end_inset = 0; // Ignore outer wall
-            direction = -1;
+            if (use_one_extruder)
+            {
+                return wallsToBeAdded(paths.rbegin(), paths.rend()); // Complete wall with one extruder
+            }
+            return wallsToBeAdded(paths.rbegin(), std::prev(paths.rend())); // Ignore outer wall
```
rburema, line 94: ... good, except now the comments are switched around as well.

I'll change that too.

rburema (approved):
Code approved for now. (I'll make the largely cosmetic changes requested later myself.) Let's do a component test to see if it holds up!

jellespijker:
Thnx for your help last night and for implementing the `const` correctness and making sure the comments match up.

rburema:
... I see that reading my mail took a back-seat last week. You're welcome! :-)

PR #1662, opened 2022-05-16
Fixes "can't decrement vector iterator before begin"

ThomasRahm:
Occurs when compiled in debug mode in the first iteration.


rburema:
Hi @ThomasRahm Sorry it took this long to pick this up. We should have been better on this.

It seems that the change of this PR has been superseded by the one mentioned above. As such, I'm closing it.

PR #1664, opened 2022-05-18
[CURA-9085] Unretracted moves on top layer of top skin. 

Joeydelarago:
This fixes travels being unretracted on the top most surface skin layer(s) when using CombingMode::NO_OUTER_SURFACES.

The top_most_surface_fill polygon being used for the combing boundary was defined incorrectly.
skin_part.top_most_surface_fill = skin_part.skin_fill.difference(no_air_above);
The skin_part.skin_fill was empty on the surface skin layers. Causing this statement to return an empty polygon for combing.
This is because the skin_fill is the fill for the bottom skin rather than the top.

This fix is to use roofing_fill, which has the correct polygon for the top surface skin fill.

CURA-9083

### DIFF
--- a/src/skin.cpp
+++ b/src/skin.cpp
@@ -648,7 +648,7 @@ void SkinInfillAreaComputation::generateTopAndBottomMostSkinSurfaces(SliceLayerP
 
     for (SkinPart& skin_part : part.skin_parts) {
         Polygons no_air_above = generateNoAirAbove(part, 1);
-        skin_part.top_most_surface_fill = skin_part.roofing_fill.difference(no_air_above);
+        skin_part.top_most_surface_fill = skin_part.outline.difference(no_air_above);
 
         Polygons no_air_below = generateNoAirBelow(part, 1);
         skin_part.bottom_most_surface_fill = skin_part.skin_fill.difference(no_air_below);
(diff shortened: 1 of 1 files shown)


casperlamboo (approved):

src/skin.cpp
```diff
@@ -648,7 +648,7 @@ void SkinInfillAreaComputation::generateTopAndBottomMostSkinSurfaces(SliceLayerP
 
     for (SkinPart& skin_part : part.skin_parts) {
         Polygons no_air_above = generateNoAirAbove(part, 1);
-        skin_part.top_most_surface_fill = skin_part.skin_fill.difference(no_air_above);
+        skin_part.top_most_surface_fill = skin_part.roofing_fill.difference(no_air_above);
```
casperlamboo, line 651: Not sure if this is expected behavior, but if we set `top surface skin layers = 0` then we do **not** get retractions anymore on the top surface layer. If this is intended behavior then disregard my comment. If this is not intended behavior and we do want retraction on the layer that is exposed to air disregarding if it is a `top surface skin layer` or not then I expect we should apply the following change.
```suggestion
        skin_part.top_most_surface_fill = skin_part.outline.difference(no_air_above);
```

PR #1666, opened 2022-05-22
CMake: Correcting minimum version for Boost

thopiekar:
This is the wrong position in find_package. Components are listed after the REQUIRED statement.


Joeydelarago:
We have had a substantial overhaul of the build system since this PR was created so the fix no longer applies. Apologies for not reviewing this sooner!

PR #1667, opened 2022-05-22
CMake: Correcting alias'ing of Boost::boost

thopiekar:
add_library takes <name> and as option after ALIAS the <target>.
Let's do it the correct way! (If I get the point of the section correct..)


Joeydelarago:
We made some pretty substantial changes to our build system since this PR was opened so these changes no longer apply. As I said in the other PR, apologies for not reviewing this sooner!

PR #1668, opened 2022-05-22
CMake: Correcting alias'ing of protobuf::protobuf

thopiekar:
Same as for Boost. Alias'ing a library just goes the other way around.

- https://cmake.org/cmake/help/latest/command/add_library.html#alias-libraries


thopiekar:
Ah, wait. My mistake, I guess.
We already have **protobuf::protobuf** but want to provide **Protobuf::protobuf**..

PR #1671, opened 2022-05-26
Don't increase support area by infill overlap.

### LINKED ISSUE #1670: [Bug] "Infill Overlap" increases support areas.
**Application Version**
5.0.0

**Actual Results**
When the support walls are calculated  [in AreaSupport::generateGradualSupport](https://github.com/Ultimaker/CuraEngine/blob/d002c2c331427ef2f0c29eb24541389626328244/src/support.cpp#L211), the support areas will be offset by overlap in `Infill::generateWallToolPaths`.
This causes the support area (not just the support infill area) to become larger.
While this does not matter most of the time, when the overlap is large, or the support area is small that can cause issues.
Examples:

High Overlap Bugged

Small area bugged

**Expected results**

When setting the infill overlap used to calculate the walls in `AreaSupport::generateGradualSupport` to 0, the overlap is still correctly added, but the support area does not increase. See #1671.
This results in the previous examples to look like this instead:

High Overlap Fixed

The support infill will still overlap with the support walls:
High Overlap Fixed lower

Small area fixed

ThomasRahm:
Don't increase support area by infill overlap. Fixes #1670.

### DIFF
--- a/src/support.cpp
+++ b/src/support.cpp
@@ -175,7 +175,6 @@ void AreaSupport::generateGradualSupport(SliceDataStorage& storage)
 
     const coord_t wall_count = infill_extruder.settings.get<size_t>("support_wall_count");
     const coord_t wall_width = infill_extruder.settings.get<coord_t>("support_line_width");
-    const coord_t overlap = infill_extruder.settings.get<coord_t>("infill_overlap_mm");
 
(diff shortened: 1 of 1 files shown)


rburema:
Devs: Assigned internal ticket CURA-9937

PR #1674, opened 2022-05-31
CURA-9295 - CURA-9295_Overextrusion_when_printing_with_gradual_infill_

casperlamboo:
Zig zaggify lines are located on the boundary of the density area polygons. If boundaries of different densities are
aligned then it is possible for the Zig zaggify lines to be printed in the same location. By offsetting each density
we prevent this.

CURA-9295


Ghostkeeper:
While this prevents the over-extrusion, it also makes the infill on lower-density parts disconnect from the walls. This makes the infill effectively useless for strength and in many cases makes it float more than necessary.

image

Perhaps this could be fixed by inverting the offsets' relation with the density layer? Or by making it behave like in 4.13, where only the lowest density layer would get connected infill lines.

Ghostkeeper (approved):

Ghostkeeper:
I think the offsets are correct according to what you designed now, but the effect is still that most of the infill lines don't connect to the edge. This is what happens when slicing the 3MF file from our Jira ticket about it:

image

What you see there is the lowest density line connecting into the corner. The second-lowest density line is the line perpendicular to that protruding from the infill in the bottom right. As you can see, in most cases, only the lowest-density infill pattern connects with the side, making the rest of the infill almost useless. The other densities only connect to the sides if the Connect Infill Lines feature happens to make a connecting line there.

There is also a secondary problem, where it seems that the lowest two densities don't get their infill lines connected at all.
image
It's correct in this picture that the second-lowest density has broken-up infill lines. They are broken up because there's geometry in some parts above this infill that makes the ceiling higher, such that the infill density needs to be lower in those parts. But I'd expect the bits that connect to the sides (or to the almost-sides for the second-lowest density) to be connecting to each other here.

I now think the infill not hitting the walls is going to be an inherent problem with this solution. Did you try making only the lowest infill density connect and then not making insets from the walls?

rburema:
See https://github.com/Ultimaker/CuraEngine/pull/1683 for the new proposed solution.

PR #1676, opened 2022-06-03
Fix edge case in "Remove Raft Inside Corners"

nallath:
Previously the attached zipped STL would cause a non convex raft to be generated even if Remove Raft Inside Corners was enabled.

This was caused by very small segments in the corners, which could cause rounding errors. 

[Simple_reproduction_CURA-9041.zip](https://github.com/Ultimaker/CuraEngine/files/8832086/Simple_reproduction_CURA-9041.zip)

CURA-9041


src/utils/polygon.cpp
```diff
@@ -101,25 +101,33 @@ void Polygons::makeConvex()
...
             }
-            convexified.path->push_back(current);
-        }
-        //remove last vertex as the starting vertex is added in the last iteration of the loop
-        convexified.path->pop_back();
+        };
+
+        Polygon convexified;
+
+        std::sort (poly.begin(), poly.end(), [](Point a, Point b) { return a.X == b.X ? a.Y < b.Y : a.X < b.X; });
```
Ghostkeeper, line 126: Knowing both the [Monotone Chain](https://en.wikibooks.org/wiki/Algorithm_Implementation/Geometry/Convex_hull/Monotone_chain) and the [Graham Scan for simple polygons](https://en.wikipedia.org/wiki/Convex_hull_of_a_simple_polygon#Algorithms), why is it necessary here to sort the whole set and not just find the leftmost vertex? I.e. why use Monotone Chain instead of the Graham Scan?

It seems that finding the leftmost vertex here should already be sufficient. And if not, the Monotone Chain shouldn't need to backtrack.

Ghostkeeper (changes requested):

src/utils/polygon.cpp
casperlamboo: If the previous algorithm was supposed to be Graham's scan it was implemented incorrectly. In Grahams scan you're supposed to find the vertex with lowest x-coord. Lets call this point `p_0`. Then after all remaining points should be sorted on "polar angle" relative to `p_0`.

This last step went wrong, we didn't sort but we traversed the points starting from `p_0` in the same order as the original polygon. This order of the points isn't guaranteed to be in order of polar angle.

So we either sort the vertices on polar angle before traversing the vertices, or implement the monotone chain algorithm. In my opinion the monotone chain is easier to understand/implement.

| Monotone Chain shouldn't need to backtrack.
You mean that we should only add points that are "right turning", and its not needed to traverse backwards to remove "left turning" vertices? Yes we do, here is a counter example to that claim

Untitled Diagram drawio

vertices `<p_0,p_1,p_2>, <p_1,p_2,p_3>, <p_2,p_3,p_4>` are all right turning and would thus initially not be rejected. Only after backtracking from `p_5` we see that `p_1 ... p_4` should be rejected.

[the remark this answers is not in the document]
casperlamboo: https://github.com/Ultimaker/CuraEngine/pull/1676/commits/345cf7c542efa61746a1e48a1d12c2c11c60f2b8

Ghostkeeper:
I found the root of the problem and solved that in the original algorithm. The case in this example was as follows, with black the original polygon and blue the convex hull.
test

During the intermediary stage at vertex 127, it finds itself in a position where the convex hull, following the original shape further, keeps turning right (in the SVG visualisation it's right, in the g-code it's left due to flipped Y axis). It's locally still turning right, so the algorithm things everything is turning right and so the polygon must be convex. But in reality it's making 2 windings, not 1.

The fix is to check the winding while backtracking to see if the vertex is still right-turning if the convex hull would be closed up right then and there, towards the start vertex. If the vertex is concave w.r.t. the previous vertex and the start vertex, then it can't be in the convex hull and it must be removed too.

Joeydelarago (approved):

PR #1677, opened 2022-06-03
New polygonal decimation algorithm

Ghostkeeper:
This is a new algorithm for simplifying polygons. It is based on the Visvalingam–Whyatt algorithm.

The objectives of this new algorithm are:
* Remove forward-propagating effects. If one edge is moved, this can propagate forward along the polygon, causing all vertices to be moved and details to be unnecessarily removed.
* Use an industry-standard algorithm that has been subject to more research than an algorithm we wrote ourselves. Although this algorithm has been adjusted to fit our needs.
* More robustly remove zero-deviation vertices.
* Reduce code complexity and maintenance burden by having one algorithm for polylines and polygons, for Polygons and ExtrusionLines.

For a more detailed explanation of the algorithm, I refer to the documentation I added in this PR too: https://github.com/Ultimaker/CuraEngine/blob/CURA-8636_rdp/docs/simplify.md

Implements CURA-8636.
Fixes https://github.com/Ultimaker/Cura/issues/9979 .


src/utils/Simplify.h
```diff
@@ -0,0 +1,465 @@
...
+            to_delete[vertex] = true;
+            if(length2_before <= length2_after)
+            {
+                polygon[before] = createIntersection(polygon[before], intersection, polygon[after]);
+            }
+            else
+            {
+                polygon[after] = createIntersection(polygon[before], intersection, polygon[after]);
+            }
+        }
```
rburema, line 337: Can this (if block above here -- specifically, the one that replaces a vertex with an intersection) create a situation where the displaced vertex (before or after, depending) is already out of the importance queue (`by_importance`) in the calling method, but now, due to being placed at the intersection, in a position where a next simplify with the same parameters would remove this (or alter something else)?

src/LayerPlan.cpp
```diff
@@ -1104,7 +1105,7 @@ void LayerPlan::addLinesByOptimizer
         }
         boundary.add(comb_boundary_minimum.offset(dist));
         // simplify boundary to cut down processing time
-        boundary.simplify(MM2INT(0.1), MM2INT(0.1));
+        boundary = Simplify(MM2INT(0.1), MM2INT(0.1), 0).polygon(boundary);
```
rburema, line 1108: I know these where originally pretty arbitrary, but now the parameters aren't the squares of those values anymore, which is a big change even so.

src/SkirtBrim.cpp
```diff
@@ -80,8 +81,8 @@ void SkirtBrim::getFirstLayerOutline(SliceDataStorage& storage, const size_t pri
     first_layer_outline = first_layer_outline.offset(join_distance).offset(-join_distance); // merge adjacent models into single polygon
     constexpr coord_t smallest_line_length = 200;
     constexpr coord_t largest_error_of_removed_point = 50;
-    first_layer_outline.simplify(smallest_line_length, largest_error_of_removed_point); // simplify for faster processing of the brim lines
-    if (first_layer_outline.size() == 0)
+    first_layer_outline = Simplify(smallest_line_length, largest_error_of_removed_point, 0).polygon(first_layer_outline);
```
rburema, line 84: Are you sure these arguments weren't meant as squares originally? I know it _says_ length, not len2 or something, but they seem relatively big.

rburema (commented):
None of the things I found seem serious enough to send it back over (at least not before the component test), but I think you'll want to take anyway.

One last thing is that I'm not sure how we should handle the `0` value that is the third argument when dealing with polygons that do not have a width. On one hand I can sort of live with it, on the other hand, when you go look up the function (because you might want to know what that 'mysterious' zero is for in the third argument), the name `max_area_deviation` might be a bit confusing to the casual reader, as it's easily confused with the square of `max_deviation`, even though they measure completely different things (one the deviation of a position of a vertex, the other the deviation of the extrusion-width).

Anyway good work! This was a joy to read, I hardly even needed the comments.

Ghostkeeper:
> One last thing is that I'm not sure how we should handle the `0` value that is the third argument when dealing with polygons that do not have a width. On one hand I can sort of live with it, on the other hand, when you go look up the function (because you might want to know what that 'mysterious' zero is for in the third argument), the name `max_area_deviation` might be a bit confusing to the casual reader, as it's easily confused with the square of `max_deviation`, even though they measure completely different things (one the deviation of a position of a vertex, the other the deviation of the extrusion-width).

Would it help to give that argument a default value? Then we can leave out the `0` from the calling side at least. The danger of that of course is that it could be that the caller doesn't provide it even if it should, and then finds that the polygon doesn't get simplified at all since it's not allowed to deviate the line width.

src/utils/Simplify.h
```diff
@@ -0,0 +1,465 @@
...
+        {
+            to_delete[vertex] = true;
+            if(length2_before <= length2_after)
+            {
+                polygon[before] = createIntersection(polygon[before], intersection, polygon[after]);
+            }
+            else
+            {
+                polygon[after] = createIntersection(polygon[before], intersection, polygon[after]);
+            }
```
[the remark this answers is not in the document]
Ghostkeeper: I edited this to fix the indentation, and then committed it. Thanks for the suggestion!

src/utils/Simplify.h
Ghostkeeper: For the fixed-width case I can prove that this can't happen in the current implementation.

Crucially, it will only move the vertex such that two conditions hold:
* It will never decrease the deviation of that vertex. The deviation can only increase or remain the same.
* It will never reduce the length of the edges incident to the vertex. The length can only increase.

If a vertex is processed through the importance queue, it means that either it got deleted (and won't be seen again by `nextNotDeleted` etc.) or it had to stay because of its deviation or the length of its incident edges. Since the deviation can only increase, if it had to stay for the previous check it'll still have to stay. Since the edge length can only increase, if it had to stay due to not being below Max Resolution it'll still have to stay.
There's a third case where the vertex initially has two short edges incident, but due to moving the vertex it gets a long edge that is not allowed to shift (so it would need to calculate the intersection again and move the vertex). However this case can't happen for vertices that have already been processed because then it would have to have started with two short edges and so the vertex is already removed unless it has too much deviation.

For the variable-width case I'm less sure of it, because that adds the additional constraint of the max area deviation. The moved vertex gets the width that was halfway the line segment that got removed. But because the two vertices of the edge are also removed, there is no such constraint on the area deviation of the moved vertex. It might increase (w.r.t. either of the edge's vertices) or it might decrease.
I guess it could happen that a vertex replacing an edge gets a lower area deviation, if the vertices adjacent to the edge's incident vertices have a width that happens to be around the average of the width of the removed edge. If they are closer to the average than the edge's vertices themselves are. However I think that this case is rare enough and harmless enough that we can accept it as a shortcoming.

src/LayerPlan.cpp
Ghostkeeper: No, this was part of the confusion of the original function, really.

`PolygonRef::simplify` indeed takes squared distances.
`Polygons::simplify` does NOT take squared distances, but squares the value itself before passing it to `PolygonRef::simplify` for each of its sub-polygons.

src/SkirtBrim.cpp
Ghostkeeper: No, this was part of the confusion of the original function, really.

`PolygonRef::simplify` indeed takes squared distances.
`Polygons::simplify` does NOT take squared distances, but squares the value itself before passing it to `PolygonRef::simplify` for each of its sub-polygons.

This confusion is why I opted for simply not accepting squared distances at all any more. Keeping it simple!

src/utils/Simplify.h
```diff
@@ -0,0 +1,465 @@
...
+            //Even if the adjacent line segments are very long.
+            to_delete[vertex] = true;
+            return;
+        }
+
+        const size_t before = previousNotDeleted(vertex, to_delete);
+        const size_t after = nextNotDeleted(vertex, to_delete);
+        const Point vertex_position = getPosition(polygon[vertex]);
+        const Point before_position = getPosition(polygon[before]);
+        const Point after_position = getPosition(polygon[after]);
```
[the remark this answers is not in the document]
Ghostkeeper: This isn't so hard to change. Just add an `&` here and there. And in the case where it gets stored in a `const Point` it may save some performance.

src/utils/Simplify.h
```diff
@@ -0,0 +1,465 @@
...
+     * This will be edited in-place.
+     * \param vertex The index of the vertex to remove.
+     * \param deviation2 The previously found deviation for this vertex.
+     * \param is_closed Whether we're working on a closed polygon or an open
+     * polyline.
+     */
+    template<typename Polygonal>
+    void remove(Polygonal& polygon, std::vector<bool>& to_delete, const size_t vertex, const coord_t deviation2, const bool is_closed) const
+    {
+        if(deviation2 <= min_resolution * min_resolution)
```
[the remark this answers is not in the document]
Ghostkeeper: Since `min_resolution` is constexpr, this is guaranteed to be optimised by the compiler.

I think adding new variables makes it harder to follow, and changing the existing one to be squared is easy to misunderstand, so I'd rather keep it this way.

src/utils/Simplify.h
```diff
@@ -0,0 +1,465 @@
...
+            //The importance may have changed since this vertex was inserted. Re-compute it now.
+            //If it doesn't change, it's safe to process.
+            vertex_importance = importance(result, to_delete, vertex.first, is_closed);
+            if(vertex_importance != vertex.second)
+            {
+                by_importance.emplace(vertex.first, vertex_importance); //Re-insert with updated importance.
+                continue;
+            }
+
+            if(vertex_importance <= max_deviation * max_deviation)
```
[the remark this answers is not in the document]
Ghostkeeper: I'd prefer not to, for two reasons:
* In this particular case, the compiler will optimise it. While the C++ language doesn't guarantee that, I know that GCC will look for values in loops that can be pre-computed (and most likely Clang has something similar). And even if they didn't, CPU pipelining would likely make this operation free anyway. This is not an argument for pre-computing the `max_deviation` and other properties in the constructor (which the compiler will not optimise). However...
* The automated tests make use of the `max_deviation`, `max_resolution` and `max_area_deviation` properties here to validate the tests. They are public properties at the moment. We could change the fields to be squared, but to prevent confusion I'd also make them protected then. We'd have to sqrt there, or we'd have to change the tests to give them the values of the fields there. Neither of them looks nice.

src/utils/Simplify.h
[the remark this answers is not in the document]
Ghostkeeper: Done here: https://github.com/Ultimaker/CuraEngine/pull/1677/commits/acfe3acc9c4771c40caf1797768f2d352f60360d

src/utils/Simplify.cpp
```diff
@@ -0,0 +1,175 @@
...
+Simplify::Simplify(const coord_t max_resolution, const coord_t max_deviation, const coord_t max_area_deviation)
+    : max_resolution(max_resolution)
+    , max_deviation(max_deviation)
+    , max_area_deviation(max_area_deviation)
+{}
+
+Simplify::Simplify(const Settings& settings)
+    : max_resolution(settings.get<coord_t>("meshfix_maximum_resolution"))
+    , max_deviation(settings.get<coord_t>("meshfix_maximum_deviation"))
+    , max_area_deviation(settings.get<coord_t>("meshfix_maximum_extrusion_area_deviation"))
```
Piezoid, line 21: This divides the `meshfix_maximum_extrusion_area_deviation` value by 1000. This value is already in um^2.

PR #1682, opened 2022-06-16
Cura 9365 fix building cura main

jellespijker:
Doc to come

Workflow and builds don't work since our free tier JFrog subscription has exhausted it's bandwidth.
We're in the process of Upgrading to JFrog Artifactory Enterprise to host our artifacts


conanfile.py
```diff
@@ -139,5 +139,6 @@ def build(self):
     def package(self):
         packager = files.AutoPackager(self)
         packager.run()
-
+        if self.settings.os in ["Macos", "Linux"]:
```
jellespijker, line 142: No need to check against OS, Windows won't produce a file called CuraEngine (without extension), so it won't copy anything.

PR #1683, opened 2022-06-18
[CURA-7848] [CURA-9295] Also cut infill wall-toolpaths w.r.t. gradual infill.

rburema:
Infill wall-toolpaths where generated for each density, which caused them to overlap in case of zig-zaggified and gradual infill.

First identified as a feature-gap in CURA-7848, which then was on the backlog long enough to be reported as the separate bug CURA-9295.


rburema:
Re-discussed, declined in favour of #1685.

PR #1685, opened 2022-06-28
Don't fill gaps more than once when using gradual infill

Ghostkeeper:
When using gradual infill or gradual support, the infill algorithm is called multiple times, once for each density index. If the infill pattern is one that keeps half-line-width distance from the boundary (such as concentric, or connected infill lines) then it will fill gaps smaller than 1 line width with a variable-width path. This gap filling was occurring for each density, because there was no check on whether it was already done by a previous density index.

I added a parameter to the infill constructor (the 19th parameter! such params, much config, wowzers) that tells the infill generator to fill gaps or not. For the infill pattern usage that has multiple densities (infill and support), this is only enabled for the lowest density parameter, so that the gap filling only happens once for all of the densities. This prevents overextruding in those areas.

Unfortunately, if the infill gaps are not filled, it still needs to calculate the area for those gaps, in order to leave them open specifically. Otherwise the normal infill pattern will still fill those areas and fill them badly that way.

Contributes to issue CURA-9295.


rburema:
We should probably drop the old TODO comment as well then, since this change'd take care of that (and leaving internal ticket nrs in OS code is a bit ugly).

rburema (approved):

PR #1687, opened 2022-06-30
[CURA-9225] 'Prime' for mutliple extruders in Raft

rburema:
__The problem:__
Since the raft can have multiple extruders now, the also need to be primed. However, extending the prime tower to the bottom would be problematic as well.

It was decided that printing a prime-tower raft first (on each layer) would sufficiently prime each nozzle. However, the raft before this PR was monolithic, meaning it was calculated for the entire build-plate at once.

(Priming with a blob was also a problem, since for certain materials the time between the blob and, say, the printing of the top layer, given that the bottom layers of the raft use an other extruder was too large.)

__The resolution:__
This PR splits the raft into a 'model' and a 'prime tower' part. Also, it guarantees printing the prime tower part of the raft if the prime tower is enabled, but not required for the actual model on top of the raft _if_ there are different extruders used for different layers of the raft.

_A note with regards to the raft convexity setting/requirement for certain materials:_
The 'model', as well as the raft _as a whole_ can still be made convex with the 'remove inside corners' option as required. The convexity specifically of the 'prime tower part' of the raft is only guaranteed if the rafts don't overlap in that case. If they do overlap, it is hoped that the convexity of the raft as a whole is enough to sufficiently reduce shearing effects and the like.

_A note w.r.t. code review:_
Hiding whitespace changes may make this easier. (One of the options when you click on the gear-menu in github when viewed with 'files changed', or `-w` on the command-line when viewed with `git diff`, so `git diff -w`.)

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -754,7 +754,7 @@ void FffGcodeWriter::processRaft(const SliceDataStorage& storage)
         }
         raft_outline_paths.emplace_back(storage.raftOutline);
 
-        for (const auto& raft_outline_path : raft_outline_paths)
+        for (const Polygons& raft_outline_path : raft_outline_paths)
         {
             Infill infill_comp(
                 EFillMethod::LINES, zig_zaggify_infill, connect_polygons, raft_outline_path, gcode_layer.configs_storage.raft_base_config.getLineWidth(), line_spacing,
@@ -846,7 +846,7 @@ void FffGcodeWriter::processRaft(const SliceDataStorage& storage)
         constexpr int zag_skip_count = 0;
         constexpr coord_t pocket_size = 0;
 
-        for (const auto& raft_outline_path : raft_outline_paths)
+        for (const Polygons& raft_outline_path : raft_outline_paths)
         {
             Infill infill_comp(
                 EFillMethod::ZIG_ZAG, zig_zaggify_infill, connect_polygons, raft_outline_path, infill_outline_width, interface_line_spacing,
@@ -921,7 +921,7 @@ void FffGcodeWriter::processRaft(const SliceDataStorage& storage)
         constexpr size_t zag_skip_count = 0;
         constexpr coord_t pocket_size = 0;
 
-        for (const auto& raft_outline_path : raft_outline_paths)
+        for (const Polygons& raft_outline_path : raft_outline_paths)
         {
             Infill infill_comp(
                 EFillMethod::ZIG_ZAG, zig_zaggify_infill, connect_polygons, raft_outline_path, infill_outline_width, surface_line_spacing,
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -747,24 +747,37 @@ void FffGcodeWriter::processRaft(const SliceDataStorage& storage)
         const coord_t max_resolution = base_settings.get<coord_t>("meshfix_maximum_resolution");
         const coord_t max_deviation = base_settings.get<coord_t>("meshfix_maximum_deviation");
 
-        Infill infill_comp(
-            EFillMethod::LINES, zig_zaggify_infill, connect_polygons, storage.raftOutline, gcode_layer.configs_storage.raft_base_config.getLineWidth(), line_spacing,
-            fill_overlap, infill_multiplier, fill_angle, z, extra_infill_shift,
-            max_resolution, max_deviation,
-            wall_line_count, infill_origin, skip_stitching, fill_gaps, connected_zigzags, use_endpieces, skip_some_zags, zag_skip_count, pocket_size
-            );
-        std::vector<VariableWidthLines> raft_paths;
-        infill_comp.generate(raft_paths, raft_polygons, raftLines, base_settings);
-        if(!raft_paths.empty())
+        std::vector<Polygons> raft_outline_paths;
+        if (storage.primeRaftOutline.area() > 0)
         {
-            const GCodePathConfig& config = gcode_layer.configs_storage.raft_base_config;
-            const ZSeamConfig z_seam_config(EZSeamType::SHORTEST, gcode_layer.getLastPlannedPositionOrStartingPosition(), EZSeamCornerPrefType::Z_SEAM_CORNER_PREF_NONE, false);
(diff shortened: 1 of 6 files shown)


src/raft.cpp
```diff
@@ -22,30 +22,53 @@ void Raft::generate(SliceDataStorage& storage)
...
     }
     else
     {
         const coord_t smoothing = settings.get<coord_t>("raft_smoothing");
-        storage.raftOutline = storage.raftOutline.offset(smoothing, ClipperLib::jtRound).offset(-smoothing, ClipperLib::jtRound); // remove small holes and smooth inward corners
+        global_raft_outlines = global_raft_outlines.offset(smoothing, ClipperLib::jtRound).offset(-smoothing, ClipperLib::jtRound); // remove small holes and smooth inward corners
+    }
+
+    constexpr bool dont_include_prime_tower = false;
+    Polygons raw_raft_without_prime{ storage.getLayerOutlines(0, include_support, dont_include_prime_tower).offset(distance, ClipperLib::jtRound) };
```
Ghostkeeper, line 52: Could we maybe use this one to start with, and only combine in `global_raft_outlines` with `storage.primeTower.outer_poly_first_layer`? Would prevent calling `getLayerOutlines` multiple times and doing that offset multiple times.

src/FffGcodeWriter.cpp
```diff
@@ -888,15 +921,21 @@ void FffGcodeWriter::processRaft(const SliceDataStorage& storage)
         constexpr size_t zag_skip_count = 0;
         constexpr coord_t pocket_size = 0;
 
-        Infill infill_comp(
-            EFillMethod::ZIG_ZAG, zig_zaggify_infill, connect_polygons, raft_outline_path, infill_outline_width, surface_line_spacing,
-            fill_overlap, infill_multiplier, fill_angle, z, extra_infill_shift,
-            surface_max_resolution, surface_max_deviation,
-            wall_line_count, infill_origin, skip_stitching, fill_gaps, connected_zigzags, use_endpieces, skip_some_zags, zag_skip_count, pocket_size
+        for (const auto& raft_outline_path : raft_outline_paths)
```
Ghostkeeper, line 924: Going by the variable name, I'd think that this would be a variable-width path or a toolpath.

Writing down the type would clarify this. Alternatively, we can consider renaming it to `_centerline`.
```suggestion
        for (const Polygons& raft_outline_path : raft_outline_paths)
```

src/FffGcodeWriter.cpp
```diff
@@ -826,15 +846,21 @@ void FffGcodeWriter::processRaft(const SliceDataStorage& storage)
         constexpr int zag_skip_count = 0;
         constexpr coord_t pocket_size = 0;
 
-        Infill infill_comp(
-            EFillMethod::ZIG_ZAG, zig_zaggify_infill, connect_polygons, raft_outline_path, infill_outline_width, interface_line_spacing,
-            fill_overlap, infill_multiplier, fill_angle, z, extra_infill_shift,
-            interface_max_resolution, interface_max_deviation,
-            wall_line_count, infill_origin, skip_stitching, fill_gaps, connected_zigzags, use_endpieces, skip_some_zags, zag_skip_count, pocket_size
+        for (const auto& raft_outline_path : raft_outline_paths)
```
Ghostkeeper, line 849: Going by the variable name, I'd think that this would be a variable-width path or a toolpath.

Writing down the type would clarify this. Alternatively, we can consider renaming it to `_centerline`.
```suggestion
        for (const Polygons& raft_outline_path : raft_outline_paths)
```

src/FffGcodeWriter.cpp
```diff
@@ -747,24 +747,37 @@ void FffGcodeWriter::processRaft(const SliceDataStorage& storage)
...
-            const ZSeamConfig z_seam_config(EZSeamType::SHORTEST, gcode_layer.getLastPlannedPositionOrStartingPosition(), EZSeamCornerPrefType::Z_SEAM_CORNER_PREF_NONE, false);
-            InsetOrderOptimizer wall_orderer(*this, storage, gcode_layer, base_settings, base_extruder_nr,
-                                             config, config, config, config,
-                                             retract_before_outer_wall, wipe_dist, wipe_dist, base_extruder_nr, base_extruder_nr, z_seam_config, raft_paths);
-            wall_orderer.addToLayer();
+            raft_outline_paths.emplace_back(storage.primeRaftOutline);
+        }
+        raft_outline_paths.emplace_back(storage.raftOutline);
+
+        for (const auto& raft_outline_path : raft_outline_paths)
```
Ghostkeeper, line 757: Going by the variable name, I'd think that this would be a variable-width path or a toolpath.

Writing down the type would clarify this. Alternatively, we can consider renaming it to `raft_outlines` or `raft_areas`.
```suggestion
        for (const Polygons& raft_outline_path : raft_outline_paths)
```

src/support.cpp
```diff
@@ -500,8 +500,8 @@ Polygons AreaSupport::join(const SliceDataStorage& storage, const Polygons& supp
                 break;
             case EPlatformAdhesion::RAFT:
             {
-                const ExtruderTrain& raft_extruder = mesh_group_settings.get<ExtruderTrain&>("adhesion_extruder_nr");
-                adhesion_size = raft_extruder.settings.get<coord_t>("raft_margin");
+                const ExtruderTrain& raft_base_extruder = mesh_group_settings.get<ExtruderTrain&>("raft_base_extruder_nr");
+                adhesion_size = raft_base_extruder.settings.get<coord_t>("raft_margin");
```
Ghostkeeper, line 504: The `raft_margin` setting gets set to `limit_to_extruder = adhesion_extruder_nr`, so you'd expect `adhesion_extruder_nr` to be the correct extruder?

Ghostkeeper (changes requested):

Ghostkeeper (approved):

Ghostkeeper:
Works well for the most part, but I found one bug.

If the prime tower does need to actually get generated for the print, it now generates through the raft, intersecting with it and overextruding.
image

This is a project file (in a zip) that reproduces this case:
[prime_tower_intersects_raft.zip](https://github.com/Ultimaker/CuraEngine/files/9047477/prime_tower_intersects_raft.zip)

rburema:
@Ghostkeeper Not the actual prime-tower it seems ... only the adhesion for it. That would also explain why I haven't seen it yet in my own tests.

casperlamboo (approved):

PR #1701, opened 2022-07-27
Cura 9031 add and send slice UUID

jellespijker:
This PR is the backend part of Ultimaker/Cura#12881

Generate a UUID for a slice add it to the Griffin header in the GCode and make sure that the front-end is aware of that UUID,
by sending an newly created Arcus message.

### DIFF
--- a/Cura.proto
+++ b/Cura.proto
@@ -139,5 +139,9 @@ message GCodePrefix {
     bytes data = 2; //Header string to be prepended before the rest of the g-code sent from the engine.
 }
 
+message SliceUUID {
+    string slice_uuid = 1; //The UUID of the slice.
+}
+
 message SlicingFinished {
 }
--- a/Cura.proto
+++ b/Cura.proto
@@ -140,7 +140,7 @@ message GCodePrefix {
 }
 
 message SliceUUID {
-    string slice_uuid = 1; //The UUID of the slice.
+    string slice_uuid = 1; //An ID to link the slice in the printer to a statistical data point in the anonymous slice data.
 }
 
 message SlicingFinished {
--- a/src/gcodeExport.cpp
+++ b/src/gcodeExport.cpp
@@ -1,5 +1,5 @@
-//  Copyright (c) 2021-2022 Ultimaker B.V.
-//  CuraEngine is released under the terms of the AGPLv3 or higher.
+// Copyright (c) 2022 Ultimaker B.V.
+// CuraEngine is released under the terms of the AGPLv3 or higher.
 
 #include <assert.h>
 #include <cmath>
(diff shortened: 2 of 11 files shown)


src/gcodeExport.h
```diff
@@ -218,23 +217,29 @@ class GCodeExport : public NoCopy
...
+
+    void setSliceUUID(const std::string& slice_uuid);
 
     void setLayerNr(unsigned int layer_nr);
 
     void setOutputStream(std::ostream* stream);
 
-    bool getExtruderIsUsed(const int extruder_nr) const; //!< return whether the extruder has been used throughout printing all meshgroup up till now
+    bool getExtruderIsUsed(
+      const int extruder_nr) const; //!< return whether the extruder has been used throughout printing all meshgroup up till now
```
Ghostkeeper, line 239: Why did you break it up here, rather than at the comment?

I'd suggest to write the comment as an actual documentation block instead so that Doxygen can pick up on it.

tests/arcus/MockCommunication.h
```diff
@@ -17,14 +17,29 @@ namespace cura
  */
 class MockCommunication : public Communication
 {
-public:
+  public:
```
Ghostkeeper, line 20: Indent with 4 spaces, not 2.

Cura.proto
```diff
@@ -139,5 +139,9 @@ message GCodePrefix {
     bytes data = 2; //Header string to be prepended before the rest of the g-code sent from the engine.
 }
 
+message SliceUUID {
+    string slice_uuid = 1; //The UUID of the slice.
```
Ghostkeeper, line 143: Superfluous documentation here. I'd suggest leaving it out, or actually documenting what it is.

```suggestion
    string slice_uuid = 1; //An ID to link the slice in the printer to a statistical data point in the anonymous slice data.
```

Ghostkeeper (changes requested):

src/gcodeExport.cpp
```diff
@@ -1,5 +1,5 @@
-//Copyright (c) 2021 Ultimaker B.V.
-//CuraEngine is released under the terms of the AGPLv3 or higher.
+//  Copyright (c) 2021-2022 Ultimaker B.V.
```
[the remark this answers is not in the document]
jellespijker: Yes that was still done with my previous Clion copyright template, will change this.

src/gcodeExport.h
```diff
@@ -1,5 +1,5 @@
-//Copyright (c) 2021 Ultimaker B.V.
-//CuraEngine is released under the terms of the AGPLv3 or higher.
+//  Copyright (c) 2021-2022 Ultimaker B.V.
```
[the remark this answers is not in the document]
jellespijker: Se comment above

src/gcodeExport.h
jellespijker: Done we the current clang-format in this branch. Will change this manually for now to 4 spaces so I don't get a merge conflict if the fix failing tests is merged

Ghostkeeper (approved):

PR #1713, opened 2022-08-23
[CURA-9367] Add forced concentric skin in removed areas.

rburema:
Areas removed by Skin Removal Width get filled in with 100% Concentric Fill.


rburema:
Superseded. Devs, see internal tickets CURA-10670 and CURA-10829 for the actual implementation.

PR #1717, opened 2022-08-26
[CURA-9515] Fix over-extrusion due to premature optimization in monotonic.

rburema:
The first case (adjacent line-segments and only one polyline-string) was creating problems, but there's no reason to not just solve it by the simpler more general case of adjacent line-segments and multiple polyline-strings, where multiple can now also be 1 in addition to 0 and more than 1.


casperlamboo (approved):

jellespijker:
closing and reopening this as a test for the workflow

PR #1730, opened 2022-09-11
Avoid moving iterators out of range

### LINKED ISSUE #1663: CuraEngine crashes under Windows when compiled in debug mode.
**Application Version**
5.0.0 Beta
**Platform**
Windows

I noticed some potential issues while trying to run a CuraEngine compiled under windows using the debug configuration. Both were `can't decrement vector iterator before begin` errors. 
 
1. In SkeletalTrapezoidation::generateExtraRibs in the [line 1311](https://github.com/Ultimaker/CuraEngine/blob/ac0aae947fbbead4861830e8ba65b6cd1de2098b/src/SkeletalTrapezoidation.cpp#L1311) there is an exception thrown, because in the first iteration std::prev is called with graph.edges.begin().
#1662 should fix this one.
    
2. Similarly, in `Polygons::removeSmallAreas`, the iterator `it`  can be decreased before the begin iterator when it reaches [`i--` in line 708 ](https://github.com/Ultimaker/CuraEngine/blob/ac0aae947fbbead4861830e8ba65b6cd1de2098b/src/utils/polygon.cpp#L708) in the first iteration.

I don't know if there are more cases where a crash in debug mode is caused by an iterator to be decremented before begin, as i wasn't sure how to fix the second one.

Piezoid:
Superseeds #1593 and #1662, resolve #1663.

As noticed in discussion of #1432, this PR introduced another case of iterators decremented before `begin()`.
While it is not an issue with release builds on platforms using pointer wrapper as iterators, it is undefined behavior after all. More practically, debug builds catches this which is annoying (iterator checking is enabled by default on mscv).

This PR resolves two cases: one in `Polygons::removeSmallAreas` and one in `SkeletalTrapezoidation::isEndOfCentral`.

I cherry-picked the `removeSmallAreas` tests from #1593 by @plaintoothpaste. I found that the implementation was working as intended but the tests were not expecting reordering of the `Polygons` path entries.

@rburema I investigated the erase-remove idiom in this application but found that it likely doesn't work for the case `remove_holes=false`. The reference given to the predicate is not stable since even kept elements can be shifted. This forbids keeping a reference to small holes for latter filtering. 
Beside, without the order-preserving requirement of `remove_if`, we can do a move per removed element instead of shifting the tail of the vector like `remove_if` does. This might be slightly more efficient for the common case with lot of polygons but few filtered out.

### DIFF
--- a/src/SkeletalTrapezoidation.cpp
+++ b/src/SkeletalTrapezoidation.cpp
@@ -1321,8 +1321,8 @@ bool SkeletalTrapezoidation::isEndOfCentral(const edge_t& edge_to) const
 
 void SkeletalTrapezoidation::generateExtraRibs()
 {
-    auto end_edge_it = --graph.edges.end(); // Don't check newly introduced edges
-    for (auto edge_it = graph.edges.begin(); std::prev(edge_it) != end_edge_it; ++edge_it)
+    const auto end_edge_it = graph.edges.end(); // Don't check newly introduced edges
+    for (auto edge_it = graph.edges.begin(); edge_it != end_edge_it; ++edge_it)
     {
         edge_t& edge = *edge_it;
 
--- a/src/utils/polygon.cpp
+++ b/src/utils/polygon.cpp
@@ -527,33 +527,35 @@ void Polygons::removeEmptyHoles_processPolyTreeNode(const ClipperLib::PolyNode& 
 void Polygons::removeSmallAreas(const double min_area_size, const bool remove_holes)
 {
     auto new_end = paths.end();
-    if(remove_holes)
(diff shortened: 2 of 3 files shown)


rburema (approved):

rburema:
Hi @Piezoid Sorry it took so long to get back. I should have at least replied if I'm mentioned, but I see that didn't even happen.

Should it still matter; I'm fine with using a more verbose way of writing if that's more correct or has other advantages like the (probable) optimization that you mentioned.

In any case, thanks again for your contributions! (... and keeping the timeline honest with proper cherry-picks 🙂)

PR #1738, opened 2022-10-05
Simplify: fix scaling of maximum area deviation setting

Piezoid:
`meshfix_maximum_extrusion_area_deviation` is already in millimeters. Using `get<coord_t>` divides the value by 1000, instead the setting should be acceded with `get<size_t>` or `get<int>`.

I've already signaled and fixed this in #1710, but maybe this simple fix has a chance of making it into the 5.2 release?

The frontend commits that recently changed the default value Ultimaker/Cura@03c88c18e6523ad370f094730d43f85e90d3a390 should probably be re-evaluated.

### DIFF
--- a/src/utils/Simplify.cpp
+++ b/src/utils/Simplify.cpp
@@ -18,7 +18,7 @@ Simplify::Simplify(const coord_t max_resolution, const coord_t max_deviation, co
 Simplify::Simplify(const Settings& settings)
     : max_resolution(settings.get<coord_t>("meshfix_maximum_resolution"))
     , max_deviation(settings.get<coord_t>("meshfix_maximum_deviation"))
-    , max_area_deviation(settings.get<coord_t>("meshfix_maximum_extrusion_area_deviation"))
+    , max_area_deviation(settings.get<size_t>("meshfix_maximum_extrusion_area_deviation"))
 {}
 
 Polygons Simplify::polygon(const Polygons& polygons) const


jellespijker:
You're right that this is handled incorrectly. Looking at the front end we use even use `μm²`

I will create a front-end PR to  make sure that the values in the profiles, match the changes in this PR.

I myself wasn't fully aware that when we grab a `coord_t` type from the settings we multiply by a factor of thousand. I will do a scan one of these days to double check all `settings.get<coord_t>` against their intended use.

jellespijker:
On second thought I'm not sure if the values in the front-end need to change
```python
>>> from pint import UnitRegistry
>>> u = UnitRegistry()
>>> width = 400 * u.um
>>> old_default = 50000 * u.um**2 * 1000
>>> old_length_dev = old_default / width
>>> old_length_dev.to("mm")
<Quantity(125.0, 'millimeter')>
```
I will contact @rijkvanmanen or @pkuiper-ultimaker to discuss how they feel about changing the default values in the front-end profiles

jellespijker:
Discussed this with @rijkvanmanen it looks like we use this as a cap in the simplify function; This never came into play due to the insanely high value because of that multiplication with a factor 1000. Even without that factor the impact of this change seems limited but it is good defensive programming imo and correct usage of the intended value.

I'm gonna merge this to main, thanks @Piezoid for contributing this fix

jellespijker (approved):

rburema:
This shouldn't be in 5.2 I think. I've merged it to main now, but we should probably undo it in the 5.2 branch, or maybe even force push it back to where it was before.

PR #1739, opened 2022-10-07
Cura 9710

casperlamboo:
Due to a bug in the [`getOverlappingLines`](https://github.com/Ultimaker/CuraEngine/blob/9d6056b24e2fb58fa3497ffd142950dc265438aa/include/PathOrderMonotonic.h#L399) function some very close lines would not be classified as "overlapping". These overlapping lines are marked as possible connections to travel between when printing the monotonic infill. 
https://github.com/Ultimaker/CuraEngine/blob/9d6056b24e2fb58fa3497ffd142950dc265438aa/include/PathOrderMonotonic.h#L167-L170
After all these "connections" are made between overlapping lines the final ordering of monotonic paths is made. This is done by finding a long string of overlapping lines, each visited line is marked so we can skip already selected paths.
https://github.com/Ultimaker/CuraEngine/blob/9d6056b24e2fb58fa3497ffd142950dc265438aa/include/PathOrderMonotonic.h#L130-L133
As one of the overlapping lines is not added to the sequence is not added to the visited paths it will be handed in a next iteration. This new line is connected to successive paths so these successive paths are printed double.

The bug in [`getOverlappingLines`](https://github.com/Ultimaker/CuraEngine/blob/9d6056b24e2fb58fa3497ffd142950dc265438aa/include/PathOrderMonotonic.h#L399) introduced because we did not take the line width into account for part of the computations.
Untitled Diagram drawio
To check if two paths path was intersecting we projected both end points of both paths onto the monotonic vector and onto the perpendicular vector. If in both these vector spaces the line segment ranges would intersect we would classify the paths as overlapping. For the projections on the monotonic vector we would already take the line width into account, but we did not do that yet in the perpendicular vector space. This is added in https://github.com/Ultimaker/CuraEngine/pull/1739/commits/9d6056b24e2fb58fa3497ffd142950dc265438aa.

### DIFF
--- a/include/PathOrderMonotonic.h
+++ b/include/PathOrderMonotonic.h
@@ -80,7 +80,8 @@ public:
             else
             {
                 polylines.push_back(&path);
-                polylines.back()->start_vertex = polylines.back()->converted->size(); //Assign an invalid starting vertex to indicate we don't know the starting point yet.
+                // Assign an invalid starting vertex to indicate we don't know the starting point yet.
+                polylines.back()->start_vertex = polylines.back()->converted->size();
             }
         }
 
@@ -100,11 +101,7 @@ public:
         SparsePointGridInclusive<Path*> line_bucket_grid(MM2INT(2)); //Grid size of 2mm.
         for(Path* polyline : polylines)
         {
-            if(polyline->converted->empty())
-            {
-                continue;
-            }
-            else
+            if(!polyline->converted->empty())
             {
                 line_bucket_grid.insert(polyline->converted->front(), polyline);
                 line_bucket_grid.insert(polyline->converted->back(), polyline);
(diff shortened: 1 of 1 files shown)


include/PathOrderMonotonic.h
```diff
@@ -168,9 +165,24 @@ class PathOrderMonotonic : public PathOrder<PathType>
                     starting_lines.insert(*polyline_it); //This is a starting point then.
                 }
                 const std::vector<Path*> overlapping_lines = getOverlappingLines(polyline_it, perpendicular, polylines);
-                for(Path* overlapping_line : overlapping_lines)
+                if(overlapping_lines.size() == 1) //If we're not a string of polylines, but adjacent to only one other polyline, create a sequence of polylines.
+                {
+                    connections[*polyline_it] = overlapping_lines[0];
+                    if(connected_lines.find(overlapping_lines[0]) != connected_lines.end()) //This line was already connected to.
```
jellespijker, line 171:
```suggestion
                    if(connected_lines.contains(overlapping_lines[0])) //This line was already connected to.
```

include/PathOrderMonotonic.h
```diff
@@ -100,11 +101,7 @@ class PathOrderMonotonic : public PathOrder<PathType>
         SparsePointGridInclusive<Path*> line_bucket_grid(MM2INT(2)); //Grid size of 2mm.
         for(Path* polyline : polylines)
         {
-            if(polyline->converted->empty())
-            {
-                continue;
-            }
-            else
+            if(!polyline->converted->empty())
```
jellespijker, line 104:
```suggestion
            if(! polyline->converted->empty())
```

jellespijker (changes requested):
Some minor suggestion

jellespijker (approved):

PR #1740, opened 2022-10-11
Use drop_last to loop over ranges

jellespijker:
More robust method to loop over the graph.edges

While helping out @rijkvanmanen with a crash in one of his mini-spike branches we ran across a situation where the Engine crashed, when the graph.edges weren't filled as expected. It is unlikely that this would happen in the wild. But it can't hurt to do some defensive programming imo.

> [views::drop_last](https://ericniebler.github.io/range-v3/structranges_1_1views_1_1drop__last__fn.html)
Given a source range and an integral count, return a range consisting of all but the last count elements from the source range, or an empty range if it has fewer elements.

### DIFF
--- a/src/SkeletalTrapezoidation.cpp
+++ b/src/SkeletalTrapezoidation.cpp
@@ -4,7 +4,6 @@
 #include "SkeletalTrapezoidation.h"
 
 #include <spdlog/spdlog.h>
-#include <range/v3/view/drop_last.hpp>
 
 #include "settings/types/Ratio.h"
 #include <functional>
--- a/src/SkeletalTrapezoidation.cpp
+++ b/src/SkeletalTrapezoidation.cpp
@@ -4,6 +4,7 @@
 #include "SkeletalTrapezoidation.h"
 
 #include <spdlog/spdlog.h>
+#include <range/v3/view/drop_last.hpp>
 
 #include "settings/types/Ratio.h"
 #include <functional>
@@ -1321,21 +1322,19 @@ bool SkeletalTrapezoidation::isEndOfCentral(const edge_t& edge_to) const
 
 void SkeletalTrapezoidation::generateExtraRibs()
 {
-    auto end_edge_it = --graph.edges.end(); // Don't check newly introduced edges
-    for (auto edge_it = graph.edges.begin(); std::prev(edge_it) != end_edge_it; ++edge_it)
+    for (auto edge : graph.edges | ranges::views::drop_last(1))
     {
-        edge_t& edge = *edge_it;
-
         if (! edge.data.isCentral() || shorterThen(edge.to->p - edge.from->p, discretization_step_size) || edge.from->data.distance_to_boundary >= edge.to->data.distance_to_boundary)
         {
             continue;
         }
 
-
         std::vector<coord_t> rib_thicknesses = beading_strategy.getNonlinearThicknesses(edge.from->data.bead_count);
 
         if (rib_thicknesses.empty())
+        {
             continue;
+        }
 
         // Preload some variables before [edge] gets changed
         node_t* from = edge.from;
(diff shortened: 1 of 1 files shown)


src/SkeletalTrapezoidation.cpp
```diff
@@ -1321,21 +1322,19 @@ bool SkeletalTrapezoidation::isEndOfCentral(const edge_t& edge_to) const
 
 void SkeletalTrapezoidation::generateExtraRibs()
 {
-    auto end_edge_it = --graph.edges.end(); // Don't check newly introduced edges
-    for (auto edge_it = graph.edges.begin(); std::prev(edge_it) != end_edge_it; ++edge_it)
+    for (auto edge : graph.edges | ranges::views::drop_last(1))
```
casperlamboo, line 1325: No remarks about the updated ranges-style code, but I find the original code here really weird. As such I'm not so sure if the suggested changes are equivalent to the original code.

Let's try to rewrite the iterator to a super basic index-based for loop.

```cpp
auto end_edge_it = --graph.edges.end();
```
Is rewritten to
```cpp
auto end_edge_index = graph.edges.size() - 1;
```
And the for loop
```cpp
for (auto edge_it = graph.edges.begin(); std::prev(edge_it) != end_edge_it; ++edge_it)
```
Translates to
```cpp
for (auto edge_index = 0; edge_index - 1 != end_edge_index; ++edge_index)
```

Lets simplify the for loop a bit
```cpp
for (auto edge_index = 0; edge_index - 1 != graph.edges.size() - 1; ++edge_index)
for (auto edge_index = 0; edge_index != graph.edges.size(); ++edge_index)
for (auto edge_index = 0; edge_index < graph.edges.size(); ++edge_index)
```

All edges are visited according to this logic; evalualting the following code
```cpp
#include <iostream>
#include <vector>

int main() {
  std::vector<int> v = { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 };
  auto end_edge_it = --v.end();
  for (auto edge_it = v.begin(); std::prev(edge_it) != end_edge_it; ++edge_it)
  {
    int value = *edge_it;
    std::cout << "value: " << value << std::endl;
  }
}
```
produces
```
value: 1
value: 2
value: 3
value: 4
value: 5
value: 6
value: 7
value: 8
value: 9
value: 10
```

Please verify my logic, but in my opinion the `drop_last(1)` should not be added if we don't want to change the result of the program. 

We can still argue about the intention and validity of the original code.

```suggestion
    for (auto edge : graph.edges)
```

casperlamboo (changes requested):

src/SkeletalTrapezoidation.cpp
jellespijker: > Please verify my logic, but in my opinion the drop_last(1) should not be added if we don't want to change the result of the program.

Damn you and your logic, it holds!

src/SkeletalTrapezoidation.cpp
```diff
@@ -4,6 +4,7 @@
 #include "SkeletalTrapezoidation.h"
 
 #include <spdlog/spdlog.h>
+#include <range/v3/view/drop_last.hpp>
```
casperlamboo, line 7: Import is redundant now
```suggestion
```

jellespijker:
@MariMakes found out that his is probably solved already by @rburema when she did here first stacktrace

jellespijker:
You can close this PR I think

jellespijker:
@casperlamboo It seems that we broke something with this PR? @MariMakes will share a stacktrace and the model

rburema:
... not _solved_ exactly, ahemhem ... I took a photo so I would remember to look into it. Which is a lesson for the future, don't _just_ take a photo, also write it into my TODO's.

PR #1741, opened 2022-10-11
Add Point fmt formatter

jellespijker:
This allows spdlog and/or fmt to output a single point as [X, Y]. With the inclusion of the fmt/ranges.h a container of points will be: [[X_0, Y_0], [X_1, Y_1], ..., [X_n, Y_n]] Usage as:

```c++
std::vector<Point> points {{100, 200}, {300, 400}, {500, 600}};
spdlog:debug("points: {}", points};
auto fmt::format("{}", points};
```

Easy debugging commence!

Created while working on: CURA-9377 💥 Slicing error with Top/Bottom line width, which I'm dropping in favor of other tickets.

Although the file currently isn't used, I still believe we should merge, because it is a helpful class to say the least.

### DIFF
--- a/include/utils/format.h
+++ b/include/utils/format.h
@@ -1,0 +1,40 @@
+// Copyright (c) 2022 Ultimaker B.V.
+// CuraEngine is released under the terms of the AGPLv3 or higher.
+
+#ifndef CURAENGINE_UTILS_FORMAT_H
+#define CURAENGINE_UTILS_FORMAT_H
+
+#include <fmt/format.h>
+#include <fmt/ranges.h>
+
+namespace ClipperLib {
+    class IntPoint; // Forward declaration
+}
+
+template<>
+struct [[maybe_unused]] fmt::formatter<ClipperLib::IntPoint> {
+    constexpr auto parse(fmt::format_parse_context &ctx) -> decltype(ctx.begin()) {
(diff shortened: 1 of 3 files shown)


casperlamboo:
I have a slight preference for the `(X, Y)` over the `[X, Y]` notation as brackets denote ranges in mathematics while parenthesis denote tuples. Tuples describe points better in my opinion.

jellespijker:
I'm fine with both

rburema:
Well, if it's about points, those are (even) closer to vectors (than tuples), and then I'd prefer `<x, y>` ;-p -- At least row vectors, for _column_ vectors, you'd use matrix notation, and those _are_ square brackets.

Given that this is quite arbitrary:

However, Jelle told me in the thread on Slack that the reason was easy(er) parsing (specifically for Python). I think an actual _use_ is more important then what we'd prefer as notation.

TL;DR: Keep as is.

rburema:
Fails for me on windows with

`C:\.conan\9e8431\1\include\fmt/core.h(2716): error C2039: 'parse': is not a member of 'fmt::v9::detail::fallback_formatter<stripped_type,char_type,void>'`

(even if I do include the IntPoint header).

jellespijker:
Did you do a `conan install .` Looks like you don't have the used `fmt` version

rburema:
pretty sure I did yeah

PR #1742, opened 2022-10-11
Cura 7271 slow lines individually minimum layer time

rijkvanmanen:
This definitely needs some code quality checks, but the expected behaviour is realized without generating errors.


src/LayerPlan.cpp
```diff
@@ -1466,46 +1454,69 @@ void LayerPlan::spiralizeWallSlice(const GCodePathConfig& config, ConstPolygonRe
...
+
+    double slowestPathSpeed = DBL_MAX;
+    for (GCodePath& path : paths)
+    {
+        total_extrude_time_at_minimum_speed += path.estimates.extrude_time_at_minimum_speed;
+        total_extrude_time_at_slowest_speed += path.estimates.extrude_time_at_slowest_speed;
+        slowestPathSpeed = std::min(path.config->getSpeed().value * path.speed_factor, slowestPathSpeed);
+    }
+
+    if (totalTime < minTime - 0.01 && extrudeTime > 0.0)
```
rburema, line 1468:
```suggestion
    constexpr double epsilon = 0.01;
    if (totalTime < minTime - epsilon && extrudeTime > 0.0)
```

src/LayerPlan.cpp
```diff
@@ -1466,46 +1454,69 @@ void LayerPlan::spiralizeWallSlice(const GCodePathConfig& config, ConstPolygonRe
...
-            if (path.isTravelPath())
-                continue;
-            double speed = path.config->getSpeed() * path.speed_factor * factor;
-            if (speed < minimalSpeed)
-                factor = minimalSpeed / (path.config->getSpeed() * path.speed_factor);
+            // Even at cool min speed extrusion is not taken enough time. So speed is set to cool min speed.
+            for (GCodePath& path : paths)
+            {
+                if (path.isTravelPath())
+                    continue;
```
rburema, line 1477:
```suggestion
                if (path.isTravelPath())
                {
                    continue;
                }
```

src/LayerPlan.cpp
```diff
@@ -1466,46 +1454,69 @@ void LayerPlan::spiralizeWallSlice(const GCodePathConfig& config, ConstPolygonRe
...
+            if (minTime - total_extrude_time_at_minimum_speed - travelTime > 0.1)
+            {
+                extraTime = minTime - total_extrude_time_at_minimum_speed - travelTime;
+            }
         }
-
-        // Only slow down for the minimal time if that will be slower.
-        assert(getExtrudeSpeedFactor() == 1.0); // The extrude speed factor is assumed not to be changed yet
-        if (factor < 1.0)
+        else if (minExtrudeTime >= total_extrude_time_at_slowest_speed && abs(total_extrude_time_at_minimum_speed - total_extrude_time_at_slowest_speed) >= 0.01)
```
rburema, line 1488:
```suggestion
        else if (minExtrudeTime >= total_extrude_time_at_slowest_speed && std::abs(total_extrude_time_at_minimum_speed - total_extrude_time_at_slowest_speed) >= epsilon)
```

src/LayerPlan.cpp
```diff
@@ -1466,46 +1454,69 @@ void LayerPlan::spiralizeWallSlice(const GCodePathConfig& config, ConstPolygonRe
...
         {
-            setExtrudeSpeedFactor(factor);
+            // Slowing down to the slowest path speed is not sufficient, need to slow down further to the minimum speed.
+            // Linear interpolate between total_extrude_time_at_slowest_speed and total_extrude_time_at_minimum_speed
+            double factor = (total_extrude_time_at_minimum_speed - minExtrudeTime) / (total_extrude_time_at_minimum_speed - total_extrude_time_at_slowest_speed);
+            double target_speed = minimalSpeed * (1-factor) + slowestPathSpeed * factor;
```
rburema, line 1493:
```suggestion
            double target_speed = minimalSpeed * (1.0 - factor) + slowestPathSpeed * factor;
```

src/LayerPlan.cpp
```diff
@@ -1466,46 +1454,69 @@ void LayerPlan::spiralizeWallSlice(const GCodePathConfig& config, ConstPolygonRe
...
+            for (GCodePath& path : paths)
+            {
+                if (path.isTravelPath())
+                    continue;
```
rburema, line 1497:
```suggestion
                if (path.isTravelPath())
                {
                    continue;
                }
```

src/LayerPlan.cpp
```diff
@@ -1466,46 +1454,69 @@ void LayerPlan::spiralizeWallSlice(const GCodePathConfig& config, ConstPolygonRe
...
-        {
-            path.estimates.extrude_time *= inv_factor;
-        }
-
-        if (minTime - (extrudeTime * inv_factor) - travelTime > 0.1)
-        {
-            extraTime = minTime - (extrudeTime * inv_factor) - travelTime;
+            // Slowing down to the slowest_speed is sufficient to respect the minimum layer time.
+            // Linear interpolate between extrudeTime and total_extrude_time_at_slowest_speed
+            double factor = (total_extrude_time_at_slowest_speed - minExtrudeTime) / (total_extrude_time_at_slowest_speed - extrudeTime);
```
rburema, line 1508:
```suggestion
            const double factor = (total_extrude_time_at_slowest_speed - minExtrudeTime) / (total_extrude_time_at_slowest_speed - extrudeTime);
```

src/LayerPlan.cpp
```diff
@@ -1466,46 +1454,69 @@ void LayerPlan::spiralizeWallSlice(const GCodePathConfig& config, ConstPolygonRe
...
+            for (GCodePath& path : paths)
+            {
+                if (path.isTravelPath())
+                    continue;
```
rburema, line 1512:
```suggestion
                if (path.isTravelPath())
                {
                    continue;
                }
```

src/LayerPlan.cpp
```diff
@@ -1514,12 +1525,28 @@ TimeMaterialEstimates ExtruderPlan::computeNaiveTimeEstimates(Point starting_pos
     TimeMaterialEstimates ret;
     Point p0 = starting_position;
 
+    double min_path_speed = fan_speed_layer_time_settings.cool_min_speed;
+
+    //TODO: This slowest_path_speed is causing warnings for empty layers and looks a bit uckly to me anyway.
+    //      Maybe it can also be returned, because it is now calculated again in the forceMinimalLayerTime function.
+    double slowest_path_speed = DBL_MAX;
```
rburema, line 1532:
```suggestion
    double slowest_path_speed = std::numeric_limits<double>::max();
```

rburema (changes requested):
Hi!

Didn't get to actually testing this yet, but I do have a ton of remarks/suggestions/observations about code-style.

(Though actually there are only 14 messages ... the rest is because you can't add a code suggestion if you've already made the comment -- so I'd have to cancel the comment to make the code suggestion -- if I could make one I just added it instead.)

Another thing, if you want to use the `numeric_limits` suggestions, you may have to import the `<limits>` header. (In general a good moment to emphasise that I didn't test my own suggestions yet either -- use with care.)

Lastly, the code suggestions are not exhaustive everywhere to prevent (excessive) spam -- like the magic values, 'const'ness, and floating point notation (`1` -> `1.0`).

cheers, Remco.

include/pathPlanning/TimeMaterialEstimates.h
```diff
@@ -17,6 +17,8 @@ class TimeMaterialEstimates
     friend class ExtruderPlan; // cause there the naive estimates are calculated
 private:
     double extrude_time; //!< Time in seconds occupied by extrusion
+    double extrude_time_at_slowest_speed; //!< Time in seconds occupied by extrusion assuming paths are printed at slowest path speed
+    double extrude_time_at_minimum_speed; //!< Time in seconds occupied by extrusion assuming paths are printed at minimum path speed
```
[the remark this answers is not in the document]
rijkvanmanen: Totally agree this is vague, I tried to improve but I am still not sure.

src/LayerPlan.cpp
```diff
@@ -1466,46 +1454,69 @@ void LayerPlan::spiralizeWallSlice(const GCodePathConfig& config, ConstPolygonRe
...
+            for (GCodePath& path : paths)
+            {
+                if (path.isTravelPath())
+                    continue;
+                path.speed_factor = minimalSpeed / (path.config->getSpeed() * path.speed_factor);
+                path.estimates.extrude_time /= path.speed_factor;
+            }
+            // Update stored naive time estimates
+            estimates.extrude_time = total_extrude_time_at_minimum_speed;
+            if (minTime - total_extrude_time_at_minimum_speed - travelTime > 0.1)
```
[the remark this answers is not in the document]
rijkvanmanen: Fixed

src/LayerPlan.cpp
```diff
@@ -1466,46 +1454,69 @@ void LayerPlan::spiralizeWallSlice(const GCodePathConfig& config, ConstPolygonRe
...
+            for (GCodePath& path : paths)
+            {
+                if (path.isTravelPath())
+                    continue;
+                double target_speed = slowestPathSpeed * (1-factor) + (path.config->getSpeed() * path.speed_factor) * factor;
+                path.speed_factor = target_speed / (path.config->getSpeed() * path.speed_factor);
+                path.estimates.extrude_time /= path.speed_factor;
+            }
+            // Update stored naive time estimates
+            estimates.extrude_time = minExtrudeTime;
```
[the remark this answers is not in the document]
rijkvanmanen: I agree this is quite some repeating code. But not sure how to write the proper function for this, can you suggest something?

src/LayerPlan.cpp
```diff
@@ -1514,12 +1525,28 @@ TimeMaterialEstimates ExtruderPlan::computeNaiveTimeEstimates(Point starting_pos
...
+
+    //TODO: This slowest_path_speed is causing warnings for empty layers and looks a bit uckly to me anyway.
+    //      Maybe it can also be returned, because it is now calculated again in the forceMinimalLayerTime function.
+    double slowest_path_speed = DBL_MAX;
+    for (GCodePath& path : paths)
+    {
+        slowest_path_speed = std::min(slowest_path_speed, path.config->getSpeed().value * path.speed_factor);
+    }
+
     bool was_retracted = false; // wrong assumption; won't matter that much. (TODO)
```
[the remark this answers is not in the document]
rijkvanmanen: I also tried to understand, but I decided to leave it

src/LayerPlan.cpp
```diff
@@ -1514,12 +1525,28 @@ TimeMaterialEstimates ExtruderPlan::computeNaiveTimeEstimates(Point starting_pos
...
     for (GCodePath& path : paths)
     {
         bool is_extrusion_path = false;
         double* path_time_estimate;
         double& material_estimate = path.estimates.material;
+
+        //TODO: I don't understand why it is needed to initialize these path estimates at zero. But if I don't do it,
```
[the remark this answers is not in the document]
rijkvanmanen: Thanks for the explanation!

src/LayerPlan.cpp
```diff
@@ -1514,12 +1532,28 @@ TimeMaterialEstimates ExtruderPlan::computeNaiveTimeEstimates(Point starting_pos
     TimeMaterialEstimates ret;
     Point p0 = starting_position;
 
+    double min_path_speed = fan_speed_layer_time_settings.cool_min_speed;
+
+    //TODO: This slowest_path_speed is causing warnings for empty layers and looks a bit uckly to me anyway.
```
rijkvanmanen, line 1537: Any suggestions for how to fix this todo?

if it was python i would have written: slowest_path_speed = min([path.config->getSpeed().value * path.speed_factor for path in paths if not path.isTravelPath()])

rburema: I think you forgot to check for `isTravelPath()` in the loop.

I've fixed that, as well as the recalculation. The value is now just a class variable.

I've also used the 'accumulate' function to be a bit closer to what python is doing. There is another way to get even closer (using 'ranges'), but I found that in this case it was more hassle than it's worth. It doesn't matter that much in a simple case like this anyway.

rburema (approved):

PR #1743, opened 2022-10-11
[CURA-7800] Properly detect and handle 'conflicting' TreeNodes.

rburema:
This was not been done properly since the move to pointer-based nodes (instead of value based). The comparison to the value in the 'find' would never compare to true, since while the value could be equal in the eyes of the overriden == operator in Node itself, it would instead compare pointers, which would _always_ be different, since all Nodes _are_ different instances here.

Since Nodes where considered to be unique w.r.t. their position, the non-resolution lead to hard-to diagnose problems.

Now that conflicting nodes _are_ resolved properly however, they should also 'play-nice' with the pruning mechanism that was introduced at the same time the conflict resolution became defunct by mistake. (W.r.t. 'parents' and 'merged neighbours'.

### DIFF
--- a/src/TreeSupport.cpp
+++ b/src/TreeSupport.cpp
@@ -334,7 +334,7 @@ void TreeSupport::dropNodes(std::vector<std::vector<Node*>>& contact_nodes)
 
                     const bool to_buildplate = ! volumes_.getAvoidance(branch_radius_node, layer_nr - 1).inside(next_position);
                     Node* next_node = new Node(next_position, new_distance_to_top, node.skin_direction, new_support_roof_layers_below, to_buildplate, p_node);
-                    insertDroppedNode(contact_nodes[layer_nr - 1], next_node); // Insert the node, resolving conflicts of the two colliding nodes.
+                    next_node = insertDroppedNode(contact_nodes[layer_nr - 1], next_node); // Insert the node, resolving conflicts of the two colliding nodes.
 
                     // Make sure the next pass doesn't drop down either of these (since that already happened).
                     node.merged_neighbours.push_front(neighbour);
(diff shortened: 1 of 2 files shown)


src/TreeSupport.cpp
```diff
@@ -596,18 +596,32 @@ void TreeSupport::generateContactPoints(const SliceMeshStorage& mesh, std::vecto
     }
 }
 
-void TreeSupport::insertDroppedNode(std::vector<Node*>& nodes_layer, Node* p_node)
+TreeSupport::Node* TreeSupport::insertDroppedNode(std::vector<Node*>& nodes_layer, Node* p_node)
 {
-    std::vector<Node*>::iterator conflicting_node_it = std::find(nodes_layer.begin(), nodes_layer.end(), p_node);
+    const std::function<bool(Node*)> node_ptr_eq{ [&p_node](Node* p_other) { return p_node->position == p_other->position; } };
+    std::vector<Node*>::iterator conflicting_node_it = std::find_if(nodes_layer.begin(), nodes_layer.end(), node_ptr_eq);
```
jellespijker, line 602:
```suggestion
    auto node_ptr_eq{ [&p_node](Node* p_other) { return p_node->position == p_other->position; } };
    auto conflicting_node_it = std::find_if(nodes_layer.begin(), nodes_layer.end(), node_ptr_eq);
```

jellespijker (dismissed):

src/TreeSupport.cpp
rburema: Yup. I started putting in the types when it kept failing, only to discover I didn't use `find_if` but had kept `find` instead. Then forgot to put back the `auto` when I fixed it.

rburema:
Quote jellespijker:

> Looks good, if you apply (or discard) the review suggestion feel free to merge. I need to get in an actual auto now

PR #1748, opened 2022-10-19
[CURA-9625] Top skin does not apply skin overlap.

Joeydelarago:
We were previously not applying skin\_overlap offsets to roofing areas. I assume this was because it was causing doubled overlap areas on adjacent roofing and skin.

This is fixed by applying offset areas to roofing, but only where it would not intersect with offset areas from skin.

### DIFF
--- a/src/skin.cpp
+++ b/src/skin.cpp
@@ -363,14 +363,9 @@ void SkinInfillAreaComputation::generateRoofing(SliceLayerPart& part)
         else if (!skin_part.skin_fill.empty() && !skin_part.roofing_fill.empty())
         {
             // Find the intersection of roofing and skin offset (skin_overlap_mm) areas.
-            Polygons skin_overlap_area = skin_part.skin_fill.offset(skin_overlap).difference(skin_part.skin_fill); // skin_overlap offset area around the skin
-            Polygons roofing_overlap_area = skin_part.roofing_fill.offset(skin_overlap).difference(skin_part.roofing_fill); // skin_overlap offset area around the roofing
-            Polygons skin_roofing_overlap_intersection = skin_overlap_area.intersection(roofing_overlap_area);
-
-            // If we offset to both the roofing_fill and skin_fill, when adjacent they would have a doubled offset area. Since they would both offset towards each other.
-            // We only apply the intersecting offset areas to skin_fill.
-            skin_part.roofing_fill = skin_part.roofing_fill.unionPolygons(roofing_overlap_area.difference(skin_roofing_overlap_intersection));
-            skin_part.skin_fill = skin_part.skin_fill.unionPolygons(skin_roofing_overlap_intersection);
+            skin_part.skin_fill = skin_part.skin_fill.offset(skin_overlap);
+            skin_part.roofing_fill =
+ skin_part.roofing_fill.offset(skin_overlap).difference(skin_part.skin_fill);
         }
     }
 }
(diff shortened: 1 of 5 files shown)


src/skin.cpp
```diff
@@ -350,6 +351,27 @@ void SkinInfillAreaComputation::generateRoofing(SliceLayerPart& part)
...
+        {
+            // Find the intersection of roofing and skin offset (skin_overlap_mm) areas.
+            Polygons skin_overlap_area = skin_part.skin_fill.offset(skin_overlap).difference(skin_part.skin_fill); // skin_overlap offset area around the skin
+            Polygons roofing_overlap_area = skin_part.roofing_fill.offset(skin_overlap).difference(skin_part.roofing_fill); // skin_overlap offset area around the roofing
+            Polygons skin_roofing_overlap_intersection = skin_overlap_area.intersection(roofing_overlap_area);
+
+            // If we offset to both the roofing_fill and skin_fill, when adjacent they would have a doubled offset area. Since they would both offset towards each other.
+            // We only apply the intersecting offset areas to skin_fill.
+            skin_part.roofing_fill = skin_part.roofing_fill.unionPolygons(roofing_overlap_area.difference(skin_roofing_overlap_intersection));
+            skin_part.skin_fill = skin_part.skin_fill.unionPolygons(skin_roofing_overlap_intersection);
```
casperlamboo, line 373: I think this whole part can be simplified to

```suggestion
            skin_part.skin_fill = skin_part.skin_fill.offset(skin_overlap);
            skin_part.roofing_fill =
 skin_part.roofing_fill.offset(skin_overlap).difference(skin_part.skin_fill);
```

Consider the following example
Untitled Diagram drawio

The blue and orange shapes are the roofing and skin areas respectively. Then the green and and red shapes are the offset shapes generated from the roofing and skin areas. 

Then here is where our approaches differ.
You subtract original shape from the offset shape to only get the "offset area". From both offset areas you take the intersection to get the purple shape. Then the final roofing and skin areas is created by doing the following
```
roofing = roofing + (roofing_offset - intersection_offsets)
skin = skin + skin_offset
```

This is a bit verbose and the same behavior can be expressed using less operations.

The difference operation takes in two shapes; A and B. In my head I consider this operation subtraction; A - B. Here it does not matter if B contains more space then A, only the area's where A and B intersect are removed from B. There is no need to limit B to the space of A before executing the difference operation.
```
skin = skin_offset
roofing = roofing_offset - skin_offset
```

Joeydelarago: These operations are much simpler. I will commit the suggestion and add some comments to explain it 👌

Joeydelarago: These operations are much simpler. I will commit the suggestion and add some comments to explain it 👌

casperlamboo (commented):
Should the same logic be applied here regarding the offsetting of the skin and roofing?

https://github.com/Ultimaker/CuraEngine/blob/3dd5e1795757212953d8455bd548310b4836b6f0/src/skin.cpp#L450-L451

casperlamboo:
> Should the same logic be applied here regarding the offsetting of the skin and roofing?
> 
> https://github.com/Ultimaker/CuraEngine/blob/3dd5e1795757212953d8455bd548310b4836b6f0/src/skin.cpp#L450-L451

https://github.com/Ultimaker/CuraEngine/pull/1748/commits/7963ec1fe993a0ac726694e1244378777d880713

PR #1757, opened 2022-11-02
Never increase speeds/accels/jerks with initial layer slowdown

richfelker:
Fixes https://github.com/Ultimaker/Cura/issues/13681


rburema:
Thanks for your contribution! It does indeed also affect speed.

PR #1759, opened 2022-11-02
Fix disconnected support due to support horizontal expansion

rijkvanmanen:
By checking of the model is splitting the support areas, this undesired behaviour can be easily identified. By expanding the model with the same amount as the horizontal expansion, it can be found which part(s) need to be removed.

CURA-9844

### DIFF
--- a/src/support.cpp
+++ b/src/support.cpp
@@ -1013,7 +1013,25 @@ void AreaSupport::generateSupportAreasForMesh(SliceDataStorage& storage,
                 layer_above = &empty;
                 layer_this = layer_this.unionPolygons(storage.support.supportLayers[layer_idx].support_mesh);
             }
-            layer_this = AreaSupport::join(storage, *layer_above, layer_this, smoothing_distance).difference(model_mesh_on_layer);
+            layer_this = AreaSupport::join(storage, *layer_above, layer_this, smoothing_distance);
+            const int layer_size = layer_this.size();
+
+            Polygons layer_this_without_expansion = layer_this.difference(model_mesh_on_layer.offset(extension_offset));
+            layer_this = layer_this.difference(model_mesh_on_layer);
+            if (layer_size < layer_this.size())
+            {
+                // the support areas in this layer are split into parts by the model itself
+                int i = 0;
+                for (PolygonsPart poly : layer_this.splitIntoParts())
+                {
+                    if (poly.intersection(layer_this_without_expansion).empty())
+                    {
+                        // this area is expanded to the other side of the model and not connected to the original area that needed support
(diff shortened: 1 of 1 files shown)


src/support.cpp
```diff
@@ -1013,7 +1013,25 @@ void AreaSupport::generateSupportAreasForMesh(SliceDataStorage& storage,
...
+                // the support areas in this layer are split into parts by the model itself
+                int i = 0;
+                for (PolygonsPart poly : layer_this.splitIntoParts())
+                {
+                    if (poly.intersection(layer_this_without_expansion).empty())
+                    {
+                        // this area is expanded to the other side of the model and not connected to the original area that needed support
+                        layer_this.remove(i);
+                    }
+                    i++;
```
smartavionics, line 1032: Do you really want to increment i when you have just removed the the i'th element ?

src/support.cpp
```diff
@@ -1013,7 +1013,25 @@ void AreaSupport::generateSupportAreasForMesh(SliceDataStorage& storage,
                 layer_above = &empty;
                 layer_this = layer_this.unionPolygons(storage.support.supportLayers[layer_idx].support_mesh);
             }
-            layer_this = AreaSupport::join(storage, *layer_above, layer_this, smoothing_distance).difference(model_mesh_on_layer);
+            layer_this = AreaSupport::join(storage, *layer_above, layer_this, smoothing_distance);
+            const int layer_size = layer_this.size();
```
Joeydelarago, line 1017: You should give this a more descriptive name, it isn't clear a few lines down what the difference is between layer\_size and layer\_this.size()

src/support.cpp
```diff
@@ -1013,7 +1013,25 @@ void AreaSupport::generateSupportAreasForMesh(SliceDataStorage& storage,
...
+
+            Polygons layer_this_without_expansion = layer_this.difference(model_mesh_on_layer.offset(extension_offset));
+            layer_this = layer_this.difference(model_mesh_on_layer);
+            if (layer_size < layer_this.size())
+            {
+                // the support areas in this layer are split into parts by the model itself
+                int i = 0;
+                for (PolygonsPart poly : layer_this.splitIntoParts())
```
Joeydelarago, line 1025: This could also do with a more descriptive name. Something like layer\_part or support\_part

Joeydelarago (changes requested):

rijkvanmanen:
A different solution for the same problem is proposed in CURA-9996. I will come back to this PR after we finished the work for CURA-9996.

rijkvanmanen:
A better solution is now implemented with CURA-10152. So this PR will be closed

PR #1761, opened 2022-11-03
CURA-9551 order of sequential walls

jellespijker:
Fixes Ultimaker/Cura#12977


jellespijker:
It seems that we need to level up the gcc compiler. Just look in the workflows (in cura repo, conan-package-create how it is installed for the modern Linux variant

jellespijker:
@casperlamboo I bumped up the GCC version to 12, This should also help with @Piezoid his PR for the usage of `ranges::span`

src/InsetOrderOptimizer.cpp
```diff
@@ -329,5 +263,55 @@ std::unordered_set<std::pair<const ExtrusionLine*, const ExtrusionLine*>> InsetO
...
+        {
+            view = paths | ranges::views::take_exactly(1);
+        }
+    }
+    return view | ranges::views::join | ranges::views::remove_if(ranges::empty) | ranges::to_vector;
+}
+
+void InsetOrderOptimizer::dfs(auto node, isGraph auto& dag, isSet auto& visited)
+{
+    if (ranges::contains(visited, node))
```
jellespijker, line 305: @casperlamboo if you decided on using a "shadow" `unordered_set` for the visited, you should also use the member function `find()` instead of the ranges. Which is much slower for unordered maps and sets. In order to keep this also working for Vectors, you would then also need to update the concept `isSet` and create use an `if constexpr(std::is_same<decltype(visited), ....>)` to create the different execution branches for the contains logic

jellespijker: same goes for the `push_back` I think.

include/utils/actions/dfs_sort.h
```diff
@@ -0,0 +1,37 @@
...
+        return;
+    }
+    visited.push_back(node);
+    const auto& [children_begin, children_end] = dag.equal_range(node);
+    auto children = ranges::make_subrange(children_begin, children_end);
+    for (const auto& [_, child] : children)
+    {
+        dfs_sort(child, dag, visited);
+    }
+}
```
casperlamboo, line 34: As we are working with a directed a cyclic graph we know that there cannot be any cycles. As such the visited check is redundant.

The order which we try to calculated is called the [topological sort](https://en.wikipedia.org/wiki/Topological_sorting#Depth-first_search) of the nodes

Also note that the algorithm below doesn't work on arbitrary DAG's as stated in the documentation of the code; only because w know that we are working with rooted trees rather then arbitrary DAG's and we start traversing each tree from the known root the algorithm works. (we are working with tree's which is a subset of a dag; $\text{tree} \subset \text{trees} \subset \text{dag} \subset \text{graph}$).
```suggestion
/* # topolocial_sort
 *
 * Walks through a directed graph using depth-first-search and adds the nodes in that order to the the order list argument
 * \param node the current visited node, should be convertable or have the same type as the nodes in dag and visited (currently not constrained by a `require`)
 * \param dag Directed Acyclic Graph as defined by \concept isGraph
 * \param order nodes as defined by concept \isList; _note: nodes are ordered according to the their topological sorting order`
 */
constexpr void topological_sort(const auto& node, const isGraph auto& dag, isList auto& order)
{
    order.push_back(node);
    const auto& [children_begin, children_end] = dag.equal_range(node);
    auto children = ranges::make_subrange(children_begin, children_end);
    for (const auto& [_, child] : children)
    {
        topological_sort(child, dag, order);
    }
}
```

casperlamboo: Please note that the above code suggestion is just to convey my intention; I haven't really tested it and it will ofcourse not work as the function calls calling `dfs_sort` should also change to `topological_sort`

casperlamboo (commented):
Posting my intermediate review. Not finished yet but now we can work somewhat in parallel

src/InsetOrderOptimizer.cpp
```diff
@@ -159,142 +145,111 @@ bool InsetOrderOptimizer::addToLayer()
...
-            }
-            else
-            {
-                assert((nearby->inset_idx > here->inset_idx) == outer_to_inner);
-                order_requirements.emplace(std::make_pair(here, nearby));
+                roots.erase(node);
             }
+            roots.emplace(loco);
         }
     }
```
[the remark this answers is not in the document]
jellespijker: GitHub says No. Apparently I can't apply the suggestion on deleted lines. feel free to commit it yourself locally

src/InsetOrderOptimizer.cpp
```diff
@@ -1,13 +1,36 @@
...
+#include <range/v3/view/concat.hpp>
+#include <range/v3/view/drop.hpp>
+#include <range/v3/view/drop_last.hpp>
+#include <range/v3/view/join.hpp>
+#include <range/v3/view/map.hpp>
+#include <range/v3/view/remove_if.hpp>
+#include <range/v3/view/reverse.hpp>
+#include <range/v3/view/take_exactly.hpp>
+#include <range/v3/view/transform.hpp>
+#include <range/v3/view/zip.hpp>
```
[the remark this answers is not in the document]
jellespijker: Reverted this comment since these are actually used

include/utils/polygon.h
```diff
@@ -299,6 +299,19 @@ class ConstPolygonRef
...
+    bool inside(const auto& polygon) const
+    {
+        for (const auto& point : *path)
+        {
+            const auto res = ClipperLib::PointInPolygon(point, *polygon.path);
+            if (res < 1)
+            {
+                return false;
+            }
+        }
```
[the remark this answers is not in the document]
jellespijker: Benchmark                                    |   Time   |  CPU    | Iterations |
:---|---:|---:|---:|
WallsComputationTest/centerpoint_boundingbox |   6563 ns|  6562 ns|      106735|
WallsComputationTest/all_inside              |   7622 ns|  7622 ns|       87355|
WallsComputationTest/intersection_area       | 158592 ns|158562 ns|        4500|


The quickest method is testing against the center point of the boundingbox

src/InsetOrderOptimizer.cpp
```diff
@@ -159,142 +145,111 @@ bool InsetOrderOptimizer::addToLayer()
...
+                      | ranges::views::transform(
+                            [](const auto& loco)
+                            {
+                                const auto poly = std::get<1>(loco);
+                                return Locator{
+                                    .line = std::get<0>(loco),
+                                    .poly = poly,
+                                    .area = poly.area(),
+                                };
+                            });
```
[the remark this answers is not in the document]
casperlamboo: Let me try this. If it compiles I'll commit it myself

include/utils/polygon.h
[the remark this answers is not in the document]
jellespijker: `all_inside` seems to behave the best and is quicker then intersection

src/InsetOrderOptimizer.cpp
```diff
@@ -159,142 +143,111 @@ bool InsetOrderOptimizer::addToLayer()
...
+    // Sort the extrusion lines from small to big
+    rg::sort( windings[0], [](const auto& lhs, const auto& rhs) { return std::abs(lhs) < std::abs(rhs); }, &Locator::area);
+    rg::sort( windings[1], [](const auto& lhs, const auto& rhs) { return std::abs(lhs) < std::abs(rhs); }, &Locator::area);
+
+    // Build the forest, depending on the winding
+    value_type order;
+    auto windings_view = rv::concat(windings[0], windings[1]); // Make sure we always have initial root even if one of the partitions resulted in an empty vector
+    std::unordered_set<Locator*> roots{ &rg::front(windings_view) };
+
+    for (const auto& locator : windings_view | rv::addressof)
```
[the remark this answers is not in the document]
jellespijker: Correct

src/InsetOrderOptimizer.cpp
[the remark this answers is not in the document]
casperlamboo: in that case let's leave it like this

casperlamboo (approved):

include/utils/views/dfs.h
```diff
@@ -0,0 +1,71 @@
...
+
+template <typename Node, typename State>
+constexpr void dfs(
+    const Node& current_node,
+    const isGraph auto& dag,
+    const State& state,
+    std::function<State(const Node, const State)> handle_node,
+    isSet auto& visited)
+{
+    if (ranges::contains(visited, current_node))
```
jellespijker, line 32: Please note that the `unordered_***.find()` is faster then the `ranges::contains()` maybe use an `if constexpr` to check against the actual type used and deviate if needed.

This function is also a good candidate for a benchmark

include/utils/views/dfs.h
```diff
@@ -0,0 +1,71 @@
...
+        return;
+    }
+    visited.emplace(current_node);
+
+    //    const auto handle_node_result = handle_node(current_node);
+    //    switch (handle_node_result)
+    //    {
+    //        case DFSAction::StopBranch: return DFSAction::ContinueSearch;
+    //        case DFSAction::StopSearch: return DFSAction::StopSearch;
+    //    }
```
jellespijker, line 44: Why is this commented code here?

include/utils/views/dfs.h
```diff
@@ -0,0 +1,71 @@
...
+    const isGraph auto& dag,
+    const State& state,
+    std::function<State(const Node, const State)> handle_node,
+    isSet auto& visited)
+{
+    if (ranges::contains(visited, current_node))
+    {
+        //        return DFSAction::ContinueSearch;
+        return;
+    }
```
jellespijker, line 36: Does it have loops? Is there a possibility that we can revisit a node? Do we need this early-out in the first place

src/InsetOrderOptimizer.cpp
```diff
@@ -159,142 +145,175 @@ bool InsetOrderOptimizer::addToLayer()
...
+                                return Locator{
+                                    .line = std::get<0>(locator),
+                                    .poly = poly,
+                                    .area = poly.area(),
+                                };
+                            })
+                      | rg::to_vector;
+
+    // sort polygons on increasing area
+    rg::sort( locator_view, [](const auto& lhs, const auto& rhs) { return AABB(lhs).area() < AABB(rhs).area(); }, &Locator::poly);
```
jellespijker, line 178: I think you get better performance if you cache the `AABB` in the `locater_view` instead of creating 2 new instances for each check

jellespijker (commented):
A couple of small remarks made adhoc

include/InsetOrderOptimizer.h
```diff
@@ -106,77 +100,42 @@ class InsetOrderOptimizer
...
+     * highest and iterate back.
+     * Otherwise, if the wall is partially printed with the current extruder we need to move forward for the outer wall extruder and iterate
+     * back for the inner wall extruder
+     *
+     * \param use_one_extruder boolean stating that we are using a single extruder.
+     * \param current_extruder_is_wall_x boolean stating if the current extruder is used for the inner walls.
+     * \param outer_to_inner boolean which should be set to true if we're printing from an outside to the inside
+     *
+     * \return a bool if the paths should be printed in reverse or not
+     */
```
Joeydelarago, line 119: The comment above is good but these params and return information should be at the function declaration. With a short comment.

include/InsetOrderOptimizer.h
```diff
@@ -106,77 +100,42 @@ class InsetOrderOptimizer
...
+    constexpr bool shouldReversePath(const bool use_one_extruder, const bool current_extruder_is_wall_x, const bool outer_to_inner);
+
+    /* Flattens the `paths` and sorts the walls that should be printed added depending on if it is a single outer wall or the inner wall(s),
+     * The order can be reversed if required.
+     *
+     * \param reverse boolean stating if the order of the wall should be revered or not
+     * \param use_one_extruder lean stating that we are using a single extruder.
+     *
+     * \return A vector of ExtrusionLines with walls that should be printed
+     */
```
Joeydelarago, line 129: Same as above here

include/utils/polygon.h
```diff
@@ -1,5 +1,5 @@
-//Copyright (c) 2022 Ultimaker B.V.
-//CuraEngine is released under the terms of the AGPLv3 or higher.
+// Copyright (c) 2022 Ultimaker B.V.
```
Joeydelarago, line 1:
```suggestion
// Copyright (c) 2022 UltiMaker B.V.
```

Not sure, but maybe we should start changing these?

include/utils/polygon.h
```diff
@@ -299,6 +299,19 @@ class ConstPolygonRef
         return res == 1;
     }
 
+    bool inside(const auto& polygon) const
+    {
+        for (const auto& point : *path)
+        {
+            const auto res = ClipperLib::PointInPolygon(point, *polygon.path);
+            if (res < 1)
```
Joeydelarago, line 307: I think either you should change "res" to a more descriptive name, or do this. in one line.

```suggestion
            if (!ClipperLib::PointInPolygon(point, *polygon.path))
```

tests/WallsComputationTest.cpp
```diff
@@ -149,12 +149,12 @@ TEST_F(WallsComputationTest, WallToolPathsGetWeakOrder)
     walls_computation.generateWalls(&layer);
 
     const bool outer_to_inner = false;
-    std::vector<const ExtrusionLine*> all_paths;
+    std::vector<ExtrusionLine> all_paths;
     for (auto& inset : part.wall_toolpaths)
     {
         for (auto& line : inset)
```
Joeydelarago, line 155: Untested of course, but I think we could do something like this here. Not a required change, I just think it is cool.

```suggestion
    for (auto& line : part.wall_toolpaths | views::join)
    {
```

Since this is a vector\<vector\<ExtrusionLine>>

Or even better

```suggestion
    std::vector<const ExtrusionLine*> all_paths = part.wall_toolpaths | action::join;
```

jellespijker: The return type of `actions::join` is a view not a `vector`. Use `auto` for complex types

include/utils/views/dfs.h
casperlamboo: Oops, this was some left over code for an idea of providing a possibility to break early in the dfs. Should remove this.

include/utils/views/dfs.h
casperlamboo: Yes, I made the edges bi-directional
https://github.com/Ultimaker/CuraEngine/blob/0e2ac875c67b5d6599a2a946002f3b113e212f18/src/InsetOrderOptimizer.cpp#L189-L190

CMakeLists.txt
```diff
@@ -139,7 +139,7 @@ set(engine_SRCS # Except main.cpp.
         src/utils/ToolpathVisualizer.cpp
         src/utils/VoronoiUtils.cpp
         src/utils/VoxelUtils.cpp
-        )
+        include/utils/concepts/graph.h include/utils/views/dfs.h)
```
jellespijker, line 142:
```suggestion
```

This is not needed if the include is src file

include/InsetOrderOptimizer.h
```diff
@@ -1,13 +1,13 @@
-//Copyright (c) 2021 Ultimaker B.V.
-//CuraEngine is released under the terms of the AGPLv3 or higher.
+// Copyright (c) 2022 Ultimaker B.V.
```
jellespijker, line 1:
```suggestion
// Copyright (c) 2022 UltiMaker B.V.
```

include/utils/views/dfs.h
```diff
@@ -0,0 +1,71 @@
...
+            current_state,
+            handle_node,
+            visited
+        );
+
+        //        const auto dfs_result = dfs(child_node, current_node, dag, visited);
+        //        switch (dfs_result)
+        //        {
+        //            case DFSAction::StopSearch: return DFSAction::StopSearch;
+        //        }
```
jellespijker, line 64:
```suggestion
```

src/InsetOrderOptimizer.cpp
```diff
@@ -159,142 +145,175 @@ bool InsetOrderOptimizer::addToLayer()
...
+    if (input.empty()) // Early out
+    {
+        return {};
     }
-    if (max_line_w == 0u)
-        return order_requirements;
 
-    struct LineLoc
+    // Cache the polygons and get the signed area of each extrusion line and store them mapped against the pointers for those lines
+    struct Locator
```
jellespijker, line 156: Rename struct to `LineLoc` to minimize the diff
```suggestion
    struct LineLoc
```

src/InsetOrderOptimizer.cpp
```diff
@@ -159,142 +145,175 @@ bool InsetOrderOptimizer::addToLayer()
...
     };
-    struct Locator
+    auto poly_views = input | views::convert<Polygon>(&ExtrusionLine::toPolygon);
+    auto pointer_view = input | rv::addressof;
+    auto locator_view = rv::zip(pointer_view, poly_views)
+                      | rv::transform(
+                            [](const auto& locator)
+                            {
+                                const auto poly = std::get<1>(locator);
+                                return Locator{
```
jellespijker, line 169: See comment above
```suggestion
                                return LineLoc {
```

src/InsetOrderOptimizer.cpp
```diff
@@ -159,142 +145,175 @@ bool InsetOrderOptimizer::addToLayer()
...
+
+    // sort polygons on increasing area
+    rg::sort( locator_view, [](const auto& lhs, const auto& rhs) { return AABB(lhs).area() < AABB(rhs).area(); }, &Locator::poly);
+
+    std::unordered_multimap<const Locator*, const Locator*> graph;
+    std::unordered_set<Locator*> roots{ &rg::front(locator_view) };
+    for (const auto& locator : locator_view | rv::addressof | rv::drop(1))
     {
-        Point operator()(const LineLoc& elem)
+        std::vector<Locator*> erase;
```
jellespijker, line 184:
```suggestion
        std::vector<LineLoc*> erase;
```

src/InsetOrderOptimizer.cpp
```diff
@@ -159,142 +145,175 @@ bool InsetOrderOptimizer::addToLayer()
...
+                                    .area = poly.area(),
+                                };
+                            })
+                      | rg::to_vector;
+
+    // sort polygons on increasing area
+    rg::sort( locator_view, [](const auto& lhs, const auto& rhs) { return AABB(lhs).area() < AABB(rhs).area(); }, &Locator::poly);
+
+    std::unordered_multimap<const Locator*, const Locator*> graph;
+    std::unordered_set<Locator*> roots{ &rg::front(locator_view) };
```
jellespijker, line 181: see above
```suggestion
    std::unordered_multimap<const LineLoc*, const LineLoc*> graph;
    std::unordered_set<LineLoc*> roots{ &rg::front(locator_view) };
```

src/InsetOrderOptimizer.cpp
```diff
@@ -159,142 +145,175 @@ bool InsetOrderOptimizer::addToLayer()
...
     {
-        Point operator()(const LineLoc& elem)
+        std::vector<Locator*> erase;
+        for (const auto& root : roots)
         {
-            return elem.j.p;
+            if (root->poly.inside(locator->poly))
+            {
+                graph.emplace(locator, root);
+                graph.emplace(root, locator);
```
jellespijker, line 190: Some documentation might help, why you add the key, value and add them both to the unordered multimap. I assume it is because you need bi-directional

include/utils/views/dfs.h
casperlamboo: https://github.com/Ultimaker/CuraEngine/pull/1761/commits/b5b05345087442cd05261c8af2e6f513f30bea97

src/InsetOrderOptimizer.cpp
casperlamboo: https://github.com/Ultimaker/CuraEngine/pull/1761/commits

include/utils/views/dfs.h
casperlamboo: https://github.com/Ultimaker/CuraEngine/pull/1761/commits/b5b05345087442cd05261c8af2e6f513f30bea97

include/utils/views/dfs.h
casperlamboo: https://github.com/Ultimaker/CuraEngine/pull/1761/commits/cb267ffbba9d5f073255989933d7e599b00a6426

tests/WallsComputationTest.cpp
casperlamboo: https://github.com/Ultimaker/CuraEngine/pull/1761/commits/2944616106d8e100504d35b3a6c18a843f18100e

src/InsetOrderOptimizer.cpp
casperlamboo: https://github.com/Ultimaker/CuraEngine/pull/1761/commits/0401b7d63190242340fdd384d06781804bea2971

include/InsetOrderOptimizer.h
casperlamboo: https://github.com/Ultimaker/CuraEngine/pull/1761/commits/8acbcd26935d28b82aa36a8484d6a28c3a9eef53

include/InsetOrderOptimizer.h
casperlamboo: https://github.com/Ultimaker/CuraEngine/pull/1761/commits/8acbcd26935d28b82aa36a8484d6a28c3a9eef53

src/InsetOrderOptimizer.cpp
```diff
@@ -159,142 +142,199 @@ bool InsetOrderOptimizer::addToLayer()
...
+    auto locator_view = rv::zip(pointer_view, poly_views)
+                      | rv::transform(
+                            [](const auto& locator)
+                            {
+                                const auto poly = std::get<1>(locator);
+                                const auto line = std::get<0>(locator);
+                                return LineLoc {
+                                    .line = line,
+                                    .poly = poly,
+                                    .area = line->is_closed ? poly.area() : 0,
```
jellespijker, line 170:
```suggestion
                                    .area = line->is_closed ? poly.area() : 0.0,
```

src/InsetOrderOptimizer.cpp
```diff
@@ -159,142 +142,199 @@ bool InsetOrderOptimizer::addToLayer()
...
                 {
-                    order_requirements.emplace(std::make_pair(here, nearby));
-                }
-            }
-            else if ((nearby->inset_idx < here->inset_idx) == outer_to_inner)
+                    min_node[current_node] = root;
+                    min_dist[current_node] = dist;
+
+                    // find hole roots, these are leaves of the tree structure
+                    // as odd walls are also lease we filter them out by adding a non-zero area check
```
jellespijker, line 222:
```suggestion
                    // find hole roots (defined by a possitive area in clipper1), these are leaves of the tree structure
                    // as odd walls are also leaves we filter them out by adding a non-zero area check
```

jellespijker (commented):
Looks good!

As discussed I, will implement the suggested changes myself, mostly code-style and handling of types.

jellespijker:
LGTM @casperlamboo 
Proud of you.

Now we can also measure the impact on the performance using the benchmark dashboard https://ultimaker.github.io/CuraEngine/dev/bench/

PR #1768, opened 2022-11-10
CURA-8890-hole-horizontal-expansion-max-diameter

rijkvanmanen:
The hole horizontal expansion is only applied to holes which are smaller then the hole_xy_offset_max_diameter. For these small holes the expansion is proportional to their size to ensure smooth transition between small and large holes.

See also front-end PR: https://github.com/Ultimaker/Cura/pull/13785


jellespijker (changes requested):
Looks good but I think a couple of changes might be needed. In my opinion it makes more sense to check against area instead of circumference. This could either be done in the front-end, by changing the setting to maximum hole area (mm^2); Although from a user perspective you probably want to keep this as `approximated diameter` since this feature is most likely used to expand circular/screw holes.

Also the whole `xy_offset_hole` block can be simplified, see code below, we only want to apply the offset on holes smaller then `max_hole_area` Since `orientation` calculates the area to determine the winding (pos = hole, neg = outline) In the proposition below the area is now only calculated once.

don't forget to `#include <numbers>` at the top


```c++
if (xy_offset_hole != 0)
{
    Polygons small_holes;
    Polygons outline;

    const auto max_hole_area = std::numbers::pi / 4 * static_cast<double>(hole_offset_max_diameter * hole_offset_max_diameter);

    for (const auto& poly : layers[layer_nr].polygons)
    {
        auto area = ClipperLib::Area(poly);
        if (area < 0 && std::abs(area) < max_hole_area)
        {
            small_holes.add(Polygon{ poly }.offset(static_cast<int>(xy_offset_hole * (1. - area / max_hole_area)), ClipperLib::JoinType::jtRound));
        }
        else
        {
            outline.add(poly);
        }
    }
    layers[layer_nr].polygons = outline.difference(small_holes.unionPolygons());
}
```

casperlamboo:
Unfortunately the code posted by @jellespijker does not produce the desired result; inner polygons of nested polygons will incorrectly be removed. See benchy example below.
Screenshot 2023-03-14 at 19 48 16
I see two solutions for this problem;
- sort polygons on value `abs(area)` from largest to smallest and maintain a polygon `ret` for the result. When a clockwise polygon is encountered add said polygon the `ret`, when a hole is encountered apply the required offsets and `difference` the hole from `ret`
- reintroduce the polygon parts, these will produce the individual polygons. When removing holes from just the polygon part the issue of removing nested polygongs is circumvented .

jellespijker:
best to reintroduce the parts then

PR #1774, opened 2022-11-16
Benchmarking

jellespijker:
See the result on the [CuraEngine development portal](https://ultimaker.github.io/CuraEngine/), check the Benchmark link in the top right of the home page, which links to [Benchmark results](https://ultimaker.github.io/CuraEngine/dev/bench/)

The action run the Google Benchmark test-suite and publish the result on the https://github.com/Ultimaker/CuraEngine/tree/gh-pages branch. Which should update accordingly. If the results need to be cleared or changed, It can be done here https://github.com/Ultimaker/CuraEngine/blob/gh-pages/dev/bench/data.js

It will be a cron job which runs 3 times per working day. That frequencies should give a more honest view, filtering out variations due to the GitHub runner virtualization.

Needs to be merged after https://github.com/Ultimaker/CuraEngine/pull/1761

image


jellespijker:
Closing this PR in favor of https://github.com/Ultimaker/CuraEngine/pull/1775 which is based on main and can be merged without a dependency.

PR #1775, opened 2022-11-16
CURA-9906 benchmark CuraEngine

jellespijker:
See the result on the [CuraEngine development portal](https://ultimaker.github.io/CuraEngine/), check the Benchmark link in the top right of the home page, which links to [Benchmark results](https://ultimaker.github.io/CuraEngine/dev/bench/)

The action run the Google Benchmark test-suite and publish the result on the https://github.com/Ultimaker/CuraEngine/tree/gh-pages branch. Which should update accordingly. If the results need to be cleared or changed, It can be done here https://github.com/Ultimaker/CuraEngine/blob/gh-pages/dev/bench/data.js

It will be a cron job which runs 3 times per working day. That frequencies should give a more honest view, filtering out variations due to the GitHub runner virtualization.

Needs to be merged after https://github.com/Ultimaker/CuraEngine/pull/1761

image


benchmark/wall_benchmark.h
```diff
@@ -0,0 +1,137 @@
...
+    void TearDown(const ::benchmark::State& state)
+    {
+    }
+};
+
+BENCHMARK_DEFINE_F(WallTestFixture, generateWalls)(benchmark::State& st)
+{
+    for (auto _ : st)
+    {
+        walls_computation.generateWalls(&layer);
```
Joeydelarago, line 92: Again, I think this is being run multiple times for each state parameter but is running the same computation.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,130 @@
...
+            -   name: Store benchmark result
+                uses: benchmark-action/github-action-benchmark@v1
+                with:
+                    name: C++ Benchmark
+                    tool: 'googlecpp'
+                    output-file-path: build/Release/benchmark/benchmark_result.json
+                    github-token: ${{ secrets.GITHUB_TOKEN }}
+                    auto-push: true
+                    # Show alert with commit comment on detecting possible performance regression
+                    alert-threshold: '200%'
```
Joeydelarago, line 128: I'm not sure how much the runners performance affects this, but I feel like going for a slightly lower threshold would be good and we can up it if we have false positives. 

```suggestion
                    alert-threshold: '150%'
```

jellespijker: According to the doc the runners can vary somewhere around the 20%. That is why I would first get it to run on each push commit to main, to see the actual variance.

I'm fine with a lower value, btw that would allow us to test that  functionality as well

PR #1776, opened 2022-11-18
[CURA-9921] Correct bottom-layer(s) behaviour for exclusive tolerance.

### LINKED ISSUE #1606: Exclusive slicing_tolerance produces empty first layer
Setting `slicing_tolerance` to `"exclusive"` usually (always? unless `mesh_position_z` < 0?) produces an empty first layer which gets removed. This both prevents `initial_bottom_layers` from getting used correctly, and produces a print that's one layer too short in the Z direction. The culprit seems to be `slicer.cpp` lines 857-863:

https://github.com/Ultimaker/CuraEngine/blob/b5c84d0d368df7d4a96387d5aa37ca9fb3ae8a23/src/slicer.cpp#L857-L863

whereby points exactly on the slice boundary get moved one unit above it. I suspect this should be changed to move them one unit below it, at least on the first layer. Or maybe it's wrong to be doing this at all; the motivation is not clear to me.

Lowering the model by 0.01 or so works around the issue, but ends up making the print one layer short at the top (assuming the entire model height was an integral number of layers to begin with).

rburema:
Since all points where increased by one, this lead the exclusive tolerance, in cases where the intial layer height is smaller or equal to the 'normal' layer height, to be missing a layer, since the model would be on z=0. Since the initial layer is not on the build-plate, this causes the number of bottom layers to not apply anymore.

should fix #1606

### DIFF
--- a/src/slicer.cpp
+++ b/src/slicer.cpp
@@ -833,9 +833,9 @@ void Slicer::buildSegments(const Mesh& mesh, const std::vector<std::pair<int32_t
                                // Compensate for points exactly on the slice-boundary, except for 'inclusive', which already handles this correctly.
                                if (slicing_tolerance != SlicingTolerance::INCLUSIVE)
                                {
-                                   p0.z += static_cast<int>(p0.z == z);
-                                   p1.z += static_cast<int>(p1.z == z);
-                                   p2.z += static_cast<int>(p2.z == z);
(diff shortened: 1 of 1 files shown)


jellespijker (approved):
LGTM; even though it is indeed a bit hacky

PR #1777, opened 2022-11-18
[CURA-9920] Small feature settings should only affect outer wall.

### LINKED ISSUE #1631: Small feature size speed is applied to Arachne inner-wall loops
In narrow regions where width varies such that short loops of inner-wall get produced, the small feature size speed limits get applied to these loops. This makes the print extremely slow if there are lots of them, and it does not serve any useful purpose since these are not "features" but simply fill between the outer walls.

The logic for slowing down "small features" should only apply to types of extrusion paths for which it makes sense: outer walls, support outer walls, and maybe a few others. Not for extrusion paths that are completely enclosed within and constrained by another outer structure where printing them slowly will not contribute to dimensional accuracy and surface quality of the part.

rburema:
Since the stated purpose of this feture is dimensional accuracy, infill and inner walls matter a lot less. With the introduction of Arachne, small loops not in the outer walls had become more common.

should fix #1631

### DIFF
--- a/src/LayerPlan.cpp
+++ b/src/LayerPlan.cpp
@@ -998,7 +998,7 @@ void LayerPlan::addWall(const ExtrusionLine& wall,
 
     bool first_line = true;
     const coord_t small_feature_max_length = settings.get<coord_t>("small_feature_max_length");
-    const bool is_small_feature = (small_feature_max_length > 0) && cura::shorterThan(wall, small_feature_max_length);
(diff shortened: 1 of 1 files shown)


smartavionics:
> Since the stated purpose of this feture is dimensional accuracy

If my memory serves me right, the primary reason for creating this feature was to help adhesion of small diameter features on the bottom layer so only applying it to outer walls may possibly cause prints to fail if people are relying on a slowdown to be applied to all walls and not only the outer wall. Just saying.

jellespijker (approved):
LGTM

rburema:
> LGTM

TIL: LGTM

PR #1778, opened 2022-11-21
[CURA-9564] Remove support on top of interface layers

Joeydelarago:
Remove support sitting on top of interface by subtracting the interface polygon from the above layers.

CURA-9564


casperlamboo (commented):
had two very minor remarks, implementers choice if it's better. I would like a comment though explaining why this is needed

casperlamboo (approved):
Looking good!

PR #1794, opened 2023-01-03
[CURA-9957] Change Ordering Constraints

Joeydelarago:
Added parent to dfs handle\_node function. I chose this over passing parent as the state variable because it makes it clearer what we are doing. Also, I think this can be useful in the future when using dfs.

Added comments in InsetOrderOptimizer.cpp

Removed sibling constraint in ordering and added parent->child constraint instead.

CURA-9957

### DIFF
--- a/include/InsetOrderOptimizer.h
+++ b/include/InsetOrderOptimizer.h
@@ -18,7 +18,7 @@ class LayerPlan;
 class InsetOrderOptimizer
 {
 public:
-    using value_type = std::unordered_set<std::pair<const ExtrusionLine*, const ExtrusionLine*>>;
+    using value_type = std::unordered_multimap<const ExtrusionLine*, const ExtrusionLine*>;
 
     /*!
      * Constructor for inset ordering optimizer.
--- a/include/PathOrderOptimizer.h
+++ b/include/PathOrderOptimizer.h
@@ -92,12 +92,12 @@ public:
      * \param combing_boundary Boundary to avoid when making travel moves.
      */
     PathOrderOptimizer(const Point start_point, const ZSeamConfig seam_config = ZSeamConfig(), const bool detect_loops = false, const Polygons* combing_boundary = nullptr, const bool reverse_direction = false, const std::unordered_set<std::pair<PathType, PathType>>& order_requirements = no_order_requirements)
-    : start_point(start_point)
-    , seam_config(seam_config)
-    , combing_boundary((combing_boundary != nullptr && !combing_boundary->empty()) ? combing_boundary : nullptr)
-    , detect_loops(detect_loops)
-    , reverse_direction(reverse_direction)
-    , order_requirements(&order_requirements)
+        : start_point(start_point)
+        , seam_config(seam_config)
+        , combing_boundary((combing_boundary != nullptr && !combing_boundary->empty()) ? combing_boundary : nullptr)
+        , detect_loops(detect_loops)
+        , reverse_direction(reverse_direction)
+        , order_requirements(&order_requirements)
     {
     }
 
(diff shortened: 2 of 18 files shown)


include/PathOrderOptimizer.h
```diff
@@ -291,30 +294,23 @@ class PathOrderOptimizer
 
             std::vector<OrderablePath> available_candidates;
             available_candidates.reserve(nearby_candidates.size());
-            for(OrderablePath candidate : nearby_candidates)
+            for(OrderablePath candidate : nearby_candidates | rv::filter(isPicked))
```
casperlamboo: nice! love the range operation here

include/utils/concepts/generic.h
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) 2023 Ultimaker B.V.
```
casperlamboo, line 1:
```suggestion
// Copyright (c) 2023 UltiMaker
```

include/utils/concepts/graph.h
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2022 Ultimaker B.V.
+// Copyright (c) 2023 Ultimaker B.V.
```
casperlamboo, line 1:
```suggestion
// Copyright (c) 2023 UltiMaker
```

include/utils/views/dfs.h
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2022 Ultimaker B.V.
+// Copyright (c) 2023 Ultimaker B.V.
```
casperlamboo, line 1:
```suggestion
// Copyright (c) 2023 UltiMaker
```

jellespijker (changes requested):

include/utils/views/dfs.h
```diff
@@ -14,20 +14,32 @@ namespace cura::actions
...
 constexpr void dfs(
     const Node& current_node,
-    const isGraph auto& dag,
-    const State& state,
+    const Graph& graph,
     std::function<State(const Node, const State)> handle_node,
-    isSet auto& visited)
+    const State& state = nullptr,
+    std::unordered_set<Node> visited = std::unordered_set<Node>(),
+    std::function<std::vector<Node>(const Node, const Graph&)> get_neighbours = [](const Node current_node, const Graph& graph){
```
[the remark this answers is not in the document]
jellespijker: See the partial patch in the JIRA ticket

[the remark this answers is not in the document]
jellespijker: See the partial patch in the JIRA ticket

include/PathOrderOptimizer.h
```diff
@@ -1,5 +1,5 @@
-//Copyright (c) 2022 Ultimaker B.V.
-//CuraEngine is released under the terms of the AGPLv3 or higher.
+// Copyright (c) 2023 Ultimaker B.V.
```
casperlamboo, line 1:
```suggestion
// Copyright (c) 2023 UltiMaker
```

src/PathOrderPath.cpp
```diff
@@ -1,44 +1,44 @@
 //Copyright (c) 2022 Ultimaker B.V.
```
casperlamboo, line 1:
```suggestion
//Copyright (c) 2023 UltiMaker
```

include/InsetOrderOptimizer.h
```diff
@@ -18,7 +18,7 @@ class LayerPlan;
 class InsetOrderOptimizer
 {
 public:
-    using value_type = std::unordered_set<std::pair<const ExtrusionLine*, const ExtrusionLine*>>;
+    using value_type = std::unordered_multimap<const ExtrusionLine*, const ExtrusionLine*>;
```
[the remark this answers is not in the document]
casperlamboo: somehow I cannot remove the const from the key

include/PathOrderOptimizer.h
```diff
@@ -265,6 +266,14 @@ class PathOrderOptimizer
      */
     bool reverse_direction;
 
+    /*
+     * Whether to print all outer walls in a group, one after another.
+     *
+     * If this is enabled outer walls will be printer first and then all other
+     * walls will be printer. If reversed they will be printer last.
```
casperlamboo, line 273:
```suggestion
     * If this is enabled outer walls will be printed first and then all other
     * walls will be printed. If reversed they will be printed last.
```

include/PathOrderOptimizer.h
```diff
@@ -421,53 +430,65 @@ class PathOrderOptimizer
...
-
-        // We want all outer walls to be printed first (For outside in ordering) or last (for inside out ordering)
-        std::unordered_set<Path> outer_walls = reverse_direction ? leaves : roots;
-
-        if (reverse_direction)
+        if (group_outer_walls)
         {
-            visited.insert(outer_walls.begin(), outer_walls.end());
+            // We want all outer walls to be printed first (For outside in ordering) or last (for inside out ordering)
+            std::unordered_set<Path> outer_walls = reverse_direction ? leaves : roots;
```
casperlamboo, line 436: Minor remark but I would place this in the individual branches of the next if-statement. I think it would improve readability of the code as you can write
```cpp
if (reverse_direction)
{
   // as the order is reverse the leaves of the order requirement are the outer walls
   std::unordered_set<Path> outer_walls = leaves;
   .....
}
else
{
   // as the order is not reverse the roots of the order requirement are the outer walls
   std::unordered_set<Path> outer_walls = roots;
   .....
}
```

include/PathOrderOptimizer.h
```diff
@@ -365,15 +266,316 @@ class PathOrderOptimizer
...
+                Path best_candidate = findClosestPathVertices(local_current_position, candidates);
+
+                candidates.erase(best_candidate);
+                order.push_back(best_candidate);
+
+                // update local_current_position
+                for (auto& path : paths)
+                {
+                    if (path.vertices == best_candidate)
+                    {
```
casperlamboo, line 386: Didn't we have a map from one type of path ordering to another type of path ordering so that we don't have a `O(n)` slow search to find the correct corresponding path type?

include/PathOrderOptimizer.h
```diff
@@ -421,53 +430,65 @@ class PathOrderOptimizer
...
+                {
+                    Path root = findClosestPathVertices(current_position, roots);
+                    roots.erase(root);
+                    actions::dfs(root, order_requirements, handle_node, visited, nullptr, get_neighbours);
+                }
+
+                // Add all outer walls
+                while (! outer_walls.empty())
+                {
+                    Path wall = findClosestPathVertices(current_position, outer_walls);
```
casperlamboo, line 453: I'm seeing this a couple of times in the newly added code and I have some concerns about this. I think we should update the `current_position` once we picked a path. Don't think that is happening right now.

casperlamboo: Oh no nevermind I see it is updated in the `handle_node` function

casperlamboo (approved):

PR #1795, opened 2023-01-05
CURA-10127-bug-fix-minimum-layer-time-speed-factor

rijkvanmanen:
The required speed is inversely proportional to the time. CURA-10127
frontend PR: https://github.com/Ultimaker/Cura/pull/14213


jellespijker (approved):
LGTM

PR #1796, opened 2023-01-05
CURA-10129-Improve-minimum-layer-time-for-dual-extrusion

rijkvanmanen:
Now the minimum layer time slow down is applied to the last extruder instead of to the last extruder plan. And the time spend on other extruder plans also taken into account.


src/LayerPlan.cpp
```diff
@@ -1703,11 +1703,27 @@ void ExtruderPlan::processFanSpeedAndMinimalLayerTime(bool force_minimal_layer_t
 
 void LayerPlan::processFanSpeedAndMinimalLayerTime(Point starting_position)
 {
+    // the minimum layer time behaviour is only applied to the last extruder.
+    size_t max_extruder_nr = 0;
+    for (unsigned int extr_plan_idx = 0; extr_plan_idx < extruder_plans.size(); extr_plan_idx++)
+    {
+        max_extruder_nr = std::max(extruder_plans[extr_plan_idx].extruder_nr, max_extruder_nr);
+    }
```
jellespijker, line 1711:
```suggestion
    const size_t max_extruder_nr = ranges::max_element(extruder_plans, [](const ExtruderPlan& a, const ExtruderPlan& b) { return a.extruder_nr < b.extruder_nr; })->extruder_nr;
```

Don't forget to include max_element.hpp at the top
```
#include <range/v3/algorithm/max_element.hpp>
```

src/LayerPlan.cpp
```diff
@@ -1703,11 +1703,27 @@ void ExtruderPlan::processFanSpeedAndMinimalLayerTime(bool force_minimal_layer_t
...
+    // determine the time spend on the other extruder plans in this layer.
+    double time_other_extr_plans = 0;
+    for (unsigned int extr_plan_idx = 0; extr_plan_idx < extruder_plans.size(); extr_plan_idx++)
+    {
+        if (extr_plan_idx != max_extruder_nr)
+        {
+            time_other_extr_plans += extruder_plans[extr_plan_idx].computeNaiveTimeEstimates(starting_position).getTotalTime();
+        }
+    }
```
jellespijker, line 1720:
```suggestion
    auto other_extr_plans = extruder_plans 
                          | ranges::views::filter([max_extruder_nr](const ExtruderPlan& extruder_plan) { return extruder_plan.extruder_nr != max_extruder_nr; })
                          | ranges::views::transform([starting_position](ExtruderPlan& extruder_plan) { return extruder_plan.computeNaiveTimeEstimates(starting_position).getTotalTime(); });
    const double time_other_extr_plans = ranges::accumulate(other_extr_plans, 0.0);
```

Don't forget to include at the top:
```
#include <range/v3/numeric/accumulate.hpp>
#include <range/v3/view/transform.hpp>
#include <range/v3/view/filter.hpp>
```

jellespijker (changes requested):
Made some suggestions how you can modernize the `for` loops that you use to determine the maximum extruder and calculate the time other extruder plans.

src/LayerPlan.cpp
rijkvanmanen: outdated

jellespijker (approved):
LGTM;
image

jellespijker:
Breaks with 3 or more extruders:
```
CuraEngine: /home/jspijker/cura_wp/CuraEngine/src/LayerPlanBuffer.cpp:416: void cura::LayerPlanBuffer::insertFinalPrintTempCommand(std::vector<cura::ExtruderPlan*>&, unsigned int): Assertion `(time_window >= 0 || last_extruder_plan.estimates.getMaterial() == 0) && "Time window should always be positive if we actually extrude"' failed.

```

rburema:
Just to be complete, I'll add it here too: 'Break with >3 extruders also happens in main in the same way, so it’s not connected to this PR.'

PR #1799, opened 2023-01-11
Fix gaps between skin and wall

casperlamboo:
Use join type miter for morphological open operation. I think the reason this is needed is best shown through an example:

### When using join type round
Untitled Diagram drawio

### When using join type miter
Untitled Diagram drawio

Before we used round, which cases the skin to no longer be attached to walls in corners.
211524708-6c0680d9-d7dd-4a03-8f02-f2494de08aa0-1

Any artifacts caused by the positive offsetting using joint type miter will be mitigated because the final shapes are limited to the area of the original skin: https://github.com/Ultimaker/CuraEngine/blob/dbe17822ab036e5ccb490214ef07abe8e46139d4/src/skin.cpp#L276-L283

Fixes [11881](https://github.com/Ultimaker/Cura/issues/11881)
contributes to CURA-9204


jellespijker (approved):
LGTM

PR #1803, opened 2023-01-16
CURA-9658 Clean up support generation

casperlamboo:
When generating support normally a layer would look something like.

In this image the colors have the following meaning:
- the **orange** polygon represents the area where overhang is not allowed to be places,
- the **bue** polygon represents the area where non-overhang is allowed to be placed and
- the **green* polygon represents the outline of the model.
The area describing the disallowed overhang support areas would sometimes have very small "specs"

With this PR all these tiny specs are removed from the layers. These specs cause a lot of issues such floating support and as well as these "hairs" on the inside of the support.

Screenshot 2023-01-16 at 14 22 41

Previously we tried to remove these small specs by offsetting the polygon.
https://github.com/Ultimaker/CuraEngine/blob/2f2e068574f691ad30e91668aef8050c7310572e/src/support.cpp#L895

However this was only mitigating the problem; there is are still polygon differences for where even after offsetting the areas would roughly over lap causing these tiny specs artifacts. Instead, I propose a solution where these specs are removed by removing _all_ polygons with a small circumference.
https://github.com/Ultimaker/CuraEngine/blob/75af1ab7e5301af71d841630178da1091bc2bc5b/src/support.cpp#L1310-L1313

| Before  | After |
| Screenshot 2023-01-16 at 15 13 39 | Screenshot 2023-01-16 at 14 59 42 |
| Screenshot 2023-01-16 at 15 57 54 | Screenshot 2023-01-16 at 15 56 49 |
| Screenshot 2023-01-16 at 16 04 16 | Screenshot 2023-01-16 at 16 05 39 |

Contributes to CURA-9658

### DIFF
--- a/src/support.cpp
+++ b/src/support.cpp
@@ -926,7 +926,7 @@ void AreaSupport::generateSupportAreasForMesh(SliceDataStorage& storage,
                                                    // "specs" has a small area).
                                                    const auto nozzle_diameter = mesh_group_settings.get<coord_t>("machine_nozzle_size");
                                                    const coord_t min_circumference = nozzle_diameter * M_PI;
-                                                   const double min_area = INT2MM2(10 * (nozzle_diameter * nozzle_diameter) / (4 * M_PI));
+                                                   const double min_area = INT2MM2((nozzle_diameter * nozzle_diameter) / 4 * M_PI);
                                                    constexpr bool remove_holes = true;
                                                    larger_area_below.removeSmallAreaCircumference(min_area, min_circumference, remove_holes);
                                                }
@@ -1303,7 +1303,7 @@ std::pair<Polygons, Polygons> AreaSupport::computeBasicAndFullOverhang(const Sli
     {
         const auto nozzle_diameter = mesh.settings.get<coord_t>("machine_nozzle_size");
         const coord_t min_circumference = nozzle_diameter * M_PI;
-        const double min_area = INT2MM2(10 * (nozzle_diameter * nozzle_diameter) / (4 * M_PI));
+        const double min_area = INT2MM2((nozzle_diameter * nozzle_diameter) / 4 * M_PI);
         constexpr bool remove_holes = true;
         basic_overhang.removeSmallAreaCircumference(min_area, min_circumference, remove_holes);
     }
--- a/src/support.cpp
+++ b/src/support.cpp
@@ -674,12 +674,6 @@ void AreaSupport::generateSupportAreas(SliceDataStorage& storage)
     {
         Polygons& support_areas = global_support_areas_per_layer[layer_idx];
         support_areas = support_areas.unionPolygons();
-
-//        AABB aabb(support_areas);
-//        aabb.expand(1000);
-//        std::string name = std::string("tmp/global_support_area") + std::to_string(layer_idx) + std::string(".svg");
-//        SVG svg(name, aabb);
-//        svg.writePolygons(support_areas, SVG::Color::ORANGE, 3.0);
     }
 
     // handle support interface
--- a/src/support.cpp
+++ b/src/support.cpp
@@ -891,7 +891,6 @@ void AreaSupport::generateSupportAreasForMesh(SliceDataStorage& storage,
                                            Polygons larger_area_below; // the areas in the layer below that protrude beyond the area of the current layer
                                            if (layer_idx > 1)
                                            {
-                                               // shrink a little so that areas that only protrude very slightly are ignored
                                                larger_area_below = mesh.layers[layer_idx - 1].getOutlines().difference(mesh.layers[layer_idx].getOutlines());
 
                                                if (!larger_area_below.empty())
--- a/src/support.cpp
+++ b/src/support.cpp
@@ -674,6 +674,12 @@ void AreaSupport::generateSupportAreas(SliceDataStorage& storage)
     {
         Polygons& support_areas = global_support_areas_per_layer[layer_idx];
         support_areas = support_areas.unionPolygons();
+
+//        AABB aabb(support_areas);
+//        aabb.expand(1000);
+//        std::string name = std::string("tmp/global_support_area") + std::to_string(layer_idx) + std::string(".svg");
+//        SVG svg(name, aabb);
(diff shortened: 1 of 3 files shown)


src/support.cpp
```diff
@@ -906,14 +906,30 @@ void AreaSupport::generateSupportAreasForMesh(SliceDataStorage& storage,
...
+                                                   // areas are desired). However, sometimes these areas become so narrow that only tiny dots are present. As the area
+                                                   // of both the rings and the specs are very small we cannot use the `removeSmallAreas` utility function. Instead, we
+                                                   // use the `removeSmallAreaCircumference` which removes polygons if both the area and circumference exceed a certain
+                                                   // threshold; we want a small circumference to differentiate between "rings" and "specs" (both have small area but
+                                                   // specs also have small circumference) and we want a small area to differentiate between larger blobs and "specs" in
+                                                   // polygon (both have a circumference but only "specs" have a "larger blobs" both have a small circumference but only
+                                                   // "specs" has a small area).
+                                                   constexpr size_t min_circumference = 500;
```
rijkvanmanen, line 928: Really nice improvement! One remark: can't we couple this min_circumference to the nozzle size? I would expect that any region with a circumference smaller than nozzle_diameter * pi can be removed. Similarly the min_area can be nozzle_diameter ** 2 / 4 * pi. That saves you two magic numbers.

src/support.cpp
```diff
@@ -892,9 +892,9 @@ void AreaSupport::generateSupportAreasForMesh(SliceDataStorage& storage,
                                            if (layer_idx > 1)
                                            {
                                                // shrink a little so that areas that only protrude very slightly are ignored
```
casperlamboo, line 894: This comment needs to be removed; the shrinking logic in the next line is removed
```suggestion
```

src/support.cpp
casperlamboo: Good point, coupling these to the nozzle size does indeed make alot of sense.

But just "naively" coupling to a value from the settings in not always a good idea imo. Example being this line.
https://github.com/Ultimaker/CuraEngine/blob/e941d95fb717a652372f632225f1cdeeca632d4d/src/support.cpp#L895
Here we shrink one of the polygons to remove tiny areas. However to shrink this polygon we use 1/10th of the layer height. The layer height and polygon size have little to no correlation, and thus using layer height for the shrinkage also feels kind of like using magic numbers to me.

casperlamboo: https://github.com/Ultimaker/CuraEngine/pull/1803/commits/a94e60f1a10c4a5db30c5b8610464df39ab08eb6

src/support.cpp
```diff
@@ -924,8 +924,9 @@ void AreaSupport::generateSupportAreasForMesh(SliceDataStorage& storage,
                                                    // specs also have small circumference) and we want a small area to differentiate between larger blobs and "specs" in
                                                    // polygon (both have a circumference but only "specs" have a "larger blobs" both have a small circumference but only
                                                    // "specs" has a small area).
-                                                   constexpr size_t min_circumference = 500;
-                                                   constexpr double min_area = 0.1;
+                                                   const auto nozzle_diameter = mesh_group_settings.get<coord_t>("machine_nozzle_size");
+                                                   const coord_t min_circumference = nozzle_diameter * M_PI;
+                                                   const double min_area = INT2MM2(10 * (nozzle_diameter * nozzle_diameter) / (4 * M_PI));
```
rijkvanmanen, line 929: I think this should be (so mulitply with pi instead of dividing by pi):
INT2MM2((nozzle_diameter * nozzle_diameter) / 4 * M_PI);

casperlamboo: https://github.com/Ultimaker/CuraEngine/pull/1803/commits/7340a0fcd8eea310c7380640ac64dad06289415c

src/utils/polygon.cpp
```diff
@@ -597,6 +597,81 @@ void Polygons::removeSmallAreas(const double min_area_size, const bool remove_ho
...
+            // keep hole-polygon if we do not remove holes, or if its
+            // circumference is bigger then the minimum circumference size
+            new_polygon.add(poly);
+        }
+    }
+
+    *this = new_polygon;
+}
+
+void Polygons::removeSmallAreaCircumference(const double min_area_size, const coord_t min_circumference_size, const bool remove_holes)
```
jellespijker, line 637: I see a lot of code duplication. Please mkae this function a specialization of `removeSmallAreaCircumference`

jellespijker (changes requested):

src/utils/polygon.cpp
jellespijker: Ofcourse I mean the function above ;-)

jellespijker (changes requested):

src/support.cpp
```diff
@@ -906,14 +905,31 @@ void AreaSupport::generateSupportAreasForMesh(SliceDataStorage& storage,
...
+                                                   // use the `removeSmallAreaCircumference` which removes polygons if both the area and circumference exceed a certain
+                                                   // threshold; we want a small circumference to differentiate between "rings" and "specs" (both have small area but
+                                                   // specs also have small circumference) and we want a small area to differentiate between larger blobs and "specs" in
+                                                   // polygon (both have a circumference but only "specs" have a "larger blobs" both have a small circumference but only
+                                                   // "specs" has a small area).
+                                                   const auto nozzle_diameter = mesh_group_settings.get<coord_t>("machine_nozzle_size");
+                                                   const coord_t min_circumference = nozzle_diameter * M_PI;
+                                                   const double min_area = INT2MM2((nozzle_diameter * nozzle_diameter) / 4 * M_PI);
```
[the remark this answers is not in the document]
casperlamboo: After some preliminary testing we discovered that the code will be optimized to a more optimal constant calculation anyways making this change redundant.

src/utils/polygon.cpp
casperlamboo: https://github.com/Ultimaker/CuraEngine/pull/1803/commits/f0f49c17ac53216d7dc77fcbc63d688c5aca5a13

jellespijker (approved):

PR #1810, opened 2023-01-26
CURA-10201_fill-narrow-skin-with-walls

rijkvanmanen:
So far only skin with a width of one line was printed with a wall instead the normal pattern. But with this contribution the width of the skin regions that need to be filled with walls is exposed to the user interface.

Small skin region cause jerky motions for a normal lines pattern. Especially when printing at higher speeds you get more vibrations in the printer and the visual quality of the top layer is reduced. Since the printer needs to slow down for every turn also productivity is impacted.

frontend: https://github.com/Ultimaker/Cura/pull/14386

### DIFF
--- a/src/infill.cpp
+++ b/src/infill.cpp
@@ -82,9 +82,45 @@ void Infill::generate(std::vector<VariableWidthLines>& toolpaths,
 
     inner_contour = generateWallToolPaths(toolpaths, outer_contour, wall_line_count, infill_line_width, infill_overlap, settings);
 
-    // Apply a half-line-width offset if the pattern prints partly alongside the walls, to get an area that we can simply print the centreline alongside the edge.
-    // The lines along the edge must lie next to the border, not on it.
-    // This makes those algorithms a lot simpler.
+    // It does not make sense to print a pattern in a narrow region. So the infill region
+    // is split into a narrow region that will be filled with walls and the normal region
+    // that will be filled with the pattern. This split of regions is not needed if the
+    // infill pattern is concentric.
+    if (pattern != EFillMethod::CONCENTRIC)
+    {
+        // Split the infill region in a narrow region and the normal region.
+        Polygons narrow_infill = inner_contour;
+        inner_contour = inner_contour.offset(-small_area_width / 2).offset(small_area_width / 2);
+        narrow_infill = narrow_infill.intersection(narrow_infill.difference(inner_contour));
+        narrow_infill = Simplify(max_resolution, max_deviation, 0).polygon(narrow_infill);
+
+        // Small corners of a bigger area should not be considered narrow and are therefore added to the bigger area again.
+        for (PolygonsPart& narrow_infill_part : narrow_infill.splitIntoParts())
+        {
+            if (narrow_infill_part.offset(-infill_line_width / 2).offset(infill_line_width / 2).area() < infill_line_width * infill_line_width * 10)
+            {
+                if (! inner_contour.intersection(narrow_infill_part.offset(infill_line_width / 4)).empty())
+                {
+                    narrow_infill = narrow_infill.difference(narrow_infill_part);
+                    inner_contour = inner_contour.unionPolygons(narrow_infill_part);
+                }
+            }
+        }
+
+        // Fill narrow area with walls.
+        const size_t narrow_wall_count = std::round(small_area_width / infill_line_width) + 1;
+        WallToolPaths wall_toolpaths(narrow_infill, infill_line_width, narrow_wall_count, 0, settings);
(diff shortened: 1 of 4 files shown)


src/infill.cpp
```diff
@@ -82,59 +82,55 @@ void Infill::generate(std::vector<VariableWidthLines>& toolpaths,
...
+
+        // Small corners of a bigger area should not be considered narrow and are therefore added to the bigger area again.
+        for (PolygonsPart& small_infill_part : small_infill.splitIntoParts())
         {
-            VariableWidthLines thin_walls_only;
-            for (const auto& extrusion : var_width_line)
+            if (small_infill_part.offset(-infill_line_width / 2).offset(infill_line_width / 2).area() < infill_line_width * infill_line_width * 10)
             {
-                if (extrusion.is_odd && extrusion.inset_idx == 0)
+                if (! inner_contour.intersection(small_infill_part.offset(infill_line_width / 4)).empty())
```
casperlamboo, line 102: Why not concatenate these `if`'s using `&&`? Less indentation is better indentation.

src/infill.cpp
```diff
@@ -82,59 +82,55 @@ void Infill::generate(std::vector<VariableWidthLines>& toolpaths,
...
-                    }
-                    if (path.polygonLength() >= infill_line_width * 4) // Don't fill gaps that are very small (with paths less than 2 line widths long, 4 back and forth).
-                    {
-                        gap_filled_areas.add(path);
-                        if (fill_gaps)
-                        {
-                            thin_walls_only.push_back(extrusion);
-                        }
-                    }
+                    small_infill = small_infill.difference(small_infill_part);
```
casperlamboo, line 104: I think you can be done smarter by omitting the difference operation as this is quite an expensive operation. I assume this is done so the `small_infill` contains all regions that are not printed yet. As you only print a subset of parts in this loop it is possible that some parts are not yet printed. Is this correct?

But if that is the case, why not do something like:
```cpp
auto small_infill_parts = small_infill.splitIntoParts();
small_infill.clear();
for (PolygonsPart& small_infill_part : small_infill_parts)
{
    if (condition_to_add_part)
    {
         // do stuff
    }
    else
    {
         small_infill.add(small_infill_part);
    }
}
```
Benefit of this is that no expensive difference is executed, and there is no chance of introducing small polygon-shilfers due to boolean operation artifacts.

src/infill.cpp
```diff
@@ -82,59 +82,55 @@ void Infill::generate(std::vector<VariableWidthLines>& toolpaths,
...
                 }
             }
-            if (! thin_walls_only.empty())
+        }
+
+        // Fill narrow area with walls.
+        const size_t narrow_wall_count = std::round(small_area_width / infill_line_width) + 1;
+        WallToolPaths wall_toolpaths(small_infill, infill_line_width, narrow_wall_count, 0, settings);
+        std::vector<VariableWidthLines> small_infill_paths = wall_toolpaths.getToolPaths();
+        if (! small_infill_paths.empty())
```
casperlamboo, line 114: You can loose this if-statement, if it's empty then the for loop will not be executed anyways.

src/infill.cpp
```diff
@@ -82,59 +82,55 @@ void Infill::generate(std::vector<VariableWidthLines>& toolpaths,
...
+        inner_contour = inner_contour.offset(-small_area_width / 2).offset(small_area_width / 2);
+        small_infill = small_infill.intersection(small_infill.difference(inner_contour));
+        small_infill = Simplify(max_resolution, max_deviation, 0).polygon(small_infill);
+
+        // Small corners of a bigger area should not be considered narrow and are therefore added to the bigger area again.
+        for (PolygonsPart& small_infill_part : small_infill.splitIntoParts())
         {
-            VariableWidthLines thin_walls_only;
-            for (const auto& extrusion : var_width_line)
+            if (small_infill_part.offset(-infill_line_width / 2).offset(infill_line_width / 2).area() < infill_line_width * infill_line_width * 10)
```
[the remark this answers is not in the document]
rijkvanmanen:
```suggestion
            if (small_infill_part.offset(-infill_line_width / 2).offset(infill_line_width / 2).area() < small_area_width * small_area_width)
```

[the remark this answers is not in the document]
rijkvanmanen: The idea is that if you have a small corner that is connected to a bigger area you don't want to fill the small corner with walls. But that requires a definition of a "small". `infill_line_width * infill_line_width * 10` is a magic number I found useful.

This is an exaggerated example. The top part is a bit more sharp, causing the narrow area to be big enough to be considered a separate area from the main body. So it is filled with walls. The bottom part is less sharp, so the narrow is a bit smaller. So it is considered to be part of the main body.
image

PR #1821, opened 2023-02-08
[CURA-10050] First expand for both, then do difference.

rburema:
Last small fix before beta?

Solve conflict with other fix.

### DIFF
--- a/src/InterlockingGenerator.cpp
+++ b/src/InterlockingGenerator.cpp
@@ -71,10 +71,13 @@ std::pair<Polygons, Polygons> InterlockingGenerator::growBorderAreasPerpendicula
     Polygons from_border_a = a.difference(total_shrunk);
     Polygons from_border_b = b.difference(total_shrunk);
 
+    Polygons temp_a, temp_b;
     for (auto _ : ranges::view::iota(0, (detect / min_line) + 2))
     {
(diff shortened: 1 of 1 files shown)


src/InterlockingGenerator.cpp
```diff
@@ -71,10 +71,13 @@ std::pair<Polygons, Polygons> InterlockingGenerator::growBorderAreasPerpendicula
     Polygons from_border_a = a.difference(total_shrunk);
     Polygons from_border_b = b.difference(total_shrunk);
 
+    Polygons temp_a, temp_b;
     for (auto _ : ranges::view::iota(0, (detect / min_line) + 2))
```
jellespijker, line 75:
```suggestion
for (auto _ : ranges::views::iota(0, (detect / min_line) + 2))
```

`view` namespace is deprecated use `views` instead

jellespijker (changes requested):
LGTM; only use the `views` different namespace

src/InterlockingGenerator.cpp
jellespijker: the includes do not change only the namespace
image

jellespijker: applied the review comment myself such that this can be merged

rburema: thanks!

PR #1836, opened 2023-02-17
[CURA-10310] support fixes Friday

rburema:
Prevent jitters by:
- Thin strips (unlike small areas) should first be propagated first before any unprintable spots are removed.
- Stop propagation of material that will be cut off by the 'minimum x/y distance' (as opposed to normal x/y distance).
- _(In front-end, see [the frontend PR here](https://github.com/Ultimaker/Cura/pull/14591): Infill patterns don't stay the same from layer to layer. This could cause jitters, but also wholly missing or floating support. -- You can easily see this when brim is enabled, and brim is generated for an area of support that just ... _stops_ somewhere at layer whatever.)_

---

_EDIT: Converted to draft for safety w.r.t. multiple possible solutions, but otherwise 'finished'._


nallath (approved):

rburema:
Closed. Superseded by #1840 and/or #1842

PR #1837, opened 2023-02-17
[CURA-9903] Move documentation to wiki

Joeydelarago:
CURA-9903

# Description

Move Documentation to Wiki

## Type of change

- [x] Documentation


jellespijker:
@Joeydelarago did you check the effects it would have on the benchmark workflow and the gh-pages?

jellespijker (approved):

PR #1840, opened 2023-02-19
Varying x/y distance

casperlamboo:
### Problem description
When x/y-distance priority is set to z overrides x/y we have 3 different support distance values that compete, namely
- z-distance, a vertical distance from the model
- x/y-distance, a horizontal distance from the model
- minimum x/y-distance, a horizontal distance, usually smaller then the x/y-distance from the model

These different distances are visualized in the figure below.

xx

As the priority is set to z distance the z-distance takes priority over the x/y-distance. However, by blindly following this z-distance the support will at some point attach to the model. Here the minimum x/y-distance comes into play to at least leave small gap between the support and the model. Once the boundary of the model is perfectly vertical only the x/y-distance is used.

The problem we are experiencing is when some of the walls are approximately vertical, we cannot reliably detect this, causing the disallowed distance to jitter between the x/y-distance and the minimum x/y-distance. This is exceptionally noticeable in organic models.

xxxasdf

### Solution
This PR aims to solve this issue by introducing a smooth transition from the minimum x/y-distance to the x/y-distance. Support placed due to overhangs have an implicit x/y-distance, this distance can be derived from the overhang angle and the z-distance (we call this distance the natural x/y-distance).

Screenshot 2023-02-20 at 11 28 00

### Known issues

There is still an issue with discontinuous between adjacent triangles. This can seen in the support example above. As there is a sudden change in wall angle there is also a hard jump in x/y distance; solving a 

For smoothing in the z-direction this can easily be solved by increasing the `layer_index_offset` here
https://github.com/Ultimaker/CuraEngine/blob/2fa64d3e67559b65232c4615aa4e3efd11cbcf8a/src/support.cpp#L822

However there is also an issue for the x/y distances. If two adjacent triangles have different overhang angles their offset can differ per offset. This can result in some very ugly looking disallowed areas. See image below. Note that this is done with an extreme x/distance of 3.0mm; default value is 0.7mm for ultimaker printers (the issue is virtually non-existent when using this default value).

A solution to the issue would be to smooth the x/y offset distance with the surrounding x/y offset distances. This would mimic the `layer_index_offset` for the z-direction in the x/y direction.

filename_114

CURA-10297

### DIFF
--- a/src/support.cpp
+++ b/src/support.cpp
@@ -6,12 +6,14 @@
 #include <fstream> // ifstream.good()
 #include <utility> // pair
 
+#include <range/v3/numeric/accumulate.hpp>
 #include <range/v3/view/drop.hpp>
 #include <range/v3/view/drop_last.hpp>
 #include <range/v3/view/enumerate.hpp>
 #include <range/v3/view/filter.hpp>
 #include <range/v3/view/slice.hpp>
 #include <spdlog/spdlog.h>
+#include <range/v3/view/transform.hpp>
 
 #include "Application.h" //To get settings.
 #include "ExtruderTrain.h"
@@ -30,6 +32,9 @@
 #include "utils/Simplify.h"
 #include "utils/ThreadPool.h"
 #include "utils/math.h"
+#include "SkeletalTrapezoidation.h"
+#include "utils/VoronoiUtils.h"
+#include "BoostInterface.hpp"
 
 namespace cura
 {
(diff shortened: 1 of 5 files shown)


src/support.cpp
```diff
@@ -783,6 +788,213 @@ void AreaSupport::generateOverhangAreasForMesh(SliceDataStorage& storage, SliceM
...
+    const LayerIndex layer_idx_below = std::max(layer_idx - layer_index_offset, static_cast<LayerIndex>(0));
+    if (layer_idx_below != layer_idx)
+    {
+        auto layer_below = storage.layers[layer_idx_below].getOutlines()
+                               .offset(-close_dist)
+                               .offset(close_dist)
+                               .smooth(close_dist);
+
+        z_distances_layer_deltas.emplace_back(
+            static_cast<double>(support_distance_top),
```
jellespijker, line 836:
```suggestion
            support_distance_top,
```

see earlier cast above

src/support.cpp
```diff
@@ -783,6 +788,213 @@ void AreaSupport::generateOverhangAreasForMesh(SliceDataStorage& storage, SliceM
                                });
 }
 
+Polygons AreaSupport::generateVaryingXYDisallowedArea(const SliceMeshStorage& storage, const Settings& infill_settings, const LayerIndex layer_idx)
+{
+    const auto& mesh_group_settings = Application::getInstance().current_slice->scene.current_mesh_group->settings;
+    const auto layer_thickness = mesh_group_settings.get<coord_t>("layer_height");
+    const auto support_distance_top = mesh_group_settings.get<coord_t>("support_top_distance");
+    const auto support_distance_bot = mesh_group_settings.get<coord_t>("support_bottom_distance");
```
jellespijker, line 797:
```suggestion
    const auto support_distance_top = static_cast<double>(mesh_group_settings.get<coord_t>("support_top_distance"));
        const auto support_distance_bot = static_cast<double>(mesh_group_settings.get<coord_t>("support_bottom_distance"));
```

you're only using it as a double everywhere

src/support.cpp
```diff
@@ -783,6 +788,213 @@ void AreaSupport::generateOverhangAreasForMesh(SliceDataStorage& storage, SliceM
...
+                               .smooth(close_dist);
+
+        z_distances_layer_deltas.emplace_back(
+            static_cast<double>(support_distance_top),
+            static_cast<double>(layer_index_offset * layer_thickness),
+            layer_current.difference(layer_below)
+        );
+
+        z_distances_layer_deltas.emplace_back(
+            static_cast<double>(support_distance_bot),
```
jellespijker, line 842:
```suggestion
            support_distance_bot,
```

see earlier cast above

src/support.cpp
```diff
@@ -783,6 +788,213 @@ void AreaSupport::generateOverhangAreasForMesh(SliceDataStorage& storage, SliceM
...
+        );
+    }
+
+    const LayerIndex layer_idx_above = std::min(layer_idx + layer_index_offset, static_cast<LayerIndex>(storage.layers.size() - 1));
+    if (layer_idx_above != layer_idx)
+    {
+        auto layer_above = storage.layers[layer_idx_below].getOutlines()
+                               .offset(-close_dist)
+                               .offset(close_dist)
+                               .smooth(close_dist);
```
jellespijker, line 854: Why not simplify?

src/support.cpp
```diff
@@ -783,6 +788,213 @@ void AreaSupport::generateOverhangAreasForMesh(SliceDataStorage& storage, SliceM
...
+
+        z_distances_layer_deltas.emplace_back(
+            static_cast<double>(support_distance_bot),
```
jellespijker, line 857:
```suggestion
            support_distance_bot,
```

See earlier cast above

src/support.cpp
```diff
@@ -783,6 +788,213 @@ void AreaSupport::generateOverhangAreasForMesh(SliceDataStorage& storage, SliceM
...
+            static_cast<double>(layer_index_offset * layer_thickness),
+            layer_current.difference(layer_above)
+        );
+
+        z_distances_layer_deltas.emplace_back(
+            static_cast<double>(support_distance_top),
```
jellespijker, line 863:
```suggestion
            static_cast<double>(support_distance_top),
```

See earlier cast above

src/support.cpp
```diff
@@ -783,6 +788,213 @@ void AreaSupport::generateOverhangAreasForMesh(SliceDataStorage& storage, SliceM
...
+            static_cast<double>(layer_index_offset * layer_thickness),
+            layer_above.difference(layer_current)
+        );
+    }
+
+    int i = 0;
```
jellespijker, line 869:
```suggestion
    int i = 0;
```

Is this variable here looking pretty, or do we actually use it? I couldn't find it

src/utils/polygon.cpp
```diff
@@ -360,6 +369,49 @@ Polygons Polygons::offset(int distance, ClipperLib::JoinType join_type, double m
     return ret;
 }
 
+Polygons Polygons::offset(const std::vector<int>& offset_dists) const
+{
+    // we need as many offset-dists as points
+    assert(this->pointCount() == offset_dists.size());
```
jellespijker, line 375: Almost like a contract, can't wait till c++23
https://github.com/Ultimaker/Scripta/blob/91c2fc92067d91ff3c811b13e1d59d5d655f36b2/include/scripta/contracts/machine_limits.h#L13

jellespijker (approved):
See comments

PR #1848, opened 2023-02-28
Fix support generation for models with narrow walls

casperlamboo:
This was basically a side effect from [CURA-10314](https://ultimaker.atlassian.net/browse/CURA-10314). For this feature we were looking at the layer(s) above. The idea is that the overhang needs to exist for multiple layers before support is generated. When looking above the overhang angle is limited by the area of the above layer shrunken by `tan(overhang_angle) * delta_z`. As this shrinking behavior is achieved using a negative offset narrow walls would completely be removed.

In order to keep in line with the spirit of the contributions from [CURA-10314](https://ultimaker.atlassian.net/browse/CURA-10314) we instead look to the layer(s) below rather then above. When looking below a positive offset need to be taken in order to achieve a similar result. With a positive offset these narrow features can never be removed.

| Without Fix | With Fix |
|----------|---------|
| Screenshot 2023-02-28 at 14 09 32 | Screenshot 2023-02-28 at 14 08 25 |

@rijkvanmanen FYI

CURA-10336

[CURA-10314]: https://ultimaker.atlassian.net/browse/CURA-10314?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CURA-10314]: https://ultimaker.atlassian.net/browse/CURA-10314?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/support.cpp
+++ b/src/support.cpp
@@ -1492,16 +1492,19 @@ std::pair<Polygons, Polygons> AreaSupport::computeBasicAndFullOverhang(const Sli
     // To avoids generating support for textures on vertical surfaces, a moving average
     // is taken over smooth_height. The smooth_height is currently an educated guess
     // that we might want to expose to the frontend in the future.
-    Polygons outlines_below = storage.getLayerOutlines(layer_idx - 1, no_support, no_prime_tower)
-                                                         .offset(max_dist_from_lower_layer);
+    Polygons outlines_below =
+        storage.getLayerOutlines(layer_idx - 1, no_support, no_prime_tower)
+        .offset(max_dist_from_lower_layer);
     for (int layer_idx_offset = 2; layer_idx - layer_idx_offset >= 0 && layer_idx_offset <= layers_below; layer_idx_offset ++)
     {
-        auto outlines_below_ = storage.getLayerOutlines(layer_idx - layer_idx_offset, no_support, no_prime_tower)
-                                   .offset(max_dist_from_lower_layer * layer_idx_offset);
+        auto outlines_below_ =
+            storage.getLayerOutlines(layer_idx - layer_idx_offset, no_support, no_prime_tower)
+            .offset(max_dist_from_lower_layer * layer_idx_offset);
         outlines_below = outlines_below.unionPolygons(outlines_below_);
     }
 
-    Polygons basic_overhang = outlines
+    Polygons basic_overhang =
+        outlines
         .difference(outlines_below);
 
     const SupportLayer& support_layer = storage.support.supportLayers[layer_idx];
@@ -1514,7 +1517,10 @@ std::pair<Polygons, Polygons> AreaSupport::computeBasicAndFullOverhang(const Sli
         basic_overhang = basic_overhang.difference(merged_polygons);
     }
 
-    Polygons overhang_extended = basic_overhang.offset(max_dist_from_lower_layer * layers_below + MM2INT(0.1)); // +0.1mm for easier joining with support from layer above
+    Polygons overhang_extended =
+        basic_overhang
+        // +0.1mm for easier joining with support from layer above
+        .offset(max_dist_from_lower_layer * layers_below + MM2INT(0.1));
     Polygons full_overhang = overhang_extended.intersection(outlines);
 
     return std::make_pair(basic_overhang, full_overhang);
--- a/src/support.cpp
+++ b/src/support.cpp
@@ -1479,30 +1479,30 @@ std::pair<Polygons, Polygons> AreaSupport::computeBasicAndFullOverhang(const Sli
     const Polygons outlines = mesh.layers[layer_idx].getOutlines();
     constexpr bool no_support = false;
     constexpr bool no_prime_tower = false;
-    const Polygons outlines_below = storage.getLayerOutlines(layer_idx - 1, no_support, no_prime_tower);
 
-    // To avoids generating support for textures on vertical surfaces, a moving average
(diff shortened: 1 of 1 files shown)


src/support.cpp
```diff
@@ -1479,30 +1479,30 @@ std::pair<Polygons, Polygons> AreaSupport::computeBasicAndFullOverhang(const Sli
...
-    // that we might want to expose to the frontend in the future.
     constexpr double smooth_height = 0.4; //mm
-    const auto layers_above = static_cast<LayerIndex>(std::round(smooth_height / mesh.settings.get<double>("layer_height")));
-    Polygons outlines_above;
-    for (int i = 0; layer_idx + i < mesh.layers.size() && i <= layers_above; ++i)
-    {
-        outlines_above.add(storage.getLayerOutlines(layer_idx + i, no_support, no_prime_tower));
-    }
-    outlines_above = outlines_above.unionPolygons();
+    const auto layers_below = static_cast<LayerIndex>(std::round(smooth_height / mesh.settings.get<double>("layer_height")));
```
rburema, line 1484: In the engine, we usually use `coord_t` when doing calculations, since the entire system is in those micron coords. You can use the `round_divide` utility-function we made a while back for those cases.

src/support.cpp
```diff
@@ -1479,30 +1479,30 @@ std::pair<Polygons, Polygons> AreaSupport::computeBasicAndFullOverhang(const Sli
...
     const AngleRadians support_angle = mesh.settings.get<AngleRadians>("support_angle");
     const double tan_angle = tan(support_angle) - 0.01; // The X/Y component of the support angle. 0.01 to make 90 degrees work too.
+    // overhang areas protruding less then `max_dist_from_lower_layer` don't need support
     const coord_t max_dist_from_lower_layer = tan_angle * layer_height; // Maximum horizontal distance that can be bridged.
 
+    // To avoids generating support for textures on vertical surfaces, a moving average
+    // is taken over smooth_height. The smooth_height is currently an educated guess
+    // that we might want to expose to the frontend in the future.
+    Polygons outlines_below = storage.getLayerOutlines(layer_idx - 1, no_support, no_prime_tower)
+                                                         .offset(max_dist_from_lower_layer);
```
rburema, line 1496: I'm not really a fan of these huge amounts of spaces and just arbitrarily placing the next bit vaguely under the last.

If you need multiple lines I'd say do something like this:

~~~
Type varname =
    blahblah
    .funky_name_name(aaaa_blahbloehbleh)
    .smurfsmurfsmurf(yolo - 1);
~~~

(For completeness sake, and this isn't really addressed to you in particular: This goes for those functions with lots of parameters as well that I now see all over the place. Place even the first parameter on a new line with a single added indentation instead of having the first argument on the same line as the function and then having a huge arbitrarily long indentation for each next line to line it out with the previous parameter...)

src/support.cpp
```diff
@@ -1479,30 +1479,30 @@ std::pair<Polygons, Polygons> AreaSupport::computeBasicAndFullOverhang(const Sli
...
+    for (int layer_idx_offset = 2; layer_idx - layer_idx_offset >= 0 && layer_idx_offset <= layers_below; layer_idx_offset ++)
+    {
+        auto outlines_below_ = storage.getLayerOutlines(layer_idx - layer_idx_offset, no_support, no_prime_tower)
```
rburema, line 1499: I'm sort-of fine with the name and structure as is, but if you know anything better for the temp-parameter name now would be a good time to change it.

Maybe just place it in there, since this is the only place it's needed, or, since `unionPolygons(...)` is symmetrical, maybe then swap the arguments as well so it's one flowing line. -- Although perhaps that would make the code _less_ readable instead of more.

rburema, line 1499: We aren't going to fix this now, but I think we could calculate the `outlines_below` for all layers within one go (if we subtract the one that is going 'out-of-scope' of the number of layers takin into account so to speak). That'd be outside of this function though and could complicate matters not only in that we need to program it correctly, but also testing, etc.

rburema (commented):
I think it looks good so far. All my comments are on the pedantic side (still would like you to take a look at them of course), so it doesn't block me from testing.

I'm curious how this will perform with a number of models I have in mind. Specifically, when the thing that is needed to support is near the top and bottom of the model, with varying amounts of # layers. (It'll _probably_ be not worse than what we have now even in those cases, but I wonder about in comparison to say, 5.1)

src/support.cpp
casperlamboo: Don't think that is possible unfortunately as we perform an offset that scales with the `layer_idx_offeset`; the outline for one particular _below-layer_ is not the same between layers.

rburema: Right of course, since we need to shrink it as well, got it.

casperlamboo: When programming I tried to resolve it by performing this statement in-line with in the `unionPolygons` function call below. Did not think that was an improvement over the current implementation.

I'm not a big fan of variable names like `outlines_below_temp` etc aswell, but couldn't really come up with a better name.

src/support.cpp
casperlamboo: https://github.com/Ultimaker/CuraEngine/pull/1848/commits/fbd017f54593c4f4340e6b0b41a6d9fdfe1c876a

PR #1849, opened 2023-02-28
[CURA-10268] Fix wipe for walls.

rburema:
For wiping in the walls (as opposed to infill) there was an extra requirement for the path not to be a linked path, that is to say, we don't wipe for open paths where there could be another of its exact type right after. -- In order to detect whether something is a 'linked path', we need to determine if a path is closed or not. Previously, a path was closed if and only if the end-vertex was the same as the start-vertex. That equivalence has been dropped. Fortunately, `is_closed` was added way before that, so it can now be used to determine the `is_linked` status of the path.

### DIFF
--- a/src/InsetOrderOptimizer.cpp
+++ b/src/InsetOrderOptimizer.cpp
@@ -110,7 +110,6 @@ bool InsetOrderOptimizer::addToLayer()
 
     order_optimizer.optimize();
 
-    cura::Point p_end{ 0, 0 };
     for (const PathOrdering<const ExtrusionLine*>& path : order_optimizer.paths)
     {
         if (path.vertices->empty())
@@ -127,9 +126,6 @@ bool InsetOrderOptimizer::addToLayer()
         const bool revert_layer = alternate_walls && (layer_nr % 2);
         const bool backwards = path.backwards != (revert_inset != revert_layer);
         const size_t start_index = (backwards != path.backwards) ? path.vertices->size() - (path.start_vertex + 1) : path.start_vertex;
-
-        p_end = path.backwards ? path.vertices->back().p : path.vertices->front().p;
-        const cura::Point p_start = path.backwards ? path.vertices->front().p : path.vertices->back().p;
         const bool linked_path = ! path.is_closed;
 
         gcode_writer.setExtruder_addPrime(storage, gcode_layer, extruder_nr);
--- a/src/InsetOrderOptimizer.cpp
+++ b/src/InsetOrderOptimizer.cpp
@@ -130,7 +130,7 @@ bool InsetOrderOptimizer::addToLayer()
 
         p_end = path.backwards ? path.vertices->back().p : path.vertices->front().p;
         const cura::Point p_start = path.backwards ? path.vertices->front().p : path.vertices->back().p;
-        const bool linked_path = p_start != p_end;
+        const bool linked_path = ! path.is_closed;
 
         gcode_writer.setExtruder_addPrime(storage, gcode_layer, extruder_nr);
         gcode_layer.setIsInside(true); // Going to print walls, which are always inside.


casperlamboo (changes requested):
I don't think the proposed solution from this PR is implementing what was intended for `is_linked` the parameter. From the documentation:
https://github.com/Ultimaker/CuraEngine/blob/088b7a34f782474859cd97564a290dd0a19b10b7/include/LayerPlan.h#L559

From context I think that `is_linked_path` should be true **if the start of the current path is the same as the end of the previous path**.

Reasons why I think this is true
- This is in accordance with the description of the parameter (see documentation excerpt above)
- [`p_end` is defined outside of the main loop](https://github.com/Ultimaker/CuraEngine/blob/088b7a34f782474859cd97564a290dd0a19b10b7/src/InsetOrderOptimizer.cpp#L113), which is a very common pattern to store _current end point_ at the end of each iteration to be used as the _previous end point_ for the next iteration

Reason why I do not think this was intended to be the closed property of the vertex is because there was already a parameter `is_closed` with the value `path.is_closed`. Description of this parameter reads https://github.com/Ultimaker/CuraEngine/blob/088b7a34f782474859cd97564a290dd0a19b10b7/include/LayerPlan.h#L556-L557

In order to implement this new proposed solution the only chance need to be made is moving the following [statement](https://github.com/Ultimaker/CuraEngine/blob/088b7a34f782474859cd97564a290dd0a19b10b7/src/InsetOrderOptimizer.cpp#L131) 3 lines down.

rburema:
@casperlamboo That would be nice, expect that doesn't work :-p

First something that you might not be aware of: This part of the code (where the changes where made) used to work fine in 5.2.x, when the definitions where equivalent. With that out of the way:

W.r.t. the pattern, it's true that this is a very common occurrence, and that might be the reason the loop was originally written this way, but it's clear to me that hasn't been the case for a long time: The 'old' end doesn't get used anywhere else in the loop. Since this is the same code that used to work, I think we can reasonably safely assume that this was the intended order: Get the start and end of the current path, then check for a loop; if it's _not_ a loop, it's an open path and can thus be linked.

Also, the `is_closed` property was implemented later (it's younger than the line I originally  replaced), if we _didn't_ change the end != start definition (either on purpose or accidentally) for looping paths, it should be equivalent to what I did here (plus maybe your suggested change). It's not however.

Whether the definition of a closed path is or isn't start != end anymore, this should work in any case. Come to think of it, I think the start/end can be removed, since they where only needed for `linked_path`.

It would be a good idea to bring the change to end != start for loop-detection to the eCCB, if we have an answer there, we could also either change the comments, or go on a hunt to see why it doesn't work like that anymore.

jellespijker:
breakage of this functionality might also be related to the wipe config now being able to set per-mesh. This also broke the purge configuration when switching nozzles (the so called poopies)

rburema:
@jellespijker I currently don't think that's the reason, and that my original fix suffices for restoring the wipe in particular.

casperlamboo:
Ah I see, wasn't aware about the changes regarding closed paths.

This just came to mind
```cpp
    cura::Point p_end{ 0, 0 };
    for (const PathOrdering<const ExtrusionLine*>& path : order_optimizer.paths)
    {
        ...

        const bool linked_path = p_start != p_end;

        if (path.is_closed)
        {
            p_end = path.vertices->front().p;
        }
        else
        {
            p_end = path.backwards ? path.vertices->back().p : path.vertices->front().p;
        }
        
        ...
   }
```
Would that also work? might be better inline with the "intended" changes (to the extend in which we can interpret the original intent).

rburema:
> Ah I see, wasn't aware about the changes regarding closed paths.

Well, I'm not sure the changes are intentional... (that's what I hope to find out at either the stand-up or the eCCB).

However:
- If they are _not_ intentional, then this change at least can't hurt (and fixes the problem at least for the wipe), since then `is_closed` was meant to convey the same, but more conveniently in a single variable. In that case, maybe the error that lead to this should be fixed either instead of as well.
- If they _are_ intentional, then `end == start` doesn't mean the same thing (we relied on) anymore and it _should_ be replaced by `is_closed` at least in this case.

Also, the whole `is_linked` thing is the only variable that still previously used the start/end. So I removed them in the next commit (I made after your first comment). The way it was using those, back when the code was correct (so in like 5.2, 5.1, or before) is consistent with `! is_closed`.

casperlamboo (approved):

PR #1850, opened 2023-03-01
[CURA-9463] Speed up skin-edge support detection for detailed models.

rburema:
Limit the considered areas to those that actually have infill to make 'infill skin edge support' from on the layer currently under consideration for the area determination.

This can speed up the calculation for certain models a huge amount (under a minute of slicing versus half an hour kind of ballpark).

Important, since we enabled skin edge support in the front-end by default instead of as an option.

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -2015,14 +2015,17 @@ bool FffGcodeWriter::partitionInfillBySkinAbove(Polygons& infill_below_skin, Pol
         const size_t skin_layer_nr = gcode_layer.getLayerNr() + i;
         if (skin_layer_nr < mesh.layers.size())
         {
-            for (const SliceLayerPart& part : mesh.layers[skin_layer_nr].parts)
+            for (const SliceLayerPart& part_i : mesh.layers[skin_layer_nr].parts)
             {
-                for (const SkinPart& skin_part : part.skin_parts)
+                for (const SkinPart& skin_part : part_i.skin_parts)
                 {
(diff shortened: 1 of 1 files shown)


src/FffGcodeWriter.cpp
```diff
@@ -2015,14 +2015,17 @@ bool FffGcodeWriter::partitionInfillBySkinAbove(Polygons& infill_below_skin, Pol
...
         if (skin_layer_nr < mesh.layers.size())
         {
-            for (const SliceLayerPart& part : mesh.layers[skin_layer_nr].parts)
+            for (const SliceLayerPart& part_i : mesh.layers[skin_layer_nr].parts)
             {
-                for (const SkinPart& skin_part : part.skin_parts)
+                for (const SkinPart& skin_part : part_i.skin_parts)
                 {
+                    // Limit considered areas to the ones that should have infill underneath at the current layer.
+                    const Polygons relevant_outline = skin_part.outline.intersection(part.getOwnInfillArea());
```
casperlamboo, line 2023: Isn't `part.getOwnInfillArea()` always a strict subset of `skin_part.outline`?

rburema: Not really, because these needn't be on the same layer. The variable `part` (and note that I renamed the one in the loop to something else, to prevent the shadowing that was going on) is the one for the 'current' layer, always, and the skin-part outlines are in the layers above (`i` is always > 0, so it doesn't even touch the current layer).

If you went with only the infill, the skin-support-edges wouldn't be underneath the skin, which would defeat the purpose.

casperlamboo (approved):

PR #1853, opened 2023-03-03
Fix generating interlocking interface for support blockers

casperlamboo:


src/InterlockingGenerator.cpp
```diff
@@ -5,6 +5,8 @@
 
 #include <algorithm> // max
 
+#include <spdlog/spdlog.h>
+
```
casperlamboo, line 9:
```suggestion
```

jellespijker (approved):
lgtm

PR #1855, opened 2023-03-05
Update LayerPlan.cpp Fixes #14679

ckvsoft:
Fix Typo NOMESH should be NONMESH


pedrolamas:
@ckvsoft please update your top comment and add a "Fixes #14679" so it gets linked to the issue.

rburema (approved):

rburema:
Thanks for the contrib!

rburema:
@pedrolamas Just the number alone won't work here; CuraEngine is another repo than that issue is reported against (The Cura repository is what we sometimes call the 'frontend', though it does more than that name implies. The 'actual' slicing happens in the engine.)

PR #1856, opened 2023-03-06
CURA-9826_add-skirt-height

rijkvanmanen:
Make the layer_nr known by the processSkirtBrim function. And call this function for all layers smaller then the skirt_height.

frontend: https://github.com/Ultimaker/Cura/pull/14792
CURA-9826


jellespijker (approved):
LGTM;

PR #1859, opened 2023-03-10
Fix fuzzy skin

casperlamboo:
Previously we had semantics that for all closed paths the last point was equal to the first point; `path.front() != path.back()`. With commit d31d61518b1659f6a5181573918b0141e726b3f2 the semantics changed.

Fixes https://github.com/Ultimaker/Cura/issues/14750, CURA-10368

### DIFF
--- a/src/WallToolPaths.cpp
+++ b/src/WallToolPaths.cpp
@@ -199,11 +199,16 @@ void WallToolPaths::removeSmallLines(std::vector<VariableWidthLines>& toolpaths)
 void WallToolPaths::simplifyToolPaths(std::vector<VariableWidthLines>& toolpaths, const Settings& settings)
 {
     const Simplify simplifier(settings);
-    for(size_t toolpaths_idx = 0; toolpaths_idx < toolpaths.size(); ++toolpaths_idx)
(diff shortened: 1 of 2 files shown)


src/WallToolPaths.cpp
```diff
@@ -199,11 +199,16 @@ void WallToolPaths::removeSmallLines(std::vector<VariableWidthLines>& toolpaths)
...
     const Simplify simplifier(settings);
-    for(size_t toolpaths_idx = 0; toolpaths_idx < toolpaths.size(); ++toolpaths_idx)
+    for (auto& toolpath : toolpaths)
     {
-        for(ExtrusionLine& line : toolpaths[toolpaths_idx])
+        for (auto& line : toolpath)
         {
             line = line.is_closed ? simplifier.polygon(line) : simplifier.polyline(line);
+
+            if (line.is_closed && line.front() != line.back())
```
casperlamboo, line 208: Maybe do some defensive programming here to prevent run time errors when `line` is emty
```suggestion
            if (line.is_closed && line.size() >= 2 && line.front() != line.back())
```

rburema (approved):

PR #1867, opened 2023-03-26
Cache extruder settings

sesse:
# Description

When writing GCode, there are certain settings that need to be checked over and over again. Instead of pulling them from the Settings hash table every time, read them once and then save them. This speeds up single-threaded slicing by about 3% (one thread of a 5950X; with many threads, it seems to be more like 10%, since GCode writing then takes proportionally more of the time).

There is a very large long tail of code that keeps reading these settings, in particular from mesh settings, so there is probably more to be gained here, but this patch takes out the two biggest cases.

## Type of change

Performance optimization.

# How Has This Been Tested?

Only marginally; verified that a simple slice gives the same GCode, modulo timestamps and similar. Benchmarking has been done partially with perf, but verified with hyperfine (the times are fairly consistent from run to run):

```
Benchmark 1: ./run-cura.sh ./build/Release/CuraEngine.old
  Time (mean ± σ):      3.025 s ±  0.062 s    [User: 3.017 s, System: 0.010 s]
  Range (min … max):    2.910 s …  3.121 s    100 runs
 
Benchmark 2: ./run-cura.sh ./build/Release/CuraEngine
  Time (mean ± σ):      2.974 s ±  0.036 s    [User: 2.967 s, System: 0.009 s]
  Range (min … max):    2.824 s …  3.020 s    100 runs
 
Summary
  './run-cura.sh ./build/Release/CuraEngine' ran
    1.02 ± 0.02 times faster than './run-cura.sh ./build/Release/CuraEngine.old'
```

**Test Configuration**:
* Operating System:  AMD 5950X, Debian unstable

# Checklist:


MariMakes:
Quick update from our side 👋

Due to a miscommunication, your contribution was not mentioned in the 5.5 beta release. 
I'm restoring that, and making sure you get the correct credits in the 5.5 stable release. 

Thanks again for your contribution 💖

sesse:
Thanks for being diligent about it!

PR #1869, opened 2023-03-28
[CURA-10394] Remove wire-printing options.

rburema:
Wire-printing wasn't actively maintained, and almost an entire different min-engine. It's also rarely used, and presumed broken. Remove it, so it doesn't have to be maintained or clog up other maintenance and/or refactor tasks.

See also https://github.com/Ultimaker/Cura/pull/15067


include/FffGcodeWriter.h
```diff
@@ -1,4 +1,4 @@
-//  Copyright (c) 2022 Ultimaker B.V.
+//  Copyright (c) 2023 UltiMaker B.V.
```
jellespijker, line 1:
```suggestion
// Copyright (c) 2023 UltiMaker
```

src/Scene.cpp
```diff
@@ -1,13 +1,11 @@
-// Copyright (c) 2022 Ultimaker B.V.
+// Copyright (c) 2023 UltiMaker B.V.
```
jellespijker, line 1:
```suggestion
// Copyright (c) 2023 UltiMaker
```

jellespijker (approved):
LGTM

PR #1873, opened 2023-04-05
[CURA-10464] Tree Support 2.0 speed-ups

rburema:
This is mostly just two parts:
- Upping (and fixing) the minimum step size in the safeOffset___ functions, since that does sort-of the same as a normal offset, except then in steps, and each step _also_ has a union or difference with the collison. The less you do this the better for speed, but you can't also just make it a normal offset, since that could make the support polygons 'move' through the model.
- Decreasing the amount of work done in order to get all of the circle polygons merged into one big blob. This is done by a) reducing the resolution of the individual 'circles' (now dodecagons I suppose), and b) not unioning the circles all at the same time, but instead in iteratively split groups.

I wanted to make the suggested benchmarks as well. However, since most tree-support classes worth benchmarking take the _entire slice-storage_ and we agreed on a time-window for this ticket, I'll park that for now.


src/utils/polygonUtils.cpp
```diff
@@ -1486,4 +1487,19 @@ void PolygonUtils::fixSelfIntersections(const coord_t epsilon, Polygons& thiss)
...
+    if (p.paths.size() < 8)
+    {
+        return p.unionPolygons();
+    }
+
+    Polygons a, b;
+    for (const auto& [i,path] : p.paths | ranges::views::enumerate)
+    {
+        ((i % 2 == 0) ? a : b).paths.push_back(path);
+    }
```
casperlamboo, line 1501: Ranges have a very nice util for this

```
std::vector<int> v {6,4,1,8,3};   
std::vector<int> even;
std::vector<int> odd;
 
ranges::partition_copy(v, ranges::back_inserter(even), 
                        ranges::back_inserter(odd), 
                        [] (int i) { return i % 2 == 0; });
    
std::cout << ranges::views::all(even) // [6,4,8]
          << ranges::views::all(odd); // [1,3]
```
Untested code, not sure if you think it's more readable but I imagine it looking looks something like
```suggestion
    ranges::partition_copy(
        p.paths | ranges::views::enumerate, 
        ranges::back_inserter(a), 
        ranges::back_inserter(b), 
        [] (auto& [i, _]) { return i % 2 == 0; }
    );
```

rburema: You don't have to sell me on it, I love functional programming :-D

Just though of this; Another thing we could do is reserve the sizes for the paths for a and b before the push-backs (or the inserters, in this case).

rburema: Could not make this work elegantly in the end.
- Structured bindings don't work in closures (yet?), apparently, so you can't use the `[i, _]` there.
- Since enumerate makes pairs of the things iterated over, the inserters for a and b should also have to accept pairs/tuples I think. Which then actually need to only put the relevant bit in one of the two.

I stripped all of that away and tried to do the predicate with the size of `a`, but that ended up crashing. I could maybe try with random again, to at least see that it works, but I'd like to keep this deterministic.

PR #1875, opened 2023-04-07
Prime/purge tower optimisation

wawanbreton:
# Description

When using a printer with multiple filaments, a prime/purge tower is usually required when switching from one to another. Depending on the configuration of the printer, we have multiple scenarios :
* Printer with multiple nozzles : the tower is used to ensure the nozzle is properly primed just before using it, so that the flow is ready when starting to print the actual part. Primed volume is usually very low.
* Printer with a single nozzle and many input filaments : the tower is used to purge the filament remaining inside the nozzle, and ensure that the color/material is right when starting to print the actual part. Purged volume can be very high.

And there are more technical variations to take care of :
* Some printers take a long time to switch filament, some of them even have to take the current tool down and hook up to a new one. In this case, changing filament should be proceeded as less as possible.
* Some materials don't stick well to others, so ensuring a clean tower is not always easy.

The current implementation of the prime tower is good enough for multiple nozzles, and also works for mixed nozzles, but it uses a lot of filament/time when this is not always required. The purpose of this improvement is to save time and filament, while still having a stable and robust tower in all situations.

Original discussion about the optimisation : https://github.com/Ultimaker/Cura/issues/6909
Recent improvement of the prime tower, which fixes some defects but uses even more filament : https://github.com/Ultimaker/Cura/issues/15078
I also found some issues about priming in mid-air, and some others complaining about the amount of used filament.

Associated Cura branch is here : https://github.com/wawanbreton/Cura/tree/optimized-prime-tower and PR: https://github.com/Ultimaker/Cura/pull/17971

## Proposed implementation
After some discussions, the proposed modifications are the following :
* Only prime/purge when this is strictly necessary
* When prime/purge is not necessary, print a sparse pattern which ensures that :
1. The tower remains strong, whatever its height
2. Whatever is printed on top of it will have a proper support

The proposed sparse pattern is the following :

It is kind of a "circular zigzag" which ensures that every circle printed on top of it will have minimum bridging and will then be printed properly. It is very fast to print and keeps a continuous dispensing flow. The maximum bridging distance has been made a parameter because it depends on the nozzle size and filament.

pikachu_slice
Peek 07-04-2023 17-59

## Type of change

* New CuraEngine feature: add prime tower modes to allow for printing the sparse layer when required
* New Cura feature: replace parameter "enable prime tower" with "prime tower mode", in which 3 different options are possible :
  * None: no prime tower at all 
  * Bucket: uses the sparse pattern when required, but never print with a filament over a different one
  * Sparse: implementation of the sparse pattern when required
* New Cura feature: added parameter "maximum bridging distance"

# How Has This Been Tested?

* I have made a few prints using a TiZYX EVY printer with a multi-material extruder (single nozzle, two filament inputs : direct drive+bowden). Only using PLA filaments. From what I can tell, the printed towers were very strong and saved a lot of filament+time.
* I have also made some basic tests with a UltiMaker S5: 
  * Double-PLA and "optimized" mode: looks very good !
  * PLA+PVA and "optimized-consistent" mode: PVA is quite dirty, but not more than with standard prime tower. This however confirms that the "optimized-consistent" mode is really required for this case. However, it may override the "default" mode completely. PLA+PVA configuration is for using support, and in this case we don't have many layers with a single material.

**Test Configuration**:
* Printer TiZYX EVY
* Printer UltiMaker S5
* PC on Ubuntu

## Discussions
The base of the implementation looks goods, but there is still a lot of work before it can be integrated :
- [x] Change post-wipe positions on the tower, which sometimes are happening over nothing due to the shape of the sparse pattern
- [x] Test the sparse pattern in various conditions (printers, filaments, ...)
- [x] Make some tests to see if the "optimized" algorithm could replace the actual one. That would make an easier configuration for the end-user.
- [x] Make some tests to see of the "optimized-consistent" algorithm is actually required. This needs specific filaments that I don't have.
- [x] Make tests with extruders having different lines widths (no printing, just generation of gcode)
- [x] Make tests with printer having more than 2 extruders (no printing, just generation of gcode)
- [x] Find the best combination of parameters to be set in the GUI for easy configuration (depends on previous tests)


wawanbreton:
Implementation is now fully working and clean. I will start some prints with this on UM printers to see if the tower shape is fine and strong enough.
Now we just need to refine the settings names and descriptions.

AstridLinka:
Merci du travail que tu fais sur cette Tour qui a vraiment besoin d'être optimisée surtout que la double extrusion devient de plus en plus populaire, je pense que ton modèle est très bien, avec en prime une optimisation pour les formats Y cyclops , et bien sur la fin de l'impression dans le vide de l'ext 2 grâce a ton système de Zigzag en cercle. J'ai pas ton talent et tes compétences,  mais j’espère vraiment que ta Tour soit un jour inclus dans une mise à jour de Cura , comme une avancée majeure, parceque si on prend mon exemple j'ai du changer de slicer et quitter Cura à cause de cette Tour qui faisait n'importe quoi ! Merci

wawanbreton:
@AstridLinka Merci pour cet encouragement :blush: effectivement les systèmes à multi-extrusion sont de plus en plus courants, et la tour d'origine est vraiment inefficace pour cette situation. J'ai mis un certain temps à comprendre qu'en fait elle n'avait pas du tout été conçue pour cela. J'espère que ce nouveau système sera plus efficace, et qu'il amènera à d'autres idées encore meilleures :smiley:
Entre-temps, j'ai rejoint UltiMaker et je peux donc faire un peu de lobbying en interne :laughing: mais je n'ai pas grand-chose à faire pour convaincre les gens que c'est une modification utile. Avec un peu de chance :crossed_fingers: ça sera intégré dans la prochaine version de Cura !

include/PrimeTower.h
```diff
@@ -40,7 +42,10 @@ class PrimeTower
     const unsigned int number_of_prime_tower_start_locations_ = 21; //!< The required size of \ref PrimeTower::wipe_locations
 
     MovesByExtruder prime_moves_; //!< For each extruder, the moves to be processed for actual priming.
+    std::map<size_t, std::map<size_t, Polygons>>
+        sparse_pattern_per_extruders_; //!< For each extruders combination, and for each actual extruder, the pattern to print on all layers where extruders are actually useless.
```
casperlamboo, line 46: From the variable name I would suspect the type to be `std::map<size_t, Polygons>`. I'm not sure from the comment why the type is `std::map<size_t, std::map<size_t, Polygons>>` instead. Can the comment be clarified a bit?

src/FffGcodeWriter.cpp
```diff
@@ -3714,12 +3756,15 @@ void FffGcodeWriter::setExtruder_addPrime(const SliceDataStorage& storage, Layer
...
-    if (! Application::getInstance().current_slice_->scene.current_mesh_group->settings.get<bool>("prime_tower_enable"))
+    if (! storage.primeTower.enabled_)
     {
         return;
     }
 
-    storage.primeTower.addToGcode(storage, gcode_layer, prev_extruder, gcode_layer.getExtruder());
+    LayerIndex layer_nr = gcode_layer.getLayerNr();
+    const std::vector<ExtruderUse>& extruder_order
+        = (layer_nr < 0) ? extruder_order_per_layer_negative_layers[extruder_order_per_layer_negative_layers.size() + layer_nr] : extruder_order_per_layer[layer_nr];
```
casperlamboo, line 3766: I'm a bit puzzled by this line; Am I correct to assume that the negative (raft) layers are appended to the end of the `extruder_order_per_layer_negative_layers` vec; e.g. if we were to map the values of the `extruder_order_per_layer_negative_layers` to the layernr we would get something like this (assume we have 1000 layers)
```
map(layer => layer.layer_nr,extruder_order_per_layer_negative_layers) == [0, 1, 2, 3, ... 998, 999, 1000, -3, -2, -1]
```

include/PrimeTower.h
wawanbreton: I tried to do that already, but that was a difficult exercise :grimacing: will try something else

src/FffGcodeWriter.cpp
wawanbreton: No they are not appended, we have two different vectors: `extruder_order_per_layer_negative_layers` which contains only the raft layers, and `extruder_order_per_layer` which contains only the model layers. This was actually the previously existing behavior.

wawanbreton:
@casperlamboo I have seen some minor issues when using a raft. I suggest we can use the actual version and solve them later, if needed.

casperlamboo (approved):

jellespijker:
Hi @wawanbreton,

Thank you for taking the time to contribute. We're in the progress of updating our Way-Of-Working with regards to PR's so I'm adding some label(s). That way we can prioritize our work-load and give your PR the love it deserves.

It's on the backlog and our top-developers are working on this PR as we speak

include/PrimeTower.h
```diff
@@ -42,8 +42,15 @@ class PrimeTower
     const unsigned int number_of_prime_tower_start_locations_ = 21; //!< The required size of \ref PrimeTower::wipe_locations
 
     MovesByExtruder prime_moves_; //!< For each extruder, the moves to be processed for actual priming.
-    std::map<size_t, std::map<size_t, Polygons>>
-        sparse_pattern_per_extruders_; //!< For each extruders combination, and for each actual extruder, the pattern to print on all layers where extruders are actually useless.
+
+    /*
+     *  The first index is a bitmask representing an extruder combination, e.g. 0x05 for extruders 1+3.
+     *  The second one is the used extruder index, e.g. 1
```
casperlamboo, line 48:
```suggestion
     *  The second index is the used extruder index, e.g. 1
```

casperlamboo:
All looks really nice, love the feature! Thanks 😄

AstridLinka:
Ha Merci quelle bonne nouvelle, tu vas être dans la version 5.7 , tu me l'avais dit que ça allé possiblement se faire, j'ai vu ta belle tour dans la version bêta! Bien joué Bravo, tout ton travail sur la tour c'est bien mérité , je te félicite , si je vois des quouack  je te le dirai .

wawanbreton:
@AstridLinka Merci beaucoup :wink: effectivement ça a l'air de bien marcher, ces dernières semaines j'ai enchaîné les impressions bicolores et ça fait plaisir. Mais je suis déjà en train de réfléchir à de nouvelles améliorations, j'espère que j'aurai le temps pour les ajouter dans la 5.8 :smile:

AstridLinka:
Petit retour sur ta tour , j'ai eu aucun problème avec son utilisation elle a marché direct et marche très bien, j'ai vu que tu l'as encore améliorée dans la v5.8beta j'ai pas encore utilisé ,  j'ai 2 suggestions à faire cependant , si tu pouvais faire la même chose mais en version mur sur les bords,  car mine de rien la tour grignote pas mal d'espace sur le plateau et ma 2 eme demande c'est plus sur le principe de simplifier cura pour les gens qui utilisent 2 extrudeurs et une seule buse  sur l'histoire du partage de température de la buse , y avoir dans le menu double extrusion les options qui sont dans le plugin "printer setting" , que chaque options regroupent les paramètres qu'il faut pour son utilisation ex une liste déroulante Chimére, cyclops, idex ,etc et ça règle automatiquement les paramètres qu'ils faut , et ça éviterai de devoir télécharger le plugin Printer setting  et cocher les 2 cases buses partagée qui faut pour une cyclops , car si tu sais pas qu'il faut qu'il faut télécharger ce plugin tu saura pas faire marcher ta bécane en double ext sur cura et les gens peuvent fuir cura pour un détail, moi j'ai galéré à savoir comment faire , j'ai perdu beaucoup de temps avec ça , ça va j'ai de la patience mais celui qui a pas de patience il va utiliser un autre logiciel et c'est dommage enfin voila, Bravo encore pour tout ton travail et ta fabuleuse tour qui a surement sauvé cura

wawanbreton:
Merci pour le retour :smiley: 
Effectivement dans Cura 5.8 j'ai encore pas mal amélioré la tour. Elle devrait utiliser moins de filament inutilement, ce qui dans certains cas peut faire une tour beaucoup moins large, et donc gagner de la place sur le plateau. La prochaine étape sera probablement d'utiliser le remplissage de l'objet pour purger, de manière à réduire encore la purge inutile dans la tour. Pour le placement sur le plateau, ça serait bien aussi de pouvoir la mettre dans un creu à l'intérieur de l'objet, ça pourrait sauver quelques autres cas.
Pour la gestion des 2 extrudeurs, Cura a été créé comme ça initialement puisque la grosse majorité des imprimantes multi-couleur ont plusieurs buses, même si ça tend à changer ces dernières années. Par contre, changer ce mécanisme va être compliqué, et à vrai dire je vois mal comment on pourrait faire autrement, puisqu'il y aura toujours plusieurs matériaux à gérer, donc avoir plusieurs buses "virtuelles" me paraît assez logique. À vrai dire ça ne m'a jamais posé de souci. C'est peut-être plus perturbant avec les systèmes qui ont genre 8 filaments en entrée, mais encore une fois je ne vois pas bien comment améliorer ça. Je vais le garder en tête et voir si ça revient, mais dans tous les cas ça impliquerait un changement énorme donc j'ai bien peur que le jeu n'en vaille pas la chandelle.
En revanche, je suis en train de pousser de plus en plus pour revoir la gestion des paramètres qui sont partagés ou pas. Je fais très souvent des erreurs de configuration parce que je suis sur le mauvais extrudeur, et je ne suis pas le seul. Donc j'espère bien que cela changera un jour prochain, et peut-être que ça sera suffisant.

PR #1885, opened 2023-06-02
[CURA-10500] simplify fixes

rburema:
Things that where probably causing jagged edges:
- Bridging (which is on for first-party printers) could chop up line segments almost to the 5 micron level.
- The iteration when simplifying wasn't exhaustive. Especially, there was a check to see if we where still below 2 or 3 for polylines and polygons respective, but the corresponding remove didn't always. You'll note that we _also_ _completely_ iterate a number of times. If it's determined that this diminishes the deviation checks too much, we can probably just leave it at the > 0 check (instead of > #min_nr_of_vertices) for `by_importance`.
- We where also not sure about the squared version of the line-distance, since those give different answers when a square-root was taken compared to the non-squared version.

At the moment I think this is enough to not introduce jaggies into polygons or polylines _that are not themselves fine details_. There are still some small lines that ping the checker, bit those seem a bunch of small polylines (which have fixed endpoints) that you'd need to have one or more special cases for, since you either _do_ need to remove the end-points then, or accept that in those cases some deviation is accepted (like throwing away paths that are not by much, but still, above the minimum length).

### DIFF
--- a/include/utils/Simplify.h
+++ b/include/utils/Simplify.h
@@ -4,6 +4,7 @@
 #ifndef UTILS_SIMPLIFY_H
 #define UTILS_SIMPLIFY_H
 
+#include "AABB.h"
 #include "polygon.h"
 #include "ExtrusionLine.h"
 #include "linearAlg2D.h" //To calculate line deviations and intersecting lines.
@@ -133,6 +134,27 @@ protected:
      */
     constexpr static coord_t min_resolution = 5; //5 units, regardless of how big those are, to allow for rounding errors.
 
+    template<typename Polygonal>
+    bool detectSmall(const Polygonal& polygon, const coord_t& min_size) const
+    {
+        if (polygon.size() < min_size) //For polygon, 2 or fewer vertices is degenerate. Delete it. For polyline, 1 vertex is degenerate.
+        {
+            return true;
+        }
+        if (polygon.size() == min_size)
+        {
+            const auto a = getPosition(polygon[0]);
+            const auto b = getPosition(polygon[1]);
+            const auto c = getPosition(polygon[polygon.size() - 1]);
+            if (std::max(std::max(vSize2(b - a), vSize2(c - a)), vSize2(c - b)) < min_resolution * min_resolution)
+            {
+                // ... unless they are degenetate.
+                return true;
+            }
+        }
+        return false;
+    }
+
     /*!
      * The main simplification algorithm starts here.
      * \tparam Polygonal A polygonal object, which is a list of vertices.
@@ -144,7 +166,7 @@ protected:
     Polygonal simplify(const Polygonal& polygon, const bool is_closed) const
     {
         const size_t min_size = is_closed ? 3 : 2;
-        if(polygon.size() < min_size) //For polygon, 2 or fewer vertices is degenerate. Delete it. For polyline, 1 vertex is degenerate.
+        if (detectSmall(polygon, min_size))
         {
             return createEmpty(polygon);
         }
@@ -161,8 +183,7 @@ protected:
         std::priority_queue<std::pair<size_t, coord_t>, std::vector<std::pair<size_t, coord_t>>, decltype(comparator)> by_importance(comparator);
 
         Polygonal result = polygon; //Make a copy so that we can also shift vertices.
-        size_t current_removed = 0;
-        do
+        for (int64_t current_removed = -1; (polygon.size() - current_removed) > min_size && current_removed != 0;)
         {
             current_removed = 0;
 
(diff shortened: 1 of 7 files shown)


nallath (approved):

src/WallsComputation.cpp
```diff
@@ -72,7 +72,7 @@ void WallsComputation::generateWalls(SliceLayerPart* part, SectionType section_t
         part->wall_toolpaths = wall_tool_paths.getToolPaths();
         part->inner_area = wall_tool_paths.getInnerContour();
     }
-    part->print_outline = part->outline;
+    part->print_outline = Simplify(settings).polygon(part->outline);
```
casperlamboo, line 75: I'm not super sure about this simplify; later in the pipeline we already simplify these same polygons

https://github.com/Ultimaker/CuraEngine/blob/7b90564857e4da887844b73ea7f53eb1fd5a7adc/src/WallToolPaths.cpp#L74

https://github.com/Ultimaker/CuraEngine/blob/7b90564857e4da887844b73ea7f53eb1fd5a7adc/src/WallsComputation.cpp#L116

We are simplifying the same polygon in multiple places. This doesn't hurt per-se but it does give me the feeling of being less in control of the process.

casperlamboo: I don't think we should change this now, but I think we should be aware in the future that these simplify functions are executed in multiple places, and removing one simplify doesn't un-simplify the per-se?

include/utils/Simplify.h
```diff
@@ -160,32 +182,41 @@ class Simplify
...
+                vertex_importance = importance(result, to_delete, vertex.first, is_closed);
+                if (vertex_importance != vertex.second)
+                {
+                    by_importance.emplace(vertex.first, vertex_importance); //Re-insert with updated importance.
+                    continue;
+                }
+
+                if (vertex_importance <= max_deviation * max_deviation)
+                {
+                    current_removed += remove(result, to_delete, vertex.first, vertex_importance, is_closed) ? 1 : 0;
```
casperlamboo, line 218:
```suggestion
                    auto removed = remove(result, to_delete, vertex.first, vertex_importance, is_closed);
                    if (removed)
                    {
                        current_removed +=  1;
                    }
```

include/utils/Simplify.h
```diff
@@ -160,32 +182,41 @@ class Simplify
...
-        while(by_importance.size() > min_size)
+        for (int64_t current_removed = -1; (polygon.size() - current_removed) > min_size && current_removed != 0;)
         {
-            std::pair<size_t, coord_t> vertex = by_importance.top();
-            by_importance.pop();
-            //The importance may have changed since this vertex was inserted. Re-compute it now.
-            //If it doesn't change, it's safe to process.
-            vertex_importance = importance(result, to_delete, vertex.first, is_closed);
-            if(vertex_importance != vertex.second)
+            current_removed = 0;
```
casperlamboo, line 188: We need to reset the `by_importance` queue as we are re-iterating over the same data structure.
```suggestion
            current_removed = 0;
            by_importance.clear();
```

casperlamboo (changes requested):

include/utils/Simplify.h
```diff
@@ -160,32 +182,41 @@ class Simplify
...
-        {
-            const coord_t vertex_importance = importance(polygon, to_delete, i, is_closed);
-            by_importance.emplace(i, vertex_importance);
-        }
-
-        //Iteratively remove the least important point until a threshold.
         Polygonal result = polygon; //Make a copy so that we can also shift vertices.
-        coord_t vertex_importance = 0;
-        while(by_importance.size() > min_size)
+        for (int64_t current_removed = -1; (polygon.size() - current_removed) > min_size && current_removed != 0;)
```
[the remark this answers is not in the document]
casperlamboo: https://github.com/Ultimaker/CuraEngine/pull/1885/commits/ca2a7945f798fe7e3874df618376fd227bd10f77

include/utils/Simplify.h
rburema: We don't really need to, since the thing is either empty or has few enough vertices that it'll stop immediately. It can't hurt though!

include/utils/Simplify.h
rburema: Not sure about this one. I like the compactness of the current code (too much lines is also less readable after all), but I'm not really against the change, because in a way it does make it more clear.

src/WallsComputation.cpp
rburema: Agreed. I put this here just to be sure to be honest.

casperlamboo (approved):

PR #1891, opened 2023-06-25
CURA 10724 smooth operator

jellespijker:
# Problem

The simplification of polygons sometimes leaves small segments with large differences in angles compared to the segments before and after them. This occurs because the simplification algorithm is over-constrained, and the solution space doesn't diverge to an optimal point, but to a plateau. This means that smaller segments won't be filtered out since doing so would violate other criteria, such as removing too much area, etc.

Printers with Marlin 1.x seem to handle the resulting toolpaths better than smooth motion planners such as Klipper. These printers see a sudden change in direction and will try to adhere to the requested toolpath as much as possible. Because that toolpath is now suddenly requesting a change in direction (for a very short length), the head will slow down to make the corner, but the extrusion process can't adjust at the same speed, resulting in over-extrusion at that spot.

# Fix
Since we can reasonably assume that such a sudden change in toolpath direction over a short distance, compared to the direction of the toolpath before and after, does not belong to the actual intended shape of the sliced model, we can adjust the polygons in favour of the intent and smooth motion printers.

We loop over each closed polygon, which is used to generate the Aranche toolpaths, and examine 3 segments (4 points) each time: `AB` -> `BC` -> `CD`. Sudden small segments `BC` (< maximum resolution) that deviate more than a certain angle compared to a fluid motion angle (wall transition angle) between the preceding segment `AB` or subsequent `CD` are smoothed out. This is done by either shifting point `B` or `C` towards `A` or `D` by half the maximum resolution's distance or removing them altogether if there is no room to shift them.

It also introduces some `concepts` and `std::get` functions which can be used to obtain coordinates from different Point and Junction types by either specifying the index or letter: X, Y or x, y, z

Peek 2023-06-30 16-45

TODO:
- [x] Work with Arachne Types
- [X] Apply the smoothing sparsely (Arachne Paths creation)
- [X] Write UnitTests
- [x] Fix failing integration test
- [X] User testing (build beta.2)
- [X] Add documentation in the ticket for future reference

## Type of change

# How Has This Been Tested?

- [X] Asked user experiencing this issue to test
- [X] Analyzed with Scripta

**Test Configuration**:
* Operating System:

# Checklist:


include/utils/actions/smooth.h
```diff
@@ -0,0 +1,151 @@
...
+    constexpr auto operator()(const std::integral auto max_resolution, const std::floating_point auto fluid_angle) const
+    {
+        return ranges::make_action_closure(ranges::bind_back(smooth_fn{}, max_resolution, fluid_angle));
+    }
+
+    template<class Rng>
+    requires ranges::forward_range<Rng>&& ranges::sized_range<Rng>&& ranges::erasable_range<Rng, ranges::iterator_t<Rng>, ranges::sentinel_t<Rng>> && (utils::point2d<ranges::range_value_t<Rng>> || utils::junctions<Rng>)
+    constexpr auto operator()(Rng&& rng, const std::integral auto max_resolution, const std::floating_point auto fluid_angle) const
+    {
+        const auto size = ranges::distance(rng) - 1; // TODO: implement for open paths! The value `-1` is for closed Paths, if open then subtract `0`
```
casperlamboo, line 40: Do we still want to support open paths, or is this TODO redundant now?

include/utils/actions/smooth.h
```diff
@@ -0,0 +1,151 @@
...
+
+struct smooth_fn
+{
+    constexpr auto operator()(const std::integral auto max_resolution, const std::floating_point auto fluid_angle) const
+    {
+        return ranges::make_action_closure(ranges::bind_back(smooth_fn{}, max_resolution, fluid_angle));
+    }
+
+    template<class Rng>
+    requires ranges::forward_range<Rng>&& ranges::sized_range<Rng>&& ranges::erasable_range<Rng, ranges::iterator_t<Rng>, ranges::sentinel_t<Rng>> && (utils::point2d<ranges::range_value_t<Rng>> || utils::junctions<Rng>)
```
casperlamboo, line 37: nitpicky comment, add spaces around binary operators

```suggestion
    requires ranges::forward_range<Rng> && ranges::sized_range<Rng> && ranges::erasable_range<Rng, ranges::iterator_t<Rng>, ranges::sentinel_t<Rng>> && (utils::point2d<ranges::range_value_t<Rng>> || utils::junctions<Rng>)
```

include/utils/actions/smooth.h
```diff
@@ -0,0 +1,151 @@
...
+                                          const std::floating_point auto AB_magnitude, const std::floating_point auto BC_magnitude, const std::floating_point auto CD_magnitude) const noexcept
+    {
+        if (BC_magnitude > max_resolution / 10) // TODO: make dedicated front-end setting for this
+        {
+            return true;
+        }
+        const double cos_A = std::acos(cosAngle(A, B, C, AB_magnitude, BC_magnitude));
+        const double cos_B = std::acos(cosAngle(A, B, D, AB_magnitude, CD_magnitude));
+        const auto abs_angle = std::abs(cos_A - cos_B);
+        return abs_angle < fluid_angle;
```
casperlamboo, line 144: Why calculate the angle `angle(ab, bc)` and `angle(ab, bd)` independently instead of directly check the angle `angle(ab, cd)`?

Pseudo code would be something like
```
const auto ab = b - a;
const auto cd = d - c;
const cos_angle = dot(ab, bc) / (length(ab) * length(bc));
const is_sharp_angle = cos_angle < cos(fluid_angle);
```
Not that this reduces the number of `acos`/`cos` calculations dramastically as `cos(fluid_angle)` is constant; you only need to calculate it once.

tests/utils/SmoothTest.cpp
```diff
@@ -0,0 +1,35 @@
...
+#include "utils/IntPoint.h"
+#include "utils/actions/smooth.h"
+#include "utils/polygon.h"
+
+namespace cura
+{
+
+
+TEST(SmoothTest, TestSmooth)
+{
```
nallath, line 19: Needs a bit of documentation; Why is this a good test? What is it trying to do?  (nitpicking)

nallath (changes requested):

include/utils/actions/smooth.h
jellespijker: Can be removed

include/utils/actions/smooth.h
jellespijker: The idea behind this smoothing is that the angle of BC should be within a fluid motion path, from where it comes from to where it will be going.

jellespijker: Lets discuss behind a whiteboard

include/utils/actions/smooth.h
```diff
@@ -0,0 +1,151 @@
...
+        }
+
+        using coord_type = std::remove_cvref_t<decltype(std::get<"X">(*ranges::begin(rng)))>;
+        const auto allowed_deviation = static_cast<coord_type>(max_resolution * 2 / 3); // The allowed deviation from the original path
+        const auto smooth_distance = static_cast<coord_type>(max_resolution / 2); // The distance over which the path is smoothed
+
+        auto tmp = rng; // We don't want to shift the points of the ingoing range, therefor we create a temporary copy
+        auto windows = ranges::views::concat(ranges::views::single(ranges::back(tmp)), ranges::views::concat(tmp, tmp | ranges::views::take(4))) | ranges::views::addressof;
+
+        // Smooth the path, by moving over three segments at a time. If the middle segment is shorter than the max resolution, then we try shifting those points outwards.
```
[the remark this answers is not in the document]
jellespijker: Yes, we should have variables for these coming from the front-end, but I'm reluctant to add them this close to a release. For now these are sensible values, for 5.5 we want them in the front-end

CMakeLists.txt
```diff
@@ -12,6 +12,7 @@ option(ENABLE_TESTING "Build with unit tests" OFF)
 option(EXTENSIVE_WARNINGS "Build with all warnings" ON)
 option(ENABLE_MORE_COMPILER_OPTIMIZATION_FLAGS "Enable more optimization flags" ON)
 option(USE_SYSTEM_LIBS "Use the system libraries if available" OFF)
+option(RETARDED_APPLE_CLANG "Apple Clang <= 13 used" OFF)
```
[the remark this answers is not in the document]
jellespijker: Fine with changing it. I do like to note that the usage of "retarded" wasn't used in a degenerative context but actual because it actually covered the meaning of the variable quite well.

image

include/utils/types/get.h
```diff
@@ -0,0 +1,144 @@
...
+
+/**
+ * @brief Get a coordinate of a 2d point by character.
+ *
+ * @tparam C The character of the coordinate to get ('x' or 'X' for X-coordinate, 'y' or 'Y' for Y-coordinate).
+ * @param point The point to get the coordinate from.
+ * @return auto& A reference to the requested coordinate.
+ */
+template<cura::utils::CharRangeLiteral C>
+constexpr const auto& get(const cura::utils::point2d auto& point) noexcept
```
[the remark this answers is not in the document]
jellespijker: Not a favor of these specialized functions, rather reuse `std::get` such that things keep consistent. The character are just there to improve readability and show intent, I could make enum's for this.

tests/utils/SmoothTest.cpp
jellespijker: Test was used during debugging, for quick iterative cycles. I believe it also fails currently. In will create a couple actual useful test cases, tomorrow

include/utils/types/geometry.h
```diff
@@ -0,0 +1,156 @@
...
+
+/*!
+ * @concept point2d_tuple
+ * @brief Checks whether T is a 2D point represented as a tuple with two integral elements
+ * @tparam T The type to check
+ */
+template<typename T>
+concept point2d_tuple = requires(T t)
+{
+    requires std::is_same_v<T, std::tuple<typename std::tuple_element<0, T>::type, typename std::tuple_element<0, T>::type>>;
```
casperlamboo, line 32:
```suggestion
    requires std::is_same_v<T, std::tuple<typename std::tuple_element<0, T>::type, typename std::tuple_element<1, T>::type>>;
```

include/utils/types/geometry.h
```diff
@@ -0,0 +1,156 @@
...
+
+/*!
+ * @concept point3d_tuple
+ * @brief Checks whether T is a 3D point represented as a tuple with three integral elements
+ * @tparam T The type to check
+ */
+template<typename T>
+concept point3d_tuple = requires(T t)
+{
+    requires std::is_same_v<T, std::tuple<typename std::tuple_element<0, T>::type, typename std::tuple_element<0, T>::type, typename std::tuple_element<0, T>::type>>;
```
casperlamboo, line 77:
```suggestion
    requires std::is_same_v<T, std::tuple<typename std::tuple_element<0, T>::type, typename std::tuple_element<1, T>::type, typename std::tuple_element<2, T>::type>>;
```

casperlamboo (commented):
Should you test that the various tuple components is the same. I believe the current concept only test if the first element is the same as the first element

include/utils/types/geometry.h
jellespijker: This is intentional, this forces to have both elements the same type

include/utils/types/geometry.h
jellespijker: See above

include/utils/actions/smooth.h
jellespijker: Put this on the C&C board to be discussed, for now the current check is acceptable

Issue #1897, opened 2023-07-08
Incorrect printf format string in InfillTest causes test failure on 32-bit architectures


onitake:
**Application Version**

5.0.0 until current HEAD ( ef46801bffb289cdcc3ec58b6dfff78b2889644d )

**Platform**

Linux on i686 (or any 32-bit architecture)

**Qt**

N/A

**PyQt**

N/A

**Display Driver**

N/A

**Steps to Reproduce**

1. Compile CuraEngine for a CPU architecture with a 32-bit `size_t`
2. Run unit test `InfillTest`

**Actual Results**

```
...
5: 
5: [ FATAL ] /usr/include/gtest/internal/gtest-param-util.h:585:: Condition IsValidParamName(param_name) failed. Parameterized test name 'InfillTestParameters_P2_Z0_C0_L350__-4294967296' is invalid, in ./tests/InfillTest.cpp line 241
5: 
...
96% tests passed, 1 tests failed out of 26

Total Test time (real) =  10.85 sec

The following tests FAILED:
	  5 - InfillTest (Subprocess aborted)
Errors while running CTest
make[1]: *** [Makefile:94: test] Error 8
```

**Expected results**

The unit test succeeds.

**Additional Information**

The bug lies in the a `printf` format string, where `%lld` is incorrectly used to format a `size_t` argument.
InfillTest runs with very high numbers for this argument which will cause a rollover on 32 architectures if they're interpreted as signed, but `size_t` is unsigned.

To fix the issue, the correct format string for `size_t` should be used: `%zu`
Alternatively, changing `%lld` to `%llu` will also work in most cases.

This occurred during automatic unit testing of CuraEngine 5.0.0 in Debian: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1040252
But the bug is still present in current HEAD: https://github.com/Ultimaker/CuraEngine/blob/main/tests/InfillTest.cpp#L103

PR #1898, opened 2023-07-08
Replace size_t format string in InfillTest.cpp

onitake:
# Description

The format string in tests/InfillTest.cpp is not correct for the (last) `size_t` argument, causing test failure on 32-bit architectures.
This PR replaces the `%lld` with the correct C99 format string for `size_t`: `%zu`.

It's expected that this will work with every current C standard library used for CuraEngine.
If this is not the case and C90 C library compatibility is desired, a more conservative format string can be used, such as `%llu`. This was tested successfully on i686 with gcc 12/glibc 2.37, even though it's not 100% correct for a 32-bit `size_t`.

Fixes: #1897 

## Type of change

# How Has This Been Tested?

- [ ] Build on amd64 and i686 Linux
- [ ] Run unit test InfillTest

**Test Configuration**:
* Operating System: Debian trixie/sid
* amd64 (native) AND i686 (chroot)

# Checklist:

### DIFF
--- a/tests/InfillTest.cpp
+++ b/tests/InfillTest.cpp
@@ -57,7 +57,7 @@ public:
         , connect_polygons(connect_polygons)
         , line_distance(line_distance)
     {
-        name = fmt::format("InfillParameters_{d}_{d}_{d}_{d}", static_cast<int>(pattern), zig_zagify, connect_polygons, line_distance);
+        name = fmt::format("InfillParameters_{:d}_{:d}_{:d}_{:d}", static_cast<int>(pattern), zig_zagify, connect_polygons, line_distance);
     }
 };
 
@@ -89,7 +89,7 @@ public:
         , result_lines(std::move(result_lines))
         , result_polygons(std::move(result_polygons))
     {
-        name = fmt::format("InfillTestParameters_P{d}_Z{d}_C{d}_L{d}__{d}", static_cast<int>(params.pattern), params.zig_zagify, params.connect_polygons, params.line_distance, test_polygon_id);
+        name = fmt::format("InfillTestParameters_P{:d}_Z{:d}_C{:d}_L{:d}__{:d}", static_cast<int>(params.pattern), params.zig_zagify, params.connect_polygons, params.line_distance, test_polygon_id);
     }
 
     friend std::ostream& operator<<(std::ostream& os, const InfillTestParameters& params)
--- a/tests/InfillTest.cpp
+++ b/tests/InfillTest.cpp
@@ -8,6 +8,7 @@
 #include <gtest/gtest.h>
 #include <filesystem>
 #include <utility>
+#include <fmt/format.h>
 
 #include <scripta/logger.h>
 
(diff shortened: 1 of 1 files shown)


tests/InfillTest.cpp
```diff
@@ -100,7 +100,7 @@ class InfillTestParameters
         , result_polygons(std::move(result_polygons))
     {
         // FIXME: Once we are using spdlog as logger, we'll also use fmt::format() here, see CURA-8258.
-        name = makeName("InfillTestParameters_P%d_Z%d_C%d_L%lld__%lld", static_cast<int>(params.pattern), static_cast<int>(params.zig_zagify), static_cast<int>(params.connect_polygons), params.line_distance, test_polygon_id);
+        name = makeName("InfillTestParameters_P%d_Z%d_C%d_L%lld__%zu", static_cast<int>(params.pattern), static_cast<int>(params.zig_zagify), static_cast<int>(params.connect_polygons), params.line_distance, test_polygon_id);
```
jellespijker, line 103:
```suggestion
        name = makeName(fmt::format("InfillTestParameters_P{}_Z{}_C{}_L{}__{}", static_cast<int>(params.pattern), static_cast<int>(params.zig_zagify), static_cast<int>(params.connect_polygons), params.line_distance, test_polygon_id));
```
An untested suggestion which would require `#include <fmt/format.h>` and the additional removal of the now redudant TODO on L102, also need to double check if this requires specific formatters

jellespijker (changes requested):
Hi @onitake I'm just curious. I noticed you asking some specific Linux questions libCharon now a PR for 32bit hardware. I'm guessing you're a package maintainer for one of the distro's.

We recently added fmt https://fmt.dev/latest/index.html as one of the dependencies. Since the TODO above states that this should be changed anyhow once we use fmt I would prefer it if this PR uses the fmt lib.

onitake:
> Hi @onitake I'm just curious. I noticed you asking some specific Linux questions libCharon now a PR for 32bit hardware. I'm guessing you're a package maintainer for one of the distro's.

That is correct, I'm one of the maintainers of the Debian packages: https://tracker.debian.org/pkg/cura-engine
Maintenance got a bit stuck lately due to lack of time and the switch from CMake to Conan, but we're intent to get at least 5.0 maintained in vanilla Debian.

Since Debian still includes some 32-bit release architectures, we need to ensure compatibility if possible.

> We recently added fmt https://fmt.dev/latest/index.html as one of the dependencies. Since the TODO above states that this should be changed anyhow once we use fmt I would prefer it if this PR uses the fmt lib.

Ok, I can look into that. I don't know that fmt library, but it doesn't sounds straight-forward.

jellespijker:
basically the same syntax as the Python format strings

onitake:
Ok, so I replaced `makeName` with `fmt::format`.
The format string explicitly uses decimal format `{:d}`, so I dropped the `static_cast`s - except for the `EFillMethod` type.

It looks like fmtlib won't automatically cast enum classes to int starting from v7, to be in line with C++20 semantics.
You either need to add a custom formatter (as in https://fmt.dev/latest/api.html#formatting-user-defined-types) or cast it to the underlying type yourself. See https://github.com/fmtlib/fmt/issues/2704 for an example and the reasoning.

jellespijker (approved):
LGTM;

Thank you for you contribution and finishing my todo.

PR #1901, opened 2023-07-11
[CURA-10732] Create github-action to perform and apply linting.

rburema:


jellespijker:
why not reuse the workflow from printer-linter as suggested in the ticket?

Only real difference is that you do a clang-tidy execution instead of running printer-linter: https://github.com/Ultimaker/Cura/blob/63e78d313c66588b8425d08a948fe16e3c2f2e00/.github/workflows/printer-linter-pr-diagnose.yml#L42


The poat of the result is then done in this workflow (leasts rights)

jellespijker (changes requested):
See comments in Jira ticket.

PR #1902, opened 2023-07-12
[CURA-10732] Add clang-format and clang-linting to push/PR actions.

rburema:


jellespijker (approved):
LGTM;

Gonna create some branches from these to test the actual workflow but I believe this should work

PR #1904, opened 2023-07-14
Cura 10728 skirt height for skirt outline

saumyaj3:
# Description

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* Operating System:

# Checklist:

### DIFF
--- a/src/SkirtBrim.cpp
+++ b/src/SkirtBrim.cpp
@@ -378,6 +378,7 @@ Polygons SkirtBrim::getFirstLayerOutline(const int extruder_nr /* = -1 */)
     {
         reference_extruder_nr = extruder_nr;
     }
+    Settings& extruder_settings = Application::getInstance().current_slice->scene.extruders[reference_extruder_nr].settings;
     const int primary_line_count = line_count[reference_extruder_nr];
     const bool external_only = adhesion_type == EPlatformAdhesion::SKIRT || external_polys_only[reference_extruder_nr]; // Whether to include holes or not. Skirt doesn't have any holes.
     const LayerIndex layer_nr = 0;
@@ -387,7 +388,7 @@ Polygons SkirtBrim::getFirstLayerOutline(const int extruder_nr /* = -1 */)
         const bool skirt_around_prime_tower_brim = storage.primeTower.enabled && global_settings.get<bool>("prime_tower_brim_enable");
         const bool include_prime_tower = ! skirt_around_prime_tower_brim; // include manually otherwise
 
-        const int skirt_height = global_settings.get<int>("skirt_height");
+        const int skirt_height = extruder_settings.get<int>("skirt_height");
         first_layer_outline = Polygons();
         for (int i_layer = layer_nr; i_layer <= skirt_height; ++i_layer)
         {
--- a/src/SkirtBrim.cpp
+++ b/src/SkirtBrim.cpp
@@ -391,8 +391,13 @@ Polygons SkirtBrim::getFirstLayerOutline(const int extruder_nr /* = -1 */)
         int skirt_height = 0;
         for (const auto& extruder : Application::getInstance().current_slice->scene.extruders)
         {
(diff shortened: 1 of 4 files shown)


nallath (approved):

src/SkirtBrim.cpp
```diff
@@ -387,15 +387,22 @@ Polygons SkirtBrim::getFirstLayerOutline(const int extruder_nr /* = -1 */)
...
+        int skirt_height = 0;
+        for (const auto& extruder : Application::getInstance().current_slice->scene.extruders)
         {
-            first_layer_outline =
-                first_layer_outline.unionPolygons
-                (
-                    storage.getLayerOutlines(i_layer, include_support, include_prime_tower, external_only, extruder_nr)
-                );
+            skirt_height = std::min(std::max(skirt_height, extruder.settings.get<int>("skirt_height")), static_cast<int>(storage.print_layer_count));
+        }
```
casperlamboo, line 395: I have a slight preference for something like this:
```suggestion
            skirt_height = std::max(skirt_height, extruder.settings.get<int>("skirt_height"));
        }
        skirt_height = std::min(skirt_height, static_cast<int>(storage.print_layer_count));
```
It is more readable in my opinion as you sequence the operations more. First you calculate the max skirt height of the various extrudes, then you cap this height to the layer count.

src/SkirtBrim.cpp
```diff
@@ -387,15 +387,22 @@ Polygons SkirtBrim::getFirstLayerOutline(const int extruder_nr /* = -1 */)
...
-        for (int i_layer = layer_nr; i_layer <= skirt_height; ++i_layer)
+        int skirt_height = 0;
+        for (const auto& extruder : Application::getInstance().current_slice->scene.extruders)
         {
-            first_layer_outline =
-                first_layer_outline.unionPolygons
-                (
-                    storage.getLayerOutlines(i_layer, include_support, include_prime_tower, external_only, extruder_nr)
-                );
+            skirt_height = std::min(std::max(skirt_height, extruder.settings.get<int>("skirt_height")), static_cast<int>(storage.print_layer_count));
```
casperlamboo, line 394: From the documentation: 
```
* \param extruder_nr The extruder for which to get the outlines. -1 to include outliens for all extruders
```
We should thus only include the skirt height of all extruders when this variable is set to `-1`. Taking some inspiration from this function
https://github.com/Ultimaker/CuraEngine/blob/14bb2f2959c06def09be91ae264c9681c44e8100/src/sliceDataStorage.cpp#L285
I would suggest to add something like
```suggestion
            if (extruder_nr == -1 || extruder_nr == extruder.extruder_nr)
            {
                skirt_height = std::min(std::max(skirt_height, extruder.settings.get<int>("skirt_height")), static_cast<int>(storage.print_layer_count));
            }
```

src/SkirtBrim.cpp
```diff
@@ -387,15 +387,22 @@ Polygons SkirtBrim::getFirstLayerOutline(const int extruder_nr /* = -1 */)
...
+        {
+            for (const auto& extruder : Application::getInstance().current_slice->scene.extruders)
+            {
+                first_layer_outline =
+                    first_layer_outline.unionPolygons
+                    (
+                        storage.getLayerOutlines(i_layer, include_support, include_prime_tower, external_only, extruder.extruder_nr)
+                    );
+            }
         }
```
casperlamboo, line 406: I'm not sure if this is needed; we have an `extruder_nr` parameter, the documentation for this parameter reads
```
     * \param extruder_nr The extruder for which to get the outlines. -1 to include outliens for all extruders
```
So we should only include the layer_outline for all extruders if `extruder_nr` is set to `-1`. When executing `storage.getLayerOutlines` with this `-1` argument should already return the outline of all extruders.

casperlamboo (approved):

PR #1907, opened 2023-07-18
Brim lines generation was terminated because of break statement.

saumyaj3:
CURA-10454

# Description

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* Operating System:

# Checklist:

### DIFF
--- a/src/SkirtBrim.cpp
+++ b/src/SkirtBrim.cpp
@@ -221,7 +221,10 @@ std::vector<coord_t> SkirtBrim::generatePrimaryBrim(std::vector<Offset>& all_bri
         }
         SkirtBrimLine& output_location = storage.skirt_brim[offset.extruder_nr][offset.inset_idx];
         coord_t added_length = generateOffset(offset, covered_area, allowed_areas_per_extruder, output_location);
-        if (! added_length)
+
+        const Settings& global_settings = Application::getInstance().current_slice->scene.current_mesh_group->settings;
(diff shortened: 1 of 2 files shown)


jellespijker (commented):
LGTM; still need to tom component test this

src/SkirtBrim.cpp
```diff
@@ -221,7 +221,10 @@ std::vector<coord_t> SkirtBrim::generatePrimaryBrim(std::vector<Offset>& all_bri
...
         SkirtBrimLine& output_location = storage.skirt_brim[offset.extruder_nr][offset.inset_idx];
         coord_t added_length = generateOffset(offset, covered_area, allowed_areas_per_extruder, output_location);
-        if (! added_length)
+
+        const Settings& global_settings = Application::getInstance().current_slice->scene.current_mesh_group->settings;
+        const bool support_brim_enable = global_settings.get<bool>("support_brim_enable");
+        if (! added_length && !support_brim_enable)
         { // no more place for more brim. Trying to satisfy minimum length constraint with generateSecondarySkirtBrim
             break;
         }
```
casperlamboo, line 230:
```suggestion
```
I would remove this piece of code al together. The added_length check was only there as an optimization improvement. Since we have found a counter example this check is giving false positives, and thus breaking while it shouldn't. I believe this behavior is unrelated to support brim lines, and it is only a coincidence we have found it while this setting was enabled. For this reason I believe we should remove this break condition entirely.

PR #1910, opened 2023-07-19
[CURA-10805] unify plugin slot calls

rburema:
Clean up the different ways which basically all come down to just 'slot identification' by unifying all (at least to the 'outside world' -- the non-plugin parts) by always using SlotID as an identifier.

### DIFF
--- a/benchmark/simplify_benchmark.h
+++ b/benchmark/simplify_benchmark.h
@@ -64,7 +64,7 @@ BENCHMARK_DEFINE_F(SimplifyTestFixture, simplify_slot_noplugin)(benchmark::State
         Polygons simplified;
 		for (const auto& polys : shapes)
 		{
-			benchmark::DoNotOptimize(simplified = slots::instance().invoke<plugins::slot_simplify>(polys, MM2INT(0.25), MM2INT(0.025), 50000));
+			benchmark::DoNotOptimize(simplified = slots::instance().invoke<plugins::v0::SlotID::SIMPLIFY_MODIFY>(polys, MM2INT(0.25), MM2INT(0.025), 50000));
 		}
 	}
 }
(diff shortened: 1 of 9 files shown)


jellespijker (changes requested):
I get why the changes in this PR are the way that they are now. There are however a couple of changes introduced which I think won't improve maintainability of this code.

The biggest concerns I have are with the number of changes that will be needed if we add a new slot now. These locations are scattered over multiple files and aren't really self explanatory. Which probably happened due to the way broadcast.h handles this already. Which should also probably be improved imo (and yes I'm the one who introduced that file and I'm not proud of it ;-) )

I also think that at the very least somethings can be simplified a lot already if the gRPC definitions don't send the requested broadcast subscriptions as strings but as SlotID's removing the need for the `CharRangeLiteral` altogether.

Another thing we needed to implement in this ticket was the need to enable plugins with pure listing capabality. Plugins that simply listen to a Broadcast and don't implement modify/generate. Yes, you're correct this additional requirement wasn't in the description of the ticket and wasn't communicated to you. This was my oversight. I should have added that to the description, after you picked up this work. CURA-10805 was mainly created to tidy up the work after a couple of tickets that build atop of each other and some comments that @nallath made about broadcasts being treated the same way as actual slots.

Given the additional requirement and the worries I have about maintainability I did a quick test if it possible to limit the extra maintenance and since the plugin system is a rather complex beast I had to write some stuff out. At a first glance it seems possible. So I would suggest that we two sit together comming Tuesday to see if we can get this ticket implemented in such away that it is robust, quick and maintainable.

See alternative approach PR:
- https://github.com/Ultimaker/CuraEngine/pull/1913
- https://github.com/Ultimaker/CuraEngine_grpc_defintions/pull/9

rburema:
Closed in favour of these two:
- https://github.com/Ultimaker/CuraEngine/pull/1913
- https://github.com/Ultimaker/CuraEngine_grpc_defintions/pull/9

PR #1918, opened 2023-07-27
Add string-switch type and integrate into settings retrieval

jellespijker:
This commit introduces a new type 'string-switch' to handle string comparisons in a switch-case format. This is achieved by employing a DJB2a-based hash function to convert each string to a unique, corresponding enum value. The primary motivation behind this was to refactor the existing settings retrieval process, which was heavily reliant on long if-else constructs. By replacing these with more manageable switch statements, the code readability and maintainability is significantly improved. This change also facilitates the addition of new settings by simply extending existing enums and the related switch statements.

Contributes to CURA-10702

### DIFF
--- a/include/utils/types/string_switch.h
+++ b/include/utils/types/string_switch.h
@@ -10,7 +10,7 @@ namespace cura::utils
 {
 
 // Source: https://learnmoderncpp.com/2020/06/01/strings-as-switch-case-labels/
-inline constexpr uint64_t hash_djb2a(const std::string_view value)
+constexpr inline uint64_t hash_djb2a(const std::string_view value)
 {
     uint64_t hash{ 5381 };
     for (unsigned char c : value)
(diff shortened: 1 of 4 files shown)


include/utils/types/string_switch.h
```diff
@@ -0,0 +1,41 @@
...
+{
+    constexpr uint64_t plugin_namespace_sep_location{ 6 };
+    if (value.size() > plugin_namespace_sep_location && value.at(plugin_namespace_sep_location) == ':')
+    {
+        return hash_djb2a("plugin");
+    }
+    return hash_djb2a(value);
+}
+
+constexpr inline auto operator""_sw(const char* str, size_t len)
```
rburema, line 33: In the above functions, you use `inline constexpr`, here you use `constexpr inline`. Pick a side, will ya? ;-)

rburema (approved):
Just reviewed the code, haven't tested this. Look great! Fan of this. Something like this should be in the std, really.

I was going to suggest maybe even putting this outside of the namespaces, since it seems like basic functionality, and we wouldn't have to bother with the 'using' directive before each switch-statement. However, since we will probably not be using this much outside of settings-loading (because why would we have enum like stuff stored as strings otherwise), I think this is fine.

I don't think we have to worry about performance; if anything, the old way is slower.

include/utils/types/string_switch.h
jellespijker: I will get my changes `inline`

PR #1923, opened 2023-08-04
[CURA-10670] Re-enable Small Top/Bottom Width

rburema:
Previously disabled just before the 5.4 release, now fixed and re-enabled.

See also frontend PR here: https://github.com/Ultimaker/Cura/pull/16384

Also fixes some edge cases:
- When a thick circular curve is close to the small area width, the polygonal nature of the models would make it so that it would alternate between small skin areas and normal top-fill areas within that arc. Fix this by removing small holes before doing the second part of the morphological operation (either open or close, depending on wether you view it from the inner contour or the small infill areas point of view).
- Both areas added together should completely cover the complete 'original' inner contour. Yet only one of them was simplified (at this juncture at least). In fact, there are paths through the code where the inner contour doesn't get simplified at all after this point. If it as in a simplified state before we start doing the small areas procedure, as it likely was in a lot of cases, then this could introduce new issues.


jellespijker:
Please check the GCodeAnalyzer results: https://ultimaker.github.io/CuraEngine/dev/gcodeanalyzer/

and the comments on: https://github.com/Ultimaker/CuraEngine/commit/9890a0174316f63b1e66dbcc234fd521d5a7fb53#comments

casperlamboo (changes requested):

casperlamboo (approved):

PR #1926, opened 2023-08-08
Refactor numeric types for better type safety and reduced code repeti…

jellespijker:
…tion

Numeric data types in the codebase are now defined as a numeric facade, providing greater type safety and reducing code repetition. This encompasses the `Acceleration`, `Velocity`, `Ratio`, and `LayerIndex` types.

This change updates the constructor and assignment operators, and ensures the facade can be used in a consistent manner with standard arithmetic types. It also supports user-defined type conversions to provide interoperability with built-in types.

In addition, correct type suffixes are added to numeric literals in a number of places to avoid potential bugs due to unexpected promotions or type coercions.

As part of this refactor, we also standardized the category of numeric types applied throughout the codebase, ensuring compatibility with these new types.

This refactor is needed to prep the GCodePaths/GCodePathConfigs for the plugin system. Since the changes for this refactor all over the placed, decided to make a seperate branch against main.

Contributes to CURA-10466

# Benchmark
## This PR
image
## main
image

### DIFF
--- a/include/settings/types/Ratio.h
+++ b/include/settings/types/Ratio.h
@@ -1,8 +1,11 @@
-//Copyright (c) 2018 Ultimaker B.V.
-//CuraEngine is released under the terms of the AGPLv3 or higher.
+// Copyright (c) 2023 UltiMaker
+// CuraEngine is released under the terms of the AGPLv3 or higher
 
 #ifndef RATIO_H
 #define RATIO_H
+
+#include "utils/types/numeric_facade.h"
+
 
 namespace cura
 {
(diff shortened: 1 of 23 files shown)


tests/ExtruderPlanTest.cpp
```diff
@@ -106,11 +106,11 @@ class ExtruderPlanTestPathCollection
         constexpr Ratio flow_04 = 0.4_r;
         decreasing_flow.assign
         ({
-            GCodePath(extrusion_config, mesh, SpaceFillType::Lines, flow_12, no_spiralize, speed_1),
-            GCodePath(travel_config, mesh, SpaceFillType::Lines, flow_1, no_spiralize, speed_1),
-            GCodePath(extrusion_config, mesh, SpaceFillType::Lines, flow_08, no_spiralize, speed_1),
-            GCodePath(travel_config, mesh, SpaceFillType::Lines, flow_1, no_spiralize, speed_1),
-            GCodePath(extrusion_config, mesh, SpaceFillType::Lines, flow_04, no_spiralize, speed_1)
+            GCodePath(extrusion_config, mesh, SpaceFillType::Lines, flow_12, width_1, no_spiralize, speed_1),
```
jellespijker, line 109: This why I don't like implicit conversions

include/settings/types/Ratio.h
```diff
@@ -12,104 +15,28 @@ namespace cura
...
-    {
-        value /= other;
-        return *this;
-    }
-
-    /*
-     * \brief The actual ratio, as a double.
-     */
-    double value = 0;
+    constexpr Ratio(const utils::numeric auto numerator, const utils::numeric auto divisor)
```
rburema, line 31: I get the reasoning behind this one better, but even here we could consider typedef & then having a single `make_ratio(a, b)` function.

include/settings/types/Velocity.h
```diff
@@ -12,74 +14,24 @@ namespace cura
  *
  * This is a facade. It behaves like a double, only it can't be negative.
  */
-struct Velocity
+struct Velocity : public utils::NumericFacade<double>
```
rburema, line 17: The comment above says how 'double, only it cant be negative', but I don't see any mechanism taking care of that.

We should either fix that, or it's also a candidate to be replaced by typedef.

include/settings/types/Velocity.h
```diff
@@ -12,74 +14,24 @@ namespace cura
...
+struct Acceleration : public utils::NumericFacade<double>
+{
+    using base_type = utils::NumericFacade<double>;
+    using base_type::NumericFacade;
 
-    /*
-     * \brief The actual temperature, as a double.
-     */
-    double value = 0;
+    constexpr Acceleration(const base_type& base) noexcept
```
rburema, line 31: ... and same for Acceleration.

rburema (commented):
I like the changes better than the original code.

If we're doing this however, I think we should also consider replacing most, or even all, of these classes with simple typedefs.

Since none of them have any actual constraints _implemented_ for them (though I guess you can't do a power on a `Velocity`, or do bit-operations on a `LayerIndex`), they mostly behave like their base-types anyway (which is why you could relatively easily consolidate them into `NumericFacade` in the first place...)

(The exception is the constructor for `Ratio`, but we could just have a `ratio_t make_ratio(a,b)` function.)

The only benefit I could think of (over typedefs) is that _if you remember to write down the type anyway_ it prevents you from accidentally doing the wrong type, like `auto velocity = 1` creating an `int` of some type. But in order to fix that, you'd have to remember to include the type _anyway_ (`auto velocity = Velocity{1}`) and at that point, it's not really any different from `velocity_t velocity = 1`.

This changes if we actually implement constraints for them; like having velocity actually always be positive, or a layer-number that can't go above the current model height or something (would always return the top layer if going above max, and the lowest when going below min). That's overkill w.r.t. our current workload though.

PR #1931, opened 2023-08-10
[CURA-10927] Fix? 'NumType' regression

rburema:
We had a regression right after the `NumType` introduction, which this attempts to fix, expect we cant reproduce the original error anymore.

Still, this should be a little bit of cleanup that might not be unwelcome. (And just to be sure...)


jellespijker (approved):
LGTM;

and yeas these types need additional thought

PR #1932, opened 2023-08-10
Cura 10916 fix layer index

jellespijker:
# Description

Make sure that LayerIndex is it's own integral type and don't inherit from numeric facade, A lot of conversion problems, because `LayerIndex + integral-> NumericFacade`

Als ensured that the Cura Arcus types used int64


rburema:
Seems like the fail in the analyzer doesn't have anything to do with this PR at least.

rburema:
Ah, I remember seeing the return types weren't templated. That should have rung a bell.

LGTM.

rburema (approved):

PR #1940, opened 2023-08-29
CURA-8889_fix_one_at_a_time_initial_temperature

wawanbreton:
Bed temperature was not applied after the first model being printed.
Only for Griffin-flavored GCode printers (which includes UltiMaker printers).

### DIFF
--- a/include/gcodeExport.h
+++ b/include/gcodeExport.h
@@ -461,6 +461,11 @@ private:
      */
     void writeMoveBFB(const int x, const int y, const int z, const Velocity& speed, double extrusion_mm3_per_mm, PrintFeatureType feature);
 
+    /*!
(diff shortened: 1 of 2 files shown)


rburema (approved):
Looks good to me, straightforward change.

wawanbreton:
Thanks,
I tried to investigate the benchmark crash, and it also happens on the main branch. Should we wait for a fix before we can merge this one ?

wawanbreton:
Ok, I have my answer :)

rburema:
@wawanbreton Haha, within the same minute, even. The same, unrelated, benchmark crash is happening for a lot of branches. We should fix that soon, but since this isn't something I'd expect any changes in the benchmark from, we can just ignore it for now.

PR #1941, opened 2023-08-29
CURA-10951 build plugin as bundled plugin on the GH runners

jellespijker:
WIP


jellespijker:
@casperlamboo I'm merging this in CURA-10475 since the diff is nonconsequential

PR #1948, opened 2023-09-07
heavy flow after bridging line solved

saumyaj3:
CURA-10933

### DIFF
--- a/src/LayerPlan.cpp
+++ b/src/LayerPlan.cpp
@@ -718,7 +718,7 @@ void LayerPlan::addWallLine(
 
             // flow required for the next line segment - when accelerating after a bridge segment, the flow is increased in inverse proportion to the speed_factor
             // so the slower the feedrate, the greater the flow - the idea is to get the extruder back to normal pressure as quickly as possible
-            const float segment_flow = (speed_factor < 1) ? flow * (1 / speed_factor) : flow;
+            const float segment_flow = (flow < 1) ? flow * (1 / speed_factor) : flow;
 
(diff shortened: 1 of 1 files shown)


src/LayerPlan.cpp
```diff
@@ -718,7 +718,7 @@ void LayerPlan::addWallLine(
 
             // flow required for the next line segment - when accelerating after a bridge segment, the flow is increased in inverse proportion to the speed_factor
             // so the slower the feedrate, the greater the flow - the idea is to get the extruder back to normal pressure as quickly as possible
-            const float segment_flow = (speed_factor < 1) ? flow * (1 / speed_factor) : flow;
+            const float segment_flow = (flow < 1) ? flow * (1 / speed_factor) : flow;
```
rburema, line 721: (Sorry for the repeat, but if this is bumped to monday, it might be good if other people can read it as well.)

Given the comment right above it, a fix that would make more sense to me is to have ... = (speed_factor > 1) ? ... instead of ... = (speed_factor < 1) ? ...  -- but your fix is ... = (flow < 1) ? ... . Could you show your reasoning behind choosing the flow there, instead of flipping the compare to be greater-than instead?

saumyaj3: As per my understanding, speed and flow are inversely proportional here. And both the statement would eventually mean the same thing. But I understand that your suggestion implies the fix in a better way. I was making a fix for the flow, so immediately thought of this the first thing.

rburema:
Superseded by #1949

PR #1950, opened 2023-09-14
[CURA-11019] speed-fix (ugly? raw pointer edition)

rburema:
The `make_shared` function copies the entire object, since it's in essence a sort of `emplace` for a smart-pointer version of the constructor of an object. Which means basically that the copy-constructor is called in the case where an object is inserted into `make_shared`. This is normally not so bad, but this is an object representing an entire layer, with non-pointer member-variables. It didn't happen that often, but each operation was quite expensive.

The solution presented here is _maybe_ a bit ugly, but we should either accept raw pointers (because the original object is on the stack and manages itself -- a lot less risk of dangling pointers or whatever), _or_ we go all the way and start constructing it as a smart-pointer in the first place.

(I also made some pass-by-copy into pass-by-reference changes, since smart pointers are still objects, so probably slightly bigger than a reference.)

### DIFF
--- a/include/LayerPlan.h
+++ b/include/LayerPlan.h
@@ -70,7 +70,7 @@ private:
     std::vector<bool> has_prime_tower_planned_per_extruder; //!< For each extruder, whether the prime tower is planned yet or not.
     std::optional<Point> last_planned_position; //!< The last planned XY position of the print head (if known)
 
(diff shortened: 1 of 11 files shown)


rburema:
Closed in favour of #1951

PR #1951, opened 2023-09-15
[CURA-11019] fix slowdown better solution

rburema:
Fix slowdown issue due to copy of large data.

Since `make_shared` is a constructor, the mesh data was copied over each time we started a new layer. Properly fix this by using smart-pointers from the beginning, so we only have the creation overhead, which is nearly nothing. In fact, this solution may be _faster_ than the original, perhaps due to some constness that is now also enforced.

Supersedes #1950

### DIFF
--- a/include/FffGcodeWriter.h
+++ b/include/FffGcodeWriter.h
@@ -333,7 +333,12 @@ private:
      * \param mesh_config the line config with which to print a print feature
      * \param gcode_layer The initial planning of the gcode of the layer.
      */
-    void addMeshLayerToGCode(const SliceDataStorage& storage, const std::shared_ptr<SliceMeshStorage>& mesh_ptr, const size_t extruder_nr, const MeshPathConfigs& mesh_config, LayerPlan& gcode_layer) const;
+    void addMeshLayerToGCode(
+        const SliceDataStorage& storage,
+        const std::shared_ptr<SliceMeshStorage>& mesh_ptr,
+        const size_t extruder_nr,
+        const MeshPathConfigs& mesh_config,
+        LayerPlan& gcode_layer) const;
 
     /*!
      * Add all features of the given extruder from a single part from a given layer of a mesh-volume to the layer plan \p gcode_layer.
--- a/include/FffGcodeWriter.h
+++ b/include/FffGcodeWriter.h
@@ -328,13 +328,12 @@ private:
      * This adds all features (e.g. walls, skin etc.) of this \p mesh to the gcode which are printed using \p extruder_nr
      *
      * \param[in] storage where the slice data is stored.
-     * \param mesh The mesh to add to the layer plan \p gcode_layer.
+     * \param mesh_ptr The mesh to add to the layer plan \p gcode_layer.
      * \param extruder_nr The extruder for which to print all features of the mesh which should be printed with this extruder
      * \param mesh_config the line config with which to print a print feature
      * \param gcode_layer The initial planning of the gcode of the layer.
      */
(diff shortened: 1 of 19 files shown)


include/FffGcodeWriter.h
```diff
@@ -328,13 +328,17 @@ class FffGcodeWriter : public NoCopy
...
      * \param gcode_layer The initial planning of the gcode of the layer.
      */
-    void addMeshLayerToGCode(const SliceDataStorage& storage, const SliceMeshStorage& mesh, const size_t extruder_nr, const MeshPathConfigs& mesh_config, LayerPlan& gcode_layer)
-        const;
+    void addMeshLayerToGCode(
+        const SliceDataStorage& storage,
+        const std::shared_ptr<SliceMeshStorage>& mesh_ptr,
+        const size_t extruder_nr,
+        const MeshPathConfigs& mesh_config,
+        LayerPlan& gcode_layer) const;
```
wawanbreton, line 341: Are you still willing to pass the smart pointers by reference ? IMHO it makes the code less readable (and having a "reference to a pointer to an object" is bit mindfucking :smile: ). Some people advise that you create a typedef of the smart pointer so that you can give `SliceMeshStorage::Ptr` instead of `std::shared_ptr<SliceMeshStorage>` but I'm no big fan of this either..

I rely on your wisdom to do what you think is best :pray:

include/LayerPlan.h
```diff
@@ -70,7 +70,7 @@ class LayerPlan : public NoCopy
     std::vector<bool> has_prime_tower_planned_per_extruder; //!< For each extruder, whether the prime tower is planned yet or not.
     std::optional<Point> last_planned_position; //!< The last planned XY position of the print head (if known)
 
-    std::shared_ptr<SliceMeshStorage> current_mesh; //!< The mesh of the last planned move.
+    std::shared_ptr<const SliceMeshStorage> current_mesh; //!< The mesh of the last planned move.
```
wawanbreton, line 73: :+1: :+1: :+1: 

I love consts everywhere...

src/FffGcodeWriter.cpp
```diff
@@ -106,8 +106,9 @@ void FffGcodeWriter::writeGCode(SliceDataStorage& storage, TimeKeeper& time_keep
     }
 
     size_t total_layers = 0;
-    for (SliceMeshStorage& mesh : storage.meshes)
+    for (auto& mesh_ptr : storage.meshes)
```
wawanbreton, line 109: Good practices with the `auto` keyword indicate that you should use it only when the type is clearly explicited somehow around it, which does not seem to be the case here. What do you think ?

include/FffGcodeWriter.h
rburema: It's a bit more efficient; if they don't get passed by reference, they get passed by copy. The object itself isn't very large, but the ref-counter will also be increased as well.

I don't feel very strongly about it, and probably won't even comment if I saw someone else (not) do it -- but now that it's in there, I don't feel the need to change it back.

src/FffGcodeWriter.cpp
rburema: I broadly agree with that statement (though I tend to use `auto` in `for`-loops as well, as the type can be looked up in the collection -- especially if it's a 'longer' type).

In any case, I'll change this to not use `auto`.

rburema: Resolved in 74456d083079cc4025c7d7d68507d6a0152e97f6

PR #1952, opened 2023-09-15
[CURA-11035] Plugins specify a verion-range, slots just a version.

rburema:
Not the other way around as we have now. This also adheres closer to the way we handle front-end plugins. Ideally you'd maybe want both of them to be ranges, but that's for the future.


rburema:
> LGTM; A couple of smaller remarks on the other PR's tho

Hi @jellespijker ! I keep meaning to ask today, but you keep escaping... where are the other, small remarks on the other PR's? 😅

PR #1964, opened 2023-10-05
the length of brim is correctly calculated in case of no adhesion

saumyaj3:
CURA-11097

Issue: This happens when Build-plate adhesion is set to None, but you still have the support brim enabled for Normal support.
still_happens_in_other_model (1)

### DIFF
--- a/src/SkirtBrim.cpp
+++ b/src/SkirtBrim.cpp
@@ -667,8 +667,8 @@ void SkirtBrim::generateSupportBrim()
         }
 
         storage.support_brim.add(brim_line);
-
-        const coord_t length = (adhesion_type == EPlatformAdhesion::NONE) ? skirt_brim_length : skirt_brim_length + storage.support_brim.polygonLength();
+        // In case of adhesion::NONE length of support brim is only the length of the brims formed for the support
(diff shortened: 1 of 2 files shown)


src/SkirtBrim.cpp
```diff
@@ -668,7 +668,7 @@ void SkirtBrim::generateSupportBrim()
 
         storage.support_brim.add(brim_line);
 
-        const coord_t length = skirt_brim_length + storage.support_brim.polygonLength();
+        const coord_t length = (adhesion_type == EPlatformAdhesion::NONE) ? skirt_brim_length : skirt_brim_length + storage.support_brim.polygonLength();
```
rburema, line 671: I'll add a comment here -- I think otherwise you'd get future developers changing the portion between the `?` and `:` to `0`, because they forget about the support-brim 😁

saumyaj3: Oh yes. I thought I did made a comment there.. but apparently I didn't :/
Will do it (y)

rburema: Never mind, it's literally in a 'support' function 😅

rburema: ... but the comment clears things up in any case!

rburema (approved):

PR #1967, opened 2023-10-09
[CURA-11100] scattered seam

wawanbreton:
* Corner angle calculation algorithm was rewritten to better handle differences in resolution between models.
* Adjustment of  start point scores calculation to make the seam consistant and expectable as much as possible

### DIFF
--- a/include/PathOrderOptimizer.h
+++ b/include/PathOrderOptimizer.h
@@ -4,9 +4,6 @@
 #ifndef PATHORDEROPTIMIZER_H
 #define PATHORDEROPTIMIZER_H
 
-#include <spdlog/spdlog.h>
-#include <unordered_set>
-
 #include "InsetOrderOptimizer.h" // for makeOrderIncludeTransitive
 #include "PathOrdering.h"
 #include "pathPlanning/CombPath.h" //To calculate the combing distance if we want to use combing.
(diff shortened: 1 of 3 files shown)


include/PathOrderOptimizer.h
```diff
@@ -636,18 +636,35 @@ class PathOrderOptimizer
             return vert;
         }
 
+        //Precompute segments lengths because we are going to need them multiple times
+        std::vector<coord_t> segments_sizes(path.converted->size());
+        coord_t total_length = 0;
+        for(const auto& [i, here]: **path.converted | ranges::views::enumerate)
+        {
+            const Point &next = (*path.converted)[(i + 1) % path.converted->size()];
+            coord_t segment_size = vSize(next - here);
```
casperlamboo, line 645:
```suggestion
            const coord_t segment_size = vSize(next - here);
```

include/PathOrderOptimizer.h
```diff
@@ -636,18 +636,35 @@ class PathOrderOptimizer
...
+            segments_sizes[i] = segment_size;
+            total_length += segment_size;
+        }
+
         size_t best_i;
         float best_score = std::numeric_limits<float>::infinity();
         for (const auto& [i, here] : **path.converted | ranges::views::enumerate)
```
casperlamboo, line 652: Rather then performing the if-check on line below for every iteration, you could also skip the last element using the ranges `drop_last` utility
```suggestion
        for (const auto& [i, here] : **path.converted | ranges::views::enumerate | ranges::views::drop_last(/*is_path_closed()*/true ? 1 : 0))
```

casperlamboo (commented):
Had some very minor remarks these were mostly personal preference remarks as I had not really any code-remarks

include/PathOrderOptimizer.h
wawanbreton: This is very nice, I was looking for a clean way to do this but I'm not familiar enough with `std::ranges`

casperlamboo (approved):

PR #1968, opened 2023-10-13
[CURA-10783] prime tower with raft

wawanbreton:
# Description
This new feature adds a base to the prime tower and changes its adhesion:
* When using a raft, only the first layer of the raft is used for the prime tower
* It is now possible to add a base to the prime tower so that it can grow taller without risking breaking

# Technical changes
* Technically, the main difference is that the brim of the prime tower is no more handled by the brim algorithm, but by the tower generation itself, as it is now part of the base. For the raft, we use the classic raft generation for the first layer, then the prime tower handles its optional base and the rest of the raft ignores it.
* When the base of the prime tower overlaps with the models brim/raft, it takes precedence over them. This may be discussed to do it the other way.
* On the front-end side, the test of clearance areas on the first layer has been customized to ensure that the tower can be properly placed without causing troubles.

## Settings
* The setting "prime_tower_brim_enable" has been kept for easier retro-compatibility, however it now designates whether the prime tower has a base, not only a brim. It would make sense to rename it but that would require an upgrade script.
* The setting "prime_tower_base_size" has been added to customize it independantly from the model brim/raft width. However it default to the brim width.
* The setting "prime_tower_base_height" has been added to customize the height of the base. It defaults to the layer height, so that if you don't touch it, you get a brim-like base.
* The setting "prime_tower_base_curve_magnitude" has been added to allow customizing the shape of the base, which would depend on the used materials and actual height of the prime tower. Default value has been set to something that looks good for most situation, but may change with further tests.

## Retro-compatibility
Loading previous models with these new settings should give similar results, as the setting "prime_tower_brim_enable" is reused and the "prime_tower_base_size" defaults to "brim_width". The only situation that could cause problems is one where the prime tower was located very close to the model, with a brim. With the new implementation, the tower's brim is included in the disallowed area (because it may have a base) so it could overlap with the model and not allow slicing. This point can be discussed as a possible improvement.

# Tests
I have tested and fixed various positions of the prime tower and settings, including ooze shield and draft shield that overlap with the prime tower. One-at-a-time mode is not available in multi-extruders so no troubles here :)
Some other options/settings may be further tested to make sure that they are not impacted by the tower's base.


include/PrimeTower.h
```diff
@@ -1,15 +1,16 @@
...
 #define PRIME_TOWER_H
 
-#include <vector>
-
+#include "settings/types/LayerIndex.h"
 #include "utils/polygon.h" // Polygons
 #include "utils/polygonUtils.h"
 
-namespace cura 
+#include <vector>
```
scottsha, line 11: Style:  C++ standard library headers, other libraries' headers, your project's headers.

include/PrimeTower.h
```diff
@@ -40,13 +41,15 @@ class PrimeTower
     const unsigned int number_of_prime_tower_start_locations = 21; //!< The required size of \ref PrimeTower::wipe_locations
 
     std::vector<ExtrusionMoves> pattern_per_extruder; //!< For each extruder the pattern to print on all layers of the prime tower.
-    std::vector<ExtrusionMoves> pattern_per_extruder_layer0; //!< For each extruder the pattern to print on the first layer
+    std::vector<std::vector<ExtrusionMoves>> pattern_extra_brim_per_layer; //!< For each layer of each extruder, the extra pattern to be added for adhesion and/or strength
```
scottsha, line 44: Is one of the vectors over the extruders and the other over a number of patterns? Maybe typedef those to show the intent. It's not clear what the outer and inner indexes will be.

include/PrimeTower.h
```diff
@@ -100,11 +103,47 @@ class PrimeTower
...
+     * \see PrimeTower::generatePaths
+     *
+     * Generate extra rings around the actual prime rings for a stronger base
+     *
+     * \param inset The inner circle of the rings to start generating the rings from
+     * \param rings The number of rings to add
+     * \param line_width The actual line width to distance the rings from each other
+     * \return The generated rings paths
+     */
+    static ExtrusionMoves generatePaths_base(const Polygons& inset, size_t rings, coord_t line_width);
```
scottsha, line 130: This generates a number of inset paths? Seems generically useful geometric function, does it belong in a geometry toolbox?

include/SkirtBrim.h
```diff
@@ -1,17 +1,17 @@
...
 #ifndef SKIRT_BRIM_H
 #define SKIRT_BRIM_H
 
-#include "utils/Coord_t.h"
 #include "ExtruderTrain.h"
 #include "settings/EnumSettings.h"
 #include "sliceDataStorage.h"
+#include "utils/Coord_t.h"
 
 #include <variant>
```
scottsha, line 12: style: header order

src/PrimeTower.cpp
```diff
@@ -97,23 +102,62 @@ void PrimeTower::generateGroundpoly()
...
+    would_have_actual_tower
+        = storage.max_print_height_second_to_last_extruder >= 0; // Maybe it turns out that we don't need a prime tower after all because there are no layer switches.
     if (would_have_actual_tower && enabled)
     {
         generatePaths_denseInfill();
         generateStartLocations();
     }
 }
 
+PrimeTower::ExtrusionMoves PrimeTower::generatePaths_base(const Polygons& inset, size_t rings, coord_t line_width)
```
scottsha, line 114: General radial fill pattern is not prime tower specific and helpfully reusable, suggest moving to a generic geometry tool location

src/PrimeTower.cpp
```diff
@@ -97,23 +102,62 @@ void PrimeTower::generateGroundpoly()
...
+
+    return pattern;
+}
+
 void PrimeTower::generatePaths_denseInfill()
 {
     const Scene& scene = Application::getInstance().current_slice->scene;
     const Settings& mesh_group_settings = scene.current_mesh_group->settings;
     const coord_t layer_height = mesh_group_settings.get<coord_t>("layer_height");
+    const bool base_enabled = mesh_group_settings.get<bool>("prime_tower_brim_enable");
```
scottsha, line 150: out of scope but not a fan of magic strings

src/PrimeTower.cpp
```diff
@@ -97,23 +102,62 @@ void PrimeTower::generateGroundpoly()
...
+    const coord_t base_extra_radius = scene.settings.get<coord_t>("prime_tower_base_size");
+    const bool has_raft = mesh_group_settings.get<EPlatformAdhesion>("adhesion_type") == EPlatformAdhesion::RAFT;
+    const coord_t base_height = std::max(scene.settings.get<coord_t>("prime_tower_base_height"), has_raft ? layer_height : 0);
+    const int base_curve_magnitude = mesh_group_settings.get<int>("prime_tower_base_curve_magnitude");
```
scottsha, line 154: I think "magnitude" is misleading as it suggests an amplification factor, but it is a decay exponent or a log-magnitude

src/PrimeTower.cpp
```diff
@@ -122,42 +166,50 @@ void PrimeTower::generatePaths_denseInfill()
...
 
-        //Only the most inside extruder needs to fill the inside of the prime tower
-        if (extruder_nr != extruder_order.back())
+        // The most outside extruder is used for the base
+        if (extruder_nr == extruder_order.front() && (base_enabled || has_raft) && base_extra_radius > 0 && base_height > 0)
         {
-            pattern_per_extruder_layer0 = pattern_per_extruder;
+            for (coord_t z = 0; z < base_height; z += layer_height)
+            {
+                double brim_radius_factor = std::pow((1.0 - static_cast<double>(z) / base_height), base_curve_magnitude);
```
scottsha, line 188: This is a nice implementation because it generically allows other prime tower cross section shapes, but  the only shape fully supported is a circle right?

scottsha:
Looks good to me! I have only superficial code style and naming nits

scottsha (approved):
Looks good to me! I have only superficial code style and naming suggestions

include/PrimeTower.h
wawanbreton: You're right ! I thought the clang-format was taking care of that, but it was wrong :)

include/PrimeTower.h
wawanbreton: Yes indeed, technically I should have called it  `pattern_extra_brim_per_layer_per_extruder` but that would have been even more confusing :laughing: Using typedefs is a good idea, I'll see what I can do.

include/PrimeTower.h
wawanbreton: That's the very same thing that the brim algorithm does indeed. I'll see if I can merge them.

src/PrimeTower.cpp
wawanbreton: Me neither :disappointed: but this is how the whole engine is implemented until now...

src/PrimeTower.cpp
wawanbreton: This was a bit hard to name indeed... For me, magnitude refers to the exponent order of the function, but maybe this is too French :smiley: On the other hand, some people have used it yet and everybody seemed to understand the meaning of this value (probably just by playing with it). But if you have an other word you really feel would be better, name it.

include/PrimeTower.h
wawanbreton: @scottsha Tell me if the new version makes it better

src/PrimeTower.cpp
wawanbreton: Most of it would work with a different shape indeed, except exclusion computation on the front-end and maybe wipe positions. With some minor effort that should be possible, but I _think_ the PPM team likes the circles because it makes smooth moves.

include/PrimeTower.h
wawanbreton: Didn't merge with the brim, but at least they are now in PolygonUtils

src/PrimeTower.cpp
```diff
@@ -97,23 +102,62 @@ void PrimeTower::generateGroundpoly()
...
+    return pattern;
+}
+
 void PrimeTower::generatePaths_denseInfill()
 {
     const Scene& scene = Application::getInstance().current_slice->scene;
     const Settings& mesh_group_settings = scene.current_mesh_group->settings;
     const coord_t layer_height = mesh_group_settings.get<coord_t>("layer_height");
+    const bool base_enabled = mesh_group_settings.get<bool>("prime_tower_brim_enable");
+    const coord_t base_extra_radius = scene.settings.get<coord_t>("prime_tower_base_size");
```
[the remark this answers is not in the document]
wawanbreton: During the refinement meeting we used "chamfer" which looks like "fillet". I turned it to "base" to make it more generic and easy to understand, but again, if there is a word you prefer, you tell me ! I like "buttress" because it really means what we want it to mean, but maybe it's too much technical, I didn't know it before...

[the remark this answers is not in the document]
scottsha: Ha ya buttress is a weird architecture term I like that it fits with tower. I like chamfer too. I can never keep chamfers and fillets straight there's some weird distinction I think fillet indicates it's more rounded so that it has g1 continuity or something? Nice and smooth like your tower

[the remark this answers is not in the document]
wawanbreton: Depends on the magnitude :) If you set 0 you get a block, with 1 you get a pyramid, and for larger values you have something smooth.

PR #1978, opened 2023-11-08
Fix prime tower when only raft has multi-material 

casperlamboo:
# Description

We didn't take into account that raft layers are negative

CURA-11291

# How Has This Been Tested?

🏃🏼‍♀️ running from sause

# Checklist:

### DIFF
--- a/src/PrimeTower.cpp
+++ b/src/PrimeTower.cpp
@@ -103,7 +103,7 @@ void PrimeTower::generateGroundpoly()
 void PrimeTower::generatePaths(const SliceDataStorage& storage)
 {
     would_have_actual_tower
-        = storage.max_print_height_second_to_last_extruder >= 0; // Maybe it turns out that we don't need a prime tower after all because there are no layer switches.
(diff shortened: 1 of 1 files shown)


src/PrimeTower.cpp
```diff
@@ -102,8 +102,9 @@ void PrimeTower::generateGroundpoly()
 
 void PrimeTower::generatePaths(const SliceDataStorage& storage)
 {
-    would_have_actual_tower
-        = storage.max_print_height_second_to_last_extruder >= 0; // Maybe it turns out that we don't need a prime tower after all because there are no layer switches.
+    const int raft_total_extra_layers = Raft::getTotalExtraLayers();
+    would_have_actual_tower = storage.max_print_height_second_to_last_extruder
+                           >= -raft_total_extra_layers + 1; // Maybe it turns out that we don't need a prime tower after all because there are no layer switches.
```
rburema, line 107:
```suggestion
                           >= -raft_total_extra_layers; // Maybe it turns out that we don't need a prime tower after all because there are no layer switches.
```

rburema (approved):

PR #1985, opened 2023-11-21
CURA-11364 Add Sentry Support

nallath:
# Description

This adds support for sentry in the engine, which will give us much better insights regarding slicing crashes. 

CURA-11364

Goes with: https://github.com/Ultimaker/Cura/pull/17394

## Type of change


src/main.cpp
```diff
@@ -37,7 +40,33 @@ int main(int argc, char** argv)
...
+#elif defined(__APPLE__) && defined(__MACH__)
+    config_path += getenv("HOME");
+    config_path += "/Library/Application Support/cura/";
+#elif defined(_WIN64)
+    config_path = "%APPDATA%\\cura\\";
+#endif
+    config_path += ".sentry-native";
+    sentry_options_set_database_path(options, config_path.c_str());
+    std::string version = "curaengine@";
+    version += std::string(CURA_ENGINE_VERSION);
```
casperlamboo, line 63: Why not use `fmt::format` here?
```suggestion
    std::string version = fmt::format("curaengine@"{}, CURA_ENGINE_VERSION);
```

rburema (commented):
Plz also test with different locales and non-ASCII user names (specifically on Win64)!

conanfile.py
```diff
@@ -28,19 +29,22 @@ class CuraEngineConan(ConanFile):
...
         "enable_plugins": [True, False],
+        "enable_sentry": [True, False],
         "enable_remote_plugins": [True, False],
     }
     default_options = {
         "enable_arcus": True,
         "enable_benchmarks": False,
         "enable_extensive_warnings": False,
         "enable_plugins": True,
+        "enable_sentry": False,
```
casperlamboo, line 40: Should we also update the gh runner install commands to `enable_sentry` is this already done, or should we still do this?

nallath: That's a bit of code that @jellespijker added. I think that it should work, but I haven't tried it.

casperlamboo: https://github.com/Ultimaker/Cura/commit/2f452c6581ba11c3381b44ff1d5151fc56173687

PR #1996, opened 2023-12-11
CURA-11392-including-threshold-setting-for-ppr

saumyaj3:
CURA-11392

for internal printers,
Introduce Threshold settings for Printjobs so it can be taken into account for print progress reporting

### DIFF
--- a/include/gcodeExport.h
+++ b/include/gcodeExport.h
@@ -167,7 +167,7 @@ private:
     Temperature bed_temperature_; //!< Current build plate temperature.
     Temperature build_volume_temperature_; //!< build volume temperature
     bool machine_heated_build_volume_; //!< does the machine have the ability to control/stabilize build-volume-temperature
-    bool ppr_enable_; //! if the print process reporting is enabled
+    bool ppr_enable_; //!< if the print process reporting is enabled
 
 protected:
     /*!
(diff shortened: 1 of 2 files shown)


rburema:
GCode-analyzer failing, probably because it has the wrong version of Cura checked out. Should go alright if we merge this after the Cura branch is merged.

jellespijker:
Why is this in the engine?
This could be implemented in the front-end with the formula's in the start/end gcode

rburema:
@jellespijker The rest of the Griffin-header is also done in the engine. If we want to do it that way, I'd propose moving it over wholesale.

~~~
    case EGCodeFlavor::GRIFFIN:
        prefix << ";START_OF_HEADER" << new_line_;
        prefix << ";HEADER_VERSION:0.1" << new_line_;
        prefix << ";FLAVOR:" << flavorToString(flavor_) << new_line_;
        prefix << ";GENERATOR.NAME:Cura_SteamEngine" << new_line_;
        prefix << ";GENERATOR.VERSION:" << CURA_ENGINE_VERSION << new_line_;
        prefix << ";GENERATOR.BUILD_DATE:" << Date::getDate().toStringDashed() << new_line_;
        prefix << ";TARGET_MACHINE.NAME:" << transliterate(machine_name_) << new_line_;
~~~

jellespijker:
yes, but these are also for a subset of printers that use the griffin header.

PR #2003, opened 2024-01-03
Fix hole expansion

casperlamboo:
As we were only taking the very inner contour of the true extrusion polygon. There was code that would increase this area a bit, but it wasn't properly implemented, so for certain cases we would miss parent child relationships.

# Description

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* Operating System:

# Checklist:

### DIFF
--- a/src/InsetOrderOptimizer.cpp
+++ b/src/InsetOrderOptimizer.cpp
@@ -145,7 +145,7 @@ InsetOrderOptimizer::value_type InsetOrderOptimizer::getRegionOrder(const std::v
     }
 
     // view on the extrusion lines, sorted by area
-    const auto sorted_extrusion_lines = [&extrusion_lines]()
+    const std::vector<const ExtrusionLine*> sorted_extrusion_lines = [&extrusion_lines]()
     {
         auto extrusion_lines_area = extrusion_lines | ranges::views::addressof
                                   | ranges::views::transform(
@@ -195,7 +195,7 @@ InsetOrderOptimizer::value_type InsetOrderOptimizer::getRegionOrder(const std::v
             }
         }
         // increase the size of the hole polygons by 10um to make sure we don't miss any invariant parents
-        hole_polygons.offset(10);
+        hole_polygons = hole_polygons.offset(10);
 
         // go through all the invariant parents and see if they are inside the hole polygon
         // if they are, then that means we have found a child for this extrusion line


include/utils/ExtrusionLine.h
```diff
@@ -228,47 +228,6 @@ struct ExtrusionLine
 
         return ret;
     }
-
-    /*!
-     * Create a true-extrusion area shape for the path; this means that each junction follows the bead-width
-     * set for that junction.
-     */
-    Polygons toExtrusionPolygons() const
```
casperlamboo, line 236: Why was this deleted, I think this can still come in handy, no?

casperlamboo:
The remaining issue that wall ordering was a bit odd when wall-ordering was set to "inner to outer" wasn't caused by the parent-child nesting logic, but was instead caused in the part where we use this nesting logic to create the resulting wall ordering. Below is an example of tool paths where the wall ordering went "wrong".
wall ordering

Since wall ordering is set to "inner to outer" the resulting nesting logic was (correctly) determined to be
```mermaid
graph TD;
g --> f
f --> e
e --> d
d --> b
c --> b
b --> a
```

Previously the logic was such that a dfs was performed on this nesting graph for every root node (blue nodes figure above). In this example the root nodes are $c$ and $g$ are root nodes. The initial root is chosen to be the tool path with the closest point to the current nozzle position. When this node would be $c$ the printing sequence would be $\langle c, b, a, g, f, e, d \rangle$. Which didn't feel ordered from inner to outer. Main issue is that wall toolpath $b$ is printed _before_ $f, g, e, d$ while $b$ is a parent of all these tool paths.

This lead me to add the following constraint: 
**A path $u$ can only be printed if all paths $v$ for which there is an order dependency $(u, v)$ in the order requirement is done printing.**
This is done by keeping track of of the number of incoming edges to each path. Each time a path $p$ is handled (done processing and added to the final ordering list) the number of incoming edges of all $p$'s neighbors is decreased by $1$. When finding the next wall tool path to print the candidate wall tool paths are limited to those with an incoming edges count of 0.

_Side node:_ this was only an issue in the situation where wall ordering is set to "inner to outer". When wall ordering is set to "outer to inner" the order requirements analog to a tree-graph. One property in a tree-graph is that all nodes have exactly one incoming edge (except for the root nodes, which have 0 incoming edges). When such a tree is traversed handling a node makes all neighbors valid candidates as all their incoming edges are decreased from 1 to 0. When this setting is set to "inner to outer" instead all edges are revered. When edges are reversed in a tree-graph the graph is no longer a tree and we loose this property.

include/utils/ExtrusionLine.h
jellespijker: I you decide to re-add it to the code base add the`[[maybe_unused]]` attribute so it won't result in a compiler warning

src/FffGcodeWriter.cpp
```diff
@@ -2529,98 +2529,7 @@ bool FffGcodeWriter::processInsets(
...
-                        }
-                    }
-                }
-            }
-            return false;
-        }();
-
+        auto use_roofing_config = false;
         const GCodePathConfig& inset0_config = use_roofing_config ? mesh_config.inset0_roofing_config : mesh_config.inset0_config;
         const GCodePathConfig& insetX_config = use_roofing_config ? mesh_config.insetX_roofing_config : mesh_config.insetX_config;
```
saumyaj3, line 2535:
```suggestion
        const GCodePathConfig& inset0_config = mesh_config.inset0_config;
        const GCodePathConfig& insetX_config = mesh_config.insetX_config;
```

src/FffGcodeWriter.cpp
```diff
@@ -2529,98 +2529,7 @@ bool FffGcodeWriter::processInsets(
...
-                        {
-                            return true;
-                        }
-                    }
-                }
-            }
-            return false;
-        }();
-
+        auto use_roofing_config = false;
```
saumyaj3, line 2532: This variable doesn't serve any purpose now

include/utils/ExtrusionLine.h
```diff
@@ -228,6 +228,47 @@ struct ExtrusionLine
 
         return ret;
     }
+
+    /*!
+     * Create a true-extrusion area shape for the path; this means that each junction follows the bead-width
+     * set for that junction.
+     */
+    Polygons toExtrusionPolygons() const
```
saumyaj3: maybe we should add [[maybe_unused]] for this function?

src/FffGcodeWriter.cpp
casperlamboo: We still want to implement something in place of the removed code; added a comment to reflect that d8cbf1f186a8d8142ad89d3a83495122bb4e9494

src/FffGcodeWriter.cpp
casperlamboo: d8cbf1f186a8d8142ad89d3a83495122bb4e9494

include/utils/ExtrusionLine.h
casperlamboo: https://github.com/Ultimaker/CuraEngine/pull/2003/commits/f2930878300de0baf7b115fc32f65e597292856e

saumyaj3 (approved):

PR #2009, opened 2024-01-10
[CURA-11157] Remove 'support_interface_skip_height' setting

rburema:
Was used for run-time efficiency, but it wasn't really ever set beyond one layer-height, which made it rather useless.

For the frontend PR see https://github.com/Ultimaker/Cura/pull/17961


casperlamboo:
Why not completely remove all references to skip layer height instead of defaulting it to 1 (see https://github.com/Ultimaker/CuraEngine/pull/2010). I think by leaving the references to "old unused features" we increase the complexity of the engine.

rburema:
@casperlamboo Good point -- I didn't want to alter too much in this ticket, but perhaps it can be done simply. Let me have a look.

rburema:
Done!

casperlamboo (approved):

PR #2014, opened 2024-01-13
Cura 11482 sentry

jellespijker:
# Description

Removed the Sentry Spdlog Sink, this was corrupting the Sentry data on most occasions, resulting in dropped Sentry issues. Since this was a nice to have feature I opted for dropping support at this time. If it turns out we are missing valuable information I propose that we make a ticket for this on the backlog.

Integrated Sentry uploading step in the build scripts. This will upload the debug and source files giving us quite a detailed Issue report, this should also result in better merging of Sentry events.

image

Integrated Sentry release step in the build scripts. This will mark a Sentry release for CuraEngine specifying the commit. This gives us the diff between previous releases and marks who worked on the commits. This should allow Sentry to give better hints when a bug was introduced.

image

The environmental variable `CURAENGINE_SENTRY_USER` is now the actual user name, which can either be provided on the shell environment or set by Cura if a logged in DF user opts-in for this feature. This will help us to quickly find our own crashes and those of colleagues. Or in the future DF customers

image

## Type of change

# How Has This Been Tested?

- [X] Locally on multiple OSes
- [ ] Test B

**Test Configuration**:
* Operating System: Win, Mac, Linux

# Checklist:


conanfile.py
```diff
@@ -160,13 +165,38 @@ def build(self):
...
+        if self.options.get_safe("enable_sentry", False):
+            # Upload debug symbols to sentry
+            sentry_project = self.conf.get("user.curaengine:sentry_project", "", check_type=str)
+            sentry_org = self.conf.get("user.curaengine:sentry_org", "", check_type=str)
+            if sentry_project == "" or sentry_org == "":
+                raise ConanInvalidConfiguration("sentry_project or sentry_org is not set")
+            if which("sentry-cli") is None:
+                self.output.warn("sentry-cli is not installed, skipping uploading debug symbols")
+                self.output.warn("sentry-cli is not installed, skipping release creation")
+                return
```
wawanbreton, line 177: You should not `return` here because it would skip some code potentially added after the sentry-uploading step (even though this case should not happen)

conanfile.py
```diff
@@ -160,13 +165,38 @@ def build(self):
...
+
+            if self.settings.os == "Linux":
+                self.output.info("Stripping debug symbols from binary")
+                self.run("objcopy --only-keep-debug --compress-debug-sections=zlib CuraEngine CuraEngine.debug")
+                self.run("objcopy --strip-debug --strip-unneeded CuraEngine")
+                self.run("objcopy --add-gnu-debuglink=CuraEngine.debug CuraEngine")
```
wawanbreton, line 183: I love that... However, I have a bit of a concern because it means the release executable will actually be built using debug mode, and it will not do any code optimization, so execution could be slightly less performant. Am I right ?

conanfile.py
```diff
@@ -160,13 +165,38 @@ def build(self):
...
+
+            self.output.info("Uploading debug symbols to sentry")
+            build_source_dir = self.build_path.parent.parent.as_posix()
+            self.run(f"sentry-cli debug-files upload --include-sources -o {sentry_org} -p {sentry_project} {build_source_dir}")
```
wawanbreton, line 187: I guess you checkd this, but aren't we going to overload Sentry with our debug data ? Are they removed after some time ?

src/Slice.cpp
```diff
@@ -20,7 +21,7 @@ Slice::Slice(const size_t num_mesh_groups)
 
 void Slice::compute()
 {
-    spdlog::info("All settings: {}", scene.getAllSettingsString());
+    fmt::print("All settings: {}", scene.getAllSettingsString()); // Print to console do not log, otherwise Sentry won't upload the crash because it is too big.
```
wawanbreton, line 24: As the sentry spdlog sink has been removed, you can revert it back to using spdlog

wawanbreton (commented):
Minor suggestions again.
I am also in favor of creating a ticket to fix the sentry sink (which should not be too difficult)

conanfile.py
jellespijker: bit only for needed for setting up the sentry release and bookkeeping in our Sentry server it self. having it raise an error will only force developers pr other build systems to have that sentry-cli installed.

Which turns out to be rather annoying on a dev setup

conanfile.py
jellespijker: yes we're now build this as RelWithDebug which is o2. Previous tests didn't indicate a increase in slice time.

But even if it did I would argue that we should find this acceptable since one of our main goals this year is to make the engine more robust and simply said we need that extra information for Sentry. Otherwise we basically only get an event that says crash occurred with no meaningful information what so ever.

It is also not that uncommon and sometimes even considered good practice to release with o2 instead of o3

conanfile.py
jellespijker: artifacts are only uploaded when changed and stored for 90 days after the last use. The amount of data being stored is not that much and should be within allowed usages, but if we run into actual problems we should setup a symbol server. But that will be a tomorrow problem, if it will be a problem at all.

src/Slice.cpp
jellespijker: good point, feel free to change

jellespijker:
> Minor suggestions again.
> I am also in favor of creating a ticket to fix the sentry sink (which should not be too difficult)

can you create the Jira ticket for me?

conanfile.py
wawanbreton: I agree that this should not raise an error, but the following part should be under an `else` statement. This way, if you later add more code to the `validate` method (not related to sentry), it will always be executed.

jellespijker: That would yet again add an extra indent for a huge chunk of code which I personally dislike. Also I think anyone working with python should be able to read the code how it is now.

But if you feel strongly about this feel free to change it

conanfile.py
jellespijker: That would be a noticeable drop in speed indeed

conanfile.py
wawanbreton: :+1:

wawanbreton:
> can you create the Jira ticket for me?

CURA-11549

wawanbreton (approved):

PR #2021, opened 2024-01-26
Re implement wall roofing

casperlamboo:
# Description

This "feature" allows users to set a different print configuration (speed, acceleration, jerk) for those outer walls that in the "top surface layers". This gives users more control over printed part quality, especially when tuning profiles for speed.

Previously this feature was implemented but it had some downsides. There was a lot of code duplication, and the code was running into performance issues.

This new implementation solves both this issues, and adds a better interpretation of the feature. Previously if a line segment was partially within and partially outside a top layer then the whole line segment would be printed with the top surface layer outer wall config. With the new implementation only those area's within a line segment that are within the area are printed with this config.

Screenshot 2024-01-26 at 15 19 26

CURA-11129 

## Type of change

# How Has This Been Tested?
Yes

**Test Configuration**:

# Checklist:

### DIFF
--- a/src/LayerPlan.cpp
+++ b/src/LayerPlan.cpp
@@ -648,7 +648,7 @@ void LayerPlan::addWallLine(
     const Point2LL& p0,
     const Point2LL& p1,
     const Settings& settings,
-    const GCodePathConfig& non_bridge_config,
+    const GCodePathConfig& default_config,
     const GCodePathConfig& roofing_config,
     const GCodePathConfig& bridge_config,
     double flow,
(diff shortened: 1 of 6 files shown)


src/LayerPlan.cpp
```diff
@@ -760,12 +763,80 @@ void LayerPlan::addWallLine(
         }
     };
 
-    if (bridge_wall_mask_.empty())
+    const auto use_roofing_config = [&]() -> bool
+    {
+        if (roofing_config == default_config)
+        {
+            // if the roofing config and normal config are the same any way there is no need to check what part of the line segment
```
saumyaj3, line 770:
```suggestion
            // if the roofing config and normal config are the same any way there is no need to check 
```

saumyaj3 (approved):

PR #2023, opened 2024-01-29
Fix printing rafts-holes in monotonic order

casperlamboo:
# Description

Previous implementation contains a bug where raft holes would be printed doubly.

CURA-11438

## Type of change

# How Has This Been Tested?

Yes

**Test Configuration**:

# Checklist:

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -983,24 +983,7 @@ void FffGcodeWriter::processRaft(const SliceDataStorage& storage)
             raft_outline_path = raft_outline_path.difference(storage.primeTower.getOuterPoly(layer_nr));
         }
 
-        std::vector<Polygons> raft_islands;
-        if (monotonic)
-        {
-            // When using monotonic infill, process islands separately otherwise multiple rafts
-            // will be printed in parallel in a global monotonic order, which doesn't look good
-            for (const PolygonRef raft_island : raft_outline_path)
-            {
-                Polygons island;
-                island.add(raft_island);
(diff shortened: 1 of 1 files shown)


src/FffGcodeWriter.cpp
```diff
@@ -983,24 +983,7 @@ void FffGcodeWriter::processRaft(const SliceDataStorage& storage)
...
-                raft_islands.emplace_back(island);
-            }
-        }
-        else
-        {
-            raft_islands.emplace_back(raft_outline_path);
-        }
-
-        for (const Polygons raft_island : raft_islands)
+        for (const Polygons raft_island : raft_outline_path.splitIntoParts())
```
wawanbreton, line 986: That reeeeeeeeeeeeally makes sense ! I wish I had known this method existed before :laughing: 

However, you are now treating the monotonic and normal cases both with the new behavior, which consists in treating every island separately. IMHO, this is definitely better, but it is a behavior change for the normal case. Did you do it on purpose ?

casperlamboo: Didn't really do it really on purpose, but I think it is good to treat every island separately. However, now that I think about it we're relying on the ordering of the `splitIntoParts` for the printer ordering of the different rafts. This might be a very optimized ordering. However, I don't think there are many situations where there are many many raft islands.

wawanbreton: Well technically this _could_ happen if you have a model with many separate legs for example, but this only concerns the top layer(s) of the raft so the effect on total length should be very minimum.

PR #2038, opened 2024-02-27
condition to start position while in case of prime-blob enabled

saumyaj3:
CURA-11553

It seems like the purpose of prime blob was to handle scarring in the print as well. I recommend that initial travel move not to be applied when prime blob is enabled. This solves our problem of scarring that might happed wile prime-blob.(Also during prime blob the start position of brim starts close to the blob rather than the “machine_extruder_start_pos_x, machine_extruder_start_pos_y”)


wawanbreton (approved):
Looks good, makes sense :+1:

PR #2039, opened 2024-02-29
[CURA-11537] Fix remove empty layers when layers do not equal layer height.

rburema:
This shouldn't happen, but neither should the hot-end ram into the build-plate (in fact, you want that even less).

_edit_ Unit tests are failing because of stuff outside of this ticket (prime-tower-enable), and github is not letting me update the branch for some reason.

### DIFF
--- a/src/FffPolygonGenerator.cpp
+++ b/src/FffPolygonGenerator.cpp
@@ -769,12 +769,25 @@ bool FffPolygonGenerator::isEmptyLayer(SliceDataStorage& storage, const LayerInd
 
 void FffPolygonGenerator::removeEmptyFirstLayers(SliceDataStorage& storage, size_t& total_layers)
 {
+    const auto get_highest_z = [&storage](const LayerIndex& layer_idx)
+        {
+            coord_t highest_z = 0;
+            for (std::shared_ptr<SliceMeshStorage>& mesh_ptr : storage.meshes)
+            {
+                auto& mesh = *mesh_ptr;
+                highest_z = layer_idx >= mesh.layers.size() ? highest_z : std::max(highest_z, mesh.layers[layer_idx].printZ);
+            }
+            return highest_z;
+        };
+
     size_t n_empty_first_layers = 0;
+    coord_t hightest_empty_layer = 0;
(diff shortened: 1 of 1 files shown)


src/FffPolygonGenerator.cpp
```diff
@@ -769,12 +769,25 @@ bool FffPolygonGenerator::isEmptyLayer(SliceDataStorage& storage, const LayerInd
...
+    const auto get_highest_z = [&storage](const LayerIndex& layer_idx)
+    {
+        coord_t highest_z = 0;
+        for (std::shared_ptr<SliceMeshStorage>& mesh_ptr : storage.meshes)
+        {
+            auto& mesh = *mesh_ptr;
+            highest_z = layer_idx >= mesh.layers.size() ? highest_z : std::max(highest_z, mesh.layers[layer_idx].printZ);
+        }
+        return highest_z;
+    };
```
wawanbreton, line 781: Could you avoid creating a lambda function for using it only once ?

src/FffPolygonGenerator.cpp
```diff
@@ -769,12 +769,25 @@ bool FffPolygonGenerator::isEmptyLayer(SliceDataStorage& storage, const LayerInd
 
 void FffPolygonGenerator::removeEmptyFirstLayers(SliceDataStorage& storage, size_t& total_layers)
 {
+    const auto get_highest_z = [&storage](const LayerIndex& layer_idx)
+    {
+        coord_t highest_z = 0;
+        for (std::shared_ptr<SliceMeshStorage>& mesh_ptr : storage.meshes)
```
wawanbreton, line 775: Use a const reference ?

wawanbreton (commented):
Minor remarks (unit tests are fixed in a other branch)

src/FffPolygonGenerator.cpp
rburema: Yes, I can.

src/FffPolygonGenerator.cpp
rburema: Done!

PR #2040, opened 2024-02-29
Cura 9838 brim location

wawanbreton:
This PR adds the ability to generate a brim outside a model, inside or both.
For this to work, I had to partially rewrite the brim generation because we cannot treat single holes when they are not part of an outer shape.
The brim has interactions with the draft/ooze shields, and also with supports, so those parts were also impacted.
And finally, this opened new edge-case that were previously impossible, so more fixes were required.
In the end, this is a pretty big PR with a lot of geometric logic. I added documentation where I felt it was required, but feel free to ask for more.

Comes along with https://github.com/Ultimaker/Cura/pull/18466

### DIFF
--- a/include/SkirtBrim.h
+++ b/include/SkirtBrim.h
@@ -198,6 +198,14 @@ private:
      */
     void generateSecondarySkirtBrim(Polygons& covered_area, std::vector<Polygons>& allowed_areas_per_extruder, std::vector<coord_t>& total_length);
 
+    /*!
+     * Generate the allowed areas for each extruder. Allowed areas represent where the brim/skirt is allowed to grow
+     * while adding new outset lines. This is basically the whole build plate, removing the areas where models are
+     * located, offsetted with some specific margins.
+     *
+     * \param[in] starting_outlines The previously generated starting outlines for each extruder
+     * \return The list of allowed areas for each extruder
+     */
     std::vector<Polygons> generateAllowedAreas(const std::vector<Polygons>& starting_outlines) const;
 
 public:
--- a/include/SkirtBrim.h
+++ b/include/SkirtBrim.h
@@ -65,7 +65,7 @@ private:
         bool extruder_is_used_; //!< Whether the extruder is actually used in this print
         bool outside_polys_; //!< Whether to generate brim on the outside
         bool inside_polys_; //!< Whether to generate brim on the inside
-        coord_t line_widths_; //!< The skirt/brim line width
+        coord_t line_width_; //!< The skirt/brim line width
         coord_t skirt_brim_minimal_length_; //!< The minimal brim length
         int line_count_; //!< The (minimal) number of brim lines to generate
         coord_t gap_; //!< The gap between the part and the first brim/skirt line
@@ -197,6 +197,8 @@ private:
      * \param[in,out] total_length The total length of the brim lines for each extruder.
      */
     void generateSecondarySkirtBrim(Polygons& covered_area, std::vector<Polygons>& allowed_areas_per_extruder, std::vector<coord_t>& total_length);
+
+    std::vector<Polygons> generateAllowedAreas() const;
 
 public:
     /*!
--- a/include/SkirtBrim.h
+++ b/include/SkirtBrim.h
@@ -29,14 +29,16 @@ private:
     {
         Offset(
             const std::variant<Polygons*, int>& reference_outline_or_index,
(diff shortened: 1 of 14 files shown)


wawanbreton:
The faulty case has been fixed with 7b6906a, a margin was indeed missing for supports brim

src/SkirtBrim.cpp
```diff
@@ -370,16 +315,11 @@ Polygons SkirtBrim::getFirstLayerOutline(const int extruder_nr /* = -1 */)
...
-        }
-
-        if (has_prime_tower)
-        {
-            first_layer_outline = first_layer_outline.unionPolygons(storage_.primeTower.getGroundPoly());
+            first_layer_outline = first_layer_outline.unionPolygons(storage_.getLayerOutlines(
+                i_layer,
+                /*include_support*/ true,
+                /*include_prime_tower*/ true,
+                true));
```
casperlamboo, line 322: Intention is pretty readable, but for consistency with the rest of the code base i would write something like
```suggestion
            constexpr include_support = true;
            constexpr include_prime_tower = true;
            first_layer_outline = first_layer_outline.unionPolygons(storage_.getLayerOutlines(
                i_layer,
                include_support,
                include_prime_tower,
                true));
```

include/SkirtBrim.h
```diff
@@ -186,6 +198,16 @@ class SkirtBrim
...
 
+    /*!
+     * Generate the allowed areas for each extruder. Allowed areas represent where the brim/skirt is allowed to grow
+     * while adding new outset lines. This is basically the whole build plate, removing the areas where models are
+     * located, offsetted with some specific margins.
+     *
+     * \param[in] starting_outlines The previously generated starting outlines for each extruder
+     * \return The list of allowed areas for each extruder
+     */
+    std::vector<Polygons> generateAllowedAreas(const std::vector<Polygons>& starting_outlines) const;
```
casperlamboo, line 209: I have an inkling that this functionality is already present somewhere, did you double check this is not duplicated code?

src/SkirtBrim.cpp
```diff
@@ -223,117 +213,77 @@ std::vector<coord_t> SkirtBrim::generatePrimaryBrim(std::vector<Offset>& all_bri
...
-            }
-            else
+            const coord_t offset_value = offset.offset_value_;
+            const double area = polygon.area();
+            if ((area > 0 && offset.outside_) || (area < 0 && offset.inside_))
             {
-                brim = reference_outline->offset(offset.offset_value_, ClipperLib::jtRound);
-                newly_covered = reference_outline->offset(offset.offset_value_ + line_widths_[offset.extruder_nr_] / 2, ClipperLib::jtRound);
+                brim.add(polygon.offset(area < 0 ? -offset_value : offset_value, ClipperLib::jtRound));
             }
```
casperlamboo, line 232: I have a (slight) preference for
```suggestion
            const coord_t offset_value = offset.offset_value_;
            const double area = polygon.area();
            if (area > 0 && offset.outside_)
            {
                brim.add(polygon.offset(offset_value, ClipperLib::jtRound));
            }
            else if (area < 0 && offset.inside_)
            {
                brim.add(polygon.offset(-offset_value, ClipperLib::jtRound));
            }
```

src/SkirtBrim.cpp
wawanbreton: Indeed, I have a personal preference for this syntax, but consistency takes precedence :smiley:

include/SkirtBrim.h
wawanbreton: This is actully a specific part of code that was done in the `generate` function, but as I had to add some lines, I preferred to move it to a dedicate function.

src/SkirtBrim.cpp
wawanbreton: Well, that would also duplicate the first two lines, or I can place a second `if` after them to make it a bit more explicit.

casperlamboo: well, you did write the exact same thing here https://github.com/Ultimaker/CuraEngine/pull/2040/files#diff-696277838c64d0189f2899f11ae1efe46f331113865e2725b10fa5f8f18baef6R377-R387

wawanbreton: Oh no, that was in the previous code.. I can change it then

casperlamboo (approved):

PR #2041, opened 2024-03-05
Filter out very thin infill shapes

wawanbreton:
This is a quick fix to remove some cases where we generate infill in very thin polygons. This results in lines printed multiple times very close to each other, which causes over-extrusion.

CURA-11597


rburema:
This should probably be closed in favour of #2062 -- but re-open if you disagree :-)

PR #2042, opened 2024-03-06
[CURA-11485] Fix: z-offset as a property of gcode paths

rburema:
See (also) https://github.com/Ultimaker/CuraEngine_grpc_definitions/pull/15

NOTE: I'm not sure whether an update to the version of curaengine-plugin(s) is needed (especially the bundled one), since arguably, nothing changed to their public API. It's still the same version of the code, that would also work the old way.


wawanbreton (approved):
Looks good

PR #2050, opened 2024-03-20
Cura 11742 crash for writing travel path

saumyaj3:
# Description

Added a defensive code to avoid the crash. 
I cannot reproduce it
CURAENGINE-2F
CURA-11742


wawanbreton (approved):
lgtm

PR #2054, opened 2024-03-25
Improved fractional tree support

ThomasRahm:
# Description

Improving how fractional support and fractional support roof is generated. Areas that need fractional support are no longer extrapolated from finished support calculations, but the information about which area need fractional support is stored to be later added as such. Should be more resilient.
Intended to be an alternative to https://github.com/Ultimaker/CuraEngine/pull/2052

It also contains a fix for a bug found while testing this: Missing conversion from square microns to square mm causing roof to sometimes not being generated correctly (Commit 9cfea95a4ce3cd9942228c9990a8b1ee696ef02d)

# Possible Improvements

- If the fractional support layer is very thin (e.g. _Support Top Distance_ of 0.19mm at a layer height of 0.1mm) reducing the top-most support layer by 0.01mm would be better than printing a fractional support layer of 0.01mm height. This would require more changes to the way fractional support is handled though.

- Fractional roof/support at layer 0 is not supported as of now.

# Checklist:

### DIFF
--- a/include/TreeSupport.h
+++ b/include/TreeSupport.h
@@ -45,6 +45,32 @@ constexpr coord_t SUPPORT_TREE_COLLISION_RESOLUTION = 500; // Only has an effect
 using PropertyAreasUnordered = std::unordered_map<TreeSupportElement, Polygons>;
 using PropertyAreas = std::map<TreeSupportElement, Polygons>;
 
+struct FakeRoofArea
+{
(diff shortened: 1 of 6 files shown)


wawanbreton (approved):
Looks very good to me, even though I don't understand the small details.
FYI it will be merged only after testing by QA, which is our new process.

PR #2059, opened 2024-03-29
CURA-11717 prime tower has unused material

wawanbreton:
Until now, the prime tower used all the extruders of the printers even if they were not used nor enabled.

We now check whether an extruder is actually used during the print, and completely remove it from the prime tower generation if it's not. This way, the rings for each extruder adapt to the global extruders use.

CURA-11717

### DIFF
--- a/include/FffGcodeWriter.h
+++ b/include/FffGcodeWriter.h
@@ -295,9 +295,11 @@ private:
      *
      * \param[in] storage where the slice data is stored.
      * \param current_extruder The current extruder with which we last printed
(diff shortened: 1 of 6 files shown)


casperlamboo (approved):

saumyaj3 (approved):

rburema (approved):
The code LGTM. Do you also want a component test?

wawanbreton:
Now 3 approvals total, we should be fine :smile: 
QA will make a final test, so component testing is not required IMHO. The last commit doesn't intend to do any behavioral change.

PR #2062, opened 2024-04-03
[CURA-11597] fix duplicate support extrusions

rburema:
Gradual support infill in combination with either zig-zag or connected infill lines can cause (nearly) duplicated or too-close support extrusions. This issue can be exacerbated (or become apparent) if the lines are in the exact direction of the 'walls' (or areas actually, as the worst cases are when there are 0 walls).

This is a best-effort for a patch, but not a _complete_ fix yet. (There's at least one case still happening in a model that was full of examples.)

Also, one of the fixes I made requires the densities for the gradual infill to be recalculated. This hasn't happened yet. (+ Limited only to when zig-zag and gradual (support) infill are combined...)

### DIFF
--- a/src/infill.cpp
+++ b/src/infill.cpp
@@ -751,7 +751,7 @@ void Infill::generateLinearBasedInfill(
                 assert(scanline_idx - scanline_min_idx >= 0 && scanline_idx - scanline_min_idx < int(cut_list.size()) && "reading infill cutlist index out of bounds!");
                 cut_list[scanline_idx - scanline_min_idx].push_back(y);
                 Point2LL scanline_linesegment_intersection(x, y);
-                zigzag_connector_processor.registerScanlineSegmentIntersection(scanline_linesegment_intersection, scanline_idx, line_distance / 2);
+                zigzag_connector_processor.registerScanlineSegmentIntersection(scanline_linesegment_intersection, scanline_idx, line_distance / 4);
                 crossings_per_scanline[scanline_idx - min_scanline_index].emplace_back(scanline_linesegment_intersection, poly_idx, point_idx);
             }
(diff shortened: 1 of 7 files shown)


src/infill.cpp
```diff
@@ -746,12 +746,12 @@ void Infill::generateLinearBasedInfill(
...
             {
-                int x = scanline_idx * line_distance + shift;
-                int y = p1.Y + (p0.Y - p1.Y) * (x - p1.X) / (p0.X - p1.X);
+                const int x = scanline_idx * line_distance + shift;
+                const int y = p1.Y + (p0.Y - p1.Y) * (x - p1.X) / (p0.X - p1.X);
                 assert(scanline_idx - scanline_min_idx >= 0 && scanline_idx - scanline_min_idx < int(cut_list.size()) && "reading infill cutlist index out of bounds!");
                 cut_list[scanline_idx - scanline_min_idx].push_back(y);
                 Point2LL scanline_linesegment_intersection(x, y);
-                zigzag_connector_processor.registerScanlineSegmentIntersection(scanline_linesegment_intersection, scanline_idx);
+                zigzag_connector_processor.registerScanlineSegmentIntersection(scanline_linesegment_intersection, scanline_idx, line_distance / 4);
```
wawanbreton, line 754: Shouldn't the minimum distance be based on the line width instead ? That would allow for less skipped lines, but still ensuring no overlap.
However this value doesn't seem to be easily accessible from this part of the code, which is maybe why using line_distance is preferrable.

wawanbreton (commented):
One remark, other (minor) code changes will be applied directly in a commit

wawanbreton (approved):
Approving now as the remark is just a question for which I think I have the answer, and we need to move on with this.

PR #2067, opened 2024-04-19
Cura 11227 zseam support

saumyaj3:
Add support z-seam type to inset order optimization

The implementation of the order optimizer has been updated to account for the z-seam type for support. This includes adjustment of the start point calculation and  Both 'InsetOrderOptimizer' and 'PathOrderOptimizer' classes have been modified to accommodate these changes.

CURA-11227

### DIFF
--- a/include/PathOrderOptimizer.h
+++ b/include/PathOrderOptimizer.h
@@ -684,7 +684,7 @@ protected:
                     return best_pos;
                 }
             }
-            number_of_paths_analysed +=1;
+            number_of_paths_analysed += 1;
         }
         else
         {
--- a/include/PathOrderOptimizer.h
+++ b/include/PathOrderOptimizer.h
@@ -62,7 +62,7 @@ class PathOrderOptimizer
 public:
     using OrderablePath = PathOrdering<Path>;
     /* Areas defined here are not allowed to have the start the prints */
-    Polygons disallowed_area;
+    Polygons disallowed_area_for_seams;
     /*!
      * After optimizing, this contains the paths that need to be printed in the
      * correct order.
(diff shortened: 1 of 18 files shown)


src/settings/Settings.cpp
```diff
@@ -513,6 +513,8 @@ EZSeamType Settings::get<EZSeamType>(const std::string& key) const
         return EZSeamType::SHARPEST_CORNER;
     case "plugin"_sw:
         return EZSeamType::PLUGIN;
+    case "support"_sw:
+        return EZSeamType::SUPPORT;
```
casperlamboo, line 517: `SUPPORT` is not seam strategy. I wouldn't add it like this. Support should add a condition where the seam is not allowed to be placed. This should be combined with a steam stragegy but is not a stream strategy in it's own right.

Having disallowed areas for seam placement will be very handy in the future aswell for features like "no support in overhangs" etc.

include/PathOrderOptimizer.h
```diff
@@ -604,6 +678,52 @@ class PathOrderOptimizer
...
+
+    size_t pathIfzeamSupportIsCloseToModel(size_t best_pos, const OrderablePath& path, size_t number_of_paths_analysed)
+    {
+        size_t path_size = path.converted_->size();
+        if (path_size > number_of_paths_analysed)
+        {
+            if (! mesh_paths_.empty())
+            {
+                Point2LL current_candidate = (path.converted_)->at(best_pos);
+                if (isVertexCloseToPolygonPath(current_candidate))
```
casperlamboo, line 709: I think this check can be done simpler without introducing the `isVertexCloseToPolygonPath`. 
Take the model-polygon, extend this polygon (with a clipper offset operation) with the `support_z_seam_min_distance` value. Then simply perform a `isInside` check with the candidate point.

casperlamboo (changes requested):

src/settings/Settings.cpp
saumyaj3: removed SUPPORT type z sea here: https://github.com/Ultimaker/CuraEngine/pull/2067/commits/70e33aa13c2d81b109b6bee8a4d898f4d1659724

include/PathOrderOptimizer.h
saumyaj3: I added a function https://github.com/Ultimaker/CuraEngine/pull/2067/commits/70e33aa13c2d81b109b6bee8a4d898f4d1659724 . It creates really nice offset(blue) for the polygon(green) for disallowed area.. Thanks for suggestion :D

casperlamboo (approved):

PR #2073, opened 2024-04-26
Cura 11821 crash due to permission

saumyaj3:
# Description

writing debug.html in temp folder

writing debug.html in the current folder (in case of build the program data) makes it read only and no write modifications are allowed, hence crash. Now it is written in temp folder of the computer where writing the file is also allowed.
should solve : https://github.com/Ultimaker/Cura/issues/18835

CURA-11821

### DIFF
--- a/src/utils/SVG.cpp
+++ b/src/utils/SVG.cpp
@@ -87,7 +87,7 @@ SVG::SVG(std::string filename, AABB aabb, double scale, Point2LL canvas_size, Co
     out_ = fopen(filename.c_str(), "w");
     if (! out_)
     {
-        spdlog::error("The file %s could not be opened for writing.", filename);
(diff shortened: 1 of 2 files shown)


jellespijker (changes requested):
@saumyaj3 I think the question you need to ask is why does it even tries to execute the code surrounded by the `DEBUG` guards, while the error happens on production.
It might be that this happens because the releases are now build with RelWithDebug, I'm not sure but you need to double check this, because now it feels like you are patching up a symptom and not the root-cause.

If it is indeed happening due to an added DEBUG preprocessor flag while building Cura for production. It could also explain this Sentry issue: https://ultimaker-o7.sentry.io/issues/5077178563/

casperlamboo (approved):

Issue #1712, opened 2022-08-19
Unretract before last travel move to outer wall causes ringing


richfelker:
Without the unretract before last travel move to the outer wall, the time spend unretracting suffices to let any vibration from high travel velocity settle before the actual print moves on the outer wall begin. But with the unretract having taken place just inside the part, a high acceleration (usually perpendicular to the wall, no less) short travel move causes significant ringing, making the start of the outer wall come out a lot worse than the rest of it:

![IMG_20220817_231918847](https://user-images.githubusercontent.com/10458007/185711225-27aab12e-7bf1-4496-a547-10d1eb17ce55.jpg)

Presumably this issue is somewhat understood, as it seems to have been the motivation for adding the feature to disable travel acceleration and use the acceleration of the following wall. I do not want to do that, because high travel acceleration is key to avoiding stringing and makes a huge difference to print time. But it would make sense to always, unconditionally do this for the short travel move after unretract. I have a patch to do just that:

```
commit 09d94767e36ef9905346949611592802b304887e
Author: Rich Felker <dalias@aerifal.cx>
Date:   Thu Aug 18 03:15:38 2022 +0000

    use wall accel for move to wall after unretract

diff --git a/src/LayerPlan.cpp b/src/LayerPlan.cpp
index 122e7882b..aefcfeb22 100644
--- a/src/LayerPlan.cpp
+++ b/src/LayerPlan.cpp
@@ -1877,6 +1877,11 @@ void LayerPlan::writeGCode(GCodeExport& gcode)
                 {
                     // We need to unretract before the last travel move of the path if the next path is an outer wall.
                     gcode.writeUnretractionAndPrime();
+                    // Always use acceleration of the following extrusion so as not to introduce
+                    // ringing from high travel accel into outer wall. Normally, the unretract would
+                    // give the ringing time to settle, but here the unretract happens before the travel.
+                    if(acceleration_enabled && next_extrusion_idx < paths.size())
+                        gcode.writeTravelAcceleration(paths[next_extrusion_idx].config->getAcceleration());
                 }
                 gcode.writeTravel(path.points.back(), speed);
                 continue;
```

and it fully solves the problem for me.

[...]

PR #2075, opened 2024-04-29
Add extra travel move to outer wall start position

saumyaj3:
Associated Cura PR https://github.com/Ultimaker/Cura/pull/18988
Contributes to CURA-11830


wawanbreton:
Would probably fix https://github.com/Ultimaker/CuraEngine/issues/1712

PR #2078, opened 2024-05-10
CURA-11873 interleaved prime tower bad raft

wawanbreton:
The recent addition of the "interleaved" prime tower didn't really consider the use of the raft, which can use various extruders combinations. This PR fixes cases were some prime tower layers were completely missing while printing the raft, resulting in unusable interleaved prime tower.

The first 2 commits do some code simplification that is now possible due to the changed prime tower raft a few months ago. Before that, the prime tower had a regular raft below it, which is not the case anymore.
The last 2 commits actually handle the prime tower generation, properly taking care of the raft.

CURA-11873

### DIFF
--- a/src/raft.cpp
+++ b/src/raft.cpp
@@ -168,18 +168,29 @@ coord_t Raft::getFillerLayerHeight()
     return round_divide(getZdiffBetweenRaftAndLayer0(), getFillerLayerCount());
 }
 
-
 size_t Raft::getTotalExtraLayers()
 {
+    return getBottomLayers() + getInterfaceLayers() + getSurfaceLayers() + getFillerLayerCount();
+}
+
+size_t Raft::getBottomLayers()
+{
     const Settings& mesh_group_settings = Application::getInstance().current_slice_->scene.current_mesh_group->settings;
-    const ExtruderTrain& base_train = mesh_group_settings.get<ExtruderTrain&>("raft_base_extruder_nr");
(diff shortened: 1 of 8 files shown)


src/raft.cpp
```diff
@@ -228,5 +225,17 @@ Raft::LayerType Raft::getLayerType(LayerIndex layer_index)
     }
 }
 
+size_t Raft::getLayersAmount(const std::string& extruder_nr, const std::string& layers_nr)
```
casperlamboo, line 228: I'm not a big fan if the variable naming here
```suggestion
size_t Raft::getLayersAmount(const std::string& target_extruder_nr, const std::string& target_raft_section)
```

casperlamboo (changes requested):
I remember implementing some logic where you could determine per layer nr what raft layer is printed. This PR seems to re-implement a lot of that logic.

https://github.com/Ultimaker/CuraEngine/blob/8a3debdfff010ff1f9d45d2d5c4bcc2748f45eff/src/raft.cpp#L199-L229

Does it make sense to re-use some of that code?

wawanbreton:
> Does it make sense to re-use some of that code?

Yes it does ! I did it where it was possible, but maybe there are other parts of the code where it could also be used. I will try to keep that in mind.
I also simplified your actual implementation now that I have added specific methods to retrieve the numbers of layers.

casperlamboo (approved):

PR #2082, opened 2024-05-23
Add option to command line settings to read in resolved values

casperlamboo:
# Description

This PR adds an option to command line to slice a model file with resolved settings.

The json format of the file resolved settings is the following:
```
{
   "global": [SETTINGS],
   "extruder.0": [SETTINGS],
   "extruder.1": [SETTINGS],
   "model.stl": [SETTINGS]
}
```
where `[SETTINGS]` follow the schema
```
{
   [key: string]: bool | string | number | number[] | number[][]
}
```
There can be any number of extruders (denoted with `extruder.n`) and any number of models (denoted with `[modelname].stl`). The key of the model values will also be the filename of the relevant model, when running CuraEngine with this option the model file with that same name _must_ be in the same folder as the resolved settings json.

To test this PR I've included an [example](https://github.com/Ultimaker/CuraEngine/files/15417764/files.zip) containing the required resolved values json and model. Then, run `CuraEngine` with the following command.
```
CuraEngine slice -r ultimaker_s5.json -o benchy.gcode
```

# Checklist:

NP-205

### DIFF
--- a/include/geometry/LinesSet.h
+++ b/include/geometry/LinesSet.h
@@ -10,7 +10,6 @@
 
 #include "geometry/OpenLinesSet.h"
 #include "geometry/Point2LL.h"
-#include "geometry/Polygon.h"
 
 namespace cura
 {
--- a/include/geometry/LinesSet.h
+++ b/include/geometry/LinesSet.h
@@ -9,6 +9,7 @@
 #include <range/v3/view/drop.hpp>
 
 #include "geometry/OpenLinesSet.h"
+#include "geometry/Polygon.h"
 #include "geometry/Point2LL.h"
 
 namespace cura
--- a/include/geometry/LinesSet.h
+++ b/include/geometry/LinesSet.h
@@ -9,8 +9,8 @@
 #include <range/v3/view/drop.hpp>
 
 #include "geometry/OpenLinesSet.h"
-#include "geometry/Polygon.h"
 #include "geometry/Point2LL.h"
+#include "geometry/Polygon.h"
 
 namespace cura
 {
--- a/include/settings/types/LayerIndex.h
+++ b/include/settings/types/LayerIndex.h
@@ -8,7 +8,6 @@
 
 #include <fmt/format.h>
 
-#include "geometry/Polygon.h"
 #include "utils/types/generic.h"
 
 namespace cura
--- a/include/settings/types/LayerIndex.h
+++ b/include/settings/types/LayerIndex.h
@@ -8,8 +8,8 @@
 
 #include <fmt/format.h>
 
-#include "utils/types/generic.h"
 #include "geometry/Polygon.h"
+#include "utils/types/generic.h"
 
 namespace cura
 {
--- a/include/settings/types/LayerIndex.h
+++ b/include/settings/types/LayerIndex.h
@@ -9,6 +9,7 @@
 #include <fmt/format.h>
 
 #include "utils/types/generic.h"
+#include "geometry/Polygon.h"
 
 namespace cura
 {
--- a/include/slicer.h
+++ b/include/slicer.h
@@ -11,7 +11,6 @@
 #include "geometry/LinesSet.h"
 #include "geometry/OpenLinesSet.h"
 #include "geometry/OpenPolyline.h"
-#include "geometry/Polygon.h"
 #include "geometry/Shape.h"
 #include "settings/EnumSettings.h"
 
--- a/include/slicer.h
+++ b/include/slicer.h
@@ -11,6 +11,7 @@
 #include "geometry/LinesSet.h"
 #include "geometry/OpenLinesSet.h"
 #include "geometry/OpenPolyline.h"
+#include "geometry/Polygon.h"
 #include "geometry/Shape.h"
 #include "settings/EnumSettings.h"
 
--- a/src/MeshGroup.cpp
+++ b/src/MeshGroup.cpp
@@ -293,11 +293,8 @@ bool loadMeshIntoMeshGroup(MeshGroup* meshgroup, const char* filename, const Mat
             spdlog::info("loading '{}' took {:03.3f} seconds", filename, load_timer.restart());
             return true;
         }
-        else
-        {
-            spdlog::warn("loading '{}' failed", filename);
-            return false;
-        }
+        spdlog::warn("loading '{}' failed", filename);
+        return false;
     }
     spdlog::warn("Unable to recognize the extension of the file. Currently only .stl and .STL are supported.");
     return false;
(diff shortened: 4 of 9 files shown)


nallath (changes requested):
I think we should put some of the documentation that is now in the PR about the format somewhere in the code. Otherwise you will likely lose that information if you ever want to figure out what it's trying to do (and forcing people to reverse engineer it)

casperlamboo:
> I think we should put some of the documentation that is now in the PR about the format somewhere in the code. Otherwise you will likely lose that information if you ever want to figure out what it's trying to do (and forcing people to reverse engineer it)

Added documentation here https://github.com/Ultimaker/CuraEngine/pull/2082/commits/ed9004ae312aaf32bef5b2ff71ca0bfe412364cd

src/MeshGroup.cpp
```diff
@@ -293,6 +293,11 @@ bool loadMeshIntoMeshGroup(MeshGroup* meshgroup, const char* filename, const Mat
             spdlog::info("loading '{}' took {:03.3f} seconds", filename, load_timer.restart());
             return true;
         }
+        else
+        {
+            spdlog::warn("loading '{}' failed", filename);
+            return false;
+        }
```
jellespijker, line 300: redundant `else`

```suggestion
        spdlog::warn("loading '{}' failed", filename);
        return false;
```

jellespijker (approved):
LGTM;

Small tiny change suggestion.

include/geometry/LinesSet.h
```diff
@@ -10,6 +10,7 @@
 
 #include "geometry/OpenLinesSet.h"
 #include "geometry/Point2LL.h"
+#include "geometry/Polygon.h"
```
wawanbreton, line 13: I see no reason to include this header here (Polygon class is not even used in this header). Please include it only in the cpp files where it is actually required.

include/settings/types/LayerIndex.h
```diff
@@ -8,6 +8,7 @@
 
 #include <fmt/format.h>
 
+#include "geometry/Polygon.h"
```
wawanbreton, line 11: Very same remark here, not used in this header so should not be required.

include/slicer.h
```diff
@@ -11,6 +11,7 @@
 #include "geometry/LinesSet.h"
 #include "geometry/OpenLinesSet.h"
 #include "geometry/OpenPolyline.h"
+#include "geometry/Polygon.h"
```
wawanbreton, line 14: Same for this one

src/communication/CommandLine.cpp
```diff
@@ -586,6 +712,55 @@ void CommandLine::loadJSONSettings(const rapidjson::Value& element, Settings& se
     }
 }
 
+std::optional<std::unordered_map<std::string, std::unordered_map<std::string, std::string>>> CommandLine::readResolvedJsonValues(const std::filesystem::path& json_filename)
```
wawanbreton, line 715: Maybe make an alias for this type ?

wawanbreton (commented):
Some minor remarks, otherwise makes sense.
Could you also please use less `auto` ? You can make aliases for types that are too long, but looking to a function signature to know the variable type is not really convenient.

src/MeshGroup.cpp
wawanbreton: Is there a good reason for doing this ? I always wonder which one is better... Putting the `else` **is** indeed redundant, but it also makes sure you return something intentionally in all situations, even if you later add new `else if` conditions.

jellespijker: https://llvm.org/docs/CodingStandards.html#don-t-use-else-after-a-return

wawanbreton: Cool, thanks !

casperlamboo:
> Could you also please use less auto?

I use concrete types in all my function declarations since I see this as a form of documentation. I use auto in function bodies since 
- most IDEs show type annotations anyway
- I let the compiler decide on the type
- Am more flexible in programming since when changing a function signature no auto's need to be changed
I really see no reason to chance this..

include/settings/types/LayerIndex.h
casperlamboo: https://github.com/Ultimaker/CuraEngine/pull/2082/commits/57d357d497f7e824cb0c8526799b5ed1978fb1d5

wawanbreton:
>     * most IDEs show type annotations anyway
I don't know how your IDE shows you more information, mine offers auto-completion even with `auto` but I can't know the type just by reading the code, or when reading it outside an IDE (like in a PR), and knowing the type is quite crucial to understand/check the code.



>     * I let the compiler decide on the type
>     * Am more flexible in programming since when changing a function signature no auto's need to be changed
IMHO, you should not. I agree that this can be very convenient when changing a signature, but it can lead to your compiler not using the type you actually meant, or with different attributes (const/pointer/reference), or in the worst case, generate unexpected type conversions.



> I really see no reason to change this..

I won't push too hard for you to change it, but just pointing out that this may not be the best option overall. I read once that one should use auto only when the type can be deduced very easily, for example:
```
auto super_object = new MySuperClass();
auto other_object = reinterprect_cast<MySuperClass *>(input_object);
```
So I tend to keep following this rule. I also use it for cases when the actual type is really verbose but doesn't really matter, like for iterators in a loop:
```
for(auto iterator = my_list.begin() ; iterator != my_list.end() ; ++iterator)
{
    const MySuperClass *object = *iterator;
...
```



> > but looking to a function signature to know the variable type is not really convenient.
> not sure what you mean by this, could you elaborate?

When you see a piece of code like this:
```
const auto settings = readResolvedJsonValues(...
```
The actual type of`settings` is the return type of the function (hopefully), so the best way to know the type is to have a look at what the function returns. In worst cases, the function may also return `auto` which can become very annoying :sweat_smile: In any case, it breaks your "reading flow" and really doesn't help understanding.

I can't find an "official" recommandation about this, but what I read online tends to confirm this: use `auto` when the type is obvious or doesn't really matter, but avoid it when the type can't be easily deduced.

jellespijker:
> but it can lead to your compiler not using the type you actually meant, or with different attributes (const/pointer/reference), or in the worst case, generate unexpected type conversions.

`auto` always uses the copy operation. those different attributes therefore stil exist

```
auto ...
const auto ....
const auto& ...
auto& ...
auto* ....
```

jellespijker:
> but it can lead to your compiler not using the type you actually meant, or with different attributes (const/pointer/reference), or in the worst case, generate unexpected type conversions.

`auto` always uses the copy operation. those different attributes therefore stil exist

```
auto ...
const auto ....
const auto& ...
auto& ...
auto* ....
```

jellespijker:
also interesting read by Herb Sutter C++ ISO member

https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/

jellespijker:
and if we require specific traits or functionality the `concepts` and `requires` can help.

ensuring the type has a certain trait

src/communication/CommandLine.cpp
casperlamboo: https://github.com/Ultimaker/CuraEngine/pull/2082/commits/549f4a9fa8bf5b42bcf935138962c8319687f7b4

PR #2083, opened 2024-05-27
CURA-11930 set start temperature but dont wait

wawanbreton:
Previously, the initial temperatures commands were ignored if the temperatures were already set in the header. This causes issues because we don't actually know in what state the printer will let the nozzles after its init sequence. So we now write non-wait and wait temperatures commands based only on the settings, so that one can set them properly according to the printer init sequence.

CURA-11930

### DIFF
--- a/include/gcodeExport.h
+++ b/include/gcodeExport.h
@@ -541,7 +541,14 @@ public:
 
     void writeFanCommand(double speed);
 
-    void writeTemperatureCommand(const size_t extruder, const Temperature& temperature, const bool wait = false);
+    /*!
+     * \brief Write a GCode temperature command
+     * \param extruder The extruder number
+     * \param temperature The temperature to bo set
+     * \param wait Indicates whether we should just set the temperature and keep going, or wait for the temperature to be reach before going further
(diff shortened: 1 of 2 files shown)


include/gcodeExport.h
```diff
@@ -541,7 +541,14 @@ class GCodeExport : public NoCopy
...
 
-    void writeTemperatureCommand(const size_t extruder, const Temperature& temperature, const bool wait = false);
+    /*!
+     * \brief Write a GCode temperature command
+     * \param extruder The extruder number
+     * \param temperature The temperature to bo set
+     * \param wait Indicates whether we should just set the temperature and keep going, or wait for the temperature to be reach before going further
+     * \param force_same_temperature When true, we should set the temperature command even if the actual set temperature is the same
+     */
+    void writeTemperatureCommand(const size_t extruder, const Temperature& temperature, const bool wait = false, const bool force_same_temperature = false);
```
rburema, line 551: I think to me, `force_same_temperature` implies that some temperature value is being forced, rather than that the _writing_ happens when the temperature _is_ the same. Maybe `force_write_on_equal`?

HellAholic (approved):

PR #2086, opened 2024-05-28
Add angry new variable

wawanbreton:
This should definitely trigger a formatting warning;


wawanbreton:
Tested script is disabled, giving up

PR #2089, opened 2024-06-09
Improve handling of small areas

ThomasRahm:
# Description

I looked into the issue https://github.com/Ultimaker/Cura/issues/18397 and would be interested in feedback regarding the following opinions:

1. WallToolPaths::removeSmallLines should not remove outer walls. By my understanding it was added to prevent very small lines filling holes. An outer wall will never fill a hole.
2. In SkeletalTrapezoidation there is no functional difference between a single local maxima or multiple very close to each other. If the resulting outer wall path is so short that it not even reaches its own width, maybe it would be better to combine all these local maxima into a single one and handle it accordingly. 

I am not convinced that this is the correct way to fix it, as I think arachne should not even generate such to small paths, even if the graph consists of multiple local maxima close to each other. I just want to highlight one possible solution to avoid completely missing areas.

## Type of change

# How Has This Been Tested?

- [x] Project file linked in https://github.com/Ultimaker/Cura/issues/18397

# Checklist:

### DIFF
--- a/src/SkeletalTrapezoidation.cpp
+++ b/src/SkeletalTrapezoidation.cpp
@@ -2183,6 +2183,32 @@ void SkeletalTrapezoidation::connectJunctions(ptr_vector_t<LineJunctions>& edge_
 void SkeletalTrapezoidation::generateLocalMaximaSingleBeads()
 {
     std::vector<VariableWidthLines>& generated_toolpaths = *p_generated_toolpaths;
+    constexpr bool is_odd = true;
+
+    std::function<void(Point2LL, coord_t, size_t)> addCircleToToolpath = [&](Point2LL center, coord_t width, size_t inset_index)
+    {
+        if (inset_index >= generated_toolpaths.size())
+        {
+            generated_toolpaths.resize(inset_index + 1);
+        }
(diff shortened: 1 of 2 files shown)


rburema:
Hi @ThomasRahm ! -- 

What a coincidence, we where just about to address this issue ourselves (given the excellent writeup @casperlamboo made, it seemed at least doable...)

In any case, thanks for another contribution :-)

> I am not convinced that this is the correct way to fix it, as I think arachne should not even generate such to small paths, even if the graph consists of multiple local maxima close to each other.

That's debatable. In any case, fixing that would take a lot of time and/or introduce a lot more uncertainty -- not something we can handle with the currently (hopefully temporarily) reduced Cura-'core'-team (or at least not without pausing a lot of other things for a long time).

Even if this can be considered a hack, it's built on top of an existing one that's already in the code. As such, I think it does a _lot_ more good than harm, given the results I'm seeing.

I've just made a few very small cosmetic changes, and I think I can just send this off as the current bugfix for this particular issue. (One of the very few things still bothering me slightly, is that the circle doesn't adhere to the original shape, but the solution in place before didn't do that either...)

PR #2091, opened 2024-06-11
[CURA-8076] Split off Seam Overhang Angle from Wall Overhang Angle,

rburema:
See also the frontend: https://github.com/Ultimaker/Cura/pull/19219


wawanbreton (approved):
Looks very good !

PR #2094, opened 2024-06-12
Reduce the interleaved prime tower wasted filament

wawanbreton:
# Description

The current version of the interleaved prime tower still prints a lot of material that is only required for supporting the upper layer. However, by priming always the same extruder at the same position in the tower, the required support amount is quite high and can be reduced a lot when priming each extruder on top of each other:
image
image

This also allows for reducing the diameter of the prime tower, which can save room on the building plate.

In case of multi-color printing, this also allows for decreasing the required amount of support when the number of used extruder decreases. Thus we can end with a very thin tower at the end of the print, reducing wasted filament a lot:
image

## Type of change

# How Has This Been Tested?

- [ ] Sliced multiple models with multiple extruder configurations, for multiple printers
- [ ] No actual print made yet => TBD

**Test Configuration**:
* Operating System: Ubuntu

# Checklist:


include/PrimeTower/PrimeTowerInterleaved.h
```diff
@@ -0,0 +1,32 @@
...
+
+#ifndef PRIME_TOWER_INTERLEAVED_H
+#define PRIME_TOWER_INTERLEAVED_H
+
+#include "PrimeTower/PrimeTower.h"
+
+namespace cura
+{
+
+class PrimeTowerInterleaved : public PrimeTower
```
nallath, line 12: A bit of documentation on how this is different would be nice (mostly looking at the future here. We now know what the difference is, but future developers might not)

include/PrimeTower/PrimeTowerNormal.h
```diff
@@ -0,0 +1,30 @@
...
+
+#ifndef PRIME_TOWER_NORMAL_H
+#define PRIME_TOWER_NORMAL_H
+
+#include "PrimeTower/PrimeTower.h"
+
+namespace cura
+{
+
+class PrimeTowerNormal : public PrimeTower
```
nallath, line 12: Same as with the primetower interleaved

include/utils/LayerVector.h
```diff
@@ -0,0 +1,234 @@
...
+
+#include "raft.h"
+#include "settings/types/LayerIndex.h"
+
+namespace cura
+{
+
+/*!
+ * \brief The LayerVector class mimics a std::vector but with the index being a LayerIndex, thus it can have negative
+ *        values (for raft layers). it also ensure that the first element in the list is always on the very first layer.
```
nallath, line 17:
```suggestion
 *        values (for raft layers). It also ensure that the first element in the list is always on the very first layer.
```
Nitpicking

src/FffGcodeWriter.cpp
```diff
@@ -3986,14 +3937,12 @@ void FffGcodeWriter::setExtruder_addPrime(const SliceDataStorage& storage, Layer
 
 void FffGcodeWriter::addPrimeTower(const SliceDataStorage& storage, LayerPlan& gcode_layer, const size_t prev_extruder) const
 {
-    if (! storage.primeTower.enabled_)
+    if (storage.prime_tower_)
```
nallath, line 3940: Guess it's a personal preference (it's not one that is in our code standard as far as I'm aware) but i tend to prefer the early out option (so the return if the primetower is not set). The reason for that is that you need less indented code, which especially for more complicated code means you're going less deep.

 I'm okay with leaving it the way you have it as I don't feel that strong about it.

nallath (changes requested):

src/FffGcodeWriter.cpp
wawanbreton: You're totally right, the CuraEngine code makes use of early out most often, and I am personally not used to this, so I sometimes forget it. I'll change it for consistency.

nallath (approved):

PR #2095, opened 2024-06-13
CURA-11947 improve multimaterial raft

wawanbreton:


rburema (changes requested):
I've found ~a few~ a thing ~s~ that we'd really want to change:

Tree support is printed with what seems to be 100% infill after this change for some reason:

filled_tree

---

This 2nd thing is also occurring in main (so no need to fix, at least for these tickets) I just wanted to point it out:

The other thing I noticed is with the infill overlap. Everything seems to work fine when the infill is connected, or when there are extra infill wall(s)*1, but when I disable 'connect infill lines' it seems that there's too much overlap.

With connect infill lines (🆗):
overlap_not_too_far_connected

Without connect infill lines (single outer wall only half as thick to show the effect, but the overlap percentage is 20%, so you wouldn't expect it to overtake all of the '''50%''' wall

overlap_too_far

*1) In which case the infill overlaps with the infill wall ... I'm not sure if that's what I expect from a setting called 'Infill Overlap Percentage', especially if you can have a different extruder for the infill, but the writeup in the ticket seems to imply that this is what we want, so that's OK.

rburema (approved):

PR #2096, opened 2024-06-14
Remove the debug defines when building with RelWithDebInfo

wawanbreton:
The variables `DEBUG`, `USE_CPU_TIME` and `ASSERT_INSANE_OUTPUT` were defined also when building in mode RelWithDebInfo, which is used when building the production installers. This leads to some debug code being executed by customers, with potential crashes and decreased performances.

Contributes to CURA-11821

### DIFF
--- a/include/Application.h
+++ b/include/Application.h
@@ -105,9 +105,17 @@ protected:
 #endif // ARCUS
 
     /*!
-     * \brief Print the header and license to the stderr channel.
+     * \brief Print the header to the stderr channel.
+     */
+    void printHeader() const;
+
+    /*!
+     * \brief Print the license to the stderr channel.
      */
     void printLicense() const;
+
+    /*! Print a big warning when the code has been build in debug mode, otherwise just issue a small nice message. */
+    void printDebugWarning() const;
 
     /*!
      * \brief Start slicing.
--- a/include/Application.h
+++ b/include/Application.h
@@ -114,9 +114,6 @@ protected:
      */
     void printLicense() const;
 
-    /*! Print a big warning when the code has been build in debug mode, otherwise just issue a small nice message. */
-    void printDebugWarning() const;
-
     /*!
      * \brief Start slicing.
      * \param argc The number of arguments provided to the application.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -191,9 +191,6 @@ target_compile_definitions(_CuraEngine
         $<$<CONFIG:Debug>:ASSERT_INSANE_OUTPUT>
         $<$<CONFIG:Debug>:USE_CPU_TIME>
         $<$<CONFIG:Debug>:DEBUG>
-        $<$<CONFIG:RelWithDebInfo>:ASSERT_INSANE_OUTPUT>
-        $<$<CONFIG:RelWithDebInfo>:USE_CPU_TIME>
-        $<$<CONFIG:RelWithDebInfo>:DEBUG>
 )
 
 enable_sanitizers(_CuraEngine)
--- a/src/Application.cpp
+++ b/src/Application.cpp
@@ -148,10 +148,49 @@ void Application::printHelp() const
     fmt::print("\n");
 }
 
-void Application::printLicense() const
+void Application::printHeader() const
 {
     fmt::print("\n");
     fmt::print("Cura_SteamEngine version {}\n", CURA_ENGINE_VERSION);
+
+#ifdef DEBUG
+    fmt::print("\n");
+    fmt::print(" _______   ________  _______   __    __   ______\n");
+    fmt::print("/       \\ /        |/       \\ /  |  /  | /      \\\n");
+    fmt::print("███████  |████████/ ███████  |██ |  ██ |/██████  |\n");
+    fmt::print("██ |  ██ |██ |__    ██ |__██ |██ |  ██ |██ | _██/\n");
+    fmt::print("██ |  ██ |██    |   ██    ██< ██ |  ██ |██ |/    |\n");
+    fmt::print("██ |  ██ |█████/    ███████  |██ |  ██ |██ |████ |\n");
+    fmt::print("██ |__██ |██ |_____ ██ |__██ |██ \\__██ |██ \\__██ |\n");
+    fmt::print("██    ██/ ██       |██    ██/ ██    ██/ ██    ██/\n");
+    fmt::print("███████/  ████████/ ███████/   ██████/   ██████/\n");
+    fmt::print("\n");
+    fmt::print(" __       __   ______   _______   ________\n");
+    fmt::print("/  \\     /  | /      \\ /       \\ /        |\n");
+    fmt::print("██  \\   /██ |/██████  |███████  |████████/\n");
+    fmt::print("███  \\ /███ |██ |  ██ |██ |  ██ |██ |__\n");
+    fmt::print("████  /████ |██ |  ██ |██ |  ██ |██    |\n");
+    fmt::print("██ ██ ██/██ |██ |  ██ |██ |  ██ |█████/\n");
+    fmt::print("██ |███/ ██ |██ \\__██ |██ |__██ |██ |_____\n");
+    fmt::print("██ | █/  ██ |██    ██/ ██    ██/ ██       |\n");
+    fmt::print("██/      ██/  ██████/  ███████/  ████████/\n");
+    fmt::print("\n");
+
+    fmt::print("#########################################################\n");
+    fmt::print("#########################################################\n");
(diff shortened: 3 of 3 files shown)


jellespijker (approved):
LGTM

rburema:
So are we building in that mode to get the(/more useful) Sentry info?

I'm not about to request a change, since I approve (regardless of Sentry), because it's nice to have semi-debug-able release code (especially since we're OSS and therefore have little to hide if anything), but it does make me wonder what prompted us to do that in the first place.

rburema (approved):
Nice ASCII art :-)

rburema:
Now that I'm testing this, since we often have a large output, maybe it'd be a good idea to print it at the end as well (not instead, because of crashes, just 'as well').

include/Application.h
```diff
@@ -105,10 +105,18 @@ class Application : NoCopy
...
+     */
+    void printHeader() const;
+
+    /*!
+     * \brief Print the license to the stderr channel.
      */
     void printLicense() const;
 
+    /*! Print a big warning when the code has been build in debug mode, otherwise just issue a small nice message. */
+    void printDebugWarning() const;
```
rburema, line 118: Is this used anywhere?

jellespijker:
> So are we building in that mode to get the(/more useful) Sentry info?
> 
> I'm not about to request a change, since I approve (regardless of Sentry), because it's nice to have semi-debug-able release code (especially since we're OSS and therefore have little to hide if anything), but it does make me wonder what prompted us to do that in the first place.

One word: "Windows"

That was the only way we could get sensible stack changes for 70% of our users.

include/Application.h
wawanbreton: Hem no, I added the function first but didn't use it and forgot to remove it :expressionless:

wawanbreton:
> Now that I'm testing this, since we often have a large output, maybe it'd be a good idea to print it at the end as well (not instead, because of crashes, just 'as well').

That's a good idea, so that you're quite sure not to miss it. But it may be a bit annoying in practice, because we (I) often look at the output once it has run, to see if it ended properly or if there is something important to note. With that, we would only see the message each time and have to go up to see anything interesting. So, not sure.

rburema:
> > Now that I'm testing this, since we often have a large output, maybe it'd be a good idea to print it at the end as well (not instead, because of crashes, just 'as well').
> 
> That's a good idea, so that you're quite sure not to miss it. But it may be a bit annoying in practice, because we (I) often look at the output once it has run, to see if it ended properly or if there is something important to note. With that, we would only see the message each time and have to go up to see anything interesting. So, not sure.

Yes, the same thing that makes the 2nd message useful is also the one that'll make it annoying 😅 I can see that. -- And we'd only have to check once for each release build anyway 🤔 -- so we can probably do without the 2nd message and only add it if it ever becomes a problem not to.

wawanbreton:
> And we'd only have to check once for each release build anyway 🤔

Not even sure we have to actively check it. It is here more as a safety, like when you get a user log file. If the message is present, you won't miss it and know directly that something is wrong.

rburema (approved):
~~~

Cura_SteamEngine version 5.8.0-alpha.0+source

 _______   ________  _______   __    __   ______
/       \ /        |/       \ /  |  /  | /      \
ÔûêÔûêÔûêÔûêÔûêÔûêÔûê  |ÔûêÔûêÔûêÔûêÔûêÔûêÔûêÔûê/ ÔûêÔûêÔûêÔûêÔûêÔûêÔûê  |ÔûêÔûê |  ÔûêÔûê |/ÔûêÔûêÔûêÔûêÔûêÔûê  |
ÔûêÔûê |  ÔûêÔûê |ÔûêÔûê |__    ÔûêÔûê |__ÔûêÔûê |ÔûêÔûê |  ÔûêÔûê |ÔûêÔûê | _ÔûêÔûê/
ÔûêÔûê |  ÔûêÔûê |ÔûêÔûê    |   ÔûêÔûê    ÔûêÔûê< ÔûêÔûê |  ÔûêÔûê |ÔûêÔûê |/    |
ÔûêÔûê |  ÔûêÔûê |ÔûêÔûêÔûêÔûêÔûê/    ÔûêÔûêÔûêÔûêÔûêÔûêÔûê  |ÔûêÔûê |  ÔûêÔûê |ÔûêÔûê |ÔûêÔûêÔûêÔûê |
ÔûêÔûê |__ÔûêÔûê |ÔûêÔûê |_____ ÔûêÔûê |__ÔûêÔûê |ÔûêÔûê \__ÔûêÔûê |ÔûêÔûê \__ÔûêÔûê |
ÔûêÔûê    ÔûêÔûê/ ÔûêÔûê       |ÔûêÔûê    ÔûêÔûê/ ÔûêÔûê    ÔûêÔûê/ ÔûêÔûê    ÔûêÔûê/
ÔûêÔûêÔûêÔûêÔûêÔûêÔûê/  ÔûêÔûêÔûêÔûêÔûêÔûêÔûêÔûê/ ÔûêÔûêÔûêÔûêÔûêÔûêÔûê/   ÔûêÔûêÔûêÔûêÔûêÔûê/   ÔûêÔûêÔûêÔûêÔûêÔûê/

 __       __   ______   _______   ________
/  \     /  | /      \ /       \ /        |
ÔûêÔûê  \   /ÔûêÔûê |/ÔûêÔûêÔûêÔûêÔûêÔûê  |ÔûêÔûêÔûêÔûêÔûêÔûêÔûê  |ÔûêÔûêÔûêÔûêÔûêÔûêÔûêÔûê/
ÔûêÔûêÔûê  \ /ÔûêÔûêÔûê |ÔûêÔûê |  ÔûêÔûê |ÔûêÔûê |  ÔûêÔûê |ÔûêÔûê |__
ÔûêÔûêÔûêÔûê  /ÔûêÔûêÔûêÔûê |ÔûêÔûê |  ÔûêÔûê |ÔûêÔûê |  ÔûêÔûê |ÔûêÔûê    |
ÔûêÔûê ÔûêÔûê ÔûêÔûê/ÔûêÔûê |ÔûêÔûê |  ÔûêÔûê |ÔûêÔûê |  ÔûêÔûê |ÔûêÔûêÔûêÔûêÔûê/
ÔûêÔûê |ÔûêÔûêÔûê/ ÔûêÔûê |ÔûêÔûê \__ÔûêÔûê |ÔûêÔûê |__ÔûêÔûê |ÔûêÔûê |_____
ÔûêÔûê | Ôûê/  ÔûêÔûê |ÔûêÔûê    ÔûêÔûê/ ÔûêÔûê    ÔûêÔûê/ ÔûêÔûê       |
ÔûêÔûê/      ÔûêÔûê/  ÔûêÔûêÔûêÔûêÔûêÔûê/  ÔûêÔûêÔûêÔûêÔûêÔûêÔûê/  ÔûêÔûêÔûêÔûêÔûêÔûêÔûêÔûê/

#########################################################
#########################################################
## WARNING: This version of CuraEngine has been built  ##
## in developper mode. This may impact performances,   ##
## provoke unexpected results or crashes.              ##
## If you downloaded an official version of CuraEngine ##
## and see this message, please report the issue.      ##
#########################################################
#########################################################

Copyright (C) 2024 Ultimaker
~~~



I'm going to approve, since I think the message is clear either way, but both this and the tree suffer from the fact that ASCII is only properly defined for the 1st 128 characters instead of the full 256 ... the rest are these 'codepages' that can be different.

wawanbreton:
> I'm going to approve, since I think the message is clear either way, but both this and the tree suffer from the fact that ASCII is only properly defined for the 1st 128 characters instead of the full 256 ... the rest are these 'codepages' that can be different.

Oh that really doesn't look good... If I'm correct, this character is part of the "extended ASCII codes" which I assumed would be universal enough. But maybe a UTF-8 character would be better ? Otherwise I can use a standard ASCII character, it will just be slightly less nice :smiling_face_with_tear:

rburema:
> If I'm correct, this character is part of the "extended ASCII codes" which I assumed would be universal enough.

The problem is that 'extended ASCII' can mean a number of things 😓 

According to wikipedia:

> There is no formal definition of "extended ASCII", and even use of the term is sometimes criticized,[[1]](https://en.wikipedia.org/wiki/Extended_ASCII#cite_note-1)[[2]](https://en.wikipedia.org/wiki/Extended_ASCII#cite_note-2)[[3]](https://en.wikipedia.org/wiki/Extended_ASCII#cite_note-3)

I think either solution is fine. (Note that this also affects the otherwise nice tree-representation-output you made for another purpose in the engine -- that isn't quite as disruptive as this and wasn't meant as 100% cosmetic, so I didn't mention it.)

PR #2100, opened 2024-06-19
[CURA-11974] better non vertex seams

rburema:
Split off from: https://github.com/Ultimaker/CuraEngine/pull/2087

For user defined seam locations, the last ticket left us in a situation where it was calculated to be the closest point on the polygon. While this is technically correct in a sense, this would not make the newly made vertices (made for the seam) line up 'correclty'. Instead specify a line from the origin of the model and intersect.

Also, while a new vertex was made, the rest of the algorithm would still occasionally pick other vertices. Probably for very defensible reason(s) when dealing with arbitrary vertices, but not wanted when we explicitly make a vertex at the specified location.

### DIFF
--- a/include/utils/linearAlg2D.h
+++ b/include/utils/linearAlg2D.h
@@ -79,8 +79,8 @@ public:
      *
      * \return Whether the two line segments intersect.
      */
-    static bool segmentSegmentIntersection(const Point2LL& p1, const Point2LL& p2, const Point2LL& p3, const Point2LL& p4, float* t, float* u);
-    static bool lineLineIntersection(const Point2LL& p1, const Point2LL& p2, const Point2LL& p3, const Point2LL& p4, float* t, float* u);
+    static bool segmentSegmentIntersection(const Point2LL& p1, const Point2LL& p2, const Point2LL& p3, const Point2LL& p4, float& t, float& u);
+    static bool lineLineIntersection(const Point2LL& p1, const Point2LL& p2, const Point2LL& p3, const Point2LL& p4, float& t, float& u);
 
     static bool lineLineIntersection(const Point2LL& a, const Point2LL& b, const Point2LL& c, const Point2LL& d, Point2LL& output);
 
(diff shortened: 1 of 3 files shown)


include/utils/linearAlg2D.h
```diff
@@ -67,6 +67,20 @@ class LinearAlg2D
...
+     * \param p1 The start point of the first line segment.
+     * \param p2 The end point of the first line segment.
+     * \param p3 The start point of the second line segment.
+     * \param p4 The end point of the second line segment.
+     * \param t The parameter of the intersection on the first line segment (intersection = p1 + t * (p2 - p1)).
+     * \param u The parameter of the intersection on the second line segment (intersection = p3 + u * (p4 - p3)).
+     *
+     * \return Whether the two line segments intersect.
+     */
+    static bool lineSegmentLineSegmentIntersection(const Point2LL& p1, const Point2LL& p2, const Point2LL& p3, const Point2LL& p4, float* t, float* u);
```
wawanbreton, line 82: Couldn't this be merged with `lineLineIntersection`, or make one use the other ? I would say that in your implementation, you do calculate the lines intersection, then check whether the points are inside the segments.

wawanbreton (commented):
Minor suggestions, feel free to decline them.

include/utils/linearAlg2D.h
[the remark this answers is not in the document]
rburema: Well, you can have circle-segments, for example :-)

But this will become a moot point anyway if I take your next suggestion.

rburema: I've had a bit of a closer look at `lineLineIntersection`, and now I'm not sure why I did the other thing in the 1st place 😅 

I'll adapt this.

rburema: done!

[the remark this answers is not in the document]
wawanbreton: Hmm for me that would be an arc then, no more a segment. But apparently I'm wrong, there can be such thing as a "circle segment". I'll have to change my definition :sweat_smile:

wawanbreton:
I took the liberty of making a last small modification, that IMHO makes the code more readable, and agree with [our own standards](https://github.com/Ultimaker/Meta/blob/master/languages/c%2B%2B_const_pointers_and_references.md#advise-regarding-references).

wawanbreton (approved):

PR #2110, opened 2024-07-02
Fix crash when using tangent on angles with PI/2 value

wawanbreton:
std::tan with a value of PI/2 works but returns a huge number, which will crash subsequent processings (like offsetting a polygon with this value). A function has been added to process a "safe tangent" calculation that return a big but not huge number.
There was previously a similar safety in the code, but it was completely inefficient...

CURA-11804

Note: unit tests are fixed in an other branch (not merged at the time of writing this)


rburema (approved):
lgtm :-)

PR #2111, opened 2024-07-02
CURA-6410 buggy fan speed interpolation

wawanbreton:
The checking for the fan miminum layer time was based on the wrong variable.
I also reworked the fan speed command to better handle interpolated values.


src/gcodeExport.cpp
```diff
@@ -1434,44 +1434,73 @@ void GCodeExport::writeFanCommand(double speed, std::optional<size_t> extruder)
...
+        bool write_value = true;
+        std::ostringstream new_value;
+        new_value << PrecisionedDouble{ (should_scale_zero_to_one ? static_cast<uint8_t>(2) : static_cast<uint8_t>(1)), (should_scale_zero_to_one ? speed : speed * 255) / 100 };
+        const std::string new_value_str = new_value.str();
+        if (current_fan_speed.has_value())
         {
-            *output_stream_ << " P" << fan_number;
+            std::ostringstream old_value;
+            old_value << PrecisionedDouble{ (should_scale_zero_to_one ? static_cast<uint8_t>(2) : static_cast<uint8_t>(1)),
+                                            (should_scale_zero_to_one ? current_fan_speed.value() : current_fan_speed.value() * 255) / 100 };
```
rburema, line 1475: I'm not sure I like this almost repeated line. Perhaps make this into a little closure?

src/gcodeExport.cpp
```diff
@@ -1434,44 +1434,73 @@ void GCodeExport::writeFanCommand(double speed, std::optional<size_t> extruder)
...
-        *output_stream_ << "M107";
-        if (fan_number)
+
+        if (write_value)
         {
-            *output_stream_ << " P" << fan_number;
+            // Check if the value to be written is only made with zeroes, in which case it is actually a turn off
+            std::string new_value_str_reduced = new_value_str;
+            std::erase(new_value_str_reduced, '0');
+            std::erase(new_value_str_reduced, '.');
```
rburema, line 1485: Depending on the printer, values can range from 0.0-1.0 or 0-255 -- I think this would 'reduce' something like `205` to `25`, which isn't what you'd want.

Also not a fan (pun not originally intended) of the way we handle something numeric via string-manipulation here.

rburema: Also will reduce `0.5` to just `5`.

rburema (changes requested):
I've got 2 remarks. -- Most importantly, ideally, I'd really like the string-manipulation to be gone, but currently, I don't even understand how that would work correctly and keep the right numbers 😅

The (other) remark about DRY is just a preference and I can live with it if not easy to fix.

Lastly, I do think the 1st commit cleans things up considerably though -- I can at least understand what's going on now there 😄

src/gcodeExport.cpp
wawanbreton: I can do some factorization indeed. Now what do you mean by "make this into a little closure" ?

src/gcodeExport.cpp
wawanbreton: Well, that's precisely the point. I think the variable is not properly named. The goal of this is to identify anything that would turn into a numeric "0" value, would it be "0", "0.0" or "0.00". If the numeric value is e.g. 0.02, turning it to a one-digit precision string would be "0.0" so in this case I want to issue a M107 command and not M106.

I'm also not a big fan of doing it via string manipulation, but this is a way to ensure that you will never ever end up with a "M106 0.0".

PR #2113, opened 2024-07-04
[CURA-11887] Fuzzy Skin Outside Only: Make definition of 'inside' user-settable

rburema:
Previously, for the Fuzzy Skin Outside Only setting, the definition of 'inside' was only areas that are fully enclosed within a layer.

With these changes (see also [the necessary added settings in the front-end](https://github.com/Ultimaker/Cura/pull/19324)), it is now possible to define 'sufficiently convex' areas as 'inside' as well -- skipping these areas when jittering the outer-wall.

The new settings are:

- A measure of the 'convexity' of a near-hole in an area; the ratio of the circumference of that hole 'exposed' to the outside versus the total circumference of that hole.
- A limit on how small such a hole area considered for exclusion needs to be at minimum (_given_ that the hole is exposed to the outside, fully enclosed holes will never be considered outside, no matter how small).

fuzzy_skin_8_1073

fuzzy_skin_8_117

### DIFF
--- a/src/FffPolygonGenerator.cpp
+++ b/src/FffPolygonGenerator.cpp
@@ -1081,6 +1081,7 @@ void FffPolygonGenerator::processFuzzyWalls(SliceMeshStorage& mesh)
         return;
     }
 
+    constexpr coord_t epsilon = 5; // the distance between two points that are considered to be the same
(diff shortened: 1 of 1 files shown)


wawanbreton (approved):
Lgtm

PR #2114, opened 2024-07-06
3 new Multitool configuration options

TheSin-:
# Description

- Add bool to set if the start_gcode needs to be the first gcode run.  Requires `machine_start_gcode_first` be added to the machine UI as a checkbox.  Having `T0` called before the start gcode on a changer isn't wise as the start code needs to initialize the system and level before a tool change happens.  This is an option to correct that behaviour for those that require it.
- Add a prestart_gcode to extruders, this will allow use set temperatures before the change call so we can preheat the next extruder before the change call to save change times.  Requires `machine_extruder_prestart_code` text box be added to the extruder UI page.
- Add a time duration to calculate time required to change extruders. Requires `machine_extruder_change_duration` float box to be added to the extruder UI page.
- Fix a bug in the export code that was checking `num_extruders` instead of `used_extruders`, this was adding a tool change on start to T0 when only one extruder is used which we do not want, with one extruder it should output like normal and let the user choose the extruder they want to use.

**Required for https://github.com/Ultimaker/Cura/pull/19333**

## Type of change

# How Has This Been Tested?

**NOTE not tested yet as I'm still learning how to build a custom engine into a full build**

Not tested yet, I need to slice and check the resulting gcode, it should move start gcode before the T0 call, Should add the restart gcode before all toolchanges (This one might need an exemption on the first T0).

**Test Configuration**:
* Operating System: MacOS

# Checklist:

### DIFF
--- a/src/gcodeExport.cpp
+++ b/src/gcodeExport.cpp
@@ -787,6 +787,7 @@ void GCodeExport::processInitialLayerExtrudersTemperatures(const SliceDataStorag
     };
 
     std::vector<ExtruderInitialize> all_extruders;
+    std::vector<bool> extruders_used = storage.getExtrudersUsed();
     for (size_t extruder_nr = 0; extruder_nr < extruders_used.size(); ++extruder_nr)
     {
         if (extruders_used[extruder_nr])
--- a/src/gcodeExport.cpp
+++ b/src/gcodeExport.cpp
@@ -682,6 +682,11 @@ bool GCodeExport::initializeExtruderTrains(const SliceDataStorage& storage, cons
 
     writeComment("Generated with Cura_SteamEngine " CURA_ENGINE_VERSION);
 
+    if (mesh_group_settings.get<bool>("machine_start_gcode_first"))
+    {
+        writeCode(mesh_group_settings.get<std::string>("machine_start_gcode").c_str());
+    }
+
     if (getFlavor() == EGCodeFlavor::GRIFFIN)
     {
         std::ostringstream tmp;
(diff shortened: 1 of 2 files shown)


nallath:
Looks good to me. Have you been able to test if this works already?

TheSin-:
I will be testing it this weekend, thanks for looking at it.

TheSin-:
finally figured out how to build it (not the proper way but enough for testing) and my code does not work at all

I'm still getting gcode before the START

```
;Generated with Cura_SteamEngine 5.8.0-alpha.0+source
T0
M104 S260
M105
M109 S260
PRINT_PRESTART BED_TEMP=110.0 CHAMBER_TEMP=70
PRINT_START WZ=0.2 WN=0.4 BED_TEMP=110.0 EXTRUDER_TEMP=260.0
M82 ;absolute extrusion mode
```

The M82 was moved down but the rest I need to figure out.

My Pre Change code isn't expanding

```
G92 E0
M104 S{material_print_temperature} T{extruder_nr}
T0
```

And I broke it as it always calling T0 even when it should be calling T1

SO I have stuff to figure out but at least now I can work on it, I'll push fixes as soon as I have it working.

TheSin-:
Tested and working

TheSin-:
Updated for the latest code changes and retested

nallath:
Sorry for that! I missed the notification that you fixed & tested the engine part.

TheSin-:
No problem, I noticed it wasn't in 5.9 did it just miss the Freeze?

rburema:
(Also) For future reference -- please check if the unit-tests actually work. We've got that ✔️ / ❌ for a reason.

In this case, it's because for the tests, the settings used there will need sensible default values -- in the relevant test these are defined in `TEST_F(GCodeExportTest, SwitchExtruderSimple)`.

I'll make a commit in main after I'll finish this comment.

rburema:
@TheSin- Yeah just about, we entered string & feature freeze on the 8th of October. -- Somewhat related: Please also note that for most of this year the Cura-team is working with a reduced team-size (which will hopefully clear up sometime early next year...)

TheSin-:
> (Also) For future reference -- please check if the unit-tests actually work. We've got that ✔️ / ❌ for a reason 😉
> 
> In this case, it's because for the tests, the settings used there will need sensible default values -- in the relevant test these are defined in `TEST_F(GCodeExportTest, SwitchExtruderSimple)`.
> 
> I'll make a commit in main after I'll finish this comment.

do you have a link to the preferred method of running the tests, I'm more then happy to do so but I figured the CI/CD would do that for me

rburema:
@TheSin- The CI/CD does do that -- you can click on the ❌ to see what went wrong -- I'm just saying that we should be more careful what we merge.

If you want to run any of them locally, they should already have been built in `./build/<release-mode>/tests` unless you build with `skip_tests`.

PR #2116, opened 2024-07-09
CURA-11967 fix material print temp prepend not being disabled

wawanbreton:
We would previously still append the temperatures within some conditions. Now we completely skip it if the setting is disabled, considering that it means the temperatures are handled in the start G-Code and not te be treated here at all.

CURA-11967

Fixes https://github.com/Ultimaker/Cura/issues/19204


rburema (approved):
lgtm!

PR #2118, opened 2024-07-10
[CURA-11875] Fix Tree support going through the model

rburema:
Just clip the offending part out if it intersects too much with the model.

### DIFF
--- a/src/TreeSupport.cpp
+++ b/src/TreeSupport.cpp
@@ -2043,7 +2043,8 @@ void TreeSupport::filterFloatingLines(std::vector<Shape>& support_layer_storage)
                 }
                 else if (hole.intersection(PolygonUtils::clipPolygonWithAABB(relevant_forbidden, hole_aabb)).area() > hole.length() * EPSILON)
                 {
-                    holes_resting_outside[layer_idx].emplace(idx); // technically not resting outside, also not valid, but the alternative is potentially having lines go though the model
+                    holes_resting_outside[layer_idx].emplace(
+                        idx); // technically not resting outside, also not valid, but the alternative is potentially having lines go though the model
                 }
                 else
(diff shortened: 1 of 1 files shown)


src/TreeSupport.cpp
```diff
@@ -2023,10 +2023,8 @@ void TreeSupport::filterFloatingLines(std::vector<Shape>& support_layer_storage)
                 return;
             }
 
-            Shape outer_walls = TreeSupportUtils::toPolylines(support_layer_storage[layer_idx - 1].getOutsidePolygons())
-                                    .createTubeShape(
-                                        closing_dist,
-                                        0); //.unionPolygons(volumes_.getCollision(0, layer_idx - 1, true).offset(-(config.support_line_width+config.xy_min_distance)));
+            const Shape relevant_forbidden = volumes_.getCollision(0, layer_idx, true);
```
wawanbreton, line 2026: `getCollision` returns a reference, so `relevant_forbidden` could also be a reference and save a copy.

wawanbreton (changes requested):
I don't fully understand the code, so just doing a syntax check.
Just one optimization remark, otherwise looks good.

src/TreeSupport.cpp
rburema: It's just a code sanitization thing, if we don't fix it now we'll never touch it 😛

It's incredibly low-risk -- does anyone mind if I push this change directly to 5.8? It should change nothing except speed up the tree stuff a tiny bit, since we avoid a copy.

wawanbreton: I would say make a quick test and push to 5.8 :+1:

rburema: Done!

PR #2121, opened 2024-07-17
CURA-10914 Add ability to register multiple modify plugins

wawanbreton:
The plugins will be called one after the other, in the order they were registered.

CURA-10914

### DIFF
--- a/include/plugins/slotproxy.h
+++ b/include/plugins/slotproxy.h
@@ -40,7 +40,7 @@ class SlotProxy
 {
     Default default_process{};
     using value_type = PluginProxy<SlotID, SlotVersion, Stub, ValidatorTp, RequestTp, ResponseTp>;
-    std::optional<value_type> plugin_{ std::nullopt };
+    std::vector<value_type> plugins_;
 
 public:
     static constexpr plugins::v0::SlotID slot_id{ SlotID };
@@ -53,14 +53,15 @@ public:
     SlotProxy() noexcept = default;
 
     /**
-     * @brief Constructs a SlotProxy object with a plugin.
-     *
-     * Constructs a SlotProxy object and initializes the plugin using the provided gRPC channel.
-     *
+     * @brief Adds a plugin to this proxy.
+     * @param name The fully-qualified name of the plugin
+     * @param version The full verison of the plugin
      * @param channel A shared pointer to the gRPC channel for communication with the plugin.
      */
-    SlotProxy(const std::string& name, const std::string& version, std::shared_ptr<grpc::Channel> channel)
-        : plugin_{ value_type{ name, version, channel } } {};
+    void addPlugin(const std::string& name, const std::string& version, std::shared_ptr<grpc::Channel> channel)
+    {
+        plugins_.emplace_back(name, version, channel);
+    }
 
     /**
      * @brief Executes the plugin operation.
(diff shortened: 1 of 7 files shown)


rburema (commented):

| Benchmark suite | Current: 5327c0286f1421e1f44aba57e58736acc3c77239 | Previous: a3e668b6966030b34ae7e3f8f1933f9275f6bf22 | Ratio |
|-|-|-|-|
| `SimplifyTestFixture/simplify_slot_noplugin` | `4.021668022608884` ns/iter | `1.2388981465436513` ns/iter | `3.25` |

include/plugins/slotproxy.h
```diff
@@ -75,18 +76,25 @@ class SlotProxy
...
         return std::invoke(default_process, std::forward<decltype(args)>(args)...);
     }
 
     constexpr auto modify(auto& original_value, auto&&... args)
     {
-        if (plugin_.has_value())
+        if (! plugins_.empty())
         {
-            return plugin_.value().modify(original_value, std::forward<decltype(args)>(args)...);
+            auto modified_value = original_value;
```
rburema, line 90: I think the efficiency here can be improved a bit by making this a reference (or maybe rename `original_value`), since I don't think we need the original `original_value` for anything after we're done (or during).

rburema, line 90:
```suggestion
            auto& modified_value = original_value;
```

wawanbreton: Well, then it would actually make sense to not return it at all, but input a reference and directly apply the changes on it, what do you think ?

rburema: Sort-of?

One the one hand, my suggestion is a simple change, that already works without going the full mile, since it avoids the extra copies while constructing the `modified` value and iterating over it copying the return to it each time (though the compiler _might_ optimize some of that).

On the other hand, it _is_ actually building up a new 'return' value (initially empty) in the `modify` call as well, which we might improve by moving/swapping it into the 'original' in the end -- that does require some work though, and may require re-building existing plugins, if we thread this change all the way down.

wawanbreton: I made a commit that seems to work, and actually avoids a few copies, which is a good thing because the objects in question can be very big (list of gcode paths of a whole layer).

However I have a suspicion of possible behavior change in `converters.cpp`. Tell me if you catch something obvious, otherwise I'll share more details :smiley:

src/plugins/converters.cpp
```diff
@@ -482,10 +483,8 @@ gcode_paths_modify_response::native_value_type
                           })
                     | ranges::to_vector;
 
-        paths.emplace_back(path);
+        value.push_back(path);
```
rburema, line 486: This can still be `emplace_back` right?

rburema (approved):
Looks good, I tested it with the 2 plugins again, and same sort of outputs (they both work on the gcode).

src/plugins/converters.cpp
rburema: Wasn't an _actual_ emplace before, so this is a better signaling of use.

rburema:
Closed in favour of https://github.com/Ultimaker/CuraEngine/pull/2123
which is the same, but without my requested changes.

PR #2123, opened 2024-07-24
[CURA-10914] Add ability to register multiple modify plugins

rburema:
The plugins will be called one after the other, in the order they were registered.

This version basically reverts the changes I wanted when reviewing https://github.com/Ultimaker/CuraEngine/pull/2121 since that screws up the benchmark (for non benchmark related reasons but it still shows that the API is different which scares a little).

It's safer to do it without the requested optimization.

### DIFF
--- a/include/plugins/slotproxy.h
+++ b/include/plugins/slotproxy.h
@@ -40,7 +40,7 @@ class SlotProxy
 {
     Default default_process{};
     using value_type = PluginProxy<SlotID, SlotVersion, Stub, ValidatorTp, RequestTp, ResponseTp>;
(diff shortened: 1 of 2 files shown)


rburema (commented):

| Benchmark suite | Current: 5327c0286f1421e1f44aba57e58736acc3c77239 | Previous: a3e668b6966030b34ae7e3f8f1933f9275f6bf22 | Ratio |
|-|-|-|-|
| `SimplifyTestFixture/simplify_slot_noplugin` | `3.7209843033742755` ns/iter | `1.2388981465436513` ns/iter | `3.00` |

PR #2129, opened 2024-08-05
Add simple full on functionality for aux fan

nallath:
CURA-11849

See also
https://github.com/Ultimaker/Cura/pull/19481


jellespijker (approved):
LGTM;

PR #2131, opened 2024-08-07
NP- 327 emscripten communication

jellespijker:
# Description

Add an extra communication option when using Emscripten and compiling for wasm. This is based on the cmdline communication layer and uses a callback function to be executed on the JS side, providing slice information and possibly in the future preview information.

## Type of change

# How Has This Been Tested?

- [X] Locally UT, for Arcus, Commandline and Emscripten.
- [X] In the CuraEngineJS

**Test Configuration**:
* Operating System: wasm

# Checklist:


src/communication/CommandLine.cpp
```diff
@@ -204,14 +193,12 @@ void CommandLine::sliceNext()
...
-                else if (argument.find("--progress") == 0)
+                else if (argument.starts_with("--progress_cb") || argument.starts_with("--slice_info_cb"))
                 {
-                    // Store progress handler name
                     argument_index++;
                     argument = arguments_[argument_index];
-                    progressHandler = argument;
                 }
-#endif
+#endif // __EMSCRIPTEN__
```
wawanbreton, line 201: Shouldn't this part be moved to the dedicated class ?

src/communication/EmscriptenCommunication.cpp
```diff
@@ -0,0 +1,111 @@
...
+    {
+        const double value = FffProcessor::getInstance()->getTotalFilamentUsed(static_cast<int>(extruder_nr));
+        spdlog::info("Extruder {} used {} [mm] of filament", extruder_nr, value);
+        rapidjson::Value extruder_id(fmt::format("{}", extruder_nr).c_str(), allocator);
+        rapidjson::Value extruder_material_estimate(value);
+        material_estimates_json.AddMember(extruder_id, extruder_material_estimate, allocator);
+    }
+    doc.AddMember("material_estimates", material_estimates_json, allocator);
+
+    // Set CureEngine information
```
wawanbreton, line 90:
```suggestion
    // Set CuraEngine information
```

stress_benchmark/stress_benchmark.cpp
```diff
@@ -195,7 +195,7 @@ void createAndWriteJson(const std::filesystem::path& out_file, double stress_lev
 {
     rapidjson::Document doc;
     doc.SetArray();
-    rapidjson::Document::AllocatorType& allocator = doc.GetAllocator();
+        rapidjson::Document::AllocatorType& allocator = doc.GetAllocator();
```
wawanbreton, line 198: This added indentation looks wrong :thinking:

wawanbreton (commented):
Oooooooooh
Nice improvement, some minor remarks but apart from that I do like it very much :+1:

src/Application.cpp
```diff
@@ -214,8 +215,11 @@ void Application::slice()
     {
         arguments.emplace_back(argv_[argument_index]);
     }
-
-    communication_ = new CommandLine(arguments);
+#ifdef __EMSCRIPTEN__
+    communication_ = std::make_shared<EmscriptenCommunication>(arguments);
+#else
+    communication_ = std::make_shared<CommandLine>(arguments);
+#endif
```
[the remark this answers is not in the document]
jellespijker: I don't see a reason why, there is after all a tight coupling with https://github.com/Ultimaker/CuraEngineJS/pull/5 
Maybe a third-party would want to run it with just a command line interface as wasm binary, be we won't.

src/communication/CommandLine.cpp
jellespijker: _This is me, trying very hard not to rewrite our custom command line parsing logic_. I normally agree with you. But solving this properly would be more work, this was the quickest fix.

I am tempted to use cxxopts https://github.com/jarro2783/cxxopts an open a community PR one of these days.

jellespijker: Let me know how strong you feel about letting this through

src/Application.cpp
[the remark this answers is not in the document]
wawanbreton: It's just to make sure that you won't later get into a situation where you would actually want to do it. But if it really does not make sense, then don't allow it.

src/communication/CommandLine.cpp
wawanbreton: Well it just feels sad that you took a lot of effort to split the Emscripten part out of CommandLine, and yet you still have this dirty part remaining there :smiley:
You could solve this easily by having a virtual method to parse extra arguments, that would return false by default and would parse those in the `EmscriptenCommunication` overridden definition.

jellespijker: Yes, it feels dirty, Let's see if I can manage to find some time to make the changes for a cleaner solution

0x5844:
https://github.com/Ultimaker/CuraEngine/pull/2131/commits/6697713ce16a799646e093fb10866eb70684f2d6 is commited to handle [NP-343](https://ultimaker.atlassian.net/browse/NP-343) "Slicing for a method XL produces unexpected results" where the generated gcode was empty. The code change is under #2134

wawanbreton:
There has been many commits on this PR recently, is it ready for review again ? Unit tests are failing though.

jellespijker:
@wawanbreton This now also integrate https://github.com/Ultimaker/CuraEngine/pull/2134 since we're building our Working package of this branch

include/communication/EmscriptenCommunication.h
```diff
@@ -0,0 +1,65 @@
...
+ * \brief A class for handling communication in an Emscripten environment.
+ *
+ * This class extends the CommandLine class and provides specific implementations
+ * for sending progress and handling slice information in an Emscripten environment.
+ */
+class EmscriptenCommunication : public CommandLine
+{
+private:
+    std::string progress_handler_; ///< Handler for progress messages.
+    std::string gcode_prefix_handler_; /// <Handler for getting gcode_prefix.
```
wawanbreton, line 24: The "prefix" naming is a bit confusing, if I'm correct we mostly use "header" for this, in the front-end and in the engine (although it seems inconsistent in the engine yet). Would you mind renaming the methods/variables to "header" ?

wawanbreton (commented):
Just one nitpicky remark about naming, which got me confused. Otherwise looks good.

wawanbreton (approved):

0x5844:
@wawanbreton @jellespijker I removed the type cast; I gave it additional tests and it seems that the `WASM_BIGINT` flag is sufficient to handle slicing issues with method xl.

include/communication/EmscriptenCommunication.h
```diff
@@ -0,0 +1,65 @@
...
+     * \brief Sends GcodeHeader
+     */
+    void sendGCodePrefix(const std::string& prefix) const override;
+
+    /**
+     * \brief Initiates the slicing of the next item.
+     */
+    void sliceNext() override;
+
+    bool isSequential() const
```
0x5844, line 56:
```shell
EmscriptenCommunication.h:56:10: warning: 'isSequential' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
   56 |     bool isSequential() const
      |          ^
CommandLine.h:56:10: note: overridden virtual function is here
   56 |     bool isSequential() const override;
      |          ^
1 warning generated.
```

wawanbreton:
What is the status of this PR ? There has been a few significant changes since the last approval (which I would ask you to avoid as much as possible), is it now stable enough for a final review ?

jellespijker:
Yes it can be reviewed as final

casperlamboo:
> @wawanbreton @jellespijker I removed the type cast; I gave it additional tests and it seems that the WASM_BIGINT flag is sufficient to handle slicing issues with method xl.

With these proposed changes I am still able to reproduce the method xl slicing issues. I think the cast is still needed for wasm.

0x5844:
> > @wawanbreton @jellespijker I removed the type cast; I gave it additional tests and it seems that the WASM_BIGINT flag is sufficient to handle slicing issues with method xl.
> 
> With these proposed changes I am still able to reproduce the method xl slicing issues. I think the cast is still needed for wasm. (FYI @0x5844)

That's strange, if we want to keep that cast we have to adjust it in all it's occurrences (I don't think this is the way to go about it). Currently there are multiple changes to how we compile wasm (flags passed to the compiler); I would much  rather wait for the merge and have a fresh test again. Removing the typecast while enabling the compiler flag seemed to tackle to issue (?) which seem to be no longer the case (this was tested last week).

wawanbreton (approved):

PR #2133, opened 2024-08-16
CURA-12096 move gradual flow plugin back into the engine core

wawanbreton:
This is just the code of the gradual flow plugin imported to the CuraEngine core, so that using it becomes easier. Most of the code has been adapted to match the CuraEngine coding standards, but it is not perfect.


include/LayerPlan.h
```diff
@@ -778,6 +778,12 @@ class LayerPlan : public NoCopy
      */
     void applyBackPressureCompensation();
 
+    /*!
+     * If enabled, applies the gradual flow acceleration splitting, that improves printing quality when printing at very high speed
+     * with a bowden extruder.
```
casperlamboo, line 783: Nitpicking, but direct drive also benefits from this, although to a lesser extend.

src/LayerPlan.cpp
```diff
@@ -645,8 +646,6 @@ void LayerPlan::addPolygonsByOptimizer(
 
 static constexpr double max_non_bridge_line_volume = MM2INT(100); // limit to accumulated "volume" of non-bridge lines which is proportional to distance x extrusion rate
 
-static int i = 0;
```
casperlamboo, line 648: oops, my bad

include/gradual_flow/FlowLimitedPath.h
```diff
@@ -0,0 +1,452 @@
...
+{
+    STABLE,
+    TRANSITION,
+    UNDEFINED
+};
+
+struct FlowLimitedPath
+{
+    const GCodePath* original_gcode_path_data;
+    geometry::Polyline<> points;
```
casperlamboo, line 33: Can't we use CuraEngine's path data type here. It made sense to introduce a new path data type in the GradualFlow repo since there was no path data struct defined there yet. But in the curaengine repo we could just re-use the existing one. The whole `PointContainer` class is a bit redundant here.

include/LayerPlan.h
wawanbreton: Ok, didn't know that, I'll refine the comment

include/gradual_flow/FlowLimitedPath.h
wawanbreton: Well, yes we could, and we should, but I'm afraid that will break a lot of things, and be a much bigger work... I may be wrong though, so I'll give it a try and see where it takes me.

wawanbreton: Done, it turned out to be very easy, because the geometry concepts were actually used exclusively as points sets.

casperlamboo (approved):

Issue #2005, opened 2024-01-05
Potentially incorrect function `ceil_divide_signed`


chungwong:
**Application Version**
Versions including this commit https://github.com/Ultimaker/CuraEngine/commit/3b34b69277

It looks like [this function](https://github.com/Ultimaker/CuraEngine/commit/3b34b69277) is logically incorrect?

```c++
inline unsigned int ceil_divide_signed(int dividend, int divisor) //!< Return dividend divided by divisor rounded up towards positive infinity.
{
    return (dividend / divisor) + (dividend * divisor > 0 ? 1 : 0);
}

// it returns 31 instead of 30
ceil_divide_signed(90, 3);
```
Is this beahviour expected?

0x5844:
The pr that addresses this issue is now merged!

PR #2135, opened 2024-08-21
refactor/improve-math-functions

0x5844:
# Description

This pull request addresses a critical issue in the `math.h` utility functions related to incorrect rounding behavior in the `ceil_divide_signed` function. Specifically, it fixes a bug where the function would incorrectly round up even when the division was exact, such as returning `31` instead of `30` for `ceil_divide_signed(90, 3)` -> (https://github.com/Ultimaker/CuraEngine/issues/2005)

Additionally, the PR includes the following improvements:

- Error Handling: Added exception handling to ensure division by zero is appropriately managed across all relevant functions.
- Refactoring: Refactored functions like `round_divide_signed` and `floor_divide_signed` to clarify logic, improve readability, and maintain consistency.
- Documentation: Enhanced inline documentation for better clarity and maintainability.

## Type of change

# How Has This Been Tested?

- [x] Unit Testing: Updated and ran unit tests for all modified functions.
- [x] Code Review.

**Test Configuration**:
* Operating System: Ubuntu 22.04 LTS

# Checklist:

### DIFF
--- a/include/utils/math.h
+++ b/include/utils/math.h
@@ -13,38 +13,125 @@
 namespace cura
 {
 
+/**
+ * @brief Returns the square of a value.
+ *
+ * @tparam T A multipliable type (arithmetic types such as int, float, double, etc.)
+ * @param a The value to be squared.
+ * @return T The square of the input value.
+ */
 template<utils::multipliable T>
 [[nodiscard]] T square(const T& a)
 {
     return a * a;
 }
 
-[[nodiscard]] inline int64_t round_divide_signed(const int64_t dividend, const int64_t divisor) //!< Return dividend divided by divisor rounded to the nearest integer
+/**
+ * @brief Returns the quotient of the division of two signed integers, rounded to the nearest integer.
+ *
+ * @param dividend The numerator.
+ * @param divisor The denominator (must not be zero).
+ * @return int64_t The result of the division rounded to the nearest integer.
+ * @throws std::invalid_argument If the divisor is zero.
+ */
+[[nodiscard]] inline int64_t round_divide_signed(const int64_t dividend, const int64_t divisor)
 {
-    if ((dividend < 0) ^ (divisor < 0)) // Either the numerator or the denominator is negative, so the result must be negative.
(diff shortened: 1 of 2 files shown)


rburema (approved):
Thanks for this! You know the current team-size, so we really appreciate these PR's 😄 

--

I've made a few changes, most notably, I've removed the exception throwing after a discussion with Erwan:

Throwing exceptions means you rely on the code around that to be set up for that, which it currently is not. If we only want to know where it crashed, then we can just rely on SIGFPE to be signalled on division-by-zero (which is almost the only case where FPE happens anyway).

If we want to use exceptions in the code in general, we should have that discussion, but introducing this into a PR that fixes other stuff is not a good idea (especially in the same commit). -- Even then I'm on the fence on these particular functions, which should be basic math-ops.

0x5844:
> Thanks for this! You know the current team-size, so we really appreciate these PR's 😄
> 
> --
> 
> I've made a few changes, most notably, I've removed the exception throwing after a discussion with Erwan:
> 
> Throwing exceptions means you rely on the code around that to be set up for that, which it currently is not. If we only want to know where it crashed, then we can just rely on SIGFPE to be signalled on division-by-zero (which is almost the only case where FPE happens anyway).
> 
> If we want to use exceptions in the code in general, we should have that discussion, but introducing this into a PR that fixes other stuff is not a good idea (especially in the same commit). -- Even then I'm on the fence on these particular functions, which should be basic math-ops.

That sounds great! I was on the fence about throwing exceptions but I went forward with it.

rburema:
> That sounds great! I was on the fence about throwing exceptions but I went forward with it.

That's what we have reviews for!

Dumping some info here since I'm off until next Tuesday:
- It seems some of the actions are failing because of a '/' in the branch-name.
- I've moved this to QA (who're in charge of the actual merging these days), but they might not be able to test if they've not got any builds.

0x5844:
> Dumping some info here since I'm off until next Tuesday:

Let me know if I can help with anything!

PR #2138, opened 2024-08-26
Np 351 header containing extruder info

saumyaj3:
# Description

Test


wawanbreton (approved):
In order to improve readability, I would suggest to put the result of `slice->scene.mesh_groups[mesh_group_index]` into a local variable, but that's really for the sake of making a remark :smile:

PR #2088, opened 2024-06-05
Fix Tree-support lines overlapping with the model

ThomasRahm:
# Description

Fixes https://github.com/Ultimaker/Cura/issues/18970 by keeping holes if they overlap with the model.
Technically this does cause floating lines, but that is in my opinion better than the alternative.
There could be an argument made to only apply this if the `support density` is > 0%, but I do not know if there could be other issues caused by model areas being inside of a support area.

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

- [x] [Project file in issue 18970](https://github.com/Ultimaker/Cura/issues/18970#issuecomment-2081523866)

**Test Configuration**:
* Operating System: Windows 10

# Checklist:

- [x] My code follows the style guidelines of this project as described in [UltiMaker Meta](https://github.com/Ultimaker/Meta)
- [x] I have read the [Contribution guide](https://github.com/Ultimaker/Cura/blob/main/contributing.md)
- [x] I have commented my code, particularly in hard-to-understand areas

PR #2145, opened 2024-10-09
[CURA-12153] Fix 'encompassing hole' issue (tree support).

rburema:
When a hole is large enough to wholly contain (sometimes many) other branches, those also get cut. This prevents that. The solution isn't ideal, as there is an offset happening somewhere, but it's better than just missing the top of trees.

### DIFF
--- a/src/TreeSupport.cpp
+++ b/src/TreeSupport.cpp
@@ -2093,6 +2093,17 @@ void TreeSupport::filterFloatingLines(std::vector<Shape>& support_layer_storage)
             if (! found)
             {
                 next_removed_holes_by_idx.emplace(idx);
+
+                // Individual pieces of the hole could still be valid (if the 'hole' is made by branches surrounding others' for instance).
+                for (const auto& poly : hole)
+                {
+                    if (poly.area() < 0)
+                    {
+                        auto poly_copy = poly;
+                        poly_copy.reverse();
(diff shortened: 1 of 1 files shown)


wawanbreton (approved):

ThomasRahm:
@rburema 
I can confirm the bug this pull request fixes. But I think it was caused by git!
I vaguely remember I the version with  
`else if (hole.intersection(PolygonUtils::clipPolygonWithAABB(relevant_forbidden, hole_aabb)).area() > hole.length() * EPSILON)`, but it had issues, which is why I replaced it with the slower but more accurate  
`else if (! hole.intersection(PolygonUtils::clipPolygonWithAABB(relevant_forbidden, hole_aabb)).offset(-config.xy_min_distance / 2).empty())`
Which is present in the pull request https://github.com/Ultimaker/CuraEngine/pull/2088
How the former one found its way back into the code-base I do not know.
If only the latter one is used I don't have any issues, even without this workaround.

Bonus performance improvement: 
If  
`&& ! hole.intersection(hole2).empty()) // TODO should technically be outline: Check if this is fine either way as it would save an offset`  
is replaced with  
`&& ! hole.intersection(PolygonUtils::clipPolygonWithAABB(hole2, hole_aabb)).empty()) // TODO should technically be outline: Check if this is fine either way as it would save an offset`
It is about twice as fast for me (15 instead of 53 seconds for the drawing part).

PR #2146, opened 2024-10-10
Add engine info handler for getting engine version and hash

saumyaj3:
Implemented a handler for the `--engine_info_cb` flag to handle engine information retrieval. Added functionality to create and send engine info messages, which include the Cura engine version and hash, to Emscripten. Updated CMake and conan files to pass Cura engine hash.

NP-349

### DIFF
--- a/conandata.yml
+++ b/conandata.yml
@@ -1,4 +1,4 @@
-version: "5.9.0-alpha.0"
+version: "5.9.0-beta.1"
 commit: "unknown"
 requirements:
   - "scripta/0.1.0@ultimaker/testing"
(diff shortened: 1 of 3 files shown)


jellespijker (changes requested):
LGTM;

But you might want to base it and merge it to `5.9` instead of `main` also ask @wawanbreton or @rburema for a review

wawanbreton:
Do you need this in 5.9 ? If not, it is preferable to keep it in main.

jellespijker:
> Do you need this in 5.9 ? If not, it is preferable to keep it in main.

Neoprep will be based on 5.9, so yes we need it in the release branch. @saumyaj3 needs to base it from `5.9` it that scenario.

I don't think there will be any risk for Cura with merging it since the changes are in the Emscripten communication class, which will not be compiled for desktop applications

saumyaj3:
I changed it to 5.9 branch

wawanbreton (approved):
Looks good

jellespijker:
@saumya did you also base it from 5.9? otherwise we will pull in main into 5.9

saumyaj3:
I started working from main. The package created is also from main. I can also recreate the packages

PR #2148, opened 2024-10-16
CURA-12164 Fix innerwall seam position

wawanbreton:
This PR actually contains a few different works:
* Rework of the score calculation for the seam point selection. Previously, the calculated scores were inconsistent and depending on each other when they should not, making it very difficult to add new scoring criteria. The scoring is now more consistent, hopefully easier to read, and much easier to tweak.
* The overhang avoidance for the seam placement was previously done after the actual seam placement, making it impossible to calculated the seam for inner walls related to the correct outer wall seam position. The overhang avoidance is now integrated in the global score calculation for the seam placement.
* The seam calculation strategy for inner walls is now forcibly set to "shortest" when outer walls are printed first, so that the starting point of the inner wall will be located as close as possible to the starting point of the outer wall.
* Rework of the user-set position for the seam, which is now treated as other strategies so that we can find a proper fallback point when the original one is on overhang.

:warning: This introduces some behavior changes of the existing seam selection strategy. Although they are more logical (no more depending on hidden settings) and more explainable, they do change the results of situations people probably got used to.

### DIFF
--- a/include/LayerPlan.h
+++ b/include/LayerPlan.h
@@ -299,6 +299,11 @@ public:
     void setSeamOverhangMask(const Shape& polys);
 
     /*!
+     * Get the areas that are considered having air below, which is a unite between bridge and overhang masks
+     */
+    Shape getAirBelowMask() const;
+
+    /*!
      * Set roofing_mask.
      *
      * \param polys The areas of the part currently being processed that will require roofing.
@@ -676,50 +681,6 @@ public:
         int last_seam_vertex_idx,
         const bool is_top_layer,
         const bool is_bottom_layer);
-
-
-    /*!
-     * Given a wall polygon and a start vertex index, return the index of the first vertex that is supported (is not above air)
-     *
-     * Uses bridge_wall_mask and overhang_mask to determine where there is air below
-     *
-     * \param wall The wall polygon
-     * \param start_idx The index of the starting vertex of \p wall
-     * \return The index of the first supported vertex - if no vertices are supported, start_idx is returned
-     */
-    template<typename T>
-    size_t locateFirstSupportedVertex(const T& wall, const size_t start_idx) const
-    {
-        if (bridge_wall_mask_.empty() && seam_overhang_mask_.empty())
-        {
-            return start_idx;
-        }
-
-        const auto air_below = bridge_wall_mask_.unionPolygons(seam_overhang_mask_);
-
-        size_t curr_idx = start_idx;
-
-        while (true)
-        {
-            const Point2LL& vertex = cura::make_point(wall[curr_idx]);
-            if (! air_below.inside(vertex, true))
-            {
-                // vertex isn't above air so it's OK to use
-                return curr_idx;
-            }
-
-            if (++curr_idx >= wall.size())
-            {
-                curr_idx = 0;
-            }
-
-            if (curr_idx == start_idx)
-            {
-                // no vertices are supported so just return the original index
-                return start_idx;
-            }
-        }
-    }
 
     /*!
      * Write the planned paths to gcode
--- a/include/LayerPlan.h
+++ b/include/LayerPlan.h
@@ -299,9 +299,9 @@ public:
     void setSeamOverhangMask(const Shape& polys);
 
     /*!
-     * Get the areas that are considered having air below, which is a union between bridge and overhang masks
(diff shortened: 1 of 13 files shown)


rburema (approved):
Looks good! Let's get this in beta :-)

include/PathOrderOptimizer.h
```diff
@@ -180,6 +182,23 @@ class PathOrderOptimizer
             }
         }
 
+        // Set actual used start point calculation strategy for each path
+        const OrderablePath* previous_path = nullptr;
+        for (auto& path : paths_)
+        {
+            if (use_shortest_for_inner_walls_ && previous_path && previous_path->is_outer_wall && ! path.is_outer_wall)
```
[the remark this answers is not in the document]
wawanbreton: Well I think we don't really care about starting position for anything else than outer walls, but that is a good question :thinking: I also remember we have a ticket to set the start position of infill to shortest distance, so that may actually tend to it. Let's PPM give an opinion on this.

include/LayerPlan.h
```diff
@@ -298,6 +298,11 @@ class LayerPlan : public NoCopy
      */
     void setSeamOverhangMask(const Shape& polys);
 
+    /*!
+     * Get the areas that are considered having air below, which is a unite between bridge and overhang masks
```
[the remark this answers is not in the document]
wawanbreton: Definitely better :sweat_smile:

include/PathOrderOptimizer.h
```diff
@@ -759,81 +740,124 @@ class PathOrderOptimizer
...
+        const bool calculate_distance_score
+            = ! calculate_forced_pos_score
+           && (path.seam_config_.type_ != EZSeamType::SHARPEST_CORNER || path.seam_config_.corner_pref_ == EZSeamCornerPrefType::Z_SEAM_CORNER_PREF_NONE);
+
+        const bool calculate_corner_score
+            = ! calculate_forced_pos_score
+           && (path.seam_config_.type_ == EZSeamType::SHARPEST_CORNER
+               && (path.seam_config_.corner_pref_ != EZSeamCornerPrefType::Z_SEAM_CORNER_PREF_NONE && path.seam_config_.corner_pref_ != EZSeamCornerPrefType::PLUGIN));
+
+        const bool calculate_random_score = ! calculate_forced_pos_score && (path.seam_config_.type_ == EZSeamType::RANDOM);
```
casperlamboo, line 756: I need to do a double (triple) read on the `!flag && condition` pattern. I find this simpler to read;
```suggestion
        bool calculate_distance_score = false;
        bool calculate_corner_score = false;
        bool calculate_random_score = false;
        if (! calculate_forced_pos_score)
        {
            calculate_distance_score = path.seam_config_.type_ != EZSeamType::SHARPEST_CORNER || path.seam_config_.corner_pref_ == EZSeamCornerPrefType::Z_SEAM_CORNER_PREF_NONE;

             calculate_corner_score
            = path.seam_config_.type_ == EZSeamType::SHARPEST_CORNER
               && (path.seam_config_.corner_pref_ != EZSeamCornerPrefType::Z_SEAM_CORNER_PREF_NONE && path.seam_config_.corner_pref_ != EZSeamCornerPrefType::PLUGIN);

             calculate_random_score = path.seam_config_.type_ == EZSeamType::RANDOM;
        }
```

casperlamboo (approved):

include/PathOrderOptimizer.h
wawanbreton: It is indeed, but then you loose the constness on the bools, so I'm shared

wawanbreton: I wish I could say in C++ "from now on, this variable is const" that is a use-case I regularly come up with

rburema (commented):

| Benchmark suite | Current: b0cf410da985e750e8bc506d657454886c1b1caa | Previous: 10faafa8b9dc4d7fc4fbb690f9960cec2a46a3a8 | Ratio |
|-|-|-|-|
| `SimplifyTestFixture/simplify_slot_noplugin` | `14.536420650004192` ns/iter | `4.024045969205645` ns/iter | `3.61` |

wawanbreton:
> Possible performance regression was detected for benchmark **'C++ Benchmark'**.

I do not see such a difference on my PC, and the code being run for this test hasn't been touched. For now I'll assume this was a hiccup of the runner, and will keep an eye on this.

PR #2152, opened 2024-10-22
Alternative fix for 'encompassing hole' tree support issue

ThomasRahm:
# Description

This pull request does 3 things:
1. It reverts https://github.com/Ultimaker/CuraEngine/pull/2145 and adds an alternative solution for what it intended.
I think explicitly excluding holes with branches inside them from being removes is better than re-adding the branches later.
In the best case if the branches are added back, there is no difference between the hole being removed or not. Worst case something breaks:
If I understood the change there correctly it adds all branches in a hole aka negative-holes in a positive-hole, as a positive-hole, which later is then reversed and added back to the support area as a negative-hole. This should mean that if support density is larger than 0 this fix will cause confused people wondering why suddenly the area around the branches is filled with support infill, but the branch itself is not:

grafik

This is what it looks with this pull request:

grafik

(Why the branch paths are different between these two screenshots is an investigation for another day)

2. It correctly adds the change from https://github.com/Ultimaker/CuraEngine/pull/2088
How the other solution, where I tried to solve it without an offset found its way back I do not understand. As seen by the example file given in the pull request https://github.com/Ultimaker/CuraEngine/pull/2145 it sadly does not work and can cause inconsistent hole removal.

3. While debugging I noticed that this function takes longer than I would have expected. I found a small change that drastically increases the performance in this test-case. While it is not related to the things above, I don't see any reason to not include it.

## Type of change

# How Has This Been Tested?

- [X] File from https://github.com/Ultimaker/CuraEngine/pull/2145 its changes back-ported to a 5.8 branch, as I currently have issues compiling 5.9.

# Checklist:

### DIFF
--- a/src/TreeSupport.cpp
+++ b/src/TreeSupport.cpp
@@ -2023,8 +2023,10 @@ void TreeSupport::filterFloatingLines(std::vector<Shape>& support_layer_storage)
                 return;
             }
 
-            const Shape& relevant_forbidden = volumes_.getCollision(0, layer_idx, true);
-            Shape outer_walls = TreeSupportUtils::toPolylines(support_layer_storage[layer_idx - 1].getOutsidePolygons()).createTubeShape(closing_dist, 0);
+            Shape outer_walls = TreeSupportUtils::toPolylines(support_layer_storage[layer_idx - 1].getOutsidePolygons())
+                                    .createTubeShape(
+                                        closing_dist,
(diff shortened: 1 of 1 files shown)


ThomasRahm:
The diff reads a bit confusing. It does not handle me adding an if and converting the if after it to an else if particularly well, so here a visual aid:
grafik


If the diff confuses me (who has written the code) for like 5 to 10 minutes, anyone reviewing it could also be confused by it.

rburema (approved):
Thanks! The screenshots of the diff where helpful, it made it clear how small this fix is :-)

PR #2154, opened 2024-10-23
[CURA-12065] Fix (sort-of) conflict between per-object and per-mesh retraction.

rburema:
We don't really have a set way to deal with settings that can both be per-extruder and per-mesh (sometimes called per-object in the engine) -- in general, building something like that would be a (new) feature (gap). This represents a workaround for a current issue where the overall settings are taken instead of the per-extruder ones, even though in the frontend, no per-mesh settings are overridden. Which happened because it always prefers the per-mesh settings over the per-extruder ones, and since the per-mesh settings where not overridden, it was just the 'top' ones. If you then overrode less then all of the extruders with a specific setting, it would default back to the 'top' settings (but if all extruders where overridden, that counted as not overridden again w.r.t per-extruder).

This commit 'fixes' that by assuming that if you have any setting different per extruder, you want that rather than the per mesh ones. This isn't necessarily the case (and also definitely not as granular as can be, it takes all retraction and wipe settings as one....) but it may do the job for now.

### DIFF
--- a/include/sliceDataStorage.h
+++ b/include/sliceDataStorage.h
@@ -323,6 +323,9 @@ public:
     std::shared_ptr<LightningGenerator> lightning_generator; //!< Pre-computed structure for Lightning type infill
 
     RetractionAndWipeConfig retraction_wipe_config; //!< Per-Object retraction and wipe settings.
+    bool override_extruder_retract_settings = true; //!< The 'per-object' retraction and wipe settings are currently always filled, even if not overridden.
+                                                    //!< This has potential conflicts with per-extruder retract/wipe settings.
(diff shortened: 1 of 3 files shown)


wawanbreton (commented):
This really seems hacky to me :grimacing: I have the feeling that it just moves the issue to an other place. Does it actually fix some cases ? If yes, then I would be ok to integrate it, but with a follow-up ticket to do a bigger refactoring.

rburema:
> Does it actually fix some cases?

It does fix the example case in the jira-ticket. I'm not sure if that's from an actual user though, or if QA made it.

rburema:
Mothballed.

rburema:
Closed in favour of https://github.com/Ultimaker/CuraEngine/pull/2191

PR #2155, opened 2024-10-23
CURA-12175 Fix scarf seam and coasting interaction

wawanbreton:
A few things were incorrect in the current pre-calculation of the coasting settings. I also merged the pre-calculation with the post-calculation so that we don't calculate anything twice, which should make the whole thing faster. Now the second part just applied what the first part has calculated.

CURA-12175

### DIFF
--- a/src/LayerPlan.cpp
+++ b/src/LayerPlan.cpp
@@ -1244,7 +1244,7 @@ std::vector<LayerPlan::PathCoasting>
 
         for (const auto& reversed_chunk : paths | ranges::views::enumerate | ranges::views::reverse
                                               | ranges::views::chunk_by(
-                                                  [](const auto& path_a, const auto& path_b)
+                                                  [](const auto&path_a, const auto&path_b)
                                                   {
                                                       return (! std::get<1>(path_a).isTravelPath()) || std::get<1>(path_b).isTravelPath();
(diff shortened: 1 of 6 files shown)


src/LayerPlan.cpp
```diff
@@ -1227,6 +1227,109 @@ void LayerPlan::addSplitWall(
...
+                    if (! chunk_coasting_point_reached && path_coasting.apply_coasting == ApplyCoasting::NoCoasting)
+                    {
+                        if (accumulated_volume >= coasting_volume)
+                        {
+                            const double start_pos_factor = (accumulated_volume - coasting_volume) / segment_volume;
+                            Point3LL coasting_start_pos = cura::lerp(previous_point, point, start_pos_factor);
+                            path_coasting = { ApplyCoasting::PartialCoasting, point_idx, coasting_start_pos };
+                            chunk_coasting_point_reached = true;
+                        }
+                        else if (point_idx == 0)
```
rburema, line 1296: This could do with a comment -- it's not really clear that we're looking at the end here because it's easy to miss that we reversed the path.

wawanbreton:
Performance spike got back to normal as I expected. We had a similar case last week, quite annoying :thinking:

PR #2158, opened 2024-10-28
Apply scarf seam and speed gradient in surface mode

wawanbreton:
CURA-12176

The addSplitWall method is now a template so that it can handle either ExtrusionLine or Polygon objects. It also takes a function as parameter because the way we add extrusion lines is not the same for surface mode and normal mode.
In order to simplify this function, I added a PathAdapter class to work with different types of paths with more transparency.

I am not completely satisfied of this fix. The initial decision I took of handling the scarf seam while adding the extrusion moves now appears very bad to me. This should have been a LayerPlan post-processing function similar to what gradual flow does. I suggest adding a ticket to refactor this, or I fear we will discover other bad interactions between scarf seam, speed gradient and gradual flow, and in the future if we want to add more similar behaviors, it will really be a mess.

### DIFF
--- a/src/LayerPlan.cpp
+++ b/src/LayerPlan.cpp
@@ -2482,7 +2482,7 @@ void LayerPlan::writeGCode(GCodeExport& gcode)
 
             for (const auto& reversed_chunk : paths | ranges::views::enumerate | ranges::views::reverse
                                                   | ranges::views::chunk_by(
-                                                      [](const auto& path_a, const auto& path_b)
+                                                      [](const auto&path_a, const auto&path_b)
                                                       {
                                                           return (! std::get<1>(path_a).isTravelPath()) || std::get<1>(path_b).isTravelPath();
                                                       }))


src/LayerPlan.cpp
```diff
@@ -1407,6 +1456,68 @@ void LayerPlan::addWall(
...
+        settings,
+        default_config,
+        flow_ratio,
+        always_retract,
+        is_closed,
+        is_reversed,
+        wall.inset_idx_ == 0,
+        scarf_seam,
+        smooth_speed,
+        [this, &settings, &default_config, &roofing_config, &bridge_config, &non_bridge_line_volume](
```
rburema, line 1497: At some point it may be better (or at least more clear) to just to specify `[&]` instead ('everything by reference'), especially if you already need to include `this`.

src/LayerPlan.cpp
```diff
@@ -630,20 +659,33 @@ void LayerPlan::addPolygonsByOptimizer(
     }
     orderOptimizer.optimize();
 
-    if (! reverse_order)
+    auto add_polygons
```
rburema, line 662: You can also make this function `const`, since you don't reassign this.

src/LayerPlan.cpp
```diff
@@ -1044,16 +1092,21 @@ void LayerPlan::addSplitWall(
 
     for (size_t point_idx = 1; point_idx < max_index; point_idx++)
     {
-        const ExtrusionJunction& p1 = wall[(wall.size() + start_idx + point_idx * direction) % wall.size()];
+        const size_t actual_point_index = (wall.size() + start_idx + point_idx * direction) % wall.size();
+        const Point2LL& p1 = wall.pointAt(actual_point_index);
+        const coord_t w1 = wall.lineWidthAt(actual_point_index);
 
-        if (! bridge_wall_mask_.empty())
+        if constexpr (std::is_same_v<PathType, ExtrusionLine>)
```
rburema, line 1099: I guess it makes sense that a `Polygon` is never bridging -- and that's the only other `PathType` at the moment, and we definitely don't _plan_ to add more... otherwise I'd say that you want the inversion maybe, select by it _not_ being a `Polygon`. Or maybe add a `bool` field/function to `PathType` to say if it's bridging or not.

That may be overkill given the situation though ... just maybe add some documentation here?

include/LayerPlan.h
```diff
@@ -398,11 +401,15 @@ class LayerPlan : public NoCopy
...
+        const Settings& settings,
         const GCodePathConfig& config,
         coord_t wall_0_wipe_dist = 0,
         bool spiralize = false,
         const Ratio& flow_ratio = 1.0_r,
-        bool always_retract = false);
+        bool always_retract = false,
+        bool scarf_seam = false,
+        bool smooth_speed = false,
+        bool is_candidate_small_feature = false);
```
rburema, line 412: I'm pretty sure I'm overlooking something, but where is this filled? Is it always `false`? I know it's used in the check for 'small features' below somewhere.

include/LayerPlan.h
```diff
@@ -398,11 +401,15 @@ class LayerPlan : public NoCopy
...
         int startIdx,
         const bool reverse,
+        const Settings& settings,
         const GCodePathConfig& config,
         coord_t wall_0_wipe_dist = 0,
         bool spiralize = false,
         const Ratio& flow_ratio = 1.0_r,
-        bool always_retract = false);
+        bool always_retract = false,
+        bool scarf_seam = false,
```
rburema, line 410: These last three aren't documented yet.

rburema (approved):
... I can really see why you want this whole thing refactored now 😅

Anyway, I have a couple of small remarks, otherwise the code seems solid (just likely to fall over if we add any more to it).

include/LayerPlan.h
```diff
@@ -893,7 +900,52 @@ class LayerPlan : public NoCopy
         const coord_t accelerate_length,
         const Ratio end_speed_ratio,
         const coord_t decelerate_length,
-        const bool is_scarf_closure);
+        const bool is_scarf_closure,
+        const bool compute_distance_to_bridge_start,
+        const std::function<void(
```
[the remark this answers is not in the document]
wawanbreton: Good idea, this adds some clarity and allows me to document the function parameters

src/LayerPlan.cpp
wawanbreton: That's right, I know the use of [&] is somehow discouraged so I tend to never use it, but in this case it does make sense.

src/LayerPlan.cpp
wawanbreton: I also added the `compute_distance_to_bridge_start` argument to make it a bit more explicit, but in the case of a Polygon it will just be silently ignored which is definitely not great... The documentation of this settings mentions that it will only work with ExtrusionLine though. I'll add some documentation in the function itself.

include/LayerPlan.h
wawanbreton: Oh you're right, in the end it is never actually used, I'll remove it

src/LayerPlan.cpp
rburema: I think the main reason for that is because it can pull in things like `this` unexpectedly. (And all the globals, etc.) However, in this case you're doing most of that anyway :-)

PR #2159, opened 2024-10-29
Add writeGcodeFile function to GCodeExport

saumyaj3:
# Description

Introduce a new method writeGcodeFile to the GCodeExport class, which ensures that the output stream is properly flushed. This function is also called at the end of the Gcode writing process to finalize the file.

### DIFF
--- a/src/gcodeExport.cpp
+++ b/src/gcodeExport.cpp
@@ -1726,8 +1726,13 @@ void GCodeExport::finalize(const char* endCode)
     for (int n = 1; n < MAX_EXTRUDERS; n++)
         if (getTotalFilamentUsed(n) > 0)
             spdlog::info("Filament {}: {}", n + 1, int(getTotalFilamentUsed(n)));
+}
+
+void GCodeExport::writeGcodeFile()
+{
     output_stream_->flush();
 }
+
 
 double GCodeExport::getExtrudedVolumeAfterLastWipe(size_t extruder)
 {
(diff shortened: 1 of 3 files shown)


src/gcodeExport.cpp
```diff
@@ -1726,6 +1726,10 @@ void GCodeExport::finalize(const char* endCode)
     for (int n = 1; n < MAX_EXTRUDERS; n++)
         if (getTotalFilamentUsed(n) > 0)
             spdlog::info("Filament {}: {}", n + 1, int(getTotalFilamentUsed(n)));
+}
```
rburema, line 1729: I think you should still flush here as well.

src/gcodeExport.cpp
```diff
@@ -1726,6 +1726,10 @@ void GCodeExport::finalize(const char* endCode)
...
+
+void GCodeExport::writeGcodeFile()
```
rburema, line 1731: I'm not sure the name of this function is completely descriptive of what it does.

rburema (changes requested):
Couple of small changes requested :-)

src/gcodeExport.cpp
saumyaj3: 3bb6d32 added it here

src/gcodeExport.cpp
saumyaj3: I changed the name to flushOutputStream and added a bit  of description as well.

rburema (approved):

PR #2160, opened 2024-10-29
CURA-12218 fix scattered seam for round models

wawanbreton:
The previous rework of the score calculation removed the relative comparison between vertices scores to ensure the consistency. The new strategy performed very badly on shapes where multiple points would get a similar score, e.g. finding sharp corners on a cylinder. We now process a multi-pass scoring, so that all points having a very similar score at their main criterion will now go through a second pass with different criteria.

CURA-12218


rburema (commented):

| Benchmark suite | Current: 71ac23e50ea58b7ec08a5823b0bbd5a03dec1c25 | Previous: 32a279d139094005205e20e8a95af14393805692 | Ratio |
|-|-|-|-|
| `SimplifyTestFixture/simplify_local` | `8.660006123440803` ns/iter | `1.2392275564797144` ns/iter | `6.99` |

PR #2165, opened 2024-11-04
CURA-12259 Bump grpc_definitions version for 5.9

wawanbreton:
CURA-12259

We made some significant changes between 5.8 and 5.9, so now is the time to bump the version number in order to indicate this.


rburema (approved):
All failing actions are because the dependant package not being there yet -- and the only thing we did was putting the version-no higher, so that's OK.

PR #2166, opened 2024-11-05
Fix dirty print lines and micro-segments

wawanbreton:
Yes I forgot it... :roll_eyes: 

CURA-12263


rburema (approved):
A very wide change :-p

PR #2170, opened 2024-11-08
CURA-12265 the wiping movement goes in the wrong direction

wawanbreton:
The wiping movement while adding extrusion lines didn't care of the reversed order. We now use a common method while adding either Polygons and ExtrusionLines, which takes care of the reversion.
Also fixes the wipe movement interaction with the scarf seam by taking the actual last printed point, which can now be different from the path last point due to the scarf seam closure.

CURA-12265
CURA-12266


rburema (commented):

| Benchmark suite | Current: 3e7f35af50d85b4e675bdbfcdd48385d394731ec | Previous: cae1d6ed8dd193ce3a9e0cbc07a783858bb3115c | Ratio |
|-|-|-|-|
| `SimplifyTestFixture/simplify_slot_noplugin` | `4.022321948141616` ns/iter | `1.2463209134545274` ns/iter | `3.23` |

nallath (approved):

PR #2172, opened 2024-11-11
Cura 12275 retract before outer wall only works on the initial layer

wawanbreton:
The `travel_retract_before_outer_wall` setting is settable per-mesh, but the global value was previously taken in this context. We now take the value of the setting, for the mesh we are moving to.

CURA-12275
Fixes https://github.com/Ultimaker/Cura/issues/19876


rburema (approved):
Code looks good -- though I know the code won't reach there, it still _feels_ a little risky to return `nullptr` like that and not check... 😆

PR #2173, opened 2024-11-13
Fix (some) crash-logs divide-by-zero issues

rburema:
Scoured the crash-logs for often-occurring div-by-zero issues, fixed these three.


wawanbreton (approved):
Small improvements, but full of sense !

PR #2175, opened 2024-11-22
CURA-12324 random seam is not so random

wawanbreton:
When using a random seam position, the scoring algorithm would still use the fallback criteria, which are useful for the "sharpest corner" strategy. We now don't use them, so that a really random point is always returned.
Also fixed a second bias where the best candidate could also not be returned to the profit of a nearly-best one.

CURA-12324

### DIFF
--- a/include/PathOrderOptimizer.h
+++ b/include/PathOrderOptimizer.h
@@ -784,17 +784,20 @@ protected:
         best_candidate_finder.appendCriteriaPass(main_criteria_pass);
 
         // ########## Step 2: add fallback passes for criteria with very similar scores (e.g. corner on a cylinder)
-        const AABB path_bounding_box(points);
+        if (path.seam_config_.type_ == EZSeamType::SHARPEST_CORNER)
+        {
+            const AABB path_bounding_box(points);
 
-        { // First fallback strategy is to take points on the back-most position
-            auto fallback_criterion = std::make_shared<DistanceScoringCriterion>(points, path_bounding_box.max_, DistanceScoringCriterion::DistanceType::YOnly);
-            constexpr double outsider_delta_threshold = 0.01;
-            best_candidate_finder.appendSingleCriterionPass(fallback_criterion, outsider_delta_threshold);
-        }
+            { // First fallback strategy is to take points on the back-most position
+                auto fallback_criterion = std::make_shared<DistanceScoringCriterion>(points, path_bounding_box.max_, DistanceScoringCriterion::DistanceType::YOnly);
+                constexpr double outsider_delta_threshold = 0.01;
+                best_candidate_finder.appendSingleCriterionPass(fallback_criterion, outsider_delta_threshold);
+            }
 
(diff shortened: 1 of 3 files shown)


src/utils/scoring/BestElementFinder.cpp
```diff
@@ -47,17 +49,32 @@ std::optional<size_t> cura::BestElementFinder::findBestElement(const size_t cand
                 iterator->score += weighed_criterion.criterion->computeScore(iterator->candidate_index) * weighed_criterion.weight;
             }
 
-            best_score = std::max(best_score, iterator->score);
+            if (best_candidate_iterator == end || iterator->score > best_candidate_iterator->score)
```
rburema, line 52: We could also set the initial score(s) to `numeric_limits<...>::min()` (if the type is integer, or negative infinity if the type is float). That way, we don't need the end iterator. -- But this works too.

src/utils/scoring/BestElementFinder.cpp
```diff
@@ -47,17 +49,32 @@ std::optional<size_t> cura::BestElementFinder::findBestElement(const size_t cand
...
+            {
+                best_candidate_iterator = iterator;
+            }
+        }
+
+        if (best_candidate_iterator == end)
+        {
+            // Something got wrong, we have not best candidate
```
rburema, line 60: I'd suggest changing this to "Something went wrong, we don't have the best candidate."

include/PathOrderOptimizer.h
```diff
@@ -784,17 +784,20 @@ class PathOrderOptimizer
         best_candidate_finder.appendCriteriaPass(main_criteria_pass);
 
         // ########## Step 2: add fallback passes for criteria with very similar scores (e.g. corner on a cylinder)
-        const AABB path_bounding_box(points);
+        if (path.seam_config_.type_ == EZSeamType::SHARPEST_CORNER)
```
rburema, line 787: I wonder if it's better to single out the other two instead of locking it to sharpest corner only. It does work fine in the current situation however.

rburema (commented):
Code looks good, though I have a couple of observations. I'll also need to look at why the unit-tests ❌ are failing (seems specifically the gcode export, but doesn't give much details beyond that), so I can't approve yet.

rburema (approved):
Turns out the unit-test errors aren't this PR's doings, but some other thing merged into main recently (see [here](https://github.com/Ultimaker/CuraEngine/pull/2114#issuecomment-2500172641)).

Also the component test looks good, so approved!

src/utils/scoring/BestElementFinder.cpp
wawanbreton: Well I need this iterator anyway because in the early-out I need the index. And I also stopped used min and max as markers for a specific state, I think the previous version used an std::optional

include/PathOrderOptimizer.h
wawanbreton: I assume that for the other options, seam alignment is not something the user is looking for, so just skipping it somehow makes sense. Could be debated for "shortest".

PR #2187, opened 2024-12-16
CURA-12352 different minimum layer time for layers that contain overhangs

wawanbreton:
CURA-12352

:warning:  Depends on CURA-11966, so don't review until merged

Requires https://github.com/Ultimaker/Cura/pull/20039


src/LayerPlan.cpp
```diff
@@ -557,6 +557,165 @@ void LayerPlan::addExtrusionMove(
...
+                intersection = MINIMUM_LINE_LENGTH / static_cast<float>(segment_length);
+            }
+            else if (intersection * segment_length > segment_length - MINIMUM_LINE_LENGTH)
+            {
+                intersection = (segment_length - MINIMUM_LINE_LENGTH) / static_cast<float>(segment_length);
+            }
+        }
+
+        // (Also) Avoid microsegments: Filter out pairs of intersections that are too close to each other.
+        // This should be possible because the intersections happen in the same region.
```
wawanbreton, line 637: I don't really understand why this should work. According to me, it shouldn't because in the next step, you always assume that you are moving from a region to an adjacent one.  But not you filter our intersections, which means you can move from a region to any other.

wawanbreton (commented):
Micro-segment filtering is a good addition, although I have doubts on the second part. We should take some time to discuss it.

src/FffGcodeWriter.cpp
```diff
@@ -3082,21 +3085,88 @@ bool FffGcodeWriter::processInsets(
...
+            }
+
+            speed_regions.push_back(SpeedRegion{ 90.0, overhang_speed_factors.back() }); // Final "everything else" speed region
+
+            // Now merge regions that have similar speed factors (saves calculations and avoid generating micro-segments)
+            auto merged_regions = speed_regions
+                                | ranges::views::chunk_by(
+                                      [](const auto& region_a, const auto& region_b)
+                                      {
+                                          return region_a.chunk && region_b.chunk && region_a.speed_factor == region_b.speed_factor;
```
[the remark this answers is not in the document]
wawanbreton: I'll add a comment for this

src/LayerPlan.cpp
wawanbreton: I'll fix this

rburema (commented):

| Benchmark suite | Current: 588a83856f8db7a347a7cc8f88bbeff87a746a1c | Previous: fb54c9291e2f420c09c7927d4da61dc920ee561b | Ratio |
|-|-|-|-|
| `SimplifyTestFixture/simplify_slot_noplugin` | `1.8697703139041124` ns/iter | `1.2458342521198391` ns/iter | `1.50` |

src/LayerPlan.cpp
```diff
@@ -710,9 +692,67 @@ void LayerPlan::addExtrusionMoveWithGradualOverhang(
...
+            extrusion_moves.push_back(SegmentExtrusionMove{ p.toPoint2LL(), actual_speed_region_index });
+            break;
+        }
+    }
+
+    // Filter out micro-segments
+    std::vector<SegmentExtrusionMove> extrusion_moves_filtered;
+    extrusion_moves_filtered.reserve(extrusion_moves.size());
+    Point2LL current_position = start;
+    for (const SegmentExtrusionMove& extrusion_move : extrusion_moves | ranges::views::drop_last(1))
```
rburema, line 703: I think I'd prefer it if we always keep the end-points -- (the end-point is already kept, but) so in the specific case the micro-segment happens between the first and the next, we throw away the next instead of the first.

I say this because we might treat points within 5 units as the same point, bit it can still visually look bad if the corners don't align.

All we'd have to do is also drop the first (or two, depending on the outcome) from the loop and make sure to prepend it separately.

Given we're in a bit of a time-crunch I'll try to add this myself rather than send it back, then send it on to QA.

PR #2202, opened 2025-01-27
CURA-12335 Handle flooring settings on the same basis as roofing

wawanbreton:
This duplicates all the behavior related to flooring management to flooring, so that we can treat them with different settings.

CURA-12335

Comes with https://github.com/Ultimaker/Cura/pull/20191


rburema (commented):

| Benchmark suite | Current: c841fb82b0e68c139ec143723bf0e6035c1a054e | Previous: 35a8c5843e208ef3fe3f5772e7e2190cd37d5b67 | Ratio |
|-|-|-|-|
| `SimplifyTestFixture/simplify_slot_noplugin` | `1.8674791813207994` ns/iter | `1.2446122488245035` ns/iter | `1.50` |

src/skin.cpp
```diff
@@ -324,22 +324,27 @@ void SkinInfillAreaComputation::generateInfill(SliceLayerPart& part)
...
+void SkinInfillAreaComputation::generateSkinRoofingFlooringFill(SliceLayerPart& part)
 {
     for (SkinPart& skin_part : part.skin_parts)
     {
         const size_t roofing_layer_count = std::min(mesh_.settings.get<size_t>("roofing_layer_count"), mesh_.settings.get<size_t>("top_layers"));
+        const size_t flooring_layer_count = std::min(mesh_.settings.get<size_t>("flooring_layer_count"), mesh_.settings.get<size_t>("bottom_layers"));
         const coord_t skin_overlap = mesh_.settings.get<coord_t>("skin_overlap_mm");
 
         Shape filled_area_above = generateFilledAreaAbove(part, roofing_layer_count);
+        Shape filled_area_below = generateFilledAreaBelow(part, flooring_layer_count);
```
rburema, line 336: I think these can be made `const`.

src/FffGcodeWriter.cpp
```diff
@@ -3273,51 +3320,42 @@ bool FffGcodeWriter::processSkinPart(
...
 
-void FffGcodeWriter::processRoofing(
+void FffGcodeWriter::processRoofingFlooring(
     const SliceDataStorage& storage,
     LayerPlan& gcode_layer,
     const SliceMeshStorage& mesh,
     const size_t extruder_nr,
-    const MeshPathConfigs& mesh_config,
-    const SkinPart& skin_part,
+    const std::string& setting_prefix,
```
rburema, line 3336: Using string interpolation as a method of retrieving settings seems like a recipe for unmaintainable code.

I'd propose making it into a structure you can just pass to the function instead, like;

~~~
struct RoofFloorSettingNames
{
public:
std::string extruder_nr;
<...>
};
~~~

Fill those up for the two possibilities and you'd be able to use it like:

`const size_t skin_extruder_nr = mesh.settings.get<ExtruderTrain&>(rooffloor_settings.extruder_nr, setting_prefix)).extruder_nr_;`

rburema (changes requested):
I have a remark I'd like to have addressed -- I'll continue with the component test in the meantime.

Looks good otherwise!

src/FffGcodeWriter.cpp
wawanbreton: Good idea, I'll try it

rburema (approved):

PR #2204, opened 2025-02-05
CURA-11978 Retract and unretract in a travel

wawanbreton:
CURA-11978

### DIFF
--- a/include/gcodeExport.h
+++ b/include/gcodeExport.h
@@ -12,6 +12,7 @@
 #include <sstream> // for stream.str()
 #include <stdio.h>
 
+#include "TravelAntiOozing.h"
 #include "geometry/Point2LL.h"
 #include "settings/EnumSettings.h"
 #include "settings/Settings.h" //For MAX_EXTRUDERS.
@@ -163,6 +164,7 @@ private:
     coord_t current_layer_z_;
     coord_t is_z_hopped_; //!< The amount by which the print head is currently z hopped, or zero if it is not z hopped. (A z hop is used during travel moves to avoid collision with
                           //!< other layer parts)
+    std::optional<ZHopAntiOozing> z_hop_prime_leftover_; //!< The leftover priming from a previous travel move that should be processed while z-hopping down
 
     size_t current_extruder_;
     std::map<size_t, double> current_fans_speeds_; //!< Current fan speed, by fan index. No value means the speed has never been set yet.
@@ -535,7 +537,7 @@ public:
      *
      * \param speed The speed used for moving.
      */
-    void writeZhopEnd(Velocity speed = 0.0);
+    void writeZhopEnd(Velocity speed = 0.0, const coord_t height = 0, const double prime_distance = 0.0, const Ratio& prime_ratio = 0.0_r);
 
     /*!
      * Start the new_extruder:
@@ -675,6 +677,8 @@ public:
      * \param wipe_config Config with wipe script settings.
      */
     void insertWipeScript(const WipeScriptConfig& wipe_config);
+
+    void setZHopPrimeLeftover(const ZHopAntiOozing& z_hop_prime_leftover);
 };
 
 } // namespace cura
--- a/include/gcodeExport.h
+++ b/include/gcodeExport.h
@@ -438,6 +438,7 @@ private:
      * \param y build plate y
      * \param z build plate z
      * \param speed movement speed
+     * \param retract_distance The retract disance to be reached during this travel, or nullopt to leave it unchanged
      */
     void writeTravel(const coord_t x, const coord_t y, const coord_t z, const Velocity& speed, const std::optional<double> retract_distance = std::nullopt);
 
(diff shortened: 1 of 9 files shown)


src/gcodeExport.cpp
```diff
@@ -1249,48 +1300,80 @@ void GCodeExport::writeRetraction(const RetractionConfig& config, bool force, bo
...
+
+    RetractionAmounts retraction_amounts;
+    if (retract_distance.has_value())
+    {
+        retraction_amounts = computeRetractionAmounts(extruder_attr_[current_extruder_], *retract_distance);
+    }
+
+    is_z_hopped_ = height;
+    const coord_t target_z = current_layer_z_ + is_z_hopped_;
+    // current_position_.z_ = current_layer_z_;
```
jellespijker, line 1364: Commented out code, please remove

src/LayerPlan.cpp
```diff
@@ -2152,6 +2153,238 @@ void LayerPlan::addLinesInGivenOrder(
...
+        const Ratio retract_prime_ratio = retract_duration_during_travel / total_anti_ooze_duration_during_travel;
+
+        const Duration allowed_retract_duration_during_travel = travel_durations.travel * retract_prime_ratio;
+        extra_time_still_retraction = retract_duration_during_travel - allowed_retract_duration_during_travel;
+
+        const Duration allowed_prime_duration_during_travel = travel_durations.travel * (1.0_r - retract_prime_ratio);
+        extra_time_still_prime = prime_duration_during_travel - allowed_prime_duration_during_travel;
+    }
+
+    const auto compute_anti_ooze_amounts
```
jellespijker, line 2225: If would make this an actual function.This would not only improve readability but also make it easier to write unit tests against the logic for computing the per‑segment amounts.

include/gcodeExport.h
```diff
@@ -113,6 +115,22 @@ class GCodeExport : public NoCopy
...
+        double new_e{ 0.0 }; // The new absolute retraction amount
+        double diff_e{ 0.0 }; // The difference between the new and previous amount, which is to be processed
+
+        /*!
+         * Indicates whether this retraction actually has something to process
+         */
+        inline bool has_retraction() const
+        {
+            return std::abs(diff_e) >= 0.000001;
+        }
```
jellespijker, line 131: `nodiscard` is your friend and member functions defined inside a class definition are implicitly inline.The primary improvements come from making the function `constexpr` and `noexcept`.

```suggestion
        [[nodiscard]] constexpr bool has_retraction() const noexcept
        {
            constexpr double threshold { 1e-6 };
            return std::abs(diff_e) >= threshold;
        }
```

include/utils/types/numeric_facade.h
```diff
@@ -14,7 +14,7 @@ struct NumericFacade
 {
     using value_type = T;
 
-    value_type value{};
+    value_type value{ static_cast<T>(0) };
```
jellespijker, line 17: not sure if a static_cast is needed here, for simple numerical types, the default is always a zero value
```suggestion
    value_type value{ };
```

src/LayerPlan.cpp
```diff
@@ -2152,6 +2153,238 @@ void LayerPlan::addLinesInGivenOrder(
...
+        const Velocity& travel_speed = path.config.getSpeed();
+        Point2LL start_position = gcode.getPosition().toPoint2LL();
+        for (const Point3LL& travel_point : path.points)
+        {
+            const Point2LL travel_point_2d = travel_point.toPoint2LL();
+            const coord_t travel_segment_length = vSize(travel_point_2d - start_position);
+            travel_durations.travel += (travel_segment_length / travel_speed) / 1000.0;
+            start_position = travel_point_2d;
+        }
+    }
```
jellespijker, line 2177: Untested and a suggestion, it should be a bit more performant avoids some extra calculations
```suggestion
    if (! path.points.empty())
    {
        const Velocity travel_speed = path.config.getSpeed();
        const double factor = 1.0 / travel_speed / 1000.0;
        Point2LL start = gcode.getPosition().toPoint2LL();
        auto total_distance = std::accumulate(
            path.points.begin(), path.points.end(),
            0.0,
            [&start](double acc, const Point3LL& pt)
            {
                const Point2LL current = pt.toPoint2LL();
                double seg = vSize(current - start);
                start = current;
                return acc + seg;
            });
        travel_durations.travel += total_distance * factor;
    }
```

jellespijker (changes requested):
Overal it looks good and well thought out, a couple of minor remarks and suggestions. But I do suggest adding some UT
given the complexity of splitting travel moves into multiple phases, maybe add unit and integration tests that cover:

- A travel move with enough duration for all retraction/priming to happen during travel.
- A travel move where extra stationary time is required.
- Edge cases with only one travel segment.
- Both “forward” (retraction) and “reversed” (priming) computations.

include/utils/types/numeric_facade.h
wawanbreton: Hem, unless I missed some improvement of the language, it is not. Basic types like `float` and `int` don't have an initial value and can contain anything. Or maybe just having the `{}` is good enough ?

wawanbreton: Ok, seems to be the case indeed. I thought `value_type value{};` and `value_type value;` were identical but they are not.

jellespijker: https://en.cppreference.com/w/cpp/language/zero_initialization

include/gcodeExport.h
jellespijker: Didn't you pay attention `constexpr` everything! https://www.youtube.com/watch?v=PJwd4JLYJJY

jellespijker: But yes in this case it is still depend on a dynamic value

jellespijker (approved):
LGTM

Cura.proto
```diff
@@ -83,10 +83,12 @@ message Polygon {
...
         InfillType = 6;
         SupportInfillType = 7;
-        MoveCombingType = 8;
-        MoveRetractionType = 9;
+        MoveUnretracted = 8;
+        MoveRetracted = 9;
         SupportInterfaceType = 10;
         PrimeTowerType = 11;
+        MoveWhileRetracting = 12;
+        MoveWhileUnretracting = 13;
```
HellAholic, line 91: Should we add the
    StationaryRetractUnretract = 14;
    NumPrintFeatureTypes = 15;
?

wawanbreton: Oh, definitely... I dont't see how it can work without it actually :thinking:

PR #2206, opened 2025-03-03
CURA-12369 extra wipe move with spiralize

wawanbreton:
This restores the previous behavior which was broken during scarf seam implementation. As a bonus, it also improves the display so that it now properly shows the actual spiral moving up during the print. This is much more accurate, however it emphasizes two incorrectnesses of the spiralize mode:
* Flow on bottom and top layer are not correct i.r.t actual layer thickness
* Last layer ends at approximaterly 1/2 layer height over the actual model

Those behaviors were similar in Cura 4.13.1, so I assume they can be safely ignored, and restoring the previous behavior is good enough for now.

CURA-12369

### DIFF
--- a/include/LayerPlan.h
+++ b/include/LayerPlan.h
@@ -851,7 +851,7 @@ private:
      *  @param position The start position (which is not included in the path points)
      *  @param extrude_speed The actual used extrusion speed
      */
-    void sendLineTo(const GCodePath& path, const Point3LL& position, const double extrude_speed);
+    void sendLineTo(const GCodePath& path, const Point3LL& position, const double extrude_speed, const std::optional<coord_t>& line_thickness = std::nullopt);
 
     /*!
      *  @brief Write a travel move and properly apply the various Z offsets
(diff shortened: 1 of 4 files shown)


rburema (changes requested):
Code looks good, but the component test shows this for me for the new code:

layer_under_top

rburema:
> > Code looks good, but the component test shows this for me for the new code:
> 
> Oh, that is weird, totally not what I see 🤔 can you send the 3MF ?

It's just the project attached to the ticket 😅

wawanbreton:
The fix for Windows does break the behavior on Linux :disappointed: 

image

image

The tapering spiral now shows a jump up and ends 1.5 layer-height over the model. We will have to dig and see what makes this behavior inconsistent.

rburema:
Hm, Windows also shows a similar jump, but smaller:

~~~
G1 X115.664 Y123.673 Z10.105 E183.92742
G1 X112.889 Y124.515 Z10.108 E184.02388
G1 X110 Y124.799 Z10.2 E184.11565
G1 X107.111 Y124.515 E184.20254
G1 X104.336 Y123.673 E184.28935
~~~

I copied this, and you screenshotted it; I'm not sure where the difference between `10.3` and `10.2` comes from.

Anyway, it's wrong here too -- still 1 layer above the model, rather than 1.5

wawanbreton:
Code looks good, now I'm gonna test it on Linux :drum:

PR #2210, opened 2025-03-11
Fix race condition in TreeSupportBaseCircle which may sometimes cause a crash

ThomasRahm:
# Description

The non-proper locking of base_circle has caused a crash for me in my cradle support branch. I don't know it that crash can also occur here, but either way the lock should actually do something.

**Test Configuration**:
* Operating System:

# Checklist:

### DIFF
--- a/include/TreeSupportBaseCircle.h
+++ b/include/TreeSupportBaseCircle.h
@@ -18,37 +18,27 @@ class TreeSupportBaseCircle
 {
 protected:
     inline static Polygon base_circle;
-    inline static bool circle_generated = false;
-    inline static std::mutex critical_sections;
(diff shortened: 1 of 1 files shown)


ThomasRahm:
Talked with a friend about this pull request, he suggested to instead use std::call_once as the compiler may reorder the base_circle and boolean assignments, so even my fixed version was not guaranteed to be actually thread-safe. So I changed it to use that instead.

wawanbreton (approved):
Hi @ThomasRahm, I am not familiar with std::call_once, but it seems to do exactly what you need here. There are different ways of achieveing this, but this one is very elegant.

PR #2213, opened 2025-03-18
CURA-12474 set fixed 0 gap for brim support

wawanbreton:
This requires having different starting outlines for models and supports, so that we don't apply the same gap for each

CURA-12474

### DIFF
--- a/include/SkirtBrim.h
+++ b/include/SkirtBrim.h
@@ -23,15 +23,25 @@ class SkirtBrim
 {
 private:
     /*!
+     * Store the various outlines that we want to create a brim around
+     */
+    struct Outline
+    {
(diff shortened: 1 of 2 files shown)


rburema (commented):

| Benchmark suite | Current: 9d06c928b7e2a25ad05d3918bc3482cdd4a3f24c | Previous: 61567388af3a805b8e79d9c960981f081d072ce7 | Ratio |
|-|-|-|-|
| `SimplifyTestFixture/simplify_slot_noplugin` | `1.8890690113623052` ns/iter | `1.247252371095342` ns/iter | `1.51` |

rburema (approved):

PR #2219, opened 2025-03-24
NP-861 Fix - emscripten compatibility with spdlog and fmt

rumboon:
This fixes the npm building issue due to compatibility issues between emscripten, spdlog and fmt.

I have updated spdlog and fmt and made changes according to the required changes for these new versions.

### DIFF
--- a/src/TreeSupportTipGenerator.cpp
+++ b/src/TreeSupportTipGenerator.cpp
@@ -66,7 +66,7 @@ TreeSupportTipGenerator::TreeSupportTipGenerator(const SliceMeshStorage& mesh, T
 
     if (max_overhang_speed < 2)
     {
-        max_overhang_insert_lag_ = static_cast<size_t>(std::numeric_limits<coord_t>::max());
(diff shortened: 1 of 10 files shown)


src/TreeSupportTipGenerator.cpp
```diff
@@ -66,7 +66,7 @@ TreeSupportTipGenerator::TreeSupportTipGenerator(const SliceMeshStorage& mesh, T
 
     if (max_overhang_speed < 2)
     {
-        max_overhang_insert_lag_ = std::numeric_limits<coord_t>::max();
+        max_overhang_insert_lag_ = static_cast<size_t>(std::numeric_limits<coord_t>::max());
```
casperlamboo, line 69: why not directly take the max of size_t instead of typecasting?

```suggestion
        max_overhang_insert_lag_ = std::numeric_limits<size_t>::max();
```

rburema (approved):

rburema:
Thanks! Will probably have this in 5.10.1

PR #2220, opened 2025-03-26
[CURA-12503] Remove now unused 'skip agressive merge hint' variable.

rburema:
This is no longer used anywhere.

TODO/Draft: It's still in our API as well, so that'll be next on the list. (See the FIXME.) -> Why not leave it in there as an unused boolean for now, then we don't have to change the interface again if the next thing we need _is_ a boolean.

### DIFF
--- a/include/LayerPlan.h
+++ b/include/LayerPlan.h
@@ -69,7 +69,6 @@ public:
     const PathConfigStorage configs_storage_; //!< The line configs for this layer for each feature type
     const coord_t z_;
     coord_t final_travel_z_;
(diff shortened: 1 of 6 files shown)


rburema (commented):

| Benchmark suite | Current: a9fd3b6bd7f7537f68e217940fda69e7566b4794 | Previous: 3835427a31180bafcfae95b45cf14ea4f25520d1 | Ratio |
|-|-|-|-|
| `SimplifyTestFixture/simplify_slot_noplugin` | `1.941201940430247` ns/iter | `1.2406784491736291` ns/iter | `1.56` |

PR #2224, opened 2025-04-22
[CURA-12525] Print only done on any extruders > 0 should still set extruder,

rburema:
... even if it's the only one. (So for instance if a model is sliced to only use extruder 1, the printer should still switch away from the default extruder 0 one at least --and likely at most-- once.)


wawanbreton (approved):
Makes sense !

PR #2227, opened 2025-04-30
[CURA-12446] Added settings for a different number of walls in top-/bottom-most skin.

rburema:
See also front-end PR: https://github.com/Ultimaker/Cura/pull/20565


rburema:
- bridging not OK yet
- need to check if gradual sloping walls are OK (top AND bottom), like for example a sphere

src/WallsComputation.cpp
```diff
@@ -37,7 +37,20 @@ WallsComputation::WallsComputation(const Settings& settings, const LayerIndex la
...
+        switch (part->wall_exposed)
+        {
+        case SliceLayerPart::WallExposedType::BOTTOM_0:
+            return "wall_line_count_0";
+        case SliceLayerPart::WallExposedType::TOP:
+            return "wall_line_count_top";
+        case SliceLayerPart::WallExposedType::SIDE_ONLY:
+            return "wall_line_count";
+        }
+    };
```
wawanbreton, line 51: I don't see the need for a function here, since it is called only once, could you just do the switch/case and store the result in a variable ?

src/layerPart.cpp
```diff
@@ -65,22 +65,60 @@ void createLayerWithParts(const Settings& settings, SliceLayer& storageLayer, Sl
...
-    for (auto& part : result)
+    for (auto& main_part : result)
     {
-        storageLayer.parts.emplace_back();
-        if (part.empty())
+        std::vector<std::pair<SliceLayerPart::WallExposedType, std::vector<SingleShape>>> parts_by_type = {
+            { SliceLayerPart::WallExposedType::BOTTOM_0, bottom_parts.splitIntoParts() },
+            { SliceLayerPart::WallExposedType::TOP, top_parts.difference(bottom_parts).splitIntoParts() },
+            { SliceLayerPart::WallExposedType::SIDE_ONLY, main_part.difference(bottom_parts).difference(top_parts).splitIntoParts() },
+        };
```
wawanbreton, line 74: Maybe a `std::map` would make things easier ? unless you need an ordering

src/layerPart.cpp
```diff
@@ -65,22 +65,60 @@ void createLayerWithParts(const Settings& settings, SliceLayer& storageLayer, Sl
...
         }
-        storageLayer.parts.back().outline = part;
-        storageLayer.parts.back().boundaryBox.calculate(storageLayer.parts.back().outline);
-        if (storageLayer.parts.back().outline.empty())
+    }
+}
+
+Shape getTopOrBottom(int direction, const std::string& setting_name, size_t layer_nr, const std::vector<SlicerLayer>& slayers, const Settings& settings)
+{
+    auto result = Shape();
```
wawanbreton, line 100: I'm probably old-style and not a huge fan of `auto`, but I have a slight preference for this:
```suggestion
    Shape result;
```

src/layerPart.cpp
```diff
@@ -65,22 +65,60 @@ void createLayerWithParts(const Settings& settings, SliceLayer& storageLayer, Sl
...
+    return result;
 }
 
+/*!
+ * \brief Split a layer into parts.
+ * \param settings The settings to get the settings from (whether to union or
+ * not).
+ * \param storageLayer Where to store the parts.
+ * \param layer The layer to split.
+ */
```
wawanbreton, line 121: Shouldn't the documentation be in the header ?

wawanbreton (approved):
Just a few suggestions for readability, feel free to reject them

src/WallsComputation.cpp
rburema: I thought I could save an assignment this way, but that optimization will probably be done by the compiler anyway. I'll change it.

src/layerPart.cpp
rburema: ... I'm not sure why I didn't see that myself (probably because I didn't conceptually see it as a mapping, given what I do next with it).

I don't think it makes things easier per-se, but it'd certainly make the type easier to read.

src/layerPart.cpp
rburema: Yeah makes sense. I mostly use `auto` when the type-name would otherwise double, but here you can just use the short notation instead.

src/layerPart.cpp
rburema: Whoops! That's supposed to be the documentation of that other method (not used outside the file), I accidentally placed it on this one.

rburema (commented):

| Benchmark suite | Current: aed8b1844cd926c3c9047072f92d0694e4fc2cca | Previous: 945c507eac606e99e212a5bce049d3a8918106e3 | Ratio |
|-|-|-|-|
| `SimplifyTestFixture/simplify_slot_noplugin` | `1.8856092072748627` ns/iter | `1.2446332009368046` ns/iter | `1.51` |

src/layerPart.cpp
wawanbreton: That is right, it is fundamentally not used as a key-value container, but I tend to use more std::map because it simplifies the type. This is mostly a personal preference.

wawanbreton (approved):

src/layerPart.cpp
rburema: I ended up using both a map for this and another one to replace the function/switch, so I think this PR is now an atlas :-p

rburema (commented):
I'm not sure how well this'd work for values of 'extension' less than the wall-line-width -- but it's probably OK.

rburema:
superseded by #2334

PR #2229, opened 2025-05-20
CURA-12436 Allow moving from/to outer wall with optimal combing

wawanbreton:
Previously, moving from/to an outer wall would use the "minimal" combing shape because the outer wall is clearly outside the "optimal" shape. Now we allow using the optimal shape by moving inside it before/after travel.

CURA-12436

### DIFF
--- a/src/pathPlanning/Comb.cpp
+++ b/src/pathPlanning/Comb.cpp
@@ -71,6 +71,7 @@ Comb::Comb(
     , travel_avoid_distance_(travel_avoid_distance)
     , offset_from_outlines_(comb_boundary_offset) // between second wall and infill / other walls
     , max_moveInside_distance2_(offset_from_outlines_ * offset_from_outlines_)
+    , max_move_inside_distance_enlarged2_(std::pow(offset_from_outlines_ + max_move_inside_enlarge_distance_, 2))
     , offset_from_inside_to_outside_(offset_from_outlines_ + travel_avoid_distance)
     , max_crossing_dist2_(
           offset_from_inside_to_outside_ * offset_from_inside_to_outside_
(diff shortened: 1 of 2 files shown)


src/pathPlanning/Comb.cpp
```diff
@@ -136,6 +136,48 @@ bool Comb::calc(
...
+            result_path,
+            -offset_dist_to_get_from_on_the_polygon_to_outside_,
+            max_comb_distance_ignored,
+            fail_on_unavoidable_obstacles);
+        Comb::moveCombPathInside(boundary_inside_minimum_, boundary_inside_optimal_, result_path, comb_paths.back()); // add altered result_path to combPaths.back()
+        // If the endpoint of the travel path changes with combing, then it means that we are moving to an outer wall
+        // and we should unretract before the last travel move when travelling to that outer wall
+        unretract_before_last_travel_move = comb_result && end_point != travel_end_point_before_combing;
+        return comb_result;
+    }
```
rburema, line 179: The block stopping here is sort-of repeated above. (You said there was a third instance as well as I recall.)

src/pathPlanning/Comb.cpp
```diff
@@ -136,6 +136,48 @@ bool Comb::calc(
         return combing_succeeded;
     }
 
+    // Move start and end point inside the optimal comb boundary
+    // Give more tolerancy when calculating move inside positions, because the target points in this case will be on the borders
+    const coord_t max_move_inside_distance_enlarged = max_moveInside_distance2_ * 1.05;
```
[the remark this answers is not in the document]
wawanbreton: Maybe a fixed value would be more logical indeed. Technically we should increase the non-squared distance, otherwise the original error will not be propagated the proper way. I'll take a look.

src/pathPlanning/Comb.cpp
wawanbreton: Yes indeed, but I don't understand 100% of this code so I am a bit afraid of changing the other part to factorize them :grimacing:

rburema: That's exactly why I'd like to keep it as-is for the patch-release :-)

PR #2235, opened 2025-06-06
CURA-9909 investigate replacing sort with stable sort

wawanbreton:
Replace `std::sort` by `std::stable_sort`
That makes the engine more deterministic, and no performance difference spotted on multiple projects.

CURA-9909
Originated on https://github.com/Ultimaker/Cura/issues/13833


wawanbreton:
All benchmarks passed, no single difference... Looks good to go !

rburema (approved):
LGTM -> move to component test

PR #2238, opened 2025-06-16
CURA-12515 Print bridge walls without overhanging speed

wawanbreton:
Bridge walls were still explicitely printed using overhanging settings, now we ignore them and just print bridge wall as usual.

CURA-12515


rburema (approved):
Yeah that makes sense, let's see if it works w.r.t. component test.

PR #2247, opened 2025-09-24
[CURA-12726] Don't compensate when bridge-flow > 1.0

rburema:
If we print fast enough, and thus need enormous amounts of flow w.r.t. 'normal' bridge printing, the bridge-line itself goes from something that can be ignored, to something that needs to be taken into account -- more-or-less negating the compensation we first needed (because the line itself would of course be printed half a line width away from the origin, and we where ignoring that the bridge wall could extend up to that amount _itself_ as well when calculating for which supports what).

CURA-12726


wawanbreton (approved):
Code is fine obviously, to be tested on real-life prints !

PR #2248, opened 2025-09-24
CURA-12730 Use mesh extruder if painted extruder is missing or disabled

wawanbreton:
The multi-material generator now takes care of inexisting extruders, in case the texture was created with a printer that supports more extruders than the active one.

CURA-12730

### DIFF
--- a/src/MeshMaterialSplitter.cpp
+++ b/src/MeshMaterialSplitter.cpp
@@ -6,7 +6,6 @@
 
 #include <boost/unordered/concurrent_flat_map.hpp>
 #include <boost/unordered/concurrent_flat_set.hpp>
-#include <range/v3/algorithm/contains.hpp>
 #include <range/v3/view/map.hpp>
 #include <spdlog/spdlog.h>
 #include <spdlog/stopwatch.h>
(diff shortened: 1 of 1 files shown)


src/MeshMaterialSplitter.cpp
```diff
@@ -62,12 +63,18 @@ bool operator==(const ContourKey& key1, const ContourKey& key2)
...
  * @param voxel_grid The voxels grid to be filled with mesh data
+ * @param main_extruder The main extruder of the mesh, to be used when we find a texture part that uses a disabled/inexisting extruder
  * @return True if this generated relevant data for multi-extruder, otherwise this means the mesh is completely filled with only extruder 0 and there is no need to go further on
  *         trying to calculate the modified meshes.
  */
-bool makeVoxelGridFromTexture(const Mesh& mesh, const std::shared_ptr<TextureDataProvider>& texture_data_provider, VoxelGrid& voxel_grid)
+bool makeVoxelGridFromTexture(const Mesh& mesh, const std::shared_ptr<TextureDataProvider>& texture_data_provider, VoxelGrid& voxel_grid, const size_t main_extruder)
 {
     boost::concurrent_flat_set<uint8_t> found_extruders;
+    std::vector<size_t> active_extruders;
```
casperlamboo, line 73: Very minor point, but since we're only using`active_extruders` to check if an extruder is present a `std::vector` isn't optimal. Instead we can use an `unordered_set` or an fixed length array of `[false; 16]` where only the active extruders are set to `true`.

casperlamboo (approved):

casperlamboo (approved):

PR #2261, opened 2025-10-22
CURA-12794 Use seam painting data in surface mode

wawanbreton:
CURA-12794


rburema (approved):
LGTM ... now to quickly approve it _before_ QA merges it, like with the other one :-p

PR #2264, opened 2025-11-24
Disable tbbproxy (not available on ARM64 platforms)

HellAholic:
Build requirements update:
- Disable tbbproxy in conan.py on windows - arm64 configuration

CURA-12814


conanfile.py
```diff
@@ -43,6 +43,10 @@ class CuraEngineConan(ConanFile):
         "with_cura_resources": False,
     }
 
+    @property
+    def _min_cppstd(self):
+        return 20
```
wawanbreton: There is already a similar check in `validate()`, moreover this doesn't seem to be valid conan code 🤔

wawanbreton (changes requested):

wawanbreton (approved):

rburema (approved):
LGTM

PR #2274, opened 2026-01-16
CURA- 12556 formula parser in cura engine

wawanbreton:
This PR moves the GCode-template replacement algorithm from the Cura front-end to the engine. This allows for more flexibility because the engine has more information when resolving the templates. Especially for the extruders start/end GCodes, which can now be different between each layer.

It also introduces 2 new variables that can be used in the exutrders GCode:
* `previous_extruder_nr` in the extruder start/prestart GCode, that contains the ID of the extruder we are switching from
* `next_extruder_nr` in the extruder end GCode, that contains the ID of the extruder we are switching to

:information_source: It was also discussed to replace the final substitution that is done in the front-end, to the engine. This is done once the whole GCode has been generated, to insert e.g. the total duration of the print. However, the GCode content is sent to the front-end immediatly after each layer is completed, at which point some layers are still to be processed. So given this process, it is currently impossible to make this change. It would require to send the whole GCode at once, which has other implications and may reduce global performance. This will be discussed further in a other ticket.

https://github.com/Ultimaker/Cura/pull/21306
CURA-12556


include/settings/SettingContainersEnvironmentAdapter.h
```diff
@@ -0,0 +1,40 @@
...
+
+    [[nodiscard]] std::optional<CuraFormulaeEngine::eval::Value> get(const std::string& setting_id) const override;
+
+    [[nodiscard]] bool has(const std::string& key) const override;
+
+    [[nodiscard]] std::unordered_map<std::string, CuraFormulaeEngine::eval::Value> getAll() const override;
+
+private:
+    const Settings& settings_;
+    const std::unordered_map<std::string, std::string> extra_settings_;
```
casperlamboo: I made the [`LocalEnvironment`](https://github.com/Ultimaker/CuraFormulaeEngine/blob/600b197f0a8254f1ff1f8b03b24f4c55faa61898/include/cura-formulae-engine/ast/ast.h#L59-L82) class especially for this reason. 

See example usage here (https://github.com/Ultimaker/CuraFormulaeEngine/blob/600b197f0a8254f1ff1f8b03b24f4c55faa61898/tests/parser.cpp#L717-L718).

I can image this function being used in the following way.
```cpp
const auto settings = SettingContainersEnvironmentAdapter(settings);
const auto settings_extra = CuraFormulaeEngine::env::LocalEnvironment(settings);
settings_extra.set("previous_extruder_nr", previous_extruder_nr);
settings_extra.set("next_extruder_nr", next_extruder_nr);
```

casperlamboo (approved):

include/settings/SettingContainersEnvironmentAdapter.h
wawanbreton: Oh, nice, I didn't see it, but will make the changes !

rburema (commented):

| Benchmark suite | Current: 4cd5086946a637c387f1f8f316703b38421cd432 | Previous: 2e6f6f9bf25c8cedea16a1b226d0dbfb3eba21db | Ratio |
|-|-|-|-|
| `SimplifyTestFixture/simplify_slot_noplugin` | `1.886970768092735` ns/iter | `1.2487941865752963` ns/iter | `1.51` |

casperlamboo (approved):
One small remark, the npm packaging seems to be broken due to this PR. If not too much trouble could you see if you can fix it here. Otherwise we'll fix it when we make the neoprep upgrade.

casperlamboo (approved):

PR #2276, opened 2026-01-28
[CURA-12917] Add a setting to continue bridging over support.

rburema:
This would help with keeping underside surfaces consistent, no matter where the support is generated.

@wawanbreton: For this PR, we needed to introduce a method to print skin lines with different configs along the line. For this I added an `OverrideArea` type which can be used generically to change the config used for any extrusion line. Since this is the same mechanism used for overhanging areas, this part has also be refactored to used this new system.
It cannot be used for the bridging walls calculation though, since those are more complex than just a simple area check.

See (also) front-end PR: https://github.com/Ultimaker/Cura/pull/21341
CURA-12917


casperlamboo (approved):

wawanbreton (commented):

| Benchmark suite | Current: 4ecf97efb3b4620e2a041f59da483e4f3b0e94ec | Previous: 16387565ec4e509f4132f858e3260225624ce7ee | Ratio |
|-|-|-|-|
| `SimplifyTestFixture/simplify_local` | `1.797587577548955` ns/iter | `1.1276244531308395` ns/iter | `1.59` |

wawanbreton (approved):
Approved on behalf of @rburema

PR #2286, opened 2026-02-17
Introduce bridging heuristic(s)

casperlamboo:
# Description

Apply two different behaviours in the case where the bridge area is completely supported and when it is not.

If the area is completely supported from all sides calculate the minimum oriented bounding box. The orientation of this bounding box is then the angle for the bridging.

For the case where the bridge area is not completely supported the old strategy is used where each angle between $0$ and $180$ is considered, and the best evaluated direction is used.

Screenshot 2026-02-17 at 15 58 08

CURA-12975


rburema (approved):
Looks good! Please have a look at my comments anyway.

(Haven't done the component test yet, but I don't expect big problems there either.)

PR #2297, opened 2026-02-24
CURA_13005

HellAholic:
This PR fixes a few wrong behaviors regarding travel moves:
* Remove a travel move that was added on empty layers when processing a modify plugin. Now empty layers remain actually empty. The added travel moved would actually be sometimes incorrect because the last position was unknown. The removed comment mentioned that empty layers would possibly crash the engine. I didn't see an occurence of this happening. And if it ever happens, then the crash should be fixed instead...
* Fix the Z of the first extrustion move of a fractional support layer. Previously the first point Z was at full layer height, with only the second being at the fractional height.
* Fix a possible bump into model when having multiple models and fractional support layer height (extension of CURA-11249)

**Minor Fix**
* Fixed a typo in a comment within `LayerPlanBuffer.cpp` ("toving" → "moving").

Requires https://github.com/Ultimaker/Cura/pull/21456
CURA-13005

### DIFF
--- a/src/LayerPlan.cpp
+++ b/src/LayerPlan.cpp
@@ -24,6 +24,7 @@
 #include "WipeScriptConfig.h"
 #include "arachne/SkeletalTrapezoidation.h"
 #include "arachne/SkeletalTrapezoidationGraph.h"
+#include "bridge/bridge.h"
 #include "communication/Communication.h"
 #include "geometry/OpenPolyline.h"
 #include "geometry/conversions/Point2D_Point2LL.h"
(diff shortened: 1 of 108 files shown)


rburema (approved):
LGTM. Do we want this in main or in the 5.13? (I think main is sensible in this case...)

rburema (commented):

| Benchmark suite | Current: 71feb005f6a193466d8aeecaf1c3697f143ae66d | Previous: e3393608aa02342bfe123a7d191b8837e0edddd5 | Ratio |
|-|-|-|-|
| `SimplifyTestFixture/simplify_local` | `1.9139461947963314` ns/iter | `1.2467169388264958` ns/iter | `1.54` |

src/LayerPlan.cpp
```diff
@@ -2420,7 +2421,12 @@ void LayerPlan::addLinesInGivenOrder(
         }
         else
         {
-            addTravel(start, ForceRetract::AUTOMATIC, config.z_offset);
+            if (config.z_offset != 0)
+            {
+                // Make sure we don't travel at fractional height
+                addTravel(getLastPlannedPositionOrStartingPosition());
```
casperlamboo, line 2427: Does this mean that while we're printing the fractional layer part of the layer and traveling between these part we perform a slight "z-hop". I thought the logic was that the fractional part of a layer is always printed before the rest of layer such that we could travel at the fractional height.

wawanbreton: > I thought the logic was that the fractional part of a layer is always printed before the rest of layer such that we could travel at the fractional height.

It was the original idea, but that is actually impossible to always ensure, for instance when doing a dual extrusion print. Then depending on the layer, one extruder is printed before the other. So if the second printed extruder has a fractional layer, you could travel through the printed parts of the first one.
In the end we decided to always travel to the layer height, which indeed creates a small z-hop.

PR #2301, opened 2026-03-05
CURA-12992 Fix slicing crashes

### LINKED ISSUE #20962: Will Slice on Ender 3 v2 Neo but not on Ender 3 S1
### Cura Version

5.10.2

### Operating System

Windows 11 24H2

### Printer

Ender 3 S1

### Describe your problem and add the package for technical support as a .zip here ⬇️

Simple challenge coin I designed in Tinkercad.  The file previously sliced, but I made some modifications to the model.  The STL sliced fine using Ender 3 v2 Neo printer profile, but would not slice using Ender 3 S1 profile.

[New Compressed (zipped) Folder.zip](https://github.com/user-attachments/files/22368537/New.Compressed.zipped.Folder.zip)

### LINKED ISSUE #20967: Slicing failed
### Cura Version

5.10.0

### Operating System

macOS 15.6.1

### Printer

Anycubic i3 Mega

### Describe your problem and add the package for technical support as a .zip here ⬇️

it happened when I chose raft option

### LINKED ISSUE #20978: Part 1 and part 4 wont slice
### Cura Version

5.10.1

### Operating System

windows

### Printer

voxelab aquila 2

### Describe your problem and add the package for technical support as a .zip here ⬇️

[REPAIRED-Ford 1934 Coupe Legends.zip](https://github.com/user-attachments/files/22523773/REPAIRED-Ford.1934.Coupe.Legends.zip)

### LINKED ISSUE #21128: Slicing Failed, but Only with Supports... no idea the cause
### Cura Version

5.10.2

### Operating System

Windows 11

### Printer

Ender 6

### Describe your problem and add the package for technical support as a .zip here ⬇️

Slicing Failed, but Only with Supports... no idea the cause
When I turn off supports, I can slice just fine...
It will work with supports if not "Tree" but instead Normal.

[export-20251108-115531.zip](https://github.com/user-attachments/files/23434177/export-20251108-115531.zip)

### LINKED ISSUE #21235: Slicing failed - dino pen holder
### Cura Version

5.11.0

### Operating System

Windows 10

### Printer

Ender 3 V3 SE

### Describe your problem and add the package for technical support as a .zip here ⬇️

The slicing fails while attempting to slice this model: https://www.printables.com/model/425498-chubby-t-rex-dino-pencil-holder-pot-no-supports. 

Export package:

[export-20251217-232340.zip](https://github.com/user-attachments/files/24223177/export-20251217-232340.zip)

### LINKED ISSUE #21357: slicing failed
### Cura Version

5.11.0

### Operating System

win 11 Device specifications below

### Printer

elegoo nep 4 pro

### Reproduction steps

https://github.com/user-attachments/assets/e4ba9d73-4a7c-479d-a151-7c24ee66fd67
Device specifications :

### Actual results

Image
Device specifications :
Device Name: DESKTOP-PMA93B2
Processor: 12th Gen Intel(R) Core(TM) i5-12400F (2.50 GHz)
Installed RAM: 32.0 GB (31.9 GB Used)
Device ID: CF9ED205-14CF-4630-BF6C-E68F32769489
Product ID: 00330-80196-65662-AA313
System Type: 64-bit OS, x64-based processor
Pen & Touch: No touch or pen input available for this display
Version: Windows 11 Pro
Build: 25H2
Installed: 01/06/2025
OS Build: 26200.7623
Windows Experience Pack: 1000.26100.275.0

### Expected results

normal sliceing just.... 

### Add your .zip and screenshots here ⬇️

[EN4PRO_بوصة مبرومة 8 شكل جديد.zip](https://github.com/user-attachments/files/24988647/EN4PRO_.8.zip)

### LINKED ISSUE #21375: slicing failed
### Cura Version

5.11.0

### Operating System

windows 10

### Printer

elegoo neptune 4 max

### Describe your problem and add the package for technical support as a .zip here ⬇️

other profiles work, just not my PLA+ one anymore. happened on the older version of cura too.
[cura error.zip](https://github.com/user-attachments/files/25089606/cura.error.zip)

wawanbreton:
This PRs proposes a few changes to fix the most occuring slicing crashes inside Arachne:
* Use an integer-based voronoi diagram when calculating skeletal trapezoidation: previously we used double values, which would cause unexpected behaviors, because the built diagram would consider some close points to be separate, but when treating them afterwards using a Point2LL, they would end up at the same position. Now the voronoi diagram precision is consistent with the precision of the subsequent algorithms.
* Change the edge insideness check: the previous method was based on a corner insideness check, which is extremely fast but would give a wrong result in some very tricky cases (with the initial shape contains overlapping vertices). Now we use a polygon insideness check.
* Use consistent settings when generating skeletal trapezoidation for non-walls: previously, we would always take the walls settings and apply them to arachne, but e.g. when generating a concentric infill for ironing, the line width is 0.1mm instead of 0.4mm, although the minimum authorized line width is actually 0.2mm. This case would generate impossible cases, thus wrong data and crashes. For the non-walls cases, we now use settings based on the current line width. The calculations are hard-coded to their equivalent formula from fdprinter.def.json.
* Fix a tree support crash: not arachne-related, but I also encountered a possible division by 0 within tree support, which was pretty easy to fix.

### Performance analysis
* Since we output a voronoi diagram that uses integers, it should have less points than previously, possibly speeding up subsequent calculations, and reducing the required conversions.
* The edge insideness check is now based on the polygon insideness, which is a much heavier calculation. This is only for point-based cells of the diagram though.
In practice, I have made performance checks on a complex model and did not see any significant difference. More testing could be done though.

### Risk analysis
Changing the voronoi diagram to integer type is a huge behavioral change. Although it is more consistent with the rest of the algorithm, it changes the input data for arachne of every single case. So far I did not observe any regression, although at first it raised the occurences of the insideness check failure (since it naturally generates more overlapping points).

CURA-12992

Fixes https://github.com/Ultimaker/Cura/issues/21357
Fixes https://github.com/Ultimaker/Cura/issues/21235
Fixes https://github.com/Ultimaker/Cura/issues/21128
Fixes https://github.com/Ultimaker/Cura/issues/20967
Fixes https://github.com/Ultimaker/Cura/issues/20962
Should fix https://github.com/Ultimaker/Cura/issues/20978
Should fix https://github.com/Ultimaker/Cura/issues/21375

### DIFF
--- a/include/WallToolPaths.h
+++ b/include/WallToolPaths.h
@@ -102,9 +102,9 @@ protected:
      *
      * Works on both toolpaths and inner contours simultaneously.
      *
-     * \param settings The settings as provided by the user
+     * \param stitch_distance The stitch distance to be used
      */
-    static void stitchToolPaths(std::vector<VariableWidthLines>& toolpaths, const Settings& settings);
+    static void stitchToolPaths(std::vector<VariableWidthLines>& toolpaths, const coord_t stitch_distance);
 
     /*!
      * Remove polylines shorter than half the smallest line width along that polyline, if that polyline isn't part of an outer wall.
@@ -128,8 +128,16 @@ private:
     bool print_thin_walls_; //<! Whether to enable the widening beading meta-strategy for thin features
     coord_t min_feature_size_; //<! The minimum size of the features that can be widened by the widening beading meta-strategy. Features thinner than that will not be printed
     coord_t min_bead_width_; //<! The minimum bead size to use when widening thin model features with the widening beading meta-strategy
-    double small_area_length_; //<! The length of the small features which are to be filtered out, this is squared into a surface
-    coord_t transition_length_; //<! The transitioning length when the amount of extrusion lines changes
+    const AngleRadians wall_transition_angle_;
+    const coord_t wall_transition_length_;
+    const double min_even_wall_line_width_;
+    const double wall_line_width_0_;
+    const double min_odd_wall_line_width_;
+    const double wall_line_width_x_;
+    const int wall_distribution_count_;
+    const coord_t wall_transition_filter_distance_;
+    const coord_t wall_transition_filter_deviation_;
+    const coord_t stitch_distance_;
     bool toolpaths_generated_; //<! Are the toolpaths generated
     std::vector<VariableWidthLines> toolpaths_; //<! The generated toolpaths binned by inset_idx.
     Shape inner_contour_; //<! The inner contour of the generated toolpaths
--- a/include/arachne/SkeletalTrapezoidation.h
+++ b/include/arachne/SkeletalTrapezoidation.h
@@ -249,7 +249,8 @@ protected:
         vd_t::edge_type*& starting_vd_edge,
         vd_t::edge_type*& ending_vd_edge,
         const std::vector<Point2LL>& points,
-        const std::vector<Segment>& segments);
+        const std::vector<Segment>& segments,
+        const Shape& polys);
 
     /*!
      * Compute the range of line segments that surround a cell of the skeletal
--- a/include/arachne/SkeletalTrapezoidation.h
+++ b/include/arachne/SkeletalTrapezoidation.h
@@ -20,6 +20,7 @@
 #include "utils/ExtrusionJunction.h"
 #include "utils/ExtrusionLine.h"
 #include "utils/PolygonsSegmentIndex.h"
+#include "utils/VoronoiUtils.h"
 #include "utils/section_type.h"
 
 namespace cura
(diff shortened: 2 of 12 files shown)


rburema (commented):

| Benchmark suite | Current: 8e708f1b3087ca0716a8820a8c8dd34316cf0d79 | Previous: 38595fafd6b9dd003b5638409c22a08232c6dde3 | Ratio |
|-|-|-|-|
| `SimplifyTestFixture/simplify_local` | `8.746792701811529` ns/iter | `1.2826544326631881` ns/iter | `6.82` |

rburema (approved):
That was surprisingly straightforward to review.

I don't think I have any remarks; good work! We've been searching for these for a long time.

PR #2313, opened 2026-03-27
CURA-13049 process the end variable replacement in the engine

wawanbreton:
This PR proposes a few changes to process the end-of-line variables replacement in the engine. It implements the following changes:
* When writing a piece of fixed GCode, instead of sending it to the buffer immediatly, it is kept in a `FixedGCodePart` instance
* When writing a piece of to-be-resolved GCode, instead of resolving it now and writing it, it is kept in a `ResolvedGCodePart` instance
* At the end of the slicing phase, the GCode parts are resolved and send to the front-end in a row.
* The initial extruder number can now be calculated just before resolving the GCode parts, so that we can use the actual one instead of the supposed one. The value is then sent back to the front-end.
* The GCode export now calls the `Communication` object to send the different pieces of information at the proper times, so that the proces is more repeateable across different implementations, and this also removes some facade code in the `FffProcessor` and `FffGCodeWriter`.
* We receive from the front-end some information about the material (density, spool cost and spool weight) so that we can calculate the filament usage stats, which are then inserted in to the resolved GCode, and send back to the front-end for display.

:information_source: Some technical notes:
* Since we now store the whole GCode in memory before sending it, the amout of memory used by the engine can possibly be much larger. If this is an issue with very large models, we will have to find a strategy to reduce it.
* The `EmscriptenCommunication` has been updated to reach feature parity. However it is to note that the `Communication::beginGCode` method has been removed, and its content has been placed in the constructor instead. The previous behavior can be restored if necessary.
* Since the whole GCode is resolved and sent at the end in all cases, this behavior could also be applied to the GCode header by creating the proper variables. This would avoid having a partial-resolved header in some cases and handling the sequential/non-sequential communications separately. I suggest creating a follow-up ticket.

Since this is now easily possible, it also adds a few useful variables to the resolvable GCode:
* `initial_layer_bb_{min_x, max_x, min_y, max_y, width, height}` containing the 2D bounding box of the effective print on the first layer (this includes every helper e.g. brim, skirt, prime tower, ...)
* `total_bb_{min_x, max_x, min_y, max_y, min_z, max_y, width, depth, height}` containing the 3D bounding box of the models to be printed
* `is_extruder_used` containing a list of 16 booleans, indicating which extruder is effectively used during the print

Requires https://github.com/Ultimaker/Cura/pull/21500
Requires https://github.com/Ultimaker/CuraFormulaeEngine/pull/2
CURA-13049


rburema (commented):

| Benchmark suite | Current: 13f1704298bdc201922904607deb54fcab1a3468 | Previous: 7300d0877921a84e5bb213d17c49ed757ed3a147 | Ratio |
|-|-|-|-|
| `SimplifyTestFixture/simplify_local` | `1.9423783240468322` ns/iter | `1.2484627420333818` ns/iter | `1.56` |

include/gcode_export/gcodeExport.h
```diff
@@ -613,15 +637,20 @@ class GCodeExport : public NoCopy
     void writeCode(const std::string& str);
 
     /*!
-     * Write code while temporarily ensuring absolute extrusion mode.
+     * Write a pice of resolvable GCode while temporarily ensuring absolute extrusion mode.
```
casperlamboo:
```suggestion
     * Write a piece of resolvable GCode while temporarily ensuring absolute extrusion mode.
```

src/PrintInformation.cpp
```diff
@@ -0,0 +1,21 @@
...
+#include "LayerPlan.h"
+#include "sliceDataStorage.h"
+
+
+void cura::PrintInformation::updateWithLayer(const LayerPlan* layer_plan)
+{
+    if (! initial_extruder_nr.has_value())
+    {
+        initial_extruder_nr = layer_plan->findInitialExtruderNr();
+    }
```
casperlamboo, line 16: why not always do this? either we're re-using this object but then I would argue is a bit bug prone since we do want to recalculate this between layers or we're not re-using this and then it's a superfluous check.

src/gcode_export/GcodeTemplateResolver.cpp
```diff
@@ -17,85 +15,61 @@
...
- * @param expression The raw expression text to be resolved
- * @param environment The contextual environment to be used when resolving variables
- * @return The parsed value result, or an error description
- */
-zeus::expected<cfe::eval::Value, EvaluateResult> evaluateExpression(const std::string_view& expression, const cfe::env::Environment& environment)
+    initial_extruder_nr_ = initial_extruder_nr;
+
+    // Create an environment containing all the extra global settings
+    global_environment_ = std::make_shared<cfe::env::LocalEnvironment>(&CuraFormulaeEngine::env::std_env);
+    global_environment_->add(extra_global_settings);
```
casperlamboo, line 30: Depending on the size of `extra_global_settings` you can also consider not using the `add` function (and copying all values) but instead introduce two layers of scoping.

src/PrintInformation.cpp
wawanbreton: The `PrintInformation` is about the whole print, and here we want to find the very initial extruder. Usually it will be filled at layer 0 and then never overridden, but it could be that layer 0 has no extrusion, then you want the extruder at the first layer that actually has an extrusion. The test it there to make sure that once we have found the first layer with an extrusion, we don't update it anymore.
Maybe a comment would help here ? Also I should document the function.

src/gcode_export/GcodeTemplateResolver.cpp
wawanbreton: That is true, but in practice there is always global settings when calling this method, so unless you want to re-use it somewhere else, it would just add some extra-complexity here.
Not 100% sure I understand your remark though :thinking:

casperlamboo: Not sure if we have the right "adpaters" for this, but instead of doing this
```cpp
    global_environment_ = std::make_shared<cfe::env::LocalEnvironment>(&CuraFormulaeEngine::env::std_env);
    global_environment_->add(extra_global_settings);
    global_environment_->set("initial_extruder_nr", static_cast<int64_t>(initial_extruder_nr_));
```
Do this
```cpp
    global_environment_ = std::make_shared<cfe::env::LocalEnvironment>(
        ScopedEnvironment(EnvironmentMap(extra_global_settings), CuraFormulaeEngine::env::std_env)
    );
    global_environment_->set("initial_extruder_nr", static_cast<int64_t>(initial_extruder_nr_));
```
Where `EnvironmentMap` takes two `Envionment`s and considers the second argument to be shadowing the first. Similar to `LocalEnvironment` the result is itself an `Envionment`. Benifit is that there is less overhead constructing the  `global_environment_` since we're not copying values. Downside is that lookups are more expensive since we need to go through 3 layers of environments in the worst case instead of 2.

wawanbreton: At some point I have been considering making an `EnvironmentChain` that could contain any number of chained environment, but that is not really possible in the end because the chain looks more like a graph.
About the performance, the `prepareForResolving` method is called only once in the whole engine processing, so I do think the cost of copying the extra settings is quite negligible. Also I don't see how your method avoids copying them :thinking:. If we really want to avoid that, it could be given as a `move`-able variable.

casperlamboo: You avoid copying since you're not calling `add` which copies all values from `extra_global_settings` and instead use the `extra_global_settings` as-is. But I agree benifits are limited

wawanbreton: From what I can see, the `extra_global_settings` would be copied in the `EnvironmentMap` constructor anyway. But never mind, let's keep it as is.

PR #2314, opened 2026-03-30
CURA-13064 Expand the Retract Before Outer Wall setting

wawanbreton:
We now handle more detailed behaviors regarding the retraction of the travel move when moving to an outer wall.

Requires https://github.com/Ultimaker/Cura/pull/21505
CURA-13064

### DIFF
--- a/src/InsetOrderOptimizer.cpp
+++ b/src/InsetOrderOptimizer.cpp
@@ -22,6 +22,7 @@
 
 #include "ExtruderTrain.h"
 #include "FffGcodeWriter.h"
+#include "ForceRetract.h"
 #include "LayerPlan.h"
 #include "utils/views/convert.h"
 #include "utils/views/dfs.h"
@@ -45,7 +46,7 @@ InsetOrderOptimizer::InsetOrderOptimizer(
     const GCodePathConfig& inset_X_flooring_config,
     const GCodePathConfig& inset_0_bridge_config,
     const GCodePathConfig& inset_X_bridge_config,
-    const bool retract_before_outer_wall,
+    const RetractBeforeOuterWall retract_before_outer_wall,
     const coord_t wall_0_wipe_dist,
     const coord_t wall_x_wipe_dist,
     const size_t wall_0_extruder_nr,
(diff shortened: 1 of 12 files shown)


src/InsetOrderOptimizer.cpp
```diff
@@ -160,6 +161,21 @@ bool InsetOrderOptimizer::addToLayer()
...
+        force_retract = ForceRetract::AUTOMATIC;
+        break;
+    case RetractBeforeOuterWall::RETRACTED:
+        force_retract = ForceRetract::RETRACTED;
+        break;
+    case RetractBeforeOuterWall::NOT_RETRACTED:
+    case RetractBeforeOuterWall::NOT_RETRACTED_FIRST:
+        force_retract = ForceRetract::NOT_RETRACTED;
+        break;
+    }
```
casperlamboo: I personally have a slight preference for this notation. I find it overly complex to read a whole mapping, just to discover that most values map to itself.
```suggestion
    switch (retract_before_outer_wall_)
    {
    case RetractBeforeOuterWall::NOT_RETRACTED_FIRST:
        force_retract = ForceRetract::NOT_RETRACTED;
        break;
    default:
        force_retract = retract_before_outer_wall_;
        break;
    }
```

src/InsetOrderOptimizer.cpp
```diff
@@ -202,6 +218,12 @@ bool InsetOrderOptimizer::addToLayer()
             scarf_seam,
             smooth_speed);
         added_something = true;
+
+        if (retract_before_outer_wall_ == RetractBeforeOuterWall::NOT_RETRACTED_FIRST)
+        {
+            // Only the first should be forced, the next ones use auto
+            force_retract = ForceRetract::AUTOMATIC;
+        }
```
casperlamboo: Nit picky, but why not use the same patteren as the switch statement above, this code achieves a very similar goal.

src/LayerPlanBuffer.cpp
```diff
@@ -153,10 +153,10 @@ void LayerPlanBuffer::addConnectingTravelMove(LayerPlan* prev_layer, const Layer
 
             next_mesh_retract_before_outer_wall = inset_direction == InsetDirection::OUTSIDE_IN || wall_line_count == 1;
         }
-        const bool force_retract = retract_at_layer_change || next_mesh_retract_before_outer_wall;
+        const ForceRetract force_retract = retract_at_layer_change || next_mesh_retract_before_outer_wall ? ForceRetract::RETRACTED : ForceRetract::AUTOMATIC;
```
casperlamboo: I find it hard to read inline if statements like this since I don't know order of operations by hearth. I suggest adding parenthesis to make this more clear
```suggestion
        const ForceRetract force_retract = (retract_at_layer_change || next_mesh_retract_before_outer_wall) ? ForceRetract::RETRACTED : ForceRetract::AUTOMATIC;
```

casperlamboo (approved):

src/InsetOrderOptimizer.cpp
wawanbreton: I agree, I don't really like how this piece of code ends up being so large for a so trivial behavior, but I still prefer it for 2 reasons:
* Enums are different types and could evolve separately. If you do an assignment with a cast and one of them changes in the future, you will just end up with an un-noticed bug.
* I try to never use `default` in switches, because when you add a value to your enum, you get a warning that tells you there is a new case to handle, which you should do. When using `default` you don't get this warning, so it is very likely that you will forget to update the code here.

src/InsetOrderOptimizer.cpp
wawanbreton: This treats a case for a specific value of the enum, so in this case I do allow myself to not use a switch-case.
If the enum changes in the future, either this value is changed and you get a compiler error, or you added a new value and you will have to add the proper code anyway.
But yes, technically we should use a switch-case every time we read an enum value. But for the readibility reasons you mentioned, I try to do it only when the code is not future-proof.

casperlamboo (approved):

PR #2315, opened 2026-03-30
CURA-13065 Fix polygons-based infill not respecting end close to seam

wawanbreton:
Path optimization for polygons didn't support reverting the optimized path, so for polygons-based infill we would start by the point close to seam instead of end there.
We now have an argument to reverse the paths for polygons, and call it appropriately from the infill optimizer.

CURA-13065


rburema (commented):

| Benchmark suite | Current: f28c6c20af79d4257a8dd9c6fe1ef7f721da04d1 | Previous: 7436652244fe6a8b49fc1234ffddec9b32c685f7 | Ratio |
|-|-|-|-|
| `SimplifyTestFixture/simplify_slot_noplugin` | `4.048432415593224` ns/iter | `2.6182001547576443` ns/iter | `1.55` |

rburema (approved):
LGTM!

PR #2316, opened 2026-04-01
[CURA-13024] Spike: merge thin parts of floor/roof/skin

rburema:
Merge roof/floor into skin or vice-versa, given a distance, where anything under this is considered 'too thin' and will be merged to the other areas when next to each other.

See frontend PR for the accompanying setting: https://github.com/Ultimaker/Cura/pull/21515


src/FffGcodeWriter.cpp
```diff
@@ -3131,6 +3131,17 @@ bool FffGcodeWriter::processSkin(
...
+std::tuple<Shape, Shape> mergeThinOverlap(const coord_t max_dist, const Shape& assume_bigger, const Shape& assume_smaller)
+{
+    const auto result_smaller = assume_smaller // Of the (supposedly) smaller area,
+                                    .difference(assume_bigger.offset(max_dist)) // take the difference with an offset of the bigger area,
+                                    .offset(max_dist) // then 'inflate' any leftover pieces (so, ones that are certainly big enough),
+                                    .intersection(assume_smaller); // and lastly intersect with the original area, so we don't go outside those bounds.
+    return std::make_tuple(
+        assume_bigger.unionPolygons(assume_smaller.difference(result_smaller).offset(EPSILON)), // Glue any 'not leftover' pieces to the (supposedly) bigger area.
+        result_smaller);
+}
```
wawanbreton: Since this is a (very nice) utility geometry function, can it be moved to a more accessible place, and get a proper documentation ?

src/FffGcodeWriter.cpp
```diff
@@ -3140,27 +3151,34 @@ bool FffGcodeWriter::processSkinPart(
...
-        mesh.flooring_angles,
-        added_something);
-    processTopBottom(storage, gcode_layer, mesh, extruder_nr, mesh_config, skin_part, added_something);
+
+    const coord_t decimate_roofing_flooring_distance = mesh.settings.get<coord_t>("top_bottom_skin_merge_distance");
+    Shape skin_fill, roofing_fill, flooring_fill;
+    if (decimate_roofing_flooring_distance > 0)
+    {
+        // For each of these, have flooring take precedence over roofing;
+        // while roofing is likely visually more pleasing, the flooring settings are probably more important for structural integrity.
```
wawanbreton, line 3157: Completely agree 👍

src/FffGcodeWriter.cpp
```diff
@@ -3140,27 +3151,34 @@ bool FffGcodeWriter::processSkinPart(
...
+        std::tie(flooring_fill, skin_fill) = mergeThinOverlap(decimate_roofing_flooring_distance, skin_part.flooring_fill, skin_part.skin_fill);
+        std::tie(roofing_fill, skin_fill) = mergeThinOverlap(decimate_roofing_flooring_distance, skin_part.roofing_fill, skin_fill);
+        // Then, the less likely scenario that flooring and roofing are both present, next to each other, and one is far thinner than the other.
+        std::tie(flooring_fill, roofing_fill) = mergeThinOverlap(decimate_roofing_flooring_distance, flooring_fill, roofing_fill);
+        std::tie(roofing_fill, flooring_fill) = mergeThinOverlap(decimate_roofing_flooring_distance, roofing_fill, flooring_fill);
+        // Lastly, the most likely scenario, where flooring/roofing is thinner than skin (even after merging as much as possible into either), merge into skin.
+        std::tie(skin_fill, flooring_fill) = mergeThinOverlap(decimate_roofing_flooring_distance, skin_fill, flooring_fill);
+        std::tie(skin_fill, roofing_fill) = mergeThinOverlap(decimate_roofing_flooring_distance, skin_fill, roofing_fill);
```
wawanbreton: That is quite a lot of morphological operations indeed, could we add more early-outs in the if ? There are many layers where there is no flooring and/or roofing, so that would probably avoid some unnecessary calculations.

src/FffGcodeWriter.cpp
```diff
@@ -3131,6 +3131,17 @@ bool FffGcodeWriter::processSkin(
     return added_something;
 }
 
+std::tuple<Shape, Shape> mergeThinOverlap(const coord_t max_dist, const Shape& assume_bigger, const Shape& assume_smaller)
```
wawanbreton: If am I correct, all the calls to this functions map the 2 output shapes to new versions of the input shapes, so maybe it would be more efficient/readable to give them as non-const references and modify them in-place ?

wawanbreton (changes requested):
The global implementation is very elegant and should indeed work without edge-cases. I have a few suggestions/requests though, to be challenged of course 🙂

src/FffGcodeWriter.cpp
rburema: Sure, we can put it in `polygonUtils.___` with some docs.

src/FffGcodeWriter.cpp
rburema: Yes, can do.

src/FffGcodeWriter.cpp
rburema: Yes, not sure why I didn't do that in the first place.

src/FffGcodeWriter.cpp
rburema: I ended up placing the early-out in the method itself, as otherwise, the code would become a lot more unreadable I think.

src/utils/polygonUtils.cpp
```diff
@@ -1518,6 +1518,20 @@ Shape PolygonUtils::clipPolygonWithAABB(const Shape& src, const AABB& aabb)
     return out;
 }
 
+void PolygonUtils::mergeThinOverlap(const coord_t max_dist, Shape& assume_bigger, Shape& assume_smaller)
+{
+    if (assume_bigger.area() == 0.0 || assume_smaller.area() == 0.0)
```
wawanbreton: Hmm, this is a quite heavy calculation for an early-out 😅  I was thinking of a simple `empty()`, or do you think that is not good enough ?

wawanbreton (approved):
Just one more possible optimization, otherwise looks good !

src/utils/polygonUtils.cpp
rburema: Yes that should suffice -- I'm not sure why I choose area, that can make it more slow even, since the operations themselves also check! 😅

src/utils/polygonUtils.cpp
```diff
@@ -1518,18 +1518,38 @@ Shape PolygonUtils::clipPolygonWithAABB(const Shape& src, const AABB& aabb)
...
+    // Make a morphological opening to keep only wide areas, and a difference to actually keep only the thin areas, which we are allowed to grow over
+    const Shape destination_wide_areas = destination.offset(-max_dist).offset(max_dist + EPSILON);
+    const Shape allow_grow_area = destination.difference(destination_wide_areas);
+    if (allow_grow_area.empty())
+    {
+        return;
+    }
+
+    // If necessary, remove the thin parts of the source to not allow them to grow (using the same principle as above)
+    const Shape source_grow_part = allow_thin_areas_grow ? source : source.intersection(source.offset(-max_dist).offset(max_dist + EPSILON));
```
rburema: I don't think it's worth it (since it would become less readable), but if we really wanted to, we could prevent the copy of `source` to `source_grow_parts` in the case `allow_thin_areas_grow` is true (and instead make it a reference).

src/utils/polygonUtils.cpp
```diff
@@ -1518,18 +1518,38 @@ Shape PolygonUtils::clipPolygonWithAABB(const Shape& src, const AABB& aabb)
...
+    if (source_grow_part.empty())
+    {
+        return;
+    }
+
+    // Now calculate the actual growing area, which is the intersection of the offset source with the allowed growing area
+    const Shape actual_grow_area = source_grow_part.offset(max_dist).intersection(allow_grow_area).offset(EPSILON);
```
rburema, line 1543: Since both `source_grow_part` and `allow_grow_area` already include `EPSILON`, I think you can add the epsilon to the `max_dist` instead here, thus avoiding the final offset operation (as those aren't the cheapest).

rburema (commented):
Two very small remarks (only one of which isn't a complete nitpick). I'm passing this along to QA again now since they're just small refactors, not functional changes, in any case.

src/utils/polygonUtils.cpp
wawanbreton: Hmm, I like the idea, but I think that would defeat the purpose of this offset. If I do it before the intersection, the final shape will not go outside the `allow_grow_area` that it intersects with, which is required to properly make the subsequent union with the `source`.

src/utils/polygonUtils.cpp
wawanbreton: Well, I completely agree, don't like this copy either, but I can't find a way to optimize it without making an ugly code...

src/utils/polygonUtils.cpp
```diff
@@ -1552,6 +1551,23 @@ void PolygonUtils::mergeThinOverlap(const coord_t max_dist, Shape& source, Shape
     destination = destination.difference(actual_grow_area);
 }
 
+Shape PolygonUtils::getThinAreas(const Shape& shape, const coord_t max_width)
+{
+    // Extract the wide areas, then do a difference to actually keep only the thin areas
+    return shape.difference(getRawWideAreas(shape, max_width, EPSILON));
```
casperlamboo, line 1557: why provide `EPSILON` as it is the default value for this argument?

src/utils/polygonUtils.cpp
```diff
@@ -1552,6 +1551,23 @@ void PolygonUtils::mergeThinOverlap(const coord_t max_dist, Shape& source, Shape
...
+}
+
+Shape PolygonUtils::getWideAreas(const Shape& shape, const coord_t min_width)
+{
+    // Extract the raw wide areas, then do an intersection to keep them inside the original shape
+    return shape.intersection(getRawWideAreas(shape, min_width, EPSILON));
```
casperlamboo, line 1563: why provide `EPSILON` as it is the default value for this argument?

src/skin.cpp
```diff
@@ -365,7 +366,7 @@ void SkinInfillAreaComputation::generateSkinRoofingFlooringFill(SliceLayerPart&
         if (has_roofing)
         {
             const Shape below_inside = skin_part.outline.intersection(filled_area_below);
-            skin_part.roofing_fill = below_inside.difference(filled_area_above).offset(roofing_expansion).intersection(below_inside);
+            skin_part.roofing_fill = below_inside.difference(filled_area_above).intersection(below_inside);
```
casperlamboo: We don't need the intersection anymore; it was only needed before since we did the offset.
```suggestion
            skin_part.roofing_fill = below_inside.difference(filled_area_above);
```

src/skin.cpp
```diff
@@ -376,6 +377,40 @@ void SkinInfillAreaComputation::generateSkinRoofingFlooringFill(SliceLayerPart&
...
+            const Shape expansion_grow_area = roofing_expansion > 0 ? skin_part.roofing_fill.offset(roofing_expansion).intersection(skin_part.skin_fill).offset(EPSILON) : Shape();
+            const Shape merge_grow_area = top_bottom_skin_merge_distance > 0 ? PolygonUtils::getThinAreas(skin_part.skin_fill, top_bottom_skin_merge_distance) : Shape();
+            if (! expansion_grow_area.empty() || ! merge_grow_area.empty())
+            {
+                const Shape total_roofing_grow_area = expansion_grow_area.unionPolygons(merge_grow_area);
+                if (! total_roofing_grow_area.empty())
+                {
+                    skin_part.roofing_fill = skin_part.roofing_fill.unionPolygons(total_roofing_grow_area);
+                    skin_part.skin_fill = skin_part.skin_fill.difference(total_roofing_grow_area);
+                }
```
casperlamboo: [[nitpick]] `total_` prefix is a bit nugatory
```suggestion
                const Shape roofing_grow_area = expansion_grow_area.unionPolygons(merge_grow_area);
                if (! roofing_grow_area.empty())
                {
                    skin_part.roofing_fill = skin_part.roofing_fill.unionPolygons(roofing_grow_area);
                    skin_part.skin_fill = skin_part.skin_fill.difference(roofing_grow_area);
                }
```

src/utils/polygonUtils.cpp
wawanbreton: Yeah, it is really not necessary but since the difference operation requires the areas to be offsetted by EPSILON, it makes it clear at this point that is has properbly been done (no need to theck inside the called function).

src/utils/polygonUtils.cpp
wawanbreton: Same reasoning here

src/skin.cpp
wawanbreton: Oh yes indeed, now the result can't go beyond the borders.

src/skin.cpp
wawanbreton: It is somehow a "total" in the sense that it contains 2 united areas. Would an other wording make more sense to you ?

wawanbreton (approved):

PR #2317, opened 2026-04-02
CURA-12580 paint on support

wawanbreton:
This PR adds the generation of modifier meshes based on painting data for support placement. The mesh generation is mostly based on the one made for the multi-material, in a simplified way:
* Since we don't need to propagate the voxels inside the model, we just create an initial voxels grid based on the texture data, then grow it by 1 unit to make sure we generate large enough meshes that will cover the borders of the model.
* The voxels having a value of 0 (automatic support) don't generate a modifier mesh, so they are completely ignored.

The meshes are then generated, depending on whether they force or disable support:
* Painting that disable support generate an `anti_overhang_mesh`.
* Painting that force support generate a `force_support_overhang_mesh`, which is a new type of modifier mesh, that can not be generated by the front-end.

In order to handle the new `force_support_overhang_mesh`, we also generalized the use of the `isPrinted()` and `isModelMesh`methods that actually differentiate between modifier meshes and regular model meshes. This way we can add new modifier mesh types very easily.

The `force_support_overhang_mesh` is then handled by support generation, when calculating the overhang areas and force those te be included by unioning them to the result.

CURA-12580
Requires https://github.com/Ultimaker/Cura/pull/21520

:warning: :warning: :warning: :warning: :warning: :warning: :warning: 
The PR now points to CURA-12636 because I needed some debug code from this branch. However the associated Cura PR points to main, so make sure CURA-12636 is already merged to 5.13 then 5.13 merged to main when merging this one.
:warning: :warning: :warning: :warning: :warning: :warning: :warning:


src/MeshMaterialSplitter.cpp
```diff
@@ -55,7 +55,7 @@ union ContourKey
     struct
     {
         uint16_t z;
-        uint8_t extruder;
+        uint8_t value;
```
casperlamboo: I find the name `extruder` more explanatory compared to `value`, is it used in a situation where `value` doesn't make sense? Also the above comment

> Utility structure to store values in a map that are grouped by unique z-height and extruder number combinations

Doesn't make sense for the changed code

casperlamboo: Maybe `feature` makes more sense compared to value? where `feature` can mean material/support etc?

casperlamboo (approved):

src/MeshMaterialSplitter.cpp
wawanbreton: Indeed, it can be used for either the extruder number or the support painted value, and just `value` is too much generic. `feature` makes some sense, but it makes me think that it indicates which feature is being used. So would you be ok with `feature_value` ?

wawanbreton:
Superseded by https://github.com/Ultimaker/CuraEngine/pull/2337

PR #2318, opened 2026-04-03
CURA-13072 Fix impossible ordering conditions

wawanbreton:
In some cases, adding ordering conditions with the `InsetOrderOptimizer::addFirstWallOrder` method would create unresolvable conditions and leave the wall parts not being added at all.
Since the purpose of this method was to favorize the longer wall to start, we now do it in the `PathOptimizer` when differentiating the starting candidates at the first round.

Part of the code in this PR is due to the type change in the `PathOrdering` struct: in order to calculate the length, we need a `Polyline` instead of a `PointsSet`.

CURA-13072

### DIFF
--- a/include/PathOrderOptimizer.h
+++ b/include/PathOrderOptimizer.h
@@ -184,7 +184,7 @@ public:
         // Get the vertex data and store it in the paths.
         for (auto& path : paths_)
         {
-            path.converted_ = path.getVertexData();
+            path.converted_ = &path.getVertexData();
             vertices_to_paths_.emplace(path.vertices_, &path);
         }
 
--- a/include/PathOrderOptimizer.h
+++ b/include/PathOrderOptimizer.h
@@ -99,6 +99,12 @@ public:
     static const std::unordered_multimap<Path, Path> no_order_requirements_;
 
     /*!
+     * When true and having multiple candidates for the starting path, prefer the longest one over the closest one. It will be set back to false
+     * after being used, so that it will apply only over the very first path.
+     */
+    bool prefer_longest_path_;
+
+    /*!
      * Construct a new optimizer.
      *
      * This doesn't actually optimize the order yet, so the ``paths`` field will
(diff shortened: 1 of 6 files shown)


include/PathOrderOptimizer.h
```diff
@@ -600,23 +608,48 @@ class PathOrderOptimizer
...
+    {
+        OrderablePath* best_candidate = prefer_longest_path_ ? findLongestPath(candidate_paths) : findClosestPath(start_position, candidate_paths);
+        prefer_longest_path_ = false;
+        return best_candidate;
+    }
+
+    OrderablePath* findLongestPath(std::vector<OrderablePath*> candidate_paths)
+    {
+        OrderablePath* longest_path = nullptr;
+        coord_t longest_path_length;
```
rburema: I know the `longest_path == nullptr` below takes care of it, but this will trip up linters and warning systems if left uninitialized, so I think it's still good form to have some initial value.

Come to think of it, since 'length' is always positive, and `coord_t` is a signed integer, we could put it at `-1` and not even bother with the `null` check.

rburema (approved):
LGTM! One very small pedantic request.

include/PathOrderOptimizer.h
wawanbreton: Yeah linters are a bit annoying on this case, which keeps happening. I will add an initialization, since I don't like having a negative value for a length :stuck_out_tongue: the performance degradation should not be noticeable :sweat_smile: 
At some point I should make a utility function to find a "maximum something from a list" because this is all around the code. Maybe now would be the right time.

rburema: Not just this code -- it's all around _any_ substantial code I've written. Each time I write a for loop (or list-comprehension, etc.) finding the 'most x' (usually min/max distance) candidate from a list I'm surprised there hasn't been a standardized solution in any language yet that I know of (though functional languages come close).

rburema: https://en.cppreference.com/w/cpp/algorithm/ranges/fold_left

PR #2319, opened 2026-04-08
Cura 13080

HellAholic:
CURA-13080


wawanbreton (approved):
Well, does not look that ugly 😄 actually, the mask application is now super localized so it makes the whole thing less future-bug-prone

PR #2320, opened 2026-04-09
[CURA-12961] Spike to add auto-brim option; DRAFT!

rburema:
Very W.I.P. part of a spike to investigate whether we can actually automatically detect brim-width / basis for discussion.

Largely copied from other (open) sources (then heavily rewritten) at the moment.

To run from the front-end, use this PR: https://github.com/Ultimaker/Cura/pull/21531


src/SkirtBrim.cpp
```diff
@@ -140,8 +142,191 @@ std::vector<SkirtBrim::Offset> SkirtBrim::generateBrimOffsetPlan(std::vector<Out
     return all_brim_offsets;
 }
 
+void computeOverSegments(const Polygon& poly, const std::function<void(const Point2F&, const Point2F&)>& func)
```
wawanbreton, line 145: Would it make sense that this function is a method of the `Polygon` class (or even `Polyline`) ?
Also maybe with a more explicit naming.

src/SkirtBrim.cpp
```diff
@@ -140,8 +142,191 @@ std::vector<SkirtBrim::Offset> SkirtBrim::generateBrimOffsetPlan(std::vector<Out
...
+{
+    for (auto it = poly.beginSegments(); it != poly.endSegments(); ++it)
+    {
+        const Point2F p1((*it).start.X, (*it).start.Y);
+        const Point2F p2((*it).end.X, (*it).end.Y);
+        func(p1, p2);
+    }
+}
+
+bool compShapeProperties(const Shape& polys, Point2F& total_centroid, Point2F& total_moment)
```
wawanbreton, line 155: Maybe move to `Shape` or `LinearAlg2D` ?

src/SkirtBrim.cpp
```diff
@@ -140,8 +142,191 @@ std::vector<SkirtBrim::Offset> SkirtBrim::generateBrimOffsetPlan(std::vector<Out
...
+
+    return area > EPSILON;
+}
+
+float getThermalLength(const std::vector<Mesh>& meshes)
+{
+    constexpr float min_length = 1250.0f;
+    return std::accumulate(
+        meshes.begin(),
+        meshes.end(),
```
wawanbreton, line 214:
```suggestion
    return ranges::accumulate(
        meshes,
```

src/SkirtBrim.cpp
```diff
@@ -140,8 +142,191 @@ std::vector<SkirtBrim::Offset> SkirtBrim::generateBrimOffsetPlan(std::vector<Out
...
+        min_length,
+        [](const float& thermal_length, const Mesh& mesh)
+        {
+            constexpr float default_length = 200.0f;
+            const float res = default_length; // TODO: make into setting! (depends on extruder of the mesh, which has the material)
+            return std::min(default_length, res);
```
wawanbreton, line 220: If I understand correctly, this function is a placeholder for when a setting is used, right ?

src/SkirtBrim.cpp
```diff
@@ -140,8 +142,191 @@ std::vector<SkirtBrim::Offset> SkirtBrim::generateBrimOffsetPlan(std::vector<Out
...
+        {
+            return std::accumulate(
+                speed_settings.begin(),
+                speed_settings.end(),
+                -1.0f,
+                [&mesh](const float& value, const std::string_view& setting_name)
+                {
+                    return std::max(value, static_cast<float>(mesh.settings_.get<coord_t>(setting_name.data())));
+                });
+        });
```
wawanbreton, line 250: Using `accumulate` in this case is quite confusing 🤔 could you use `ranges::max_element` instead, with a projection function ?

src/SkirtBrim.cpp
```diff
@@ -140,8 +142,191 @@ std::vector<SkirtBrim::Offset> SkirtBrim::generateBrimOffsetPlan(std::vector<Out
...
+    const float max_speed = getMaxSpeed(meshes);
+
+    // Get height of the mesh-group.
+    const AABB3D aabb = std::accumulate(
+        meshes.begin(),
+        meshes.end(),
+        AABB3D(),
+        [](const AABB3D& aabb, const Mesh& mesh)
+        {
+            return AABB3D(mesh.getAABB()).include(aabb);
```
wawanbreton, line 277: As far as I can see, `mesh.getAABB()` return a `AABB3D`, so why the conversion ?

src/SkirtBrim.cpp
```diff
@@ -140,8 +142,191 @@ std::vector<SkirtBrim::Offset> SkirtBrim::generateBrimOffsetPlan(std::vector<Out
...
+    Point2F centroid(0.0f, 0.0f);
+    Point2F second_moment(0.0f, 0.0f);
+    if (! compShapeProperties(shape, centroid, second_moment))
+    {
+        return 0.0f;
+    }
+    second_moment.x_ = INT2MM2(INT2MM2(second_moment.x_));
+    second_moment.y_ = INT2MM2(INT2MM2(second_moment.y_));
+
+    // Thermal lenght stuff.
```
wawanbreton, line 291: I keep doing this mistake too 😁 
```suggestion
    // Thermal length stuff.
```

src/SkirtBrim.cpp
```diff
@@ -140,8 +142,191 @@ std::vector<SkirtBrim::Offset> SkirtBrim::generateBrimOffsetPlan(std::vector<Out
...
+    const Point2F width_depth(aabb.max_.x_ - aabb.min_.x_, aabb.max_.y_ - aabb.min_.y_);
```
wawanbreton, line 292:
```suggestion
    const Point2F width_depth(aabb.spanX(), aabb.spanY();
```

src/SkirtBrim.cpp
```diff
@@ -140,8 +142,191 @@ std::vector<SkirtBrim::Offset> SkirtBrim::generateBrimOffsetPlan(std::vector<Out
...
+    const float thermal_length = INT2MM(width_depth.vSize());
+    const float thermal_length_ref = getThermalLength(meshes);
+
+    // Calculate the result.
+    constexpr float height_to_area_normalization = 1920.0f;
+    const float height_to_area
+        = std::max(max_height / second_moment.x_ * INT2MM(width_depth.y_), max_height / second_moment.y_ * INT2MM(width_depth.x_)) * max_height / height_to_area_normalization;
```
wawanbreton, line 299: Can either `second_moment.x_` or `second_moment.y_` be null anyhow ?

wawanbreton (commented):
Code is a bit drafty as advertised, but overall looks good. Also you are slightly abusing `std::accumulate` 😛

src/SkirtBrim.cpp
rburema: Yes and yes -- I wanted to have it self-contained for the spike, but we can move anything that is accepted to polygon/-line.

Any proposal for the name?

src/SkirtBrim.cpp
rburema: Yes, exactly!

src/SkirtBrim.cpp
rburema: Ah, I was trying to find that function 2 weeks ago. Yes, we should use that instead.

src/SkirtBrim.cpp
rburema: I think that might just be a leftover from when I had written something else there instead.

src/SkirtBrim.cpp
rburema: Hm, I think we should guard against that in any case. But that would be an object that'd require an infinitely large brim then 😄 

I think I see another (small) division by zero potential in the `accumulate_areas` above as well now that you've pointed out this one.

src/SkirtBrim.cpp
rburema: ... on second thought, this is quite specific; I think I might just need to change the name.

rburema: (I mean specifically this one -- the other one could still be moved.)

src/SkirtBrim.cpp
rburema: ... and that way I just noticed I missed the parameter in the 1st accumulate (and misnamed it), making it only return the max speed of the last mesh. So it's a good thing I'm rewriting it 😅

src/SkirtBrim.cpp
rburema: Turns out `accumulate` isn't in ranges (yet), though there's `fold_left` in ranges C++23 that does the same thing.

I'm reverting this back for now; because I'm not sure if we'd made the switch to 23 already (and apple clang will probably not have it.....).

src/SkirtBrim.cpp
rburema: ... except now I'm trying to rewrite it as `max_element`, and even when you use a projection, you get the max element out, then have to project it _again_ to get the actual value out, so I do think `accumulate` makes sense in this case (it helps if you see it as `fold` instead -- accumulate is a bit of a misnomer to begin with).

src/SkirtBrim.cpp
wawanbreton: I didn't mean the one in `std`, but in `ranges-v3` that we are using . You need to `#include <range/v3/numeric/accumulate.hpp>`

PR #2321, opened 2026-04-09
[CURA-13075] Handle LIGHTNING support pattern

HellAholic:
This pull request makes a targeted fix to the support infill processing logic in `FffGcodeWriter.cpp`. The change ensures that when retrieving trees for a given layer, the layer index is never negative, which prevents potential errors with invalid indices.

- Support infill processing:
  * Updated the call to `getTreesForLayer` to use `std::max(LayerIndex{ 0 }, gcode_layer.getLayerNr())`, ensuring the layer index passed is always zero or greater.

CURA-13075


wawanbreton (approved):
Makes sense 🙂

PR #2322, opened 2026-04-10
CURA-12636 Non tree support inside tree support

wawanbreton:
This PR fixes possible weak non-tree support towers being generated along with tree support, by adding a filter when dropping the "non-gracious" areas of tree support. The filter does the following:
* When the dropped part reaches an actual branch, which is quite likely, it now merges inside the branch without going down further.
* If the dropping ever goes down to reach the buildplate, we remove the remaining part completely because it will not provide any good support anyway.

The PR also provides a small refactoring of the tree support files and adds methods to allow saving the internal tree support data to OBJ for visualization.

CURA-12636

### DIFF
--- a/src/TreeSupportElement.cpp
+++ b/src/TreeSupportElement.cpp
@@ -1,0 +1,240 @@
+// Copyright (c) 2026 UltiMaker
+// CuraEngine is released under the terms of the AGPLv3 or higher
+
+#include "TreeSupportElement.h"
+
+#include "utils/OBJ.h"
+
+
+namespace cura
+{
+
+TreeSupportElement::TreeSupportElement(
+    coord_t distance_to_top,
+    size_t target_height,
+    Point2LL target_position,
+    bool to_buildplate,
+    bool to_model_gracious,
+    bool use_min_xy_dist,
+    size_t dont_move_until,
+    bool supports_roof,
+    bool can_use_safe_radius,
+    bool force_tips_to_roof,
+    bool skip_ovalisation,
+    bool influence_area_limit_active,
+    coord_t influence_area_limit_range)
+    : target_height_(target_height)
+    , target_position_(target_position)
+    , next_position_(target_position)
+    , next_height_(target_height)
+    , effective_radius_height_(distance_to_top)
+    , to_buildplate_(to_buildplate)
+    , distance_to_top_(distance_to_top)
+    , area_(nullptr)
+    , result_on_layer_(target_position)
+    , increased_to_model_radius_(0)
+    , to_model_gracious_(to_model_gracious)
+    , buildplate_radius_increases_(0)
+    , use_min_xy_dist_(use_min_xy_dist)
+    , supports_roof_(supports_roof)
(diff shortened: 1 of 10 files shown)


src/TreeSupportElement.cpp
```diff
@@ -0,0 +1,240 @@
...
+    {
+        std::vector<TreeSupportElement*> next_parents;
+        for (TreeSupportElement* grandparent : grandparents)
+        {
+            next_parents.insert(next_parents.end(), grandparent->parents_.begin(), grandparent->parents_.end());
+            grandparent->to_buildplate_ = new_value;
+            grandparent->to_model_gracious_ |= new_value; // If we set to_buildplate to true, update to_model_gracious
+        }
+        grandparents = next_parents;
+    }
```
casperlamboo, line 229: I think this block can be simplified, is just a simple queue based visit all procedure right? The "grandparent" variable name is a bit messing with my head because I'm thinking two steps up with each iteration, but I don't think that is what is happening.
```suggestion
    std::vector<TreeSupportElement*> queue{ parents_ };
    while (! queue.empty())
    {
        TreeSupportElement* node = queue.pop_back();
        node->to_buildplate_ = new_value;
        node->to_model_gracious_ |= new_value; // If we set to_buildplate to true, update 
        queue.insert(queue.end(), node->parents_.begin(), node->parents_.end());
    }
```

casperlamboo: also, you're not updating `to_build-late_` and `to_model_gracious_` for the current node, but I guess this is intentional based on the function name.

src/TreeSupportElement.cpp
```diff
@@ -0,0 +1,240 @@
...
+        for (Point2LL p : all_tips_)
+        {
+            Polygon circle;
+            for (Point2LL corner : TreeSupportBaseCircle::getBaseCircle())
+            {
+                circle.push_back(p + corner * influence_area_limit_range_ / double(TreeSupportBaseCircle::base_radius));
+            }
+            if (influence_area_limit_area_.empty())
+            {
+                influence_area_limit_area_ = circle.offset(0);
```
casperlamboo, line 204: why do we do an offset of 0? that does not do much right?

src/TreeSupportElement.cpp
```diff
@@ -0,0 +1,240 @@
...
+            }
+            else
+            {
+                influence_area_limit_area_ = influence_area_limit_area_.intersection(circle.offset(0));
+            }
+        }
+    }
+}
```
casperlamboo, line 212: I don't really get this function, from reading this code is re-initializing the `influence_area_limit_area_` to it's the intersection of all tips. Doesn't this always result in a empty earea since all tips are dis-joint?

wawanbreton:
@casperlamboo I don't have the answers to any of your questions :sweat_smile: the code in `TreeSupportElement.cpp` is just taken from the header and moved to a dedicated cpp file for sanity, but the only thing I wrote myself in this file is the `saveToObj` method...

casperlamboo (approved):

PR #2326, opened 2026-04-16
Update the workflow

HellAholic:
Make the workflow ticket aware across different repositories (including the GcodeAnalyzer)
Moving the script to execute the analyzer to its own repository (to do item) to keep all the connected elements in one place for the analyzer
Using the cura.jinja profile for the build to include the overrides
- Package override action, checks for existing tag before appending it. Required to be fixed to prevent a duplicated tag error https://github.com/Ultimaker/cura-workflows/commit/038066f4dee37c60483613d54b7614b228d78e22
- cura.jinja, move the override section to the bottom of the file so we can just append the overrides when required without needing to add the override tag https://github.com/Ultimaker/conan-config/commit/64a4bebfd76b4366c8156e68832252fd024ab704

CURA-13110
Requires https://github.com/Ultimaker/Cura/pull/21563

### DIFF
--- a/.github/workflows/gcodeanalyzer.yml
+++ b/.github/workflows/gcodeanalyzer.yml
@@ -48,7 +48,7 @@ jobs:
   find-packages:
     name: Find ticket-specific packages
     needs: [extract-ticket]
-    if: always() && needs.extract-ticket.outputs.ticket != ''
+    if: needs.extract-ticket.outputs.ticket != ''
     uses: ultimaker/cura-workflows/.github/workflows/find-package-by-ticket.yml@main
     with:
       jira_ticket_number: ${{ needs.extract-ticket.outputs.ticket }}
--- a/.github/workflows/gcodeanalyzer.yml
+++ b/.github/workflows/gcodeanalyzer.yml
@@ -40,7 +40,6 @@ jobs:
     name: Run GCodeAnalyzer on the engine
     needs: [check_actor, find-ticket-packages]
     if: >-
-      always() &&
       needs.check_actor.outputs.proceed == 'true' &&
       needs.find-ticket-packages.result == 'success'
     runs-on: ubuntu-latest
--- a/.github/workflows/gcodeanalyzer.yml
+++ b/.github/workflows/gcodeanalyzer.yml
@@ -31,36 +31,18 @@ jobs:
     uses: ultimaker/cura-workflows/.github/workflows/check-actor.yml@main
     secrets: inherit
 
-  extract-ticket:
-    name: Extract Jira ticket from branch
-    runs-on: ubuntu-latest
-    if: github.ref_name != 'main'
-    outputs:
-      ticket: ${{ steps.extract.outputs.ticket }}
-    steps:
-      - name: Extract ticket number
-        id: extract
-        run: |
-          branch="${{ github.head_ref || github.ref_name }}"
-          ticket=$(echo "$branch" | grep -oP '^(CURA|PP|NP)-\d+' || echo "")
-          echo "ticket=$ticket" >> $GITHUB_OUTPUT
-
-  find-packages:
-    name: Find ticket-specific packages
-    needs: [extract-ticket]
-    if: always() && needs.extract-ticket.outputs.ticket != ''
-    uses: ultimaker/cura-workflows/.github/workflows/find-package-by-ticket.yml@main
-    with:
-      jira_ticket_number: ${{ needs.extract-ticket.outputs.ticket }}
+  find-ticket-packages:
+    name: Extract ticket and find packages
+    uses: ultimaker/cura-workflows/.github/workflows/extract-ticket-and-find-packages.yml@main
     secrets: inherit
 
   gcodeanalyzer:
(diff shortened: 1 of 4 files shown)


.github/workflows/gcodeanalyzer.yml
```diff
@@ -31,20 +31,60 @@ jobs:
...
+    if: github.ref_name != 'main'
+    outputs:
+      ticket: ${{ steps.extract.outputs.ticket }}
+    steps:
+      - name: Extract ticket number
+        id: extract
+        run: |
+          branch="${{ github.head_ref || github.ref_name }}"
+          ticket=$(echo "$branch" | grep -oP '^(CURA|PP|NP)-\d+' || echo "")
+          echo "ticket=$ticket" >> $GITHUB_OUTPUT
```
wawanbreton, line 46: For more consistency and future-proofness, would it make sense to use the conan-recipe-version action instead ? Since this is the one that generates the conan packages channels, it should always give the proper result even if it iself changes.
This method is simple and much faster though, so this is merely a suggestion.

.github/workflows/gcodeanalyzer.yml
```diff
@@ -31,20 +31,60 @@ jobs:
...
+
+  find-packages:
+    name: Find ticket-specific packages
+    needs: [extract-ticket]
+    if: always() && needs.extract-ticket.outputs.ticket != ''
```
wawanbreton:
```suggestion
    if: needs.extract-ticket.outputs.ticket != ''
```
I could be missing something, but I would say this condition is good enough, right ? If the `extract-ticket` didn't run, the `ticket` output will be empty anyway...

.github/workflows/gcodeanalyzer.yml
```diff
@@ -31,20 +31,60 @@ jobs:
...
+        id: resolve_gca_branch
+        env:
+          GH_TOKEN: ${{ secrets.CURA_BENCHMARK_PAT }}
+        run: |
+          branch="${{ github.head_ref || github.ref_name }}"
+          if gh api "repos/Ultimaker/GCodeAnalyzer/branches/$branch" --silent 2>/dev/null; then
+            echo "ref=$branch" >> $GITHUB_OUTPUT
+          else
+            echo "ref=main" >> $GITHUB_OUTPUT
+          fi
```
wawanbreton, line 81: What is the point of this step ? :thinking:

.github/workflows/gcodeanalyzer.yml
```diff
@@ -63,8 +103,18 @@ jobs:
...
+        uses: ultimaker/cura-workflows/.github/actions/set-package-overrides@main
+        with:
+          package_overrides: ${{ needs.find-packages.outputs.package_overrides }}
+          profile: cura.jinja
+
       - name: Install dependencies and build CuraEngine
-        run: conan build . -s build_type=Release --build=missing --update -g VirtualRunEnv -o "curaengine/*:with_cura_resources=True"
+        env:
+          CONAN_PROFILE: ${{ steps.set_overrides.outputs.profile || 'cura.jinja' }}
+        run: conan build . --profile "$CONAN_PROFILE" -s build_type=Release --build=missing --update -g VirtualRunEnv -o "curaengine/*:with_cura_resources=True"
```
wawanbreton: You can pass the profile in the argument directly without going through an envi variable
```suggestion
        run: conan build . --profile ${{ steps.set_overrides.outputs.profile || 'cura.jinja' }} -s build_type=Release --build=missing --update -g VirtualRunEnv -o "curaengine/*:with_cura_resources=True"
```

wawanbreton (approved):
Looks good overall, some suggestion and questions

.github/workflows/gcodeanalyzer.yml
HellAholic: In this step is it redundant indeed. The goal was to make sure the workflow executes if you're on main for example, there if the step above is skipped since we're on main, the following steps would execute.
But here the results would be the same either way. So can be removed.

.github/workflows/gcodeanalyzer.yml
HellAholic: This gets the branch from analyzer so the output here is consumed by the checkout on the analyzer. This should (in theory), allow for tests to run on the branch of gcode analyzer that is the same as the one for the feature.

.github/workflows/gcodeanalyzer.yml
HellAholic: I agree, either we would consume the output of the action or run the action (at least that is my ultimate vision). But for now went with the faster option. So follow up ticket once things are working could be a nice addition.

.github/workflows/gcodeanalyzer.yml
wawanbreton: Oh, I see.. GCodeAnalyzer is on another repo and it doesn't have conan packages. Makes sense then.

wawanbreton (approved):
Looks good ! However the GCodeAnalyzer failed to execute, which is precisely the point of the PR 😄 I will take a look

PR #2327, opened 2026-04-16
Use shared ticket package extractor in CI

HellAholic:
Utilize the reusable workflow added in : https://github.com/Ultimaker/cura-workflows/pull/51

CURA-13113

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -42,7 +42,6 @@ jobs:
   benchmark:
     needs: [check_actor, find-ticket-packages]
     if: >-
-      always() &&
       needs.check_actor.outputs.proceed == 'true' &&
       needs.find-ticket-packages.result == 'success'
     uses: ultimaker/cura-workflows/.github/workflows/benchmark.yml@main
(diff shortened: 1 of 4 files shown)


.github/workflows/benchmark.yml
```diff
@@ -34,7 +34,17 @@ jobs:
...
 
+  find-ticket-packages:
+    name: Extract ticket and find packages
+    uses: ultimaker/cura-workflows/.github/workflows/extract-ticket-and-find-packages.yml@main
+    secrets: inherit
+
   benchmark:
+    needs: [check_actor, find-ticket-packages]
+    if: >-
+      always() &&
```
rburema: Why's there an `always()` here if there are two conditions more below? (... two more instances of this below btw)

HellAholic: the always() placement is a bit of a force of habit for me to ensure the condition is checked regardless of the results from previous step/jobs. It is redundant in all cases here since we need the output of the previous jobs anyway, so whether we execute the check or not, if there is no output from previous step, then the result would be the same.

rburema (approved):

PR #2328, opened 2026-04-18
fix: Temperature preheat command fires at correct time within long co…

HellAholic:
# Description

For long zigzag skin (initial layer) the temperature command was firing at the beginning of the path (start of skin) rather than at the end-ish of layer 0 (with consideration for the heat up time).

The fix ensures that at the start of each path only genuinely overdue inserts are fired immediately. Inserts assigned to the current path are held until the insertTempOnTime callback reaches their time_after_path_start, preserving the intended placement regardless of how many paths the feature is split into.

image

## Type of change

- [x] Bug fix

# How Has This Been Tested?

- [x] Create a build -> Slice -> inspect Gcode under similar conditions with both before and after having the zigzag initial layer -> check the temperature command is inserted at the correct moment

**Test Configuration**:
* Operating System: Windows 11

# Checklist:

CURA-13115

### DIFF
--- a/src/ExtruderPlan.cpp
+++ b/src/ExtruderPlan.cpp
@@ -30,8 +30,21 @@ void ExtruderPlan::insertCommand(NozzleTempInsert&& insert)
 
 void ExtruderPlan::handleInserts(const size_t path_idx, GCodeExport& gcode, const double cumulative_path_time)
 {
-    while (! inserts_.empty() && path_idx >= inserts_.front().path_idx && inserts_.front().time_after_path_start < cumulative_path_time)
-    { // handle the Insert to be inserted before this path_idx (and all inserts not handled yet)
+    while (! inserts_.empty())
+    {
+        const NozzleTempInsert& insert = inserts_.front();
+        if (insert.path_idx > path_idx)
+        {
+            // Insert is scheduled for a future path; nothing more to do for now.
+            break;
+        }
+        if (insert.path_idx == path_idx && insert.time_after_path_start > cumulative_path_time)
+        {
+            // Insert is scheduled for this path, but the required time within the path has not been reached yet.
+            break;
+        }
+        // Either the insert is overdue (insert.path_idx < path_idx) and must be fired immediately,
+        // or it is exactly on time (insert.path_idx == path_idx and time threshold reached).
         inserts_.front().write(gcode);
         inserts_.pop_front();
     }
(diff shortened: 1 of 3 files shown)


rburema (commented):

| Benchmark suite | Current: 3b6dad27e2243c7940b451bdbeda08c7f20fa2f6 | Previous: e4130606fe1a39aee6da6278ce0d04340dc4b789 | Ratio |
|-|-|-|-|
| `SimplifyTestFixture/simplify_local` | `1.9447085998230405` ns/iter | `1.246300090173324` ns/iter | `1.56` |

src/ExtruderPlan.cpp
```diff
@@ -30,8 +30,21 @@ void ExtruderPlan::insertCommand(NozzleTempInsert&& insert)
...
+        if (insert.path_idx > path_idx)
+        {
+            // Insert is scheduled for a future path; nothing more to do for now.
+            break;
+        }
+        if (insert.path_idx == path_idx && insert.time_after_path_start > cumulative_path_time)
+        {
+            // Insert is scheduled for this path, but the required time within the path has not been reached yet.
+            break;
+        }
```
rburema: I don't actually have much of a problem with this, since the verbosity also serves documenting the code, but you could also try the 'self documenting' thing here:

```suggestion
        const bool in_future_path = insert.path_idx > path_idx;
        const bool in_current_path_at_future_time = insert.path_idx == path_idx && insert.time_after_path_start > cumulative_path_time;
        if (in_future_path || in_current_path_at_future_time)
        {
            break;
        }
```

(Not that it matters here, but if you didn't know, for future reference: The `||` operator itself will short-circuit, so it won't evaluate the 2nd part here if the 1st part is already true. So we technically could make this a tiny bit more efficient, like in the original code, at the cost of having to do actual comments again.)

HellAholic: It was partly the documentation through comments and also a bit of a habit when it comes to conditions, so calculating conditions in order of occurrence separately. Which translates to if it's scheduled for future, we don't need to check the within current path, but the difference is so miniscule in reality that I don't think it matters that much 😝

rburema: I agree, just a style I prefer when writing code personally 😅 -- if we're talking about inconsequential differences; technically my way is also very slightly slower, since it calculates both booleans first (though maybe the compiler will pick up on that with the 'as-if' rule).

rburema (approved):

PR #2330, opened 2026-04-21
CURA-13070 Allow disabling prime tower wheel pattern

HellAholic:
Disable wheel pattern for prime tower when there is actually not enough space i.r.t the maximum bridging distance to allow a proper pattern. When this is the case, we just fallback to using full circles.

CURA-13070
Requires https://github.com/Ultimaker/Cura/pull/21561

### DIFF
--- a/src/PrimeTower/PrimeTower.cpp
+++ b/src/PrimeTower/PrimeTower.cpp
@@ -138,27 +138,29 @@ std::tuple<ClosedLinesSet, coord_t> PrimeTower::generatePrimeToolpaths(const siz
 ClosedLinesSet PrimeTower::generateSupportToolpaths(const size_t extruder_nr, const coord_t outer_radius, const coord_t inner_radius)
 {
     const Scene& scene = Application::getInstance().current_slice_->scene;
-    const double max_bridging_distance = static_cast<double>(scene.extruders[extruder_nr].settings_.get<coord_t>("prime_tower_max_bridging_distance"));
-    const coord_t line_width = scene.extruders[extruder_nr].settings_.get<coord_t>("prime_tower_line_width");
+    const Settings& extruder_settings = scene.extruders[extruder_nr].settings_;
+    const coord_t line_width = extruder_settings.get<coord_t>("prime_tower_line_width");
+    const double max_bridging_distance = static_cast<double>(std::max(extruder_settings.get<coord_t>("prime_tower_max_bridging_distance"), line_width));
     const coord_t radius_delta = outer_radius - inner_radius;
     const coord_t semi_line_width = line_width / 2;
 
     ClosedLinesSet toolpaths;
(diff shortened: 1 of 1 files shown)


src/PrimeTower/PrimeTower.cpp
```diff
@@ -144,13 +144,21 @@ ClosedLinesSet PrimeTower::generateSupportToolpaths(const size_t extruder_nr, co
...
 
     ClosedLinesSet toolpaths;
+    // Split annuli according to max bridging distance with constraints to ensure we can fit the wheel
+    const coord_t nb_annuli = max_bridging_distance > 0.0 ? static_cast<coord_t>(std::ceil(static_cast<double>(radius_delta) / max_bridging_distance)) : 0;
+    const coord_t actual_radius_step = nb_annuli > 0 ? radius_delta / nb_annuli : 0;
 
-    // Split annuli according to max bridging distance
-    const coord_t nb_annuli = static_cast<coord_t>(std::ceil(static_cast<double>(radius_delta) / max_bridging_distance));
-    if (nb_annuli > 0)
+    if (nb_annuli == 0 || actual_radius_step <= line_width)
```
casperlamboo:
```suggestion
    if (actual_radius_step <= line_width)
```

$\text{nb annuli} = 0$ is redundent since `const coord_t actual_radius_step = nb_annuli > 0 ? radius_delta / nb_annuli : 0;` will enforce $\text{actual radius step = 0 <= line width}$ in the case when $\text{nb annuli} = 0$

src/PrimeTower/PrimeTower.cpp
```diff
@@ -144,13 +144,21 @@ ClosedLinesSet PrimeTower::generateSupportToolpaths(const size_t extruder_nr, co
...
 
     ClosedLinesSet toolpaths;
+    // Split annuli according to max bridging distance with constraints to ensure we can fit the wheel
+    const coord_t nb_annuli = max_bridging_distance > 0.0 ? static_cast<coord_t>(std::ceil(static_cast<double>(radius_delta) / max_bridging_distance)) : 0;
+    const coord_t actual_radius_step = nb_annuli > 0 ? radius_delta / nb_annuli : 0;
 
-    // Split annuli according to max bridging distance
-    const coord_t nb_annuli = static_cast<coord_t>(std::ceil(static_cast<double>(radius_delta) / max_bridging_distance));
-    if (nb_annuli > 0)
+    if (actual_radius_step <= line_width)
```
HellAholic: We can add a filter here that if the actual_radius_step is smaller than 2*line_width or 3*line_width we fall back to the circle.

wawanbreton: I think this is done by checking the `nb_annuli` already, I will do some tests and provide a bulletproof version.

casperlamboo (approved):

PR #2332, opened 2026-05-12
[CURA-13024] Merge surface (floor, roof) to skin when relevant settings are equal.

rburema:
It'll be a bit of a hassle to (remember to) maintain this, but unfortunately we can't use the MeshPathConfigs, since they're not only generated at a later juncture, but are also not 100% overlapping with what we want to do here (some settings aren't used there that we do need, so we'd need something bespoke here anyway...).


rburema:
Linked the wrong Jira ticket -- will remake the PR with the correct info as soon as I close this one. (Link ~forthcoming~ :  #2333 .)

PR #2333, opened 2026-05-12
[CURA-13023] Merge surface (floor, roof) to skin when relevant settings are equal.

rburema:
It'll be a bit of a hassle to (remember to) maintain this, but unfortunately we can't use the MeshPathConfigs, since they're not only generated at a later juncture, but are also not 100% overlapping with what we want to do here (some settings aren't used there that we do need, so we'd need something bespoke here anyway...).


src/skin.cpp
```diff
@@ -339,6 +339,99 @@ void SkinInfillAreaComputation::generateInfill(SliceLayerPart& part)
...
+    const Settings& settings,
+    const std::unordered_map<std::string, std::string>& setting_float_names,
+    const std::unordered_map<std::string, std::string>& setting_other_names,
+    const std::vector<std::string> settings_zero_default)
+{
+    const bool equal_floats = std::ranges::all_of(
+        setting_float_names,
+        [&settings](const auto& kvp)
+        {
+            return settings.get<double>(kvp.first) == settings.get<double>(kvp.second);
```
casperlamboo: Don't want a small $\epsilon$ size diff for floating point differences? Some of these values are resolved from formula's and can have ever so slight differences.

src/skin.cpp
```diff
@@ -339,6 +339,99 @@ void SkinInfillAreaComputation::generateInfill(SliceLayerPart& part)
...
+    if (! all_zeros)
+    {
+        return false;
+    }
+
+    constexpr auto extruder_nr = "extruder_nr";
+    constexpr auto top_bottom_pattern = "top_bottom_pattern";
+    constexpr auto skin_monotonic = "skin_monotonic";
+    constexpr auto skin_angles = "skin_angles";
+    if ((settings.get<int>(setting_other_names.at(extruder_nr)) > 0 && settings.get<int>(setting_other_names.at(extruder_nr)) != settings.get<int>(extruder_nr))
```
casperlamboo: I don't get the $> 0$ part here. I can imagine that the check after is not required if the extruder is $-1$, but you still van to check extruder equality even though the extruder is $0$?

src/skin.cpp
```diff
@@ -339,6 +339,99 @@ void SkinInfillAreaComputation::generateInfill(SliceLayerPart& part)
...
+        || settings.get<EFillMethod>(setting_other_names.at(top_bottom_pattern)) != settings.get<EFillMethod>(top_bottom_pattern)
+        || settings.get<bool>(setting_other_names.at(skin_monotonic)) != settings.get<bool>(skin_monotonic)
+        || ! std::ranges::equal(settings.get<std::vector<double>>(setting_other_names.at(skin_angles)), settings.get<std::vector<double>>(skin_angles)))
```
casperlamboo: Same as comment above, don't we want a small $\epsilon$ range here?

src/skin.cpp
rburema: Oops, that should've been `> -1` or `>= 0` instead. Good catch.

src/skin.cpp
rburema: Yeah, fair enough! Do you have a proposal for the size of the epsilon?

src/skin.cpp
rburema: (Same as above!)

src/skin.cpp
casperlamboo: `std::numeric_limits<double>::epsilon()`?

PR #2334, opened 2026-05-29
CURA-12446 top bottom wall count v3

wawanbreton:
This PR implements the changes to allow more flexible wall counts on different areas: infill, skin, roofing, flooring, first layer.
All the wall counts can now have different values, so the outer contours will take the lowest value of them, and additional walls will be added when necessary.
A side-effect is that depending on the number of walls, some inner walls will end up being printed as actuall inner walls, or as skin/infill extra walls, thus being printed with different settings. This is quite necessary to maintain the consistency when having multiple areas with different wall counts on the same layer, however in some cases the result may look unexpected:

image
image

CURA-12446
Requires https://github.com/Ultimaker/Cura/pull/21634
Supersedes https://github.com/Ultimaker/CuraEngine/pull/2227


rburema (approved):
LGTM!

It bothers me a bit that I can't find anything in the engine part of a ticket we've struggled this long with, but the actual implementation of this attempt is straightforward it seems, at least from a 'hindsight' perspective.

wawanbreton:
> LGTM!
> 
> It bothers me a bit that I can't find anything in the engine part of a ticket we've struggled this long with, but the actual implementation of this attempt is straightforward it seems, at least from a 'hindsight' perspective.

Well, I did start with a much more complex solution, which really turned out into a rabbit hole... Then realised I could use this solution, which is indeed super straightforward, but not really sure it is good enough. I'm not a big fan of the walls not being printed as actual walls depending on the other wall counts. In practice this may be good enough though :shrug:

PR #2336, opened 2026-06-03
Bump curaformulaengine package

casperlamboo:
Main objective of this PR is to be able to parse the following values in gcode start-end templates
```py
min(extruderValues('adhesion_type'), key=('raft', 'brim', 'skirt', 'none').index)
```
Added support for the following syntax
- **property index** can now write `property.index`
- **key word arguments** can now write `min(..., key=...)`

NP-1333


wawanbreton (commented):
The version has not been bumped (yet ?) 🤔

HellAholic (approved):

PR #2337, opened 2026-06-03
[CURA-12580, CURA-13056] Support painting

rburema:
In addition to other kinds of painting (seam and material/extruder) introduce support painting, which is a large part of the reason the painting feature was started to begin with.

With this, 'support_enable' becomes more like an 'auto-support', that just tries to support everything w.r.t a given support angle, versus support painting, in which a UV-texture with the relevant allow/deny colours determines support. (Combining the two is possible.)

See also:
- https://github.com/Ultimaker/Cura/pull/21639 -- frontend PR
- https://github.com/Ultimaker/Uranium/pull/1041 -- small Uranium PR


src/FffGcodeWriter.cpp
```diff
@@ -405,7 +406,7 @@ size_t FffGcodeWriter::getStartExtruder(const SliceDataStorage& storage) const
     }
     else // No adhesion.
     {
-        if (mesh_group_settings.get<bool>("support_enable") && mesh_group_settings.get<bool>("support_brim_enable"))
+        if ((mesh_group_settings.get<bool>("support_enable") || mesh_group->has_support_paint) && mesh_group_settings.get<bool>("support_brim_enable"))
```
wawanbreton: Since the same test is systematically done on the `mesh_group` at different places, would it make sense to create a dedicated method in `MeshGroup`, that tests both the settings and the variable ?

include/MeshGroup.h
```diff
@@ -35,6 +35,7 @@ class MeshGroup
 
     std::vector<Mesh> meshes;
     Settings settings;
+    bool has_support_paint = false;
```
wawanbreton: It would somehow make more sense to me to call this variable `has_painted_support`, does it seems relevant ?

src/MeshMaterialSplitter.cpp
```diff
@@ -832,25 +833,18 @@ void makePaintingModifierMeshes(MeshGroup* meshgroup)
...
         delta_iterations += mesh_data.estimated_material_iterations;
         spdlog::info("Painting mesh-modifier generation for {} took {} seconds", mesh.mesh_name_, timer.elapsed().count());
     }
 
     // Add meshes to group afterwards to avoid re-allocating the meshes in the vector
-    for (Mesh& modifier_mesh : modifier_meshes)
-    {
-        meshgroup->meshes.push_back(std::move(modifier_mesh));
-    }
+    ranges::move(modifier_meshes, std::back_inserter(meshgroup->meshes));
```
wawanbreton, line 847: Nice, didn't know this one 😃

wawanbreton (approved):
Some tiny remarks, but looks good !

src/MeshMaterialSplitter.cpp
rburema: I didn't either, but I just suspected it must exist because my _isn't there something we can do to make this less clunky_ sense went off (since moving things to another vector is probably a pretty standard operation) and went 'probably ranges has something for this' 😅

include/MeshGroup.h
rburema: I think your proposed name is slightly better, so I'll rename it.

src/FffGcodeWriter.cpp
rburema: I'm not sure; `support_enable` can be set per-mesh.

If you look in the code, it doesn't always go for the mesh-group, sometimes it checks the mesh-settings instead, so I think I'd prefer to keep it this way.

wawanbreton: Oh yes indeed, I missed the subtlety

PR #2340, opened 2026-06-23
harden binary parser against integer underflow and unsafe type punning

SABITHSAHEB:
This patch improves the safety of the binary STL loader when handling malformed or truncated files.

### Changes

* Rejects STL files that are too small to contain a valid header.
* Handles `ftell()` failures and adds a missing `fopen()` null check.
* Warns when the triangle count stored in the header does not match the file size.
* Replaces raw float pointer casting with a safer `memcpy`-based implementation.
* Replaces temporary C-style arrays with `std::array`.

These changes do not affect valid STL files and preserve existing parsing behavior.

## Type of Change

* [x] Bug fix (non-breaking change which fixes an issue)

## Testing

Added tests for:

* Truncated STL files.
* Header triangle count mismatches.
* Normal STL loading (`testModel.stl`, 564 triangles).

### DIFF
--- a/src/MeshGroup.cpp
+++ b/src/MeshGroup.cpp
@@ -3,6 +3,8 @@
 
 #include "MeshGroup.h"
 
+#include <array>
+#include <cstring>
 #include <filesystem>
 #include <fstream>
 #include <limits>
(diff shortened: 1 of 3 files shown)


SABITHSAHEB:
any update?

PR #2341, opened 2026-06-23
Fix undefined behavior in texture bit-field parsing from untrusted input

SABITHSAHEB:
# Description

While reviewing the texture painting code, I noticed that bit ranges for paint features are read from PNG `Description` metadata and used directly for bit extraction.

Since these values originate from untrusted PNG files or Arcus messages, invalid ranges such as `end >= 32` or `start > end` can result in out-of-range shift operations in `TextureDataProvider::getValue`, leading to undefined behavior.

## Changes

* Added `TextureBitField::isValid()` to enforce:

  * `start <= end`
  * `end < 32`
* Validate bit ranges while parsing JSON metadata in `loadTextureFromPngData`.
* Added a defensive validation check in `TextureDataProvider::getValue`.
* Verify JSON values are unsigned integers before calling `GetUint()`.
* Invalid metadata is now logged and skipped instead of being processed.

## Testing

Added `TextureDataProviderTest.cpp` covering:

* Valid bit ranges continue to extract the expected values.
* Invalid ranges such as:

  * `[0,32]`
  * `[0,64]`
  * `[10,5]`
    are rejected safely.
* Confirmed that valid textures behave exactly as before and only malformed metadata is ignored.

### DIFF
--- a/include/TextureDataMapping.h
+++ b/include/TextureDataMapping.h
@@ -18,6 +18,14 @@ struct TextureBitField
 {
     size_t bit_range_start_index{ 0 }; // The index of the first bit of the field
     size_t bit_range_end_index{ 0 }; // The index of the last bit of the field
+
+    //! A pixel value is decoded as a 32-bit field, so a well-formed range is ordered and within [0, 31]. The ranges come from untrusted texture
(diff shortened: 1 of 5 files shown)


SABITHSAHEB:
Any Updates on this pr?

PR #2344, opened 2026-06-30
[CURA-13125] Respect cool_fan_enabled for fan overrides

HellAholic:
Decision related to larger setting structure:
- Multiple settings that would set fan speed → all of these should not be set when print cooling is disabled
- If you want to disable cooling for certain features → set fan speed to 0%
- If you want fan speed unchanged from current fan speed → use setting formula

Comes with:
https://github.com/Ultimaker/Cura/pull/21689

CURA-13125

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -3221,7 +3221,8 @@ void FffGcodeWriter::processTopBottom(
 
     double fan_speed = GCodePathConfig::FAN_SPEED_DEFAULT;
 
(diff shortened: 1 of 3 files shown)


rburema (approved):
I do think that this'll get the desired behaviour. It does raise the question if we maybe want to check the following as well;

"cool_fan_..._0"
"raft_base_fan_..."
"raft_interface_fan_..."
"support_fan_..."
"support_supported_skin_fan_..."

(In any case, LGTM for the current purpose...)

HellAholic:
Follow up work split to: CURA-13258

PR #2345, opened 2026-07-02
CURA-13226 Engine adjustments for benchmark tooling

wawanbreton:
This PR provides two minor improvements that were necessary for the new benchmarking tools:
* Improve the OBJ file loading to fully support files exported with assimp
* Export the GCode header irregardless of the communication type. Since we now resolve the whole GCode at the end of the slicing, the header can be treated the same and always be fully generated. Thus the `Communication` objects don't need the `isSequential` method anymore, and they will all get the very same GCode header. The header is also now sent as a regular GCode part, so we don't need the specific `GCodePrefix` message.

CURA-13226

### DIFF
--- a/src/communication/CommandLine.cpp
+++ b/src/communication/CommandLine.cpp
@@ -279,9 +279,9 @@ void CommandLine::sliceNext()
 
                     const auto transformation = last_settings->get<Matrix4x3D>("mesh_rotation_matrix"); // The transformation applied to the model when loaded.
 
-                    if (! loadMeshIntoMeshGroup(&slice->scene.mesh_groups[mesh_group_index], argument.c_str(), transformation, last_extruder->settings_))
+                    if (! loadMeshIntoMeshGroup(&slice->scene.mesh_groups[mesh_group_index], argument, transformation, last_extruder->settings_))
                     {
-                        spdlog::error("Failed to load model: {}. (error number {})", argument, errno);
+                        spdlog::error("Failed to load model: {} (error number {})", argument, errno);
                         exit(1);
                     }
                     else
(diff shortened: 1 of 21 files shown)


wawanbreton (commented):

| Benchmark suite | Current: 0b3f82215d5cf8cec8487e5b4487150578d59c97 | Previous: 4d5fb16ae7ac0d49ec0a6475d37db551adca739d | Ratio |
|-|-|-|-|
| `SimplifyTestFixture/simplify_local` | `8.381421742923362` ns/iter | `1.2540880137436994` ns/iter | `6.68` |

src/communication/CommandLine.cpp
```diff
@@ -381,7 +376,6 @@ void CommandLine::sliceNext()
 
                     constexpr std::string_view global_identifier = "global";
                     constexpr std::string_view extruder_identifier = "extruder.";
-                    constexpr std::string_view model_identifier = "model.";
```
casperlamboo, line 384: rather then removing this identifier I think we should add a 
```cpp
else if (key.starts_with(model_identifier)) 
```
to the last condition in
https://github.com/Ultimaker/CuraEngine/blob/7a627bc0c090657d2a849f7a7c700a6aa8b11eee/src/communication/CommandLine.cpp#L396-L411

wawanbreton: Yes indeed, I just saw an unused variable and removed it, but it does make sense to actually have it used

wawanbreton: Oh, no actually. The model name doesn't start with `model.`, so it is just "anything that is not specifically global of extruder"

casperlamboo (approved):

PR #2346, opened 2026-07-14
CURA-13250 chamfered tree base

wawanbreton:
This PR's main goal is to implement the chamfered base for the support, both tree and normal. The base makes the bottom of the support much sturdier and also improves adhesion to the buildplate, so it is now possible to use higher supports, or tree supports with less trunks.

In addition to that, it contains a few other significant changes:
* The support walls are no more generated using Arachne, but a naive insetting operation instead. Since we don't really care about dimentional accuracy nor fine details in the support, using a naive algorithm helps reducing the slicing time a lot when using especially tree support that generate a lot of tiny tips.
* Fix in the skirt generation: so far, the skirt generation was using an approximation of the convex hull of the models on the buildplate. In some situations this process would generate invalid self-intersecting polygons, so it is now replaced by the actual convex hull calculation, which gives a correct result and also happens to be way faster. This is a very visible behavior change though, because the skirt is no more rounded as it used to be, however it has more straight lines which help for adhesion while priming, and also it is now always composed of a single line enclosing the whole print, no more multiple pieces attracted to each other.
* introduction of CuraEngineViz: we now have a dedicated tool to visualize the internal data of the engine in real-time. The engine contains the interface part that sends the raw data to the visualizer, which is of course disabled by default and can be enabled with a `-o "with_curaviz=True"` at conan install time

Requires https://github.com/Ultimaker/Cura/pull/21714


wawanbreton (commented):

| Benchmark suite | Current: 10f1b81d55f0c5615627d9a80cb74cdc1bf23887 | Previous: 93f5d54c32156f2f5e8144401639e002a5d88a82 | Ratio |
|-|-|-|-|
| `InfillTest/Infill_generate_connect/0/400` | `5.766691518246707` ms/iter | `3.646253201438742` ms/iter | `1.58` |
| `HolesWallTestFixture/InsetOrderOptimizer_getInsetOrder/3` | `0.00026352653403075943` ms/iter | `0.00016906826041988748` ms/iter | `1.56` |

include/settings/types/LayerIndex.h
```diff
@@ -203,6 +203,11 @@ constexpr auto format_as(LayerIndex index)
     return index.value;
 }
 
+constexpr LayerIndex distance(const LayerIndex& lhs, const LayerIndex& rhs)
```
casperlamboo, line 206: This function feels a bit weird to me since
- distance in this case can be negative
- distance returns an `Index` type
A better name for this function would be `layerDelta`, `layersBetween` or `layersTo` in my opinion.

include/WallToolPathGenerator.h
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) 2023 UltiMaker
```
casperlamboo:
```suggestion
// Copyright (c) 2026 UltiMaker
```

include/utils/polygonUtils.h
```diff
@@ -726,6 +734,24 @@ class PolygonUtils
...
+    static InsetOutset generateOutset(const Shape& shape, const coord_t width, const coord_t line_width);
+
+    /*!
+     * Generates a series of insetting walls inside a given shape
+     * @param shape The shape to be filled
+     * @param width The maximum width to be reached, actual result may be thinner but never larger
+     * @param line_width The line width to be used to print each inset line
+     * @return The generated inset lines and the actual width of the innermost contour
+     */
+    static InsetOutset generateInset(const Shape& shape, const coord_t width, const coord_t line_width);
```
casperlamboo: Can't this be achieved using a single function where `width` can be either negative or positive, same for `generateInsetOutset`? It would also be more inline with the offset function where the `offset parameter can be positive or negative.

include/sliceDataStorage.h
```diff
@@ -420,14 +422,16 @@ class SliceDataStorage : public NoCopy
      * \param include_models Whether to include the models in the outline
      * \param external_polys_only Whether to disregard all hole polygons.
      * \param extruder_nr (optional) only give back outlines for this extruder (where the walls are printed with this extruder)
+     * \param include_support_base (optional) include the support base (implies include_support). If false, only the raw support outline is returned.
```
casperlamboo:
```suggestion
     * \param include_support_base (optional) include the support base (requires include_support). If false, only the raw support outline is returned.
```
Think you mean requires here. To me "implies" means that when this flag is set, it will auto set `include_support`

src/SkirtBrim.cpp
```diff
@@ -146,16 +148,20 @@ void SkirtBrim::generate()
...
     // with the other area calculation above, since they are either itself a simple/convex shape or relevant for brim.
     Shape covered_area = storage_.getLayerOutlines(
         0,
         /*include_support*/ true,
-        /*include_prime_tower*/ adhesion_type_ == EPlatformAdhesion::SKIRT);
+        /*include_prime_tower*/ adhesion_type_ == EPlatformAdhesion::SKIRT,
+        /*external_polys_only*/ false,
+        /*extruder_nr*/ -1,
+        /*include_models*/ true,
+        /*include_support_base*/ false);
```
casperlamboo: I think we should be using named variables for consistency.
```suggestion
    constexpr LayerIndex layer_nr = 0;
    constexpr bool include_support = true;
    const bool include_prime_tower = adhesion_type_ == EPlatformAdhesion::SKIRT;
    constexpr coord_t extruder_nr = -1;
    constexpr bool include_models = true;
    constexpr bool include_support_base = false;
    Shape covered_area = storage_.getLayerOutlines(
        layer_nr,
        include_support,
        include_prime_tower,
        external_polys_only,
        extruder_nr,
        include_models,
        include_support_base
    );
```

src/SkirtBrim.cpp
```diff
@@ -667,30 +674,70 @@ std::vector<Shape> SkirtBrim::generateAllowedAreas(const std::vector<Outline>& s
...
+            ranges::remove_if(
+                base_inset,
+                [support_brim_minimum_hole_area](const Polygon& polygon)
+                {
+                    const double area = polygon.area();
+                    return area < 0 && area > -support_brim_minimum_hole_area;
+                }),
+            base_inset.end());
+
+        if (base_inset.length() > 0)
```
casperlamboo, line 695:
```suggestion
        if (! base_inset.empty())
```

casperlamboo (commented):
I don;t fully understand the need of changing `x.getInfillArea()` to `x.outline_`

include/settings/types/LayerIndex.h
wawanbreton: Well, it has to be called `distance` to match the standard `std::distance` so that we can use LayerIndex directly in a `parallel_for` loop.
Maybe a comment would be helpful ?

src/SkirtBrim.cpp
wawanbreton: In this case, `length()` gives the contained lines total length, so it does check that the list is not empty, and also that it contains non-empty polylines. Probably a bit redundant, but slightly safer. I think I actually moved this check from somewhere else.

include/utils/polygonUtils.h
wawanbreton: That is actually what is done under the hood. There is a private `generateInsetOutset` method that is called by both. As an external API I find it nicer to have two explicit functions rather than giving an extra argument.

include/sliceDataStorage.h
wawanbreton: Yes indeed, `requires` is more correct here

wawanbreton:
> I don;t fully understand the need of changing `x.getInfillArea()` to `x.outline_`

`SupportInfillPart` was a class with only public variables. Now it is a struct, still with the same elements. I actually don't see a need for the `getInfillArea` method, since it just returns the `outline_` which is publicly accessible anyway. It was confusing that you could get the same element by 2 different means IMHO.

include/utils/polygonUtils.h
casperlamboo: That private function is using an additional argument `direction` which can be positive and negative, I'm suggestion to combine that into the width variable, if this variable is negative it's an inwards inset, if it's positive it's an outwards outset.

Main reason why I believe this to be better is that it's functionally the same as clipper offsets. Cura developers are expected to know this library very well so having function structures that follow same principles is beneficial. For clipper we also don't have separate offset functions for going outwards vs going inwards.

wawanbreton: Well, I could argue that it makes sense for an offset function, because an `offset` value can be negative or positive, whereas a `width` should only be positive. Also every time we use one of those functions, it is in a specific context where we explicitely want either an outer or an inner contour, so I think it helps in conveying the meaning.

But I can do it if you really think it adds some value :stuck_out_tongue:

casperlamboo: https://github.com/Ultimaker/CuraEngine/pull/2346/commits/a3f1565d18eef7a80da5d3f06ecaa3a845df8933

HellAholic (approved):
LGTM!

PR #2347, opened 2026-07-15
[CURA-13251] variable wall thickness for tree support

rburema:
The general idea is that you don't want a lot of walls near the tips (and bottom, since we're going to have chamfered bases there, and the size of the branches there is wider anyway), but you do want the extra sturdiness with thicker/more walls more down (but again, not all the way down) in the tree -- without losing too much speed.

To that end, have a nominal wall-thickness and an 'enlarged' wall-thickness, the number of layers from either the bottom, or the top, and a number of walls to smooth out transitions.

(For these settings, see the front-end PR: https://github.com/Ultimaker/Cura/pull/21717)

NOTE:
Since the trees' branches can overlap in the sense that the circles/ovals that are drawn can (and will frequently) get merged to larger shapes, and (because we want to keep the use of arachne to a minimum for helper structures where the shape doesn't need to be perfect) we want to keep the number of walls the same for each entire disconnected shape (which possibly contains a number of branches), there's at least one complication, for which I've made a compromise.

Most notably, the smoothing as is implemented is in essence quite simple, and can still induce more hard transitions when shapes of branches are finally split to distinct shapes. These jumps will usually not be quite as harsh, since the previous (lower) shape was the weighted average of the now distinct shapes. Furthermore, since branching happens upward, and while branches do merge, this happens more rarely, the branch more close to the tip (and therefore likely smaller), will have had less influence on the larger shape, and therefore make the larger 'jump' in wall-width; _which is mostly OK actually_ , since a smaller wall (width) can go fine on a larger one. The jump from a smaller to a larger wall-width won't be quite as pronounced.

A typical case:
gradient_and_jumps

On the bottom, there's a gradient, until the shapes become distinct. On the left-side, the jump isn't quite as noticeable as on the right side. The jump on the right side isn't too bad in practice, since it jumps from larger to smaller walls instead of the other way around.

There are ways to fix this, but it would require either doing some quite messy things (like making a new 'tree' out of where the shapes actually overlap ... with all kinds of edge-cases like sometimes two distinct shapes can merge back) -- or there will have to be a refactor that is out of scope of this ticket.


include/TreeSupport.h
```diff
@@ -1,9 +1,11 @@
-// Copyright (c) 2021 Ultimaker B.V.
+// Copyright (c) 2026 UltiMaker
 // CuraEngine is released under the terms of the AGPLv3 or higher.
 
 #ifndef TREESUPPORT_H
 #define TREESUPPORT_H
 
+#include <utility> // For std::pair
```
wawanbreton: I would suggest not adding the "reason" for including the header. I understand the purpose, and it does make sense, but it can very easily get outdated, and thus be counter-productive because you would think it is actually useful when it is no more.
Moreover, modern IDEs do a pretty good job of showing useless header inclusions.
Please tell me what you think, maybe there is a bigger story.

include/TreeSupportSettings.h
```diff
@@ -315,7 +319,12 @@ struct TreeSupportSettings
     /*!
      * \brief Amount of walls the support area will have.
      */
-    int support_wall_count;
+    coord_t support_wall_thickness;
+
+    /*!
+     * \brief ...
```
wawanbreton: This description is not very descriptive 😛

src/TreeSupport.cpp
```diff
@@ -2285,9 +2287,27 @@ void TreeSupport::finalizeInterfaceAndSupportAreas(
...
+                const coord_t search_radius = EPSILON + std::max(aabb.width(), aabb.height()) / 2;
+                const auto located_data = center_locator->getNearby(aabb.getMiddle(), search_radius);
+                const std::pair<float, float> total_area_width = std::accumulate(
+                    located_data.begin(),
+                    located_data.end(),
+                    std::make_pair(0.f, 0.f),
+                    [&support_part](const auto& res, const auto& elem)
+                    {
+                        return support_part.inside(elem.point) ? std::make_pair((elem.val.first * elem.val.second) + res.first, elem.val.second + res.second) : res;
+                    });
```
wawanbreton, line 2307: Would it be possible to do the wall thickness calculation earlier in the process ? This way you could use the TreeSupportElement structure, which contains the parent/children links between each unitary circle. That would save this (quite heavy) calculation completely, if I understand correctly.

wawanbreton, line 2307: Could you add some small comments to make the meaning of this piece of code more explicit ?

src/TreeSupport.cpp
```diff
@@ -2391,6 +2410,30 @@ void TreeSupport::drawAreas(std::vector<std::set<TreeSupportElement*>>& move_bou
...
+
+            for (const auto& elem : move_bounds[layer_idx])
+            {
+                const float lerp_mul = static_cast<float>(std::min(
+                                           config.support_line_width_layer_smooth,
+                                           std::max(
+                                               config.support_line_width_bottom_layers - static_cast<coord_t>(layer_idx),
+                                               config.support_line_width_top_layers - static_cast<coord_t>(elem->distance_to_top_))))
+                                     / config.support_line_width_layer_smooth;
+                const coord_t nominal_wall_width = std::lerp(config.support_enlarged_wall_thickness, config.support_wall_thickness, std::max(0.f, std::min(1.f, lerp_mul)));
```
wawanbreton:
```suggestion
                const coord_t nominal_wall_width = std::lerp(config.support_enlarged_wall_thickness, config.support_wall_thickness, std::clamp(lerp_mul, 0.f, 1.f));
```

wawanbreton (commented):
Some remarks and questions, maybe I also need some explanations

include/TreeSupport.h
rburema: That's fair and a good point about modern IDE's. I think it's a good idea indeed then to stop doing this -- I'll try to remember, please call out if I forget again ;-)

src/TreeSupport.cpp
rburema: I'd love to, but instead of a single 'circle' or 'oval', for the resulting shapes, in practice, you get a blob representing different overlapping part of the tree that, from a data-structure perspective, don't even have to be 'adjacent' (but are in real 3D).

Take for instance two branches seeming to merge 'back' into each other higher up; they don't actually merge when it concerns the internal representation of the tree, but the shapes overlap (and thus are one 'unioned' shape).

Since we still want the shape to have _one singular_ outline width, so the (weighted) average has to be taken somehow ... and because it can't come from the actual tree, the shapes have to be leading!

(I actually do calculate much of it earlier, just putting it into the locator; the 'nominal' wall width you see (later in the code, but earlier in the process) is just doing that, just instead.)

That said, I don't think the current solution is very elegant (for example, the way I relate the shapes to the center-points feels somewhat brittle, and as you say, it's not a cheap calculation), so please, if you have an idea for improvement, tell me.

src/TreeSupport.cpp
rburema: Ah I keep forgetting C++ has clamp now too, thank goodness.

src/TreeSupport.cpp
wawanbreton: Well, I don't see the branches much overlapping higher after a previous split. Not saying that it cannot happen, but I think it is rare enough to be neglected (could definitely be proven wrong)

I think if you do neglect this aspect, then you can indeed base the whole calculation on the pure structural shape. But maybe I am missing something. Let's have a talk with drawing after the holiday, when this thing makes no more sense to us :smile:

include/TreeSupportSettings.h
rburema: But it is brief! :-p ... in any case, done!

src/TreeSupport.cpp
rburema: Yes, let's get some paper when we're both back. Happy holidays!

rburema: Done!

src/TreeSupport.cpp
rburema: I found another instance where I used min/max like this, see the latest commit.

src/TreeSupport.cpp
```diff
@@ -2302,10 +2305,10 @@ void TreeSupport::finalizeInterfaceAndSupportAreas(
                     {
                         return support_part.inside(elem.point) ? std::make_pair((elem.val.first * elem.val.second) + res.first, elem.val.second + res.second) : res;
                     });
+                const auto weighted_average = static_cast<coord_t>(total_area_width.first / std::max(1.f, total_area_width.second));
 
-                const auto wall_thickness = std::max(
-                    config.support_wall_thickness,
-                    std::min(config.support_enlarged_wall_thickness, static_cast<coord_t>(total_area_width.first / std::max(1.f, total_area_width.second))));
+                // Calmp wall-thickness to configured values, and draw the support part with the calculated wall-thickness.
```
wawanbreton: I think that is what you mean 😃 
```suggestion
                // Clamp wall-thickness to configured values, and draw the support part with the calculated wall-thickness.
```

wawanbreton (approved):

src/TreeSupport.cpp
rburema: Ha, yes 😁 keep Calmp and carry on.

wawanbreton (commented):

| Benchmark suite | Current: 40cbe2dd19589d869cfdced3ce84f0ba2b9ecd84 | Previous: 9ec0839b6932a342f068bde168ae47d261f7fd7d | Ratio |
|-|-|-|-|
| `SimplifyTestFixture/simplify_slot_noplugin` | `14.624904656954746` ns/iter | `3.5472080572366425` ns/iter | `4.12` |

PR #2348, opened 2026-07-16
Apply fuzzy skin to the spiral wall when spiralize is enabled

### LINKED ISSUE #12850: Fuzzy skin not working with spiralize outer contour mode
### Application Version

5.1.0

### Platform

Windows 10

### Printer

Ender 3 V2

### Reproduction steps

Activate fuzzy skin mode as well as spiralize outer contour.

### Actual results

Fuzzy skin only beeing printed on the layers that have a wallthickness bigger than 1.

### Expected results

Both working together, like they did in Cura 4.13.1

### Checklist of files to include

- [ ] Log file
- [ ] Project file

### Additional information & file uploads

I printed several parts with those two modes together in Cura 4.13.1 without a problem. (using thickness .1mm and density 5 1/mm)
Cura 5.1.0 doesn't seem to allow it anymore. In the preview window only the layers with a wallstrength bigger than 1 allow fuzzy skin. 
Cura_4_13_1
Cura_5_1_0

kvanonavk:
# Description

This fixes fuzzy skin not being applied when Spiralize Outer Contour is enabled, a regression introduced with the variable-width wall toolpaths (Arachne) in Cura 5.0.

`FffPolygonGenerator::processFuzzyWalls()` only processes `part.wall_toolpaths`, but in spiralize mode the wall is stored as plain polygons in `part.spiral_wall`, so fuzzy skin was only visible on the initial bottom layers (the only layers that get wall toolpaths). Before 5.0 the spiral wall was handled explicitly (`spiralize ? part.spiral_wall : part.insets[0]`).

This restores that behavior by applying the same fuzzing to the spiral wall polygons, including the "Fuzzy Skin Outside Only" hole check and the first-layer brim exception.

Fixes Ultimaker/Cura#12850

## Type of change

# How Has This Been Tested?

- Headless CLI slices of `tests/integration/resources/cylinder1000.stl` with `magic_spiralize=true` and `magic_fuzzy_skin_enabled=true` (otherwise default fdmprinter settings), comparing the radial deviation of the wall from the layer centroid per layer:
  - Before: spiralized layers were perfect circles (radial std. dev. ≈ 0.004 mm, i.e. only simplification noise); only the initial bottom layers were fuzzed.
  - After: every spiralized layer shows the expected jitter (std. dev. ≈ 0.15 mm at the default 0.3 mm fuzzy thickness), while the bottom layers are unchanged.
- Regression check: slicing with fuzzy skin enabled but spiralize disabled is unaffected.
- Visual check of the sliced result in the Cura preview.

**Test Configuration**:
* Operating System: Windows 11

# IMPORTANT
The code was generated using Claude's Fable model. I was annoyed by this bug for quite a while (the issue on Github is almost 4 years old) so I decided to give it a go using Fable. Looks fine and works as expected but I'd really appreciate if someone could take a look at it and verify that it can be merged, thank you!

### DIFF
--- a/src/FffPolygonGenerator.cpp
+++ b/src/FffPolygonGenerator.cpp
@@ -1205,6 +1205,72 @@ void FffPolygonGenerator::processFuzzyWalls(SliceMeshStorage& mesh)
                 }
             }
             part.wall_toolpaths = result_paths;
+
+            // Also fuzz the spiral wall, if any. In spiralize mode the outer wall isn't stored in wall_toolpaths but in spiral_wall (as plain polygons),
+            // so it needs to be processed separately, like it used to be before the variable-width toolpaths existed.
+            if (part.spiral_wall.empty())
+            {
+                continue;
+            }
+            if (apply_outside_only)
+            {
+                hole_area = part.print_outline.getOutsidePolygons().offset(-line_width);
+            }
+            Shape fuzzy_spiral_walls;
+            for (const Polygon& wall : part.spiral_wall)
+            {
+                if (wall.size() < 3
+                    || (apply_outside_only
+                        && std::any_of(
+                            wall.begin(),
+                            wall.end(),
+                            [&hole_area](const Point2LL& point)
+                            {
+                                return hole_area.inside(point);
+                            })))
+                {
+                    fuzzy_spiral_walls.push_back(wall);
+                    continue;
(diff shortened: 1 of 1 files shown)


GregValiant:
@HellAholic,
This is a proposed fix for a long standing bug.

PR #2349, opened 2026-07-16
[CURA-13054] Don't take support into account w.r.t. seam overhang.

rburema:
It was using the `outlines_below` value, which included the support, as it probably should for the other thing(s) we use it for.

CURA-13054

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -2684,7 +2684,9 @@ FffGcodeWriter::InsetsPreprocessResult FffGcodeWriter::preProcessInsets(
             gcode_layer.setBridgeWallMask(Shape());
         }
 
-        const Shape model_supported_region = non_support_outlines_below.offset(-half_outer_wall_width);
+        Shape model_supported_region = non_support_outlines_below.offset(-half_outer_wall_width);
+        // remove those parts of the layer below that are narrower than a wall line width as they will not be printed
+        model_supported_region = model_supported_region.offset(-half_outer_wall_width).offset(half_outer_wall_width);
 
         const auto get_supported_region = [&model_supported_region, &layer_height](const AngleDegrees& overhang_angle) -> Shape
         {
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -2691,7 +2691,7 @@ FffGcodeWriter::InsetsPreprocessResult FffGcodeWriter::preProcessInsets(
         const auto get_supported_region = [&model_supported_region, &layer_height](const AngleDegrees& overhang_angle) -> Shape
         {
             // the overhang mask is set to the area of the current part's outline minus the region that is considered to be supported
-            // the supported region is made up of those areas that really are supported by (either) the model (or support, if 'full') on the layer below
+            // the supported region is made up of those areas that are supported by the model on the layer below
             // expanded to take into account the overhang angle, the greater the overhang angle, the larger the supported area is
             // considered to be
             if (overhang_angle < 90.0)
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -2595,7 +2595,7 @@ FffGcodeWriter::InsetsPreprocessResult FffGcodeWriter::preProcessInsets(
                 }
             }
         }
-        Shape non_support_outlines_below = outlines_below;
+        const Shape& non_support_outlines_below = outlines_below;
 
         const coord_t layer_height = mesh_config.inset0_config.getLayerThickness();
 
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -2595,6 +2595,7 @@ FffGcodeWriter::InsetsPreprocessResult FffGcodeWriter::preProcessInsets(
                 }
             }
         }
+        Shape non_support_outlines_below = outlines_below;
 
         const coord_t layer_height = mesh_config.inset0_config.getLayerThickness();
 
@@ -2684,18 +2685,19 @@ FffGcodeWriter::InsetsPreprocessResult FffGcodeWriter::preProcessInsets(
         }
 
         const Shape fully_supported_region = outlines_below.offset(-half_outer_wall_width);
+        const Shape model_supported_region = non_support_outlines_below.offset(-half_outer_wall_width);
         const Shape part_print_region = part.outline.offset(-half_outer_wall_width);
 
-        const auto get_supported_region = [&fully_supported_region, &layer_height](const AngleDegrees& overhang_angle) -> Shape
+        const auto get_supported_region = [&fully_supported_region, &model_supported_region, &layer_height](const AngleDegrees& overhang_angle, bool full = true) -> Shape
         {
             // the overhang mask is set to the area of the current part's outline minus the region that is considered to be supported
-            // the supported region is made up of those areas that really are supported by either model or support on the layer below
+            // the supported region is made up of those areas that really are supported by (either) the model (or support, if 'full') on the layer below
             // expanded to take into account the overhang angle, the greater the overhang angle, the larger the supported area is
             // considered to be
             if (overhang_angle < 90.0)
             {
                 const coord_t overhang_width = layer_height * std::tan(AngleRadians(overhang_angle));
-                return fully_supported_region.offset(overhang_width + 10);
+                return (full ? fully_supported_region : model_supported_region).offset(overhang_width + 10);
             }
 
             return Shape();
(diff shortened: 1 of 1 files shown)


src/FffGcodeWriter.cpp
```diff
@@ -2595,6 +2595,7 @@ FffGcodeWriter::InsetsPreprocessResult FffGcodeWriter::preProcessInsets(
                 }
             }
         }
+        Shape non_support_outlines_below = outlines_below;
```
casperlamboo:
```suggestion
```

src/FffGcodeWriter.cpp
```diff
@@ -2684,18 +2685,19 @@ FffGcodeWriter::InsetsPreprocessResult FffGcodeWriter::preProcessInsets(
         }
 
         const Shape fully_supported_region = outlines_below.offset(-half_outer_wall_width);
+        const Shape model_supported_region = non_support_outlines_below.offset(-half_outer_wall_width);
```
casperlamboo:
```suggestion
        const Shape model_supported_region = outlines_below.offset(-half_outer_wall_width);
```

casperlamboo (commented):
Overal code looks good, nice and concise. Just have two remarks

- For line `Shape non_support_outlines_below = outlines_below;`, I think we should be using a const ref here. Since the only operator we apply on it is an offset which does not mutate the original data we don't need to copy the original polygon. My code suggestions are related to this remark
- Second remark I'm not sure if we should implement; but now we either look at the model, or model+support outline. For seam placement, don't we want to preferably place the seam on model, if not possible on support, if not possible elsewhere.

wawanbreton:
> * For line `Shape non_support_outlines_below = outlines_below;`, I think we should be using a const ref here. Since the only operator we apply on it is an offset which does not mutate the original data we don't need to copy the original polygon. My code suggestions are related to this remark

Completely agree. Compiler will tell you if it works anyway :smiley: 

> * Second remark I'm not sure if we should implement; but now we either look at the model, or model+support outline. For seam placement, don't we want to preferably place the seam on model, if not possible on support, if not possible elsewhere.

It would be nice indeed to have 2 levels of placement. This can be easily achieved by storing the two shapes (since we do calculate them at some point anyway) and have a second `ExclusionAreaScoringCriterion` with a different weight. https://github.com/Ultimaker/CuraEngine/blob/7a627bc0c090657d2a849f7a7c700a6aa8b11eee/include/PathOrderOptimizer.h#L819

src/FffGcodeWriter.cpp
```diff
@@ -2747,7 +2749,11 @@ FffGcodeWriter::InsetsPreprocessResult FffGcodeWriter::preProcessInsets(
                 for (const auto& regions : merged_regions)
                 {
                     const SpeedRegion& last_region = *ranges::prev(regions.end());
-                    overhang_masks.push_back(LayerPlan::OverhangMask{ get_supported_region(last_region.overhang_angle), last_region.speed_factor });
+                    constexpr bool not_full = false; // Do not take support into account for the seam overhang.
```
HellAholic:
```suggestion
                    constexpr bool not_full = false; // Do not take support into account for the wall overhang.
```
this section seems to be in the wall overhang speed rather than seam 🤔 if so the comment should be wall overhang rather than seam overhang.

src/FffGcodeWriter.cpp
```diff
@@ -2758,7 +2764,8 @@ FffGcodeWriter::InsetsPreprocessResult FffGcodeWriter::preProcessInsets(
         const AngleDegrees seam_overhang_angle = mesh.settings.get<AngleDegrees>("seam_overhang_angle");
         if (seam_overhang_angle < 90.0)
         {
-            const Shape supported_region_seam = get_supported_region(seam_overhang_angle);
+            constexpr bool full = true; // Do not take support into account for the seam overhang.
+            const Shape supported_region_seam = get_supported_region(seam_overhang_angle, full);
```
HellAholic:
```suggestion
            constexpr bool full = false; // Do not take support into account for the seam overhang.
            const Shape supported_region_seam = get_supported_region(seam_overhang_angle, full);
```

Same pattern as above it seems but we're sending the value true, so in both cases we should send in not_full (false) to ignore support and return `model_supported_region`

src/FffGcodeWriter.cpp
```diff
@@ -2684,18 +2685,19 @@ FffGcodeWriter::InsetsPreprocessResult FffGcodeWriter::preProcessInsets(
...
         const Shape part_print_region = part.outline.offset(-half_outer_wall_width);
 
-        const auto get_supported_region = [&fully_supported_region, &layer_height](const AngleDegrees& overhang_angle) -> Shape
+        const auto get_supported_region = [&fully_supported_region, &model_supported_region, &layer_height](const AngleDegrees& overhang_angle, bool full = true) -> Shape
         {
             // the overhang mask is set to the area of the current part's outline minus the region that is considered to be supported
-            // the supported region is made up of those areas that really are supported by either model or support on the layer below
+            // the supported region is made up of those areas that really are supported by (either) the model (or support, if 'full') on the layer below
             // expanded to take into account the overhang angle, the greater the overhang angle, the larger the supported area is
             // considered to be
```
HellAholic, line 2696: It's a bit confusing of a description:
```
// returns the region that is considered to be supported by (either) the model (or support, if 'full') on the layer below,
// expanded to take into account the overhang angle; the greater the overhang angle, the larger the supported area is
// considered to be
```

HellAholic (changes requested):

src/FffGcodeWriter.cpp
```diff
@@ -2683,19 +2684,19 @@ FffGcodeWriter::InsetsPreprocessResult FffGcodeWriter::preProcessInsets(
             gcode_layer.setBridgeWallMask(Shape());
         }
 
-        const Shape fully_supported_region = outlines_below.offset(-half_outer_wall_width);
+        const Shape model_supported_region = non_support_outlines_below.offset(-half_outer_wall_width);
         const Shape part_print_region = part.outline.offset(-half_outer_wall_width);
```
HellAholic:
```suggestion
```
I think this part can be removed since the calculation of seam_overhang_mask no longer relies on it.

https://github.com/Ultimaker/CuraEngine/pull/2349/changes#diff-1503d7ea10a5771cbd5934a56564ffae85b86613f443dd7e4750810dd8c58f5fR2762

src/FffGcodeWriter.cpp
```diff
@@ -2595,6 +2595,7 @@ FffGcodeWriter::InsetsPreprocessResult FffGcodeWriter::preProcessInsets(
                 }
             }
         }
+        const Shape non_support_outlines_below = outlines_below;
```
HellAholic, line 2598: Wouldn't it be more effective if we do the snapshot after clean up in https://github.com/Ultimaker/CuraEngine/blob/8b39252a9d8c2acf6d883dc9705d76e4aa90e810/src/FffGcodeWriter.cpp#L2639-L2641

That does removes the lines that we're not gonna print since they are deemed too narrow, so it does change the outline in some models 🤔

src/FffGcodeWriter.cpp
```diff
@@ -2758,8 +2759,9 @@ FffGcodeWriter::InsetsPreprocessResult FffGcodeWriter::preProcessInsets(
         const AngleDegrees seam_overhang_angle = mesh.settings.get<AngleDegrees>("seam_overhang_angle");
         if (seam_overhang_angle < 90.0)
         {
-            const Shape supported_region_seam = get_supported_region(seam_overhang_angle);
-            gcode_layer.setSeamOverhangMask(part_print_region.difference(supported_region_seam).offset(10));
+            const auto seam_overhang_mask
+                = storage.getMachineBorder(mesh.settings.get<ExtruderTrain&>("wall_0_extruder_nr").extruder_nr_).difference(get_supported_region(seam_overhang_angle));
+            gcode_layer.setSeamOverhangMask(seam_overhang_mask);
```
HellAholic, line 2765: Mostly a question regarding the removal of offset in the new calculation.

src/FffGcodeWriter.cpp
```diff
@@ -2683,19 +2684,19 @@ FffGcodeWriter::InsetsPreprocessResult FffGcodeWriter::preProcessInsets(
...
         const Shape part_print_region = part.outline.offset(-half_outer_wall_width);
 
-        const auto get_supported_region = [&fully_supported_region, &layer_height](const AngleDegrees& overhang_angle) -> Shape
+        const auto get_supported_region = [&model_supported_region, &layer_height](const AngleDegrees& overhang_angle) -> Shape
         {
             // the overhang mask is set to the area of the current part's outline minus the region that is considered to be supported
-            // the supported region is made up of those areas that really are supported by either model or support on the layer below
+            // the supported region is made up of those areas that really are supported by (either) the model (or support, if 'full') on the layer below
             // expanded to take into account the overhang angle, the greater the overhang angle, the larger the supported area is
             // considered to be
```
HellAholic, line 2696: The comment here needs some updating with the changes made.

src/FffGcodeWriter.cpp
casperlamboo: Because the previous implementation had issues where the polygon edge would be on the border of the seam locations, meaning that slight imprecisions means a candidate seam location would either correctly or incorrectly be marked as overhanging. This is why a "grace" area was needed. However, with the current implementation we don't consider the model polygon so no candidate seam locations would be bordering in this case.

src/FffGcodeWriter.cpp
casperlamboo: https://github.com/Ultimaker/CuraEngine/pull/2349/commits/685c0107ff2351c43e36d3102db3569ba22ad1aa

src/FffGcodeWriter.cpp
casperlamboo: https://github.com/Ultimaker/CuraEngine/pull/2349/commits/892da8546475b3783ac2d8aa69e857714bcaa146

HellAholic (approved):
LGTM!

PR #2354, opened 2026-07-30
Fix implicit type conversions

casperlamboo:
Leading to WASM crashes when support + touching buildplate is configured

NP-1348

### DIFF
--- a/src/support.cpp
+++ b/src/support.cpp
@@ -1425,7 +1425,7 @@ void AreaSupport::moveUpFromModel(
         to_be_removed = stair_removal.unionPolygons(bottom_outline);
         if (layer_idx % bottom_stair_step_layer_count == 0)
         { // update stairs for next step
-            const Shape supporting_bottom = storage.getLayerOutlines(bottom_layer_nr - 1, no_support, no_prime_tower);
+            const Shape supporting_bottom = storage.getLayerOutlines(LayerIndex(bottom_layer_nr) - 1, no_support, no_prime_tower);
             const Shape allowed_step_width = supporting_bottom.offset(support_bottom_stair_step_width).intersection(sloped_areas);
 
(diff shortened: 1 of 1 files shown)


src/support.cpp
```diff
@@ -1453,7 +1270,7 @@ void AreaSupport::generateSupportAreasForMesh(
             conical_support_offset = boundedTan(-conical_support_angle) * layer_thickness;
         }
         const bool conical_support = infill_settings.get<bool>("support_conical_enabled") && conical_support_angle != 0;
-        for (LayerIndex layer_idx = 1; layer_idx < storage.support.supportLayers.size(); layer_idx++)
+        for (LayerIndex layer_idx = 1; layer_idx < LayerIndex(storage.support.supportLayers.size()); layer_idx++)
```
jellespijker, line 1273: Probably okay, unless you have very big numbers, but this will cast a `size_t` into a `int64_t` introducing potential unsigned-to-signed conversion

src/support.cpp
```diff
@@ -1608,7 +1425,7 @@ void AreaSupport::moveUpFromModel(
         to_be_removed = stair_removal.unionPolygons(bottom_outline);
         if (layer_idx % bottom_stair_step_layer_count == 0)
         { // update stairs for next step
-            const Shape supporting_bottom = storage.getLayerOutlines(bottom_layer_nr - 1, no_support, no_prime_tower);
+            const Shape supporting_bottom = storage.getLayerOutlines(LayerIndex(bottom_layer_nr) - 1, no_support, no_prime_tower);
```
jellespijker, line 1428: same as above

jellespijker (approved):
nitpick, because this will never be a problem

src/support.cpp
casperlamboo: Don't think we'll ever have $9,223,372,036,854,775,807$ layers

PR #2357, opened 2026-08-05
[CURA-13287] Re-fix random dots without introducing microsegments.

rburema:
Originally, an issue (see #2089 / C-9399) was fixed, where we needed to re-insert some ignored local maxima.

This fix was intended to be applied only for very small areas. Large areas didn't need to be corrected for, certainly not ones large enough to be a problem w.r.t non-convexness. (And non-connected shapes won't be treated as a single area w.r.t. SkeletalTrapezoidation.)

We found, while doing some other work, that this could cause points to appear in the middle of nowhere (see comments in #2346 / C-13250). This was fixed by taking into account that averaged points should only be merged when close together.

The real problem was however that the fix was also applied to large areas (otherwise the middle of nowhere behaviour wouldn't be possible at all). If we restrict the fix to only be for cases where the total length of the covered lines is very small (as posited, but not actually implemented, in the first mentioned PR), we get the originally intended behaviour of the hack.

... and this is not a problem, because there can't be multiple areas within a single trapezoidation (other than holes), and areas _otherwise_ large enough to contain multiple clearly separate local maxima are not the intended target for the original and now current fix anyway). (Even though the input for the trapezoidation is a Shape, not a SimpleShape ... but the outline that it receives as an input via SlicePart _is_, at the very least for the walls.)

P.S. I wanted the `combined_local_maxima_point` to be an optional, but that turned out to work a bit ugly with the +-operator.

### DIFF
--- a/src/arachne/SkeletalTrapezoidation.cpp
+++ b/src/arachne/SkeletalTrapezoidation.cpp
@@ -2215,7 +2215,13 @@ void SkeletalTrapezoidation::generateLocalMaximaSingleBeads()
         coord_t length_;
 
         // NOTE: Empty constructor; `acc_` is set to 0, so that it can be used as a neutral element in the `+=` operator.
-        LocalMaximaPoint() : p_(0, 0), width_(0), acc_(0), length_(0) {}
+        LocalMaximaPoint()
+            : p_(0, 0)
+            , width_(0)
+            , acc_(0)
+            , length_(0)
+        {
+        }
 
         LocalMaximaPoint(const Point2LL& p, coord_t width)
             : p_(p)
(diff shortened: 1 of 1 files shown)


src/arachne/SkeletalTrapezoidation.cpp
```diff
@@ -2212,11 +2212,22 @@ void SkeletalTrapezoidation::generateLocalMaximaSingleBeads()
         Point2LL p_;
         coord_t width_;
         size_t acc_;
```
wawanbreton: I know this is not part of this PR, but could this variable be renamed with a more explicit name ? Out of the context, I don't know if this is `acceleration`, `accumulation` or something else 😄

wawanbreton (approved):
Just a minor remark. Code looks good, but I must admit I don't get the global meaning...

src/arachne/SkeletalTrapezoidation.cpp
rburema: Yes, that's a good idea -- was intended to be an 'accumulator'.

rburema:
Superseded by a different fix.

PR #2360, opened 2026-08-11
CURA-13252 smooth tree support branches

wawanbreton:
Implements the tree support branches smoothing, as previously developed on the spike branch.
Smoothing is made starting from the current layer and going down only, so that branches separation does not generate position jumps.
Also the smoothing distance is capped to the distance to the top of the tree, so that smoothing decreases when moving up, and ends up ineffective on the top-most layers, that should keep their original shape to properly support the interface.

CURA-13252
Comes with https://github.com/Ultimaker/Cura/pull/21749

### DIFF
--- a/src/TreeSupport.cpp
+++ b/src/TreeSupport.cpp
@@ -2259,16 +2259,15 @@ void TreeSupport::finalizeInterfaceAndSupportAreas(
                 case InterfacePreference::SUPPORT_LINES_OVERWRITE_INTERFACE:
                 {
                     Shape tree_lines;
-                    tree_lines = tree_lines.unionPolygons(
-                        TreeSupportUtils::generateSupportInfillLines(
-                            support_layer_storage[layer_idx],
-                            config,
-                            false,
-                            layer_idx,
-                            config.support_line_distance,
-                            storage.support.cross_fill_provider,
-                            true)
-                            .offset(config.support_line_width / 2));
+                    tree_lines = tree_lines.unionPolygons(TreeSupportUtils::generateSupportInfillLines(
+                                                              support_layer_storage[layer_idx],
+                                                              config,
+                                                              false,
+                                                              layer_idx,
+                                                              config.support_line_distance,
+                                                              storage.support.cross_fill_provider,
+                                                              true)
+                                                              .offset(config.support_line_width / 2));
                     storage.support.supportLayers[layer_idx].support_roof = storage.support.supportLayers[layer_idx].support_roof.difference(tree_lines);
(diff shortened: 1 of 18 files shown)


src/TreeSupport.cpp
```diff
@@ -1713,7 +1698,7 @@ void TreeSupport::generateBranchAreas(
                     // Visualization: https://jsfiddle.net/0zvcq39L/2/
                     // Ovalizes the circle to an ellipse, that contains both old center and new target position.
                     double used_scale = (movement.second + offset) / (1.0 * config.branch_radius);
-                    Point2LL center_position = elem->result_on_layer_ + movement.first / 2;
+                    Point2LL center_position = elem->result_on_layer_ + movement.first / static_cast<coord_t>(2);
```
rburema: I recently realized that you can also just do `coord_t{<number>}` in these cases; seems a bit more readable to me.

src/TreeSupport.cpp
```diff
@@ -1926,6 +1911,49 @@ void TreeSupport::smoothBranchAreas(std::vector<std::unordered_map<TreeSupportEl
...
+                const size_t actual_smooth_window = std::min(smooth_window, element->distance_to_top_);
+                const TreeSupportElement* child = element->getChild();
+                Point2LL interlayer_position_sum = element->result_on_layer_;
+                size_t added_layers = 1;
+                while (child && child->isResultOnLayerSet() && added_layers < actual_smooth_window)
+                {
+                    interlayer_position_sum += child->result_on_layer_;
+                    child = child->getChild();
+                    ++added_layers;
+                }
```
rburema: Why not `for (added_layers = 1; child & ....; ++added_layers){ ... }`?

rburema (approved):
LGTM!

src/TreeSupport.cpp
wawanbreton: Oh, I will definitely try it. The `static_cast` syntax is really ugly.

src/TreeSupport.cpp
wawanbreton: Definitely possible indeed :smile: I'll see how readable that ends up

wawanbreton (commented):

| Benchmark suite | Current: 5cdf40b4958ecd9fff97a598055d398849d57068 | Previous: ed0bd3e9b751f0d57b6fd27198a35d7ae819dbf3 | Ratio |
|-|-|-|-|
| `SimplifyTestFixture/simplify_local` | `1.7652989915507917` ns/iter | `1.117797545340744` ns/iter | `1.58` |

PR #2362, opened 2026-08-13
[CURA-13265] Fix infill (wall) overlap for infill-support.

rburema:
Wall-width is already taken into account (but maybe not at the time of writing?), so when subtracting half a wall-width over that, it made it seem like the infill overlap setting wasn't working properly.

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -2183,7 +2183,7 @@ bool FffGcodeWriter::processSingleLayerInfill(
     {
         const auto infill_wall_line_count = static_cast<coord_t>(mesh.settings.get<size_t>("infill_wall_line_count"));
         const coord_t infill_wall_offset = -infill_wall_line_count * infill_line_width;
-        const Shape infill_contour = part.infill_area.offset(-(infill_line_width / 2) + infill_overlap + infill_wall_offset);
+        const Shape infill_contour = part.infill_area.offset(infill_overlap + infill_wall_offset);
         const LayerPlan* completed_layer_below = layer_plan_buffer.getCompletedLayerPlan(gcode_layer.getLayerNr() - 1);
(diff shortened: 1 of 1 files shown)


wawanbreton (approved):
Nice, that should improve adhesion between the skin support and the outer walls.
Sorry for the mistake 😅
When thinking about it though, I probably did that to avoid the protrusion of the infill into the outer walls. @HellAholic could you make a test print as part of QA to check the quality ?

rburema:
@wawanbreton Don't worry about it!

Also, if the quality is a problem, we should change that with parameters I think. Worst case, we can split it out to a new one.

wawanbreton (commented):

| Benchmark suite | Current: 9a6c743fe9182488cf505821471f66a756cfee12 | Previous: fbf37fc21931626b206ebb95550af90ec7b375ec | Ratio |
|-|-|-|-|
| `InfillTest/Infill_generate_connect/0/400` | `5.804878534799016` ms/iter | `3.648410709524255` ms/iter | `1.59` |
| `HolesWallTestFixture/InsetOrderOptimizer_getInsetOrder/3` | `0.0002625348999732111` ms/iter | `0.00017382738462409593` ms/iter | `1.51` |
| `HolesWallTestFixture/InsetOrderOptimizer_getInsetOrder/9999` | `0.011970930861527649` ms/iter | `0.007572545422304867` ms/iter | `1.58` |
| `SimplifyTestFixture/simplify_slot_noplugin` | `3.789785515326669` ns/iter | `2.1604040436005425` ns/iter | `1.75` |

HellAholic:
> Nice, that should improve adhesion between the skin support and the outer walls. Sorry for the mistake 😅 When thinking about it though, I probably did that to avoid the protrusion of the infill into the outer walls. HellAholic could you make a test print as part of QA to check the quality ?

Before:


After:


Using lightning infill and 0.04mm infill overlap (PLA-0.2mm)

PR #2363, opened 2026-08-13
[CURA-13266] Skin-support: No more infill-bridging over skin one layer lower.

rburema:
Skin support was still applied in some cases where there's only one layer of infill sandwiched between two layers of skin. Though this already should have worked; it turns out the bridging will 'fill in' the parts anyway. (Also even if that would have worked correctly, the normal infill wasn't applied to the 'sandwiched' area, so that would've been empty.) This should now be fixed with this commit.

### DIFF
--- a/src/FffGcodeWriter.cpp
+++ b/src/FffGcodeWriter.cpp
@@ -2173,10 +2173,11 @@ bool FffGcodeWriter::processSingleLayerInfill(
     // boundary edge
     Shape infill_below_skin;
     Shape infill_not_below_skin;
+    Shape infill_sandwiched;
     AngleDegrees skin_support_angle;
     if (gcode_layer.getLayerNr() > 0)
     {
-        partitionInfillBySkinAbove(infill_below_skin, infill_not_below_skin, gcode_layer, mesh, part, infill_line_width);
+        partitionInfillBySkinAbove(infill_below_skin, infill_not_below_skin, infill_sandwiched, gcode_layer, mesh, part, infill_line_width);
     }
 
     if (! infill_below_skin.empty())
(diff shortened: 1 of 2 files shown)


src/FffGcodeWriter.cpp
```diff
@@ -2433,22 +2436,24 @@ void FffGcodeWriter::partitionInfillBySkinAbove(
         return;
     }
 
-    Shape skin_above_combined; // skin regions on the layers above combined with small gaps between
-
-    const size_t skin_layer_nr = gcode_layer.getLayerNr() + 1;
-    if (skin_layer_nr < mesh.layers.size())
+    const auto get_combined_skin_func = [&](Shape& skin_combined, int skin_layer_nr)
```
wawanbreton: Since this piece of code is quite big, maybe move it to a separate function ?

wawanbreton (commented):
I was a bit concerned by the `sandwiched` wording at first, but it properly conveys its meaning 😃
Just one remark otherwise, to be discussed

src/FffGcodeWriter.cpp
rburema: Yes that does make sense, there isn't (even) any calculations before that would need to be captured.

wawanbreton (approved):

HellAholic:
Before
image
After
image

@GregValiant I couldn't find the issue on the Cura side reports, if you encounter the issue during your sweeps, could you please link to this PR and close it with "Implemented for 5.14 release". Many thanks.

GregValiant:
I don't recall this one.  I've been sweeping early issues to see if they were fixed/impacted since their submittal.  I'll keep this in mind.

PR #2364, opened 2026-08-17
Display engine plugins in gcode

wawanbreton:
Add the list of active engine plugins in the generated gcode. Syntax is based on the post-processing plugins: https://github.com/Ultimaker/Cura/blob/05e2c8937da42b48f8502f3ac69bcf8ed44b7307/plugins/PostProcessingPlugin/PostProcessingPlugin.py#L94

Result is something like:
```
;ENGINE PLUGINS
;  [valve (GCODE_PATHS_MODIFY)]
```

CURA-11212

### DIFF
--- a/include/plugins/slotproxy.h
+++ b/include/plugins/slotproxy.h
@@ -64,6 +64,21 @@ public:
     }
 
     /**
+     * @brief Applies the visitor function on all the loaded plugins
+     * @param function The function to be called on each plugin
+     */
+    void forEachPlugin(const std::invocable<const slot_metadata&, const plugin_metadata&> auto& function) const
+    {
+        for (const value_type& plugin : plugins_)
+        {
+            if (plugin.pluginInfo().has_value())
+            {
+                function(plugin.slotInfo(), plugin.pluginInfo().value());
+            }
+        }
+    }
+
+    /**
      * @brief Executes the plugin operation.
      *
      * This operator allows the SlotProxy object to be invoked as a callable, which delegates the
--- a/src/gcode_export/gcodeExport.cpp
+++ b/src/gcode_export/gcodeExport.cpp
@@ -25,6 +25,7 @@
 #include "gcode_export/FixedGCodePart.h"
 #include "gcode_export/GcodeTemplateResolver.h"
 #include "gcode_export/ResolvedGCodePart.h"
+#include "plugins/slots.h"
 #include "settings/types/LayerIndex.h"
 #include "sliceDataStorage.h"
 #include "utils/Date.h"
@@ -345,6 +346,21 @@ std::string GCodeExport::getFileHeader(const std::vector<bool>& extruder_is_used
         prefix << ";MAXY:" << INT2MM(total_bounding_box_.max_.y_) << new_line_;
         prefix << ";MAXZ:" << INT2MM(total_bounding_box_.max_.z_) << new_line_;
         prefix << ";TARGET_MACHINE.NAME:" << transliterate(machine_name_) << new_line_;
+    }
+
+    std::vector<std::pair<std::string, std::string>> plugins; // For each plugin: slot name, plugin name
+    slots::instance().forEachPlugin(
+        [&plugins](const plugins::slot_metadata& slot, const plugins::plugin_metadata& plugin)
+        {
+            plugins.push_back(std::make_pair(plugins::v0::SlotID_Name(slot.slot_id), plugin.plugin_name));
+        });
+    if (! plugins.empty())
+    {
+        prefix << ";ENGINE PLUGINS" << new_line_;
+        for (const std::pair<std::string, std::string>& plugin : plugins)
+        {
+            prefix << fmt::format(";  [{} ({})]", plugin.second, plugin.first) << new_line_;
+        }
     }
 
     return prefix.str();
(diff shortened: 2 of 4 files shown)


src/gcode_export/gcodeExport.cpp
```diff
@@ -25,6 +25,7 @@
 #include "gcode_export/FixedGCodePart.h"
 #include "gcode_export/GcodeTemplateResolver.h"
 #include "gcode_export/ResolvedGCodePart.h"
+#include "plugins/slots.h"
```
casperlamboo, line 28:
```suggestion
#ifdef ENABLE_PLUGINS
#include "plugins/slots.h"
#endif
```

src/gcode_export/gcodeExport.cpp
```diff
@@ -347,6 +348,21 @@ std::string GCodeExport::getFileHeader(const std::vector<bool>& extruder_is_used
...
+            plugins.push_back(std::make_pair(plugins::v0::SlotID_Name(slot.slot_id), plugin.plugin_name));
+        });
+    if (! plugins.empty())
+    {
+        prefix << ";ENGINE PLUGINS" << new_line_;
+        for (const std::pair<std::string, std::string>& plugin : plugins)
+        {
+            prefix << fmt::format(";  [{} ({})]", plugin.second, plugin.first) << new_line_;
+        }
+    }
```
casperlamboo, line 365:
```suggestion
#ifdef ENABLE_PLUGINS
    std::vector<std::pair<std::string, std::string>> plugins; // For each plugin: slot name, plugin name
    slots::instance().forEachPlugin(
        [&plugins](const plugins::slot_metadata& slot, const plugins::plugin_metadata& plugin)
        {
            plugins.push_back(std::make_pair(plugins::v0::SlotID_Name(slot.slot_id), plugin.plugin_name));
        });
    if (! plugins.empty())
    {
        prefix << ";ENGINE PLUGINS" << new_line_;
        for (const std::pair<std::string, std::string>& plugin : plugins)
        {
            prefix << fmt::format(";  [{} ({})]", plugin.second, plugin.first) << new_line_;
        }
    }
#endif
```

casperlamboo (commented):
NPM package is failing, I believe we should add `#ifdef ENABLE_PLUGINS` blocks

src/gcode_export/gcodeExport.cpp
wawanbreton: Oh yes of course, totally missed that. The header contains an `#ifdef` though, so this one should not be required.

include/plugins/slotproxy.h
```diff
@@ -63,6 +63,21 @@ class SlotProxy
...
 
+    /**
+     * @brief Applies the visitor function on all the loaded plugins
+     * @param function The function to be called on each plugin
+     */
+    void forEachPlugin(const std::invocable<const slot_metadata&, const plugin_metadata&> auto& function) const
+    {
+        for (const value_type& plugin : plugins_)
+        {
+            if (plugin.pluginInfo().has_value())
```
rburema, line 74: Would there be a case where there is a plugin without plugin-info?

If so; that partly defeats the purpose of the exercise, since then you can still have a plugin running but get gcode without any indication of that. Another problem in that case is that then, this 'foreach' function isn't really doing what it says in the tin, since it skips those without the info.

If _not_, then why do the check at all? (And then not throw/indicate some kind of error if an assumption is voided after all?)

casperlamboo (approved):

src/gcode_export/gcodeExport.cpp
```diff
@@ -347,6 +348,23 @@ std::string GCodeExport::getFileHeader(const std::vector<bool>& extruder_is_used
...
+#ifdef ENABLE_PLUGINS
+    std::vector<std::pair<std::string, std::string>> plugins; // For each plugin: slot name, plugin name
+    slots::instance().forEachPlugin(
+        [&plugins](const plugins::slot_metadata& slot, const plugins::plugin_metadata& plugin)
+        {
+            plugins.push_back(std::make_pair(plugins::v0::SlotID_Name(slot.slot_id), plugin.plugin_name));
+        });
+    if (! plugins.empty())
+    {
+        prefix << ";ENGINE PLUGINS" << new_line_;
```
rburema, line 360: This doesn't really fit with the rest of the header. I'd be fine with it, but I wonder what the firmware team thinks regarding parsing/skipping this part.

@robinmdh ? Would you have a problem with:
```
;ENGINE PLUGINS
;  [blah (bloeh)]
;  [bleh (blib blie)]
;  ...
```
in the header? Note the lack of `_` and `:` and the newlines.

rburema (dismissed):
I've got some remarks. I'd be willing to accept as-is if you can explain though :-)

include/plugins/slotproxy.h
wawanbreton: If I am correct, that happens when the proxy is empty, i.e. there is no single plugin actually running for this slot, but then there is still a pseudo-plugin that does the default implementation. I can make sure of that.

src/gcode_export/gcodeExport.cpp
wawanbreton: As mentioned in the description, this matches the way the post-processing plugins are mentioned:
```
;ENGINE PLUGINS
;  [valve (GCODE_PATHS_MODIFY)]
;POSTPROCESSED
;  [PurgeLinesAndUnload]
```

rburema: Ah sorry, not sure how I missed that 😅

PR #2365, opened 2026-08-17
Package `.wasm` file seperatly

casperlamboo:


wawanbreton:
@HellAholic or @rburema any idea what this PR could be useful to ?

PR #2366, opened 2026-08-24
CURA-9839 add bridging and overhang line type color scheme

wawanbreton:
This PR adds the possibility to set special attributes to print segments. Those attributes are sent to front-end for display purposes.

In this case, we are interested in displaying the segment that are overhanging and bridging, so we also add 2 flags for those cases and set them when appropriate.

Comes with https://github.com/Ultimaker/Cura/pull/21766

CURA-9839

### DIFF
--- a/include/PrintSegmentAttributes.h
+++ b/include/PrintSegmentAttributes.h
@@ -11,6 +11,11 @@
 namespace cura
 {
 
+/*!
+ * Print attributes are flags that can be added to some print segments to indicate that they have been processed a specific way,
+ * e.g. by using overhanging or bridging settings. They are used for display purposes.
+ * This enumeration has an equivalent in Cura/cura/PrintSegmentAttributes.py
+ */
 enum class PrintSegmentAttribute : uint8_t
 {
     None = 0,
--- a/include/PrintSegmentAttributes.h
+++ b/include/PrintSegmentAttributes.h
@@ -18,9 +18,11 @@ namespace cura
  */
 enum class PrintSegmentAttribute : uint8_t
 {
-    None = 0,
-    Overhanging = 0x1,
-    Bridging = 0x2,
+    // clang-format off
+    None        = 0b00,
+    Overhanging = 0b01,
+    Bridging    = 0b10,
+    // clang-format on
 };
 
 using PrintSegmentAttributes = Flags<PrintSegmentAttribute>;
--- a/include/PrintSegmentAttributes.h
+++ b/include/PrintSegmentAttributes.h
@@ -1,0 +1,25 @@
+// Copyright (c) 2026 UltiMaker
+// CuraEngine is released under the terms of the AGPLv3 or higher
+
+#ifndef PRINT_SEGMENT_FLAG
+#define PRINT_SEGMENT_FLAG
+
+#include <cstdint>
+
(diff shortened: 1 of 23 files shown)


include/PrintSegmentAttributes.h
```diff
@@ -0,0 +1,30 @@
...
+/*!
+ * Print attributes are flags that can be added to some print segments to indicate that they have been processed a specific way,
+ * e.g. by using overhanging or bridging settings. They are used for display purposes.
+ * This enumeration has an equivalent in Cura/cura/PrintSegmentAttributes.py
+ */
+enum class PrintSegmentAttribute : uint8_t
+{
+    None = 0,
+    Overhanging = 0x1,
+    Bridging = 0x2,
```
rburema: Not really important, but: You can do `0b0`, `0b1`, `0b10` these days (though that might get quite verbose if we have many of them) -- I guess you can also do `1 << n` though 🤔

src/LayerPlan.cpp
```diff
@@ -58,12 +58,13 @@ GCodePath* LayerPlan::getLatestPathWithConfig(
...
+    const bool travel_to_z,
+    const PrintSegmentAttributes print_attributes)
 {
     std::vector<GCodePath>& paths = extruder_plans_.back().paths_;
-    if (paths.size() > 0 && paths.back().config == config && ! paths.back().done && paths.back().flow == flow && paths.back().width_factor == width_factor
-        && paths.back().speed_factor == speed_factor && paths.back().z_offset == z_offset
-        && paths.back().mesh == current_mesh_) // spiralize can only change when a travel path is in between
+    if (! paths.empty() && paths.back().config == config && ! paths.back().done && paths.back().flow == flow && paths.back().width_factor == width_factor
+        && paths.back().speed_factor == speed_factor && paths.back().z_offset == z_offset && paths.back().mesh == current_mesh_
+        && paths.back().print_attributes == print_attributes) // spiralize can only change when a travel path is in between
```
rburema, line 67: I guess we absolutely need this to show it in the front-end properly, and in practice it won't make a difference, but _in theory_ we could have a situation where say, the bridging and overhang are printed with the same settings, which would previously be merged.

rburema (approved):
Far less intimidating than it seems on first glance; basically all of this code is just threading the attributes variable where it needs to go.

I don't really have any serious comments on this, despite its size, seems solid!

I'm going to try all of this out on Windows before I sent it to QA though -- normally I'd skip that, but this has Arcus PR's connected to it, and I'll need it working on my system anyway when we merge.

include/PrintSegmentAttributes.h
wawanbreton: Indeed there are many ways of displaying this properly. I used the `0x` because this is what I'm used to :smile: but the binary notation is more explicit indeed.

src/LayerPlan.cpp
wawanbreton: Yes indeed, this can split up `GCodePath`s that were previously in one part, to multiple parts. I remember doing that in the past and then having to fix some code that was impacted by side-effect. But on the principle, the rest of the engine should be independant of the `GCodePath`s grouping, especially for this kind of case.

PR #2368, opened 2026-09-02
[CURA-13291] 'Sharpen' (wall) bridging conditions.

rburema:
(For the newly introduced setting(s), see front-end PR: https://github.com/Ultimaker/Cura/pull/21782 )

What was originally requested as part of the ticket:

Update the default behaviour of (wall+) bridging with the following changes:
- Tool paths must be anchored by the model-area (_not_ the support area++) on the layer below.
- Curves in bridges are to be avoided, except for when within the tolerance of a newly introduced setting 'bridge maximum deviation'.
- Lines partially supported by the layer below are not to be considered bridging. Use the overhang angle as a threshold to determine 'partially supported'.

+) From the context, I've taken to mostly mean the _wall_ bridging specifically. I mostly didn't touch skin bridging, except for the changes to the bridging mask I suppose (but even that should mostly effect the wall-bridging rather than the skin-bridging, given that the changes will be mostly smaller than a wall-width).
++) Though this set of PR's includes a setting to toggle this behaviour, so the maintainers of 3rd party printers (and our own experiments) have a chance to --at least when it comes to whether or not to regard support as anchoring for bridges specifically-- the 'old' behaviour.

----

So, what was a relatively simple bundle of requests from our internal stakeholders, turned into a decent into unravelling sanity (maybe that's overstating it a _little_ bit...), after I determined that a) even  if only focussing on the walls, there where about 3-ish sources of truth where bridges where supposed to start/end, which b) made, in combination with none of these being _complete_, leading to (apparent) bugs, made it (almost?) completely impossible to even test the changes I'd made.

So halfway through I decided that enough was enough, the technical debt was out of hand, and re-implemented a large part of the wall-bridging logic. (This does mean that some of the initial commits here are overwritten to a large extent by later ones.)


wawanbreton (commented):

| Benchmark suite | Current: 85e90a655a48b5f69e6da178149d1703eb961126 | Previous: 5230d17ad341f4cf101a93b2736d3ed95552a5c7 | Ratio |
|-|-|-|-|
| `SimplifyTestFixture/simplify_local` | `1.415090308394044` ns/iter | `0.9184018838490322` ns/iter | `1.54` |

include/LayerPlan.h
```diff
@@ -1213,15 +1218,64 @@ class LayerPlan : public NoCopy
...
-     * \brief Helper function to calculate the distance from the start of the current wall line to the first bridge segment
-     * \param wall The currently processed wall
-     * \param current_index The index of the currently processed point
-     * \param min_bridge_line_len The minimum line width to allow an extrusion move to be processed as a bridge move
-     * \param direction The direction to look for, 1 to use the actual line direction, -1 to go backwards
-     * \return The distance from the start of the current wall line to the first bridge segment
+     * \brief Convert from 'bridge location' structs, which are 'per bridge' & absolute, to 'bridging subsections' tuples, which are per wall-segment and relative to the lenght of
+     * that segment. \param wall The entire wall. \param bridge_locations The input bridge-locations (one bridge-location for one complete bridge). \param out_bridging_subsections
+     * The output bridging-subsections (one vector of bridging subsections for each wall-segment). \param direction The direction in which the head moves over the wall.
+     */
```
wawanbreton, line 1241: Documentation indentation is broken, probably because of the IDE

include/LayerPlan.h
```diff
@@ -1213,15 +1218,64 @@ class LayerPlan : public NoCopy
...
+    void convertBridgeLocations(
+        const ExtrusionLine& wall,
+        const std::vector<BridgeLocation>& bridge_locations,
+        std::vector<std::vector<std::tuple<Ratio, Ratio>>>& out_bridging_subsections,
```
wawanbreton: Why not give this vector as the return value ?

include/LayerPlan.h
```diff
@@ -1213,15 +1218,64 @@ class LayerPlan : public NoCopy
...
+        const std::vector<BridgeLocation>& bridge_locations,
+        std::vector<coord_t>& out_next_bridge_dists,
+        const int direction) const;
+
+
+    /*!
+     * \brief Find the locations of each bridge for the input-wall.
+     * \param wall The currently processed wall.
+     * \param current_index The index of the currently processed point.
+     * \param min_bridge_line_len The minimum line width to allow an extrusion move to be processed as a bridge move.
```
wawanbreton:
```suggestion
     * \param min_bridge_line_len The minimum line length to allow an extrusion move to be processed as a bridge move.
 ```

File: include/LayerPlan.h [hunk 4]
```diff
@@ -1213,15 +1218,64 @@ class LayerPlan : public NoCopy
...
+     * \param min_anchor_distance The minimum distance needed before and after a bridge to consider it properly 'anchored' (and therefore not a brdige).
```
wawanbreton:
```suggestion
     * \param min_anchor_distance The minimum distance needed before and after a bridge to consider it properly 'anchored' (and therefore a bridge).
 ```

File: include/LayerPlan.h [hunk 5]
```diff
@@ -1213,15 +1218,64 @@ class LayerPlan : public NoCopy
...
+     * \param max_bridge_deviation The maximum distance a bridge can be from the 'direct route' before it's not considered a bridge anymore.
```
wawanbreton: In context I know what you mean, but maybe otherwise that would help ? Could be an other adjective.
```suggestion
     * \param max_bridge_deviation The maximum radial distance a bridge can be from the 'direct route' before it's not considered a bridge anymore.
```

include/LayerPlan.h
```diff
@@ -1213,15 +1218,64 @@ class LayerPlan : public NoCopy
...
+     * \param out_bridge_locations The output bridge locations.
+     * \param direction The direction to look for, 1 to use the actual line direction, -1 to go backwards.
+     */
+    void findBridgingSections(
+        const ExtrusionLine& wall,
+        const size_t current_index,
+        const coord_t min_bridge_line_len,
+        const coord_t min_anchor_distance,
+        const coord_t max_bridge_deviation,
+        std::vector<BridgeLocation>& out_bridge_locations,
```
wawanbreton: Why not give this vector as the return value ?

include/LayerPlan.h
```diff
@@ -1213,15 +1218,64 @@ class LayerPlan : public NoCopy
...
+
+
+    /*!
+     * \brief Find the locations of each bridge for the input-wall.
+     * \param wall The currently processed wall.
+     * \param current_index The index of the currently processed point.
+     * \param min_bridge_line_len The minimum line width to allow an extrusion move to be processed as a bridge move.
+     * \param min_anchor_distance The minimum distance needed before and after a bridge to consider it properly 'anchored' (and therefore not a brdige).
+     * \param max_bridge_deviation The maximum distance a bridge can be from the 'direct route' before it's not considered a bridge anymore.
+     * \param out_bridge_locations The output bridge locations.
```
wawanbreton:
```suggestion
     * \param[out] out_bridge_locations The output bridge locations.
```

src/LayerPlan.cpp
```diff
@@ -1353,6 +1347,27 @@ std::tuple<size_t, Point2LL> LayerPlan::addSplitWall(
     const bool compute_distance_to_bridge_start,
     const AddExtrusionSegmentFunction<PathType>& func_add_segment)
 {
+    std::vector<coord_t> distances_to_next_bridge;
+    std::vector<std::vector<std::tuple<Ratio, Ratio>>> bridging_subsections_per_segment;
+
+    { // start scope (for bridging_locations)
```
wawanbreton: I don't see a high value in this comment 🤔

src/LayerPlan.cpp
```diff
@@ -1353,6 +1347,27 @@ std::tuple<size_t, Point2LL> LayerPlan::addSplitWall(
...
+        std::vector<BridgeLocation> bridging_locations;
```
wawanbreton: Looks like this variable could be within the `if consexpr` scope, thus eliminating the need for the higher scope.

src/LayerPlan.cpp
```diff
@@ -1694,93 +1704,294 @@ std::vector<LayerPlan::PathCoasting>
...
+            ending_anchor_distance += line_distance;
+            continue;
+        }
 
-            if (bridge_wall_mask_bb_.hit(AABB({ p0.p_, p1.p_ })) && PolygonUtils::polygonCollidesWithLineSegment(bridge_wall_mask_, p0.p_, p1.p_))
-            {
-                constexpr bool restitch = false; // only a single line doesn't need stitching
-                OpenLinesSet intersections_with_bridge_mask = bridge_wall_mask_.intersection(OpenLinesSet(OpenPolyline({ p0.p_, p1.p_ })), restitch);
+        constexpr bool restitch = false; // only a single line doesn't need stitching
+        OpenLinesSet intersections_with_bridge_mask = bridge_wall_mask_.intersection(OpenLinesSet(OpenPolyline({ p0.p_, p1.p_ })), restitch);
```
wawanbreton, line 1740: There is a very similar algo to calculate the intersections of the segments with the overhanging regions. It uses the `Shape::intersectionsWithSegment` which doesn't give sub-segments but the intersections along the segment, that are easily sortable. Maybe it could help simplify this piece of code ? Execution may be longer though, because inner implementation is not really smart. Merely a FYI.
See https://github.com/Ultimaker/CuraEngine/blob/278fdb03ce078949cd91d2630fdf804d356d2928/src/LayerPlan.cpp#L657

src/LayerPlan.cpp
```diff
@@ -1694,93 +1704,294 @@ std::vector<LayerPlan::PathCoasting>
...
+
+        // Finally, loop over the intersections to find candidate bridge segments.
+        for (const OpenPolyline& intersection_segment : intersections_with_bridge_mask)
+        {
+            const auto& [b0, b1] = (vSize2(intersection_segment[0] - p0.p_) < vSize2(intersection_segment[1] - p0.p_)) ? std::tie(intersection_segment[0], intersection_segment[1])
+                                                                                                                       : std::tie(intersection_segment[1], intersection_segment[0]);
+
+            if (vSize2(b1 - b0) < EPSILON_SQUARED)
+            {
+                // don't consider neligable lenghts in any case however
```
wawanbreton:
```suggestion
                // don't consider negligble lengths in any case however
```

src/LayerPlan.cpp
```diff
@@ -1694,93 +1704,294 @@ std::vector<LayerPlan::PathCoasting>
...
+                continue;
+            }
+
+            // add a candidate bridge-segment here (don't check anchoring, distance, max deviation or distance to first yet; that's for the next step)
+            // note that they're also not merged yet
+            const coord_t start_len = vSize(b0 - p0.p_);
+            const coord_t end_len = vSize(b1 - p0.p_);
+            const coord_t from_end = vSize(p1.p_ - b1);
+            bridge_segment_candidates.emplace_back(idx_0, idx_1, start_len, end_len, line_distance - (start_len + from_end), from_end, total_distance_p0 + start_len);
```
wawanbreton, line 1802: This syntax used not to compile on our old Mac runner. Maybe it does on the updated one though, would love it 🤞

src/LayerPlan.cpp
```diff
@@ -1694,93 +1704,294 @@ std::vector<LayerPlan::PathCoasting>
...
+        {
+            if (LinearAlg2D::getDistFromLine(wall[pt_idx].p_, start_pt, end_pt) > max_bridge_deviation)
+            {
+                // don't add a bridge -- we're over the max deviation here
+                return;
+            }
+        }
+
+        out_bridge_locations.push_back(std::move(bridge));
+    };
```
wawanbreton, line 1838: Can that easily be made an external function ? Maybe the `point_index` could also be a utility function somewhere.

src/LayerPlan.cpp
```diff
@@ -1694,93 +1704,294 @@ std::vector<LayerPlan::PathCoasting>
...
+            const coord_t end_len = vSize(b1 - p0.p_);
+            const coord_t from_end = vSize(p1.p_ - b1);
+            bridge_segment_candidates.emplace_back(idx_0, idx_1, start_len, end_len, line_distance - (start_len + from_end), from_end, total_distance_p0 + start_len);
+        }
+
+        total_distance_p0 += line_distance;
+        ending_anchor_distance = bridge_segment_candidates.empty() ? 0 : bridge_segment_candidates.back().backwards_end_dist;
+    }
+
+    // append to result function -- also checks if the bridge-legnth is sufficient and if it stays under the max deviation
```
wawanbreton:
```suggestion
    // append to result function -- also checks if the bridge-length is sufficient and if it stays under the max deviation
```

src/LayerPlan.cpp
```diff
@@ -1694,93 +1704,294 @@ std::vector<LayerPlan::PathCoasting>
...
+        }
+
+        // get the endpoints of the bridge
+        const auto p00 = wall[bridge.wall_idx_start];
+        const auto p01 = wall[point_index(bridge.wall_idx_start + direction)];
+        const auto start_pt = p00.p_ + normal(p01.p_ - p00.p_, bridge.start_dist);
+
+        const auto p10 = wall[point_index(bridge.wall_idx_end - direction)];
+        const auto p11 = wall[bridge.wall_idx_end];
+        const auto end_pt = p10.p_ + normal(p11.p_ - p10.p_, bridge.end_dist);
```
wawanbreton: Maybe some of them can be turned into references ? Not sure `auto` does it when possible.

src/LayerPlan.cpp
```diff
@@ -1694,93 +1704,294 @@ std::vector<LayerPlan::PathCoasting>
...
+                return;
+            }
+        }
+
+        out_bridge_locations.push_back(std::move(bridge));
+    };
+
+    // merge bridges, check if the resulting bridges conform to the parameters, and if so, add them to the out batch
+    std::optional<BridgeLocation> current_bridge;
+    coord_t skipped_first_anchor_len
```
wawanbreton, line 1842: From what I can see, this variable can be declared in the for loop. Or maybe it is made on purpose for the comment ?

wawanbreton (commented):
The code looks much better than the previous one, the structure makes a lot of sense and I can definitely understand it more easily ❤️ 
I have some small remarks about the syntax and possible improvements, but nothing big.

src/LayerPlan.cpp
rburema: It uses min-bridge-length, max-deviation, the direction, the wall-collection, the point-index function, and it pushes the bridge to the local out-bridge-locations.

It's also very specific to this function, very probably not going to be used anywhere else.

I'm willing to do this (don't think it'll make things worse -- since on the one hand it clarifies the surrounding function, but it also removes the inner function from its original context), but I don't really see the benefit much here.

Point-index is a better candidate though, I'll do that -- it does annoy me that I repeated that so often.

src/LayerPlan.cpp
rburema: It's used over the loop though not just in one particular iteration (although it's abandoned as soon as we have a bridge-candidate with a proper begin anchor) ... it's added to in one part of the if-if (in the else-else part) and used in another iteration of the loop (in the else-if part) on the first encounter of a bridge (to check if) that actually has the required beginning anchor length.

src/LayerPlan.cpp
rburema: 🤞🏻 -- let's hope so, not changing it right now just in case it does work 😁

src/LayerPlan.cpp
rburema: I'm not sure about this -- the sorting part was 'stolen' from the earlier bridge implementation, and I didn't want to mess too much with it 😅 -- let's make a mental note of this and maybe do it when the bridging conditions are in and tested?

src/LayerPlan.cpp
rburema: Good catch. Auto doesn't (always) do this (and definitely not in this situation) -- you have to (mostly) specify modifiers yourself. I'll change this.

src/LayerPlan.cpp
wawanbreton: Sure, it doesn't have to be changed right now, especially if it works well 🙂

src/LayerPlan.cpp
wawanbreton: Looks like it does !!!! https://github.com/Ultimaker/Cura/actions/runs/33728359036

rburema: I guess they've finally updated apple clang again 🎉

wawanbreton (approved):

PR #2370, opened 2026-09-09
Fix tree support crash generation

wawanbreton:
Previously, the `TreeSupportElement` objects were stored directly in a map, and some pointers over them were used. However the maps can be reorganized in memory, so that makes these pointers very unsafe. Now we use smart pointers in the maps so that they can be referenced without issues. This also removes memory duplication of the elements in various maps, and boilerplate code to actually place the objects in the maps.

Note: the fix was made previously in the tree support improvement spike. It was just cherry-picked here.

CURA-13304

### DIFF
--- a/include/TreeSupportElement.h
+++ b/include/TreeSupportElement.h
@@ -43,9 +43,11 @@ struct AreaIncreaseSettings
     bool operator==(const AreaIncreaseSettings& other) const = default;
 };
 
-class TreeSupportElement
+class TreeSupportElement : public std::enable_shared_from_this<TreeSupportElement>
 {
 public:
+    using Ptr = std::shared_ptr<TreeSupportElement>;
+
     TreeSupportElement(
         coord_t distance_to_top,
         size_t target_height,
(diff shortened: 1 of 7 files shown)


wawanbreton (commented):

| Benchmark suite | Current: cd6991c2f85b9cf4e6e0d0938d49e0bf2fb432ca | Previous: b0efb82b10a1dbb1a0074c246922ac236dd1a60f | Ratio |
|-|-|-|-|
| `SimplifyTestFixture/simplify_slot_noplugin` | `4.05119484426529` ns/iter | `2.53327894515077` ns/iter | `1.60` |

include/TreeSupportElement.h
```diff
@@ -261,30 +234,36 @@ class TreeSupportElement
...
 
 private:
+    /*!
+     * \brief Create a new Element for one layer below the element of the pointer supplied.
+     */
+    explicit TreeSupportElement(const TreeSupportElement::Ptr& element_above);
+
+    TreeSupportElement(const TreeSupportElement::Ptr& elem, Shape* newArea);
+
+    // ONLY to be called in merge as it assumes a few assurances made by it.
```
rburema, line 244: In theory we could have a shibolet structure like `struct MergeOnly { /*empty*/ };` and then have that as an argument to this constructor -- but I think that might be overkill in this situation.

rburema (approved):
LGTM! Maybe keep an eye on the performance, but I think it'll be fine.

include/TreeSupportElement.h
wawanbreton: Well, yes about the overkill :sweat_smile: but I do agree with the idea. This is just a piece of code that I moved btw :see_no_evil:

rburema: Yeah I thought I recognized the `ONLY ...` comment from somewhere 😁

PR #2371, opened 2026-09-11
CURA-13166 Add basic instructions to customize Copilot PR review

wawanbreton:
Hopefully this can make Copilot reviews more relevant. This is still very experimental and will have to be fine-tuned over time, if ever useful.
Feel free to suggest additions if you feel something critical is missing. I do think it should remain short though.

:information_source: The files were originally based on the ones generated by Jelle, but those are more oriented for agent-generated PRs, which is not what we want, so most of it was irrelevant for our case. 

CURA-13166


.github/skills/code-review/SKILL.md
```diff
@@ -0,0 +1,16 @@
...
+
+* In your main comment, do not publish a pull request overview, file summaries or review details
+* Generated comments should be as concise as possible
+* Do not report code styling issues, we have an automated action for that
+* Create replacement code suggestions in the comment when the change you suggest is straightforward, e.g. for typos
+* Issue a warning when a piece of code is quite critical, very suitable for being unit tested, and no test has been added yet
+* Some code-related rules:
+  * All the variables and functions should have explicit names
+  * The use of the `auto` keyword is not to be enforced, but it can be suggested when extremely relevant
+  * `for` loops are to be preferred over `while` loops when possible
```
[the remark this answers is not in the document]
wawanbreton: @copilot I want to enforce the use of `for` loops whenever possible, as long as it does not break the algorithm or reduces readibility. What would be a good instruction for that ?

.github/skills/code-review/SKILL.md
```diff
@@ -0,0 +1,16 @@
...
+  * Exceptions should not be used
```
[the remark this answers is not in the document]
wawanbreton: @copilot I want to make sure new code doesn't explicitly make use of exceptions, unless that is really mandatory, e.g. when using a function from a library that does work by raising exceptions. What would be a good instruction for that ?

wawanbreton:
@copilot In the PR skill, I clearly mentioned not to generate the review details, however they are here. Is my instruction not good enough, or is there an other way I can disable this verbose and useless set of details ?

.github/skills/code-review/SKILL.md
```diff
@@ -0,0 +1,34 @@
...
+  * The use of the `auto` keyword is not to be enforced, but it can be suggested when extremely relevant
+  * Prefer `for` loops over `while` loop whenever possible and if it does not significantly reduce the readability
+  * In new code, avoid introducing explicit exception-based control flow; prefer error handling by return value, unless exceptions are mandatory (e.g. required by external library/APIs)
+  * Short comments should be present in very complex pieces of code
+  * Complex functions should be documented, but trivial ones don't need to be when their signature is already very explicit, e.g. getters
+  * The code should make use of the explicitly defined types as much as possible
+  * Most parts of the code are processed in parallel, so make sure we don't run into race-conditions and the code is entirely repeatable across consecutive executions
+  * Lambdas declared inside a function are allowed, but with the following attention points:
+    * The body of the nested function should not be longer than 30 lines
+    * Broad capturing is not allowed
```
rburema, line 37: Does this 'broad capturing' preclude capturing like this `[&]`? Because I do think we use that at the moment. I'm not necessarily against this, but (I think) it represents a change from where we are now. I'd like to weigh the arguments before and against it before putting it in a file that describes how we do things.

.github/skills/code-review/SKILL.md
```diff
@@ -0,0 +1,34 @@
+---
+name: code-review
+description: Repository-specific guidance for reviewing pull requests in CuraEngine.
+---
+# Role: Pull Request Assistant
+
+You are the Pull Request Assistant. Your primary directive is to help developers make sure the code they wrote is robust, modern and readable, for the **CuraEngine** repository.
+
```
rburema, line 8: Some more suggestions for this file:

* If, when calling a function with many parameters, for which the meaning isn't clear when viewed from the calling site, you may suggest the use of `const` or `constexpr` variables (whichever is most applicable).
* Most parameters, unless primitive, should be passed by reference, unless it's useful to do otherwise.
* Raw pointers are allowed when referring to 'existing' data, that is, there should be as little manual memory management as possbile. Smart pointers should be used when both memory management and pointers come together (or similar, such as like when a collection isn't stable during the lifetime of a pointer).
* We may compromise on some principles if it's absolutely needed for efficiencies' sake.
* If a simple change can lead to an efficiency gain, mention it (for example, pulling the calculation of a variable out of a loop).
* If something can be done simpler via ranges/views, mention it (ranges/v3 can be used, otherwise we're still on C++20).

AGENTS.md
```diff
@@ -0,0 +1,58 @@
...
+* Send the extrusion data (with metadata) to the front-end, and the final gcode alongside
+
+Since the input meshes can have very different shapes, we try to handle all the possible cases and use safe code as much as possible. We also focus very much on efficiency, since some meshes can have a very large number of triangles, or be large in physical size, which means the amount of generated extrusions is huge.
+
+## Development
+
+### Codebase
+The codebase is essentially C++. Some parts of it are quite old, and possibly written at a time where there were no strict rules. But every time we make changes, we try to upgrade it with modern standards. The one we use is C++20, so not all features of modern C++ are available to us, because we need to support old platforms that don't support modern compilers. However we try to leverage the modern features as much as possible, in order to simplify our code, make it more portable and faster.
+
+The application will be built on both Linux, Windows and Mac platforms. So we have many specific cases here and there for each platform, and it is important that they all keep working.
```
rburema:
```suggestion
The application will be built for Linux, Windows and Mac platforms. So we have many specific cases here and there for each platform, and it is important that they all keep working.
```

AGENTS.md
```diff
@@ -0,0 +1,58 @@
...
+
+It also has a few unit testing and benchmarking sub-projects that are run periodically, so it is critical that they keep working.
+
+### Computational geometry
+Given its nature, CuraEngine contains a lot of computational geometry algorithms. The main library we use is clipper, for boolean operations (union, intersections, difference) and offsetting.
+Since clipper works only with integer values, we have adopted the following conventions for numeric types:
+* By default, geometric coordinates are typed with the `coord_t` type, which is an alias to `signed long long`. Its derivates can also be used: `Point2LL` and `Point3LL`. This way we can give those elements directly to clipper. Physically, those coordinates represent micrometers. All the values that represent a distance/position in physical space should then use this type.
+* When we require floating-point calculation, we use `float` type by default, and its derivatives: `Point2F` and `Point3F`
+* When we require floating-point calculation with a specific need for precision, we use the `double` type, and its derivatives: `Point2D` and `Point3D`
+
```
rburema, line 53: * We usually consider geometric points to be effectively the same when they're 5 micron or less apart. (Or areas to be negligible when less than 25 square micron.) -- See also the EPSILON value mentioned below.

AGENTS.md
```diff
@@ -0,0 +1,58 @@
...
+* CMake for building
+* protobuf to generate messages for the front-end application
+
+It also has a few unit testing and benchmarking sub-projects that are run periodically, so it is critical that they keep working.
+
+### Computational geometry
+Given its nature, CuraEngine contains a lot of computational geometry algorithms. The main library we use is clipper, for boolean operations (union, intersections, difference) and offsetting.
+Since clipper works only with integer values, we have adopted the following conventions for numeric types:
+* By default, geometric coordinates are typed with the `coord_t` type, which is an alias to `signed long long`. Its derivates can also be used: `Point2LL` and `Point3LL`. This way we can give those elements directly to clipper. Physically, those coordinates represent micrometers. All the values that represent a distance/position in physical space should then use this type.
```
rburema, line 50:
```suggestion
* By default, geometric coordinates are typed with the `coord_t` type (most often this corresponds to 'microns' in world units), which is an alias to `signed long long`. Its derivates can also be used: `Point2LL` and `Point3LL`. This way we can give those elements directly to clipper. Physically, those coordinates represent micrometers. All the values that represent a distance/position in physical space should then use this type.
```

AGENTS.md
```diff
@@ -0,0 +1,58 @@
...
+
+Since the input meshes can have very different shapes, we try to handle all the possible cases and use safe code as much as possible. We also focus very much on efficiency, since some meshes can have a very large number of triangles, or be large in physical size, which means the amount of generated extrusions is huge.
+
+## Development
+
+### Codebase
+The codebase is essentially C++. Some parts of it are quite old, and possibly written at a time where there were no strict rules. But every time we make changes, we try to upgrade it with modern standards. The one we use is C++20, so not all features of modern C++ are available to us, because we need to support old platforms that don't support modern compilers. However we try to leverage the modern features as much as possible, in order to simplify our code, make it more portable and faster.
+
+The application will be built on both Linux, Windows and Mac platforms. So we have many specific cases here and there for each platform, and it is important that they all keep working.
+
```
rburema, line 28: Some of the libraries we work with in this repository are (though this list is not exhaustive and doesn't mention all of them); ranges/v3, fmt, (a version of) clipperlib and (some of) boost.

rburema (changes requested):
I'd like to know what you think about my remarks first before we merge this. Though I could be convinced to merge as-is.

.github/skills/code-review/SKILL.md
wawanbreton: You are correct, this is totally about the `[&]`
The first version I wrote did allow using it if the number of captured elements was more than 10, however Copilot strongly advised against it, since it can really lead to unexpected bugs, and also mentioned than having such a large number means the code should probably be refactored to avoid it from the beginning, which makes some sense. I would then leave it for now and see in practice how we can avoid it. Also remember that this will apply only on new code, and only create a copilot remark that you are totally allowed to ignore if you feel like it.

.github/skills/code-review/SKILL.md
wawanbreton: I will add that, good remarks.

> * We may compromise on some principles if it's absolutely needed for efficiencies' sake.

Again, worst case is that a non-relevant remark is generated, that you can totally ignore.

AGENTS.md
wawanbreton: Are there cases where we use this type but not for microns ?

rburema: Yes, for example, sometimes we count something, and then divide a real-world measure by that count (or similar), then it's useful to have that count be the same type as the measure.

You could say that it's always a consequence of needing to use microns though -- I'm not actually sure if it will be smart enough to 'think' that way though.

rburema (approved):
Well, let's at least try what this does, I suppose 🤞🏻 ... Which is not a line that previously belonged in a code review, but here we are 😓

Not sure about the 'All the variables of a class should be declared private' rule. But as long as it 'understands' that (simple) structs should probably be excluded from that, I'm fine to try it.

AGENTS.md
wawanbreton: Well, I would say let's try and see what happens, maybe that happens rarely enough that we can ignore/fix it

wawanbreton:
> Not sure about the 'All the variables of a class should be declared private' rule. But as long as it 'understands' that (simple) structs should probably be excluded from that, I'm fine to try it.

Well, technically the only difference between structs and classes is that classes don't have privacy, so I assume it will do the distinction :smile:

PR #2374, opened 2026-09-14
CURA-13328 Set wall computation proper arguments

wawanbreton:
By adding this argument, a different constructor is called, and the other arguments now have their proper meaning.

CURA-13328


rburema (approved):
Good find! I do see that the documentation for the parameters (in the WallToolPaths.h header) is a bit out of date (several parameters are missing).

I also wonder if we could solve the constructor confusion in this particular instance by making `WallToolpathGenerator` a mandatory, instead of optional, argument.

wawanbreton:
> Good find! I do see that the documentation for the parameters (in the WallToolPaths.h header) is a bit out of date (several parameters are missing).
> 
> I also wonder if we could solve the constructor confusion in this particular instance by making `WallToolpathGenerator` a mandatory, instead of optional, argument.

Hmm, the documentation is very poor indeed. The 2 constructors have a very different usage, and this is not really mentioned.
Making the `generator` argument non-optional would reduce the confusion a bit indeed, but I think it does make sense here because most of the called do want to use Arachne un-explicitly. On the other hand, there are very few calls of them, so we could also update them.

wawanbreton:
@rburema I have added the argument explicitization in 0c556615677b0abda0e36e4a798bf280e2bfcc45, but not sure I like it :smile: What do you think ?

rburema:
> Remco Burema (@rburema) I have added the argument explicitization in [0c55661](https://github.com/Ultimaker/CuraEngine/commit/0c556615677b0abda0e36e4a798bf280e2bfcc45), but not sure I like it 😄 What do you think ?

It's just an enum, so it would already be constant, and since the value already has a meaningful reading on it's own (unlike a number like 1) I don't think it needs the `constexpr` explaining/documenting variable before that.

PR #2375, opened 2026-09-14
CURA-13112 Apply bridging separately for flooring

wawanbreton:
This allows flooring and roofing to be printed with the proper extruder, even when the bridging is active

CURA-13112


rburema (approved):
LGTM!

PR #2376, opened 2026-09-15
CURA-13303 Add specific combing ignore distance for skin support

wawanbreton:
When printing skin support lines with combing and interlacing enabled, it adds a backward travel move for each line, which degrades the anchoring of the line. Now we have a specific minimum distance for combing when printing those lines, so that combing is mostly ignored.

CURA-13303


src/LayerPlan.cpp
```diff
@@ -2609,7 +2612,10 @@ void LayerPlan::addLinesInGivenOrder(
                 // Make sure we don't travel at fractional height
                 addTravel(getLastPlannedPositionOrStartingPosition());
             }
-            addTravel(start);
+
+            constexpr auto force_retract{ ForceRetract::AUTOMATIC };
```
rburema: I think these sorts of construction are mostly meant to help the reader understand the code, not for any sort of other advantage; with that in mind, I wonder what the thought behind making this a separate variable is? (Since implicitly, the enum is already a compile-time constant, and you have to explicitly say it's a `ForceRetract` enum specifically anyway ... so why have a separate `force_retract` constant as well?)

src/InfillOrderOptimizer.cpp
```diff
@@ -494,6 +494,10 @@ void InfillOrderOptimizer::addSkinSupportLinesToLayer(
         const coord_t max_adjacent_distance = skin_support_line_distance * 1.1;
         constexpr coord_t exclude_distance = 0;
         constexpr bool skin_support_interlaced = true;
+
+        // You really want to avoid combing because it adds a backwards travel move that prevents the bridging lines to stick to their anchors
+        const coord_t max_distance_ignore_combing = max_adjacent_distance * 4;
```
rburema, line 499: What is '4' here based on specifically? (It's fine, we do this all over the code, I just thought there might be a reason.)

src/LayerPlan.cpp
```diff
@@ -425,7 +425,7 @@ GCodePath& LayerPlan::addTravel(const Point2LL& p, const ForceRetract force_retr
 
         // Divide by 2 to get the radius
         // Multiply by 2 because if two lines start and end points places very close then will be applied combing with retractions. (Ex: for brim)
-        const coord_t max_distance_ignored = mesh_or_extruder_settings.get<coord_t>("machine_nozzle_tip_outer_diameter") / 2 * 2;
+        const coord_t max_distance_ignored = max_distance_ignore_combing.value_or(mesh_or_extruder_settings.get<coord_t>("machine_nozzle_tip_outer_diameter") / 2 * 2);
```
rburema: Maybe the `... / 2 * 2` should live outside of the brackets?

rburema (approved):
LGTM, some comments, nothing serious.

src/LayerPlan.cpp
wawanbreton: Hmm, that is entirely true, I am so much used to this syntax that I now also apply for enums :sweat_smile: I'll remove it

src/InfillOrderOptimizer.cpp
wawanbreton: Technically should be 2, which is the distance between 2 consecutive interlaced lines. In practice this is not enough because for e.g. a circle, the tips are more than 2 distances away. Maybe 3 should be enough, but because we want to really avoid this happening, 4 seems to be the right-ish number.

src/LayerPlan.cpp
wawanbreton: Hmm, you are probably right. Will change it.

PR #2377, opened 2026-09-16
[CURA-13334] Fix skin-bridge-line mis-alignment with bridge-walls.

rburema:
The scoring system we use to determine the angle of the bridging-lines for skin sometimes ended up _almost_ aligning with the walls, instead of _fully_. Giving a small bonus when grading skin-bridge-lines for alignment with the longest walls in the same direction solves this problem.

CURA-13334


src/bridge/bridge.cpp
```diff
@@ -53,6 +53,51 @@ std::vector<coord_t> shapeLineIntersections(const coord_t line_y, const Transfor
...
+    {
+        return 0.0;
+    }
+
+    // Find the 'longest' segments (in forwards and backwards directions) w.r.t. the current direction.
+    const TransformedSegment* skin_forward_longest{ nullptr };
+    const TransformedSegment* skin_backward_longest{ nullptr };
+    for (const auto& segment : transformed_skin_area.getSegments())
+    {
+        const auto vec_x{ segment.getEnd().X - segment.getStart().X };
```
wawanbreton: To be really nitpicky, I would prefer this variable to be called `length_x` instead, what do you think ?
Same on line 84

wawanbreton (approved):
Like it ! Finally that should fix this annoying issue

src/bridge/bridge.cpp
rburema: Here, sure, but on line 84 it's an actual vector -- all the more reason to rename the first one though!

rburema: Done! see 723c68e

PR #2378, opened 2026-09-16
update the flag for processing skin part

HellAholic:
This pull request makes a small correction to the `processSkinPart` function in `src/FffGcodeWriter.cpp`. The change fixes the value of the `is_not_flooring_roofing` constant.

Required to generate the small top bottom areas without setting the  Small top bottom on surface to true. 
image

CURA-13338


rburema (approved):
LGTM!
Easy to review as well, what with these two booleans right above one-another :-p

